PR #1, opened 2017-01-18
Adds lerna

talend-glorieux:
* Ran `lerna init` on the project. Generated package.json and
lerna.json
* Ran `yarn` on the project. Generated yarn.lock
* Adds .gitignore


package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "devDependencies": {
+    "lerna": "2.0.0-beta.32"
```
jmfrancois: Do we need to depends on the beta ? This is very boring to update.
> The instructions below are for Lerna 2.x which is currently in beta. We recommend using it instead of 1.x for a new Lerna project

.gitignore
```diff
@@ -0,0 +1 @@
+node_modules
```
jmfrancois: Could you please init this file using the template in tools ?

package.json
talend-glorieux: OK I'll go back to 1.x.

talend-glorieux: Well the entire documentation is based on that version.

jmfrancois: So we depend on the features of the 2.x (lerna run and so on)

PR #55, opened 2017-01-26
feat(containers): connect ObjectViewer component

jmfrancois:

The ObjectViewer component is at the moment only needed by TFD.
It only exists as a pure component

**What is the new behavior?**

Now the component only require data + displayMode.
Will be fixed later because not needed at the moment the onChange + onSubmit are at the moment not finished.

screen shot 2017-01-26 at 14 15 36


packages/containers/examples/ExampleObjectViewer.js
```diff
@@ -0,0 +1,60 @@
...
+			<IconsProvider />
+			<ObjectViewer data={data} displayMode="table" />
+		</div>
+	))
+	.addWithInfo('flat default', () => (
+		<div>
+			<IconsProvider />
+			<ObjectViewer data={data} displayMode="flat" />
+		</div>
+	));*/
```
talend-glorieux: Why did you comment all of that block?

packages/containers/src/ObjectViewer/ObjectViewer.container.js
```diff
@@ -0,0 +1,92 @@
...
+		stateWillMount(this.props);
+	}
+
+	onClick(event, data) {
+		const opened = this.props.state.get('opened');
+		const edited = this.props.state.get('edited');
+		let newState;
+		if (data.isOpened) {
+			// we should close the tree
+			newState = this.props.state.set(
```
talend-glorieux: I guess a function would make more sense than a comment. 
Plus you could reduce repeated code.

packages/generator/generators/react-component/templates/src/full.connect.js
```diff
@@ -11,38 +12,21 @@ export function getContainerInfo(ownProps) {
...
-		]),
-	};
+	const info = getContainerInfo(ownProps);
+	return getStateProps(state, info.name, info.id);
+}
+
+/* if you need it
+export function mergeProps(stateProps, dispatchProps, ownProps) {
 }
+*/
```
talend-glorieux: `if you need it`?

packages/containers/src/ObjectViewer/ObjectViewer.connect.js
```diff
@@ -0,0 +1,29 @@
+import { connect } from 'react-redux';
+import { actions } from 'react-cmf';
```
jsomsanith: `actions` is not used

packages/containers/src/ObjectViewer/ObjectViewer.connect.js
```diff
@@ -0,0 +1,29 @@
...
+	return {
+		name: 'ObjectViewer',
+		id: ownProps.id || 'ObjectViewer',
+	};
+}
+
+export function mapDispatchToProps(dispatch, ownProps) {
+	const info = getContainerInfo(ownProps);
+	const props = getStateAccessors(dispatch, info.name, info.id, DEFAULT_STATE);
+	props.dispatch = dispatch;
```
jsomsanith: `dispatch` is passed as props but never used in the container

packages/containers/src/ObjectViewer/ObjectViewer.container.js
```diff
@@ -0,0 +1,92 @@
+import React, { PropTypes } from 'react';
+import Immutable, { List, Map } from 'immutable';
```
jsomsanith: `Immutable` is imported but not used

packages/containers/src/ObjectViewer/ObjectViewer.container.js
```diff
@@ -0,0 +1,92 @@
...
+	modified: new Map(),  // Store the onChange
+});
+
+class ObjectViewer extends React.Component {
+	static displayName = 'CMFContainer(ObjectViewer)';
+	static propTypes = {
+		id: PropTypes.string,
+		data: get(Component, 'propTypes.data'),
+		displayMode: get(Component, 'propTypes.displayMode'),
+		onSubmit: get(Component, 'propTypes.onSubmit'),
```
jsomsanith: Why do you use lodash `get` on those 3 lines instead of just getting it ? `Component.propTypes.onSubmit`

packages/containers/src/ObjectViewer/ObjectViewer.test.js
```diff
@@ -0,0 +1,85 @@
...
+		obj: {
+			bool: false,
+		},
+		arrayOb: [
+			{ foo: 3.2 },
+		],
+	},
+];
+
+describe('Component ObjectViewer', () => {
```
jsomsanith: This is already tested in components, this test is not really needed here

packages/containers/src/ObjectViewer/ObjectViewer.test.js
```diff
@@ -0,0 +1,85 @@
...
+		const props = mapStateToProps(state, {});
+		expect(typeof props).toBe('object');
+	});
+	it('should map state to props', () => {
+		const dispatch = () => {};
+		const props = mapDispatchToProps(dispatch, {});
+		expect(typeof props).toBe('object');
+	});
+});
+
```
jsomsanith: There is no user action tests (`onClick`, `onChange`)

packages/containers/src/ObjectViewer/ObjectViewer.container.js
jmfrancois: because in the tests it fails (we do mock on react-talend-components)

packages/containers/examples/ExampleObjectViewer.js
jmfrancois: it was because the config loader only support a function export. fixed

packages/containers/src/ObjectViewer/ObjectViewer.container.js
```diff
@@ -0,0 +1,101 @@
...
+import { ObjectViewer as Component } from 'react-talend-components';
+import { statePropTypes, stateWillMount } from '../state';
+
+export const DEFAULT_STATE = new Map({
+	edited: new List(),  // Array of JSONPath
+	opened: new List(),  // Array of JSONPath
+	modified: new Map(),  // Store the onChange
+});
+
+export function open(path, props) {
```
talend-glorieux: Why do you pass the props as arguments when you only need the state?

PR #95, opened 2017-01-31
fix(Component): List table header sort order

jsomsanith:

On current sorted column click, the `isDescending` flag is always `true`.
This is du to a bad refacto during review where we negate an undefined value.

**What is the new behavior?**
On current sorted column click, the `isDescending` value is negated.

- [ ] Yes
- [x] No


packages/components/src/List/DisplayTable/DisplayTable.snapshot.test.js
```diff
@@ -0,0 +1,201 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
+import { mount } from 'enzyme';
```
charlesnguyen: mount  is not used.

PR #99, opened 2017-02-01
fix(containers/list): add no toolbar + displayMode as props

jmfrancois:

If you don't pass toolbar props to List you have an javascript error;
If you pass the displayMode you will have the 'table' mode because it's handle by the state.

**What is the new behavior?**

If you don't pass toolbar props to List you have a rendered list without toolbar.
If you pass the displayMode it will set as default value for the state.
Note: then render has been refactored to compute props to pass to pure List.


jmfrancois:
[stack uploaded to surge](http://talend.99.surge.sh/)

talend-glorieux:
Can't see the result working on the static page.

jmfrancois:
:octocat: [Demo is available here](http://99.talend.surge.sh)

PR #101, opened 2017-02-01
fix(components/list): do not render toolbar when empty

jsomsanith:

When we have actions but nothing else in the toolbar, we had the Action toolbar and an empty toolbar just under.

**What is the new behavior?**
When we have no item in the toolbar, there is no empty navbar rendered

- [ ] Yes
- [x] No


jmfrancois:
Why did you need that change. The list call the toolbar only if toolbar props has been passed.
{toolbar && (<Toolbar {...toolbarProps} />)} in List.component.js

PR #104, opened 2017-02-02
feat(Component): List with fixed toolbar and header

jsomsanith:

The List component doesn't manage scroll. So within a container that can scroll, the whole list (including the toolbars and table header) scroll.
The guideline specifies that only the list content should scroll in the main layout.

**What is the new behavior?**
The list is now taking its container's height. The list part (only the `<tbody>` in Table) takes the available space properly and scroll if needed. The rest stay in place, even if user scroll.

This has been fully integrated into the Layout component. Stories have been added to demonstrate that.

- [ ] Yes
- [x] No

**Other information**:

Please note that the css has no fixed height except in one place because of Safari. It requires a container with a fixed size (not in %) to render a flex-column-layout with the size of it's container. Otherwise the flex-column just take all the height of its content.

This style hack is in the layout scss, with is the less ugly place because it already set the height of the header, so the header size is known (without a magic number) and reused to calculate et set the height of the main placeholder.

Tested on the Chrome, FF, Safari 10, Edge, IE11.

Thank you @frassinier for your help !


jmfrancois:
:octocat: [Demo is available here](http://104.talend.surge.sh)

jmfrancois:
:octocat: [Demo is available here](http://104.talend.surge.sh)

jmfrancois:
:octocat: [Demo is available here](http://104.talend.surge.sh)

jmfrancois:
:octocat: [Demo is available here](http://104.talend.surge.sh)

packages/components/.storybook/head.html
```diff
@@ -16,4 +16,8 @@
     .tc-list-display-table tr > td + td {
         max-width: 400px;
     }
+
+    .list-container {
+        height: 80vh;
```
frassinier: Why this is useful?

packages/components/src/List/DisplayTable/DisplayTable.component.js
```diff
@@ -147,7 +149,14 @@ function ListHeader(props) {
 	return (
 		<tr>
 			{(isSelected && onToggleAll) && (<th />)}
-			{columns.map((column, index) => (<th key={index}>{headerContent(column, sort)}</th>))}
+			{columns.map((column, index) => (
+				<th key={index}>
+					{column.label}
+					<div aria-hidden="true" className={theme.header}>
+						{headerContent(column, sort)}
```
frassinier: Because of `aria-hidden`, if columns have `Actions`, they will be unreachable from Screen Readers :(

packages/components/src/List/DisplayTable/DisplayTable.scss
```diff
@@ -2,6 +2,20 @@
...
+$tc-list-table-header-color: $black !default;
+$tc-list-table-header-height: 36px !default;
+
+.container {
+	padding-top: $tc-list-table-header-height;
+	position: relative;
+	display: flex;
+
+	> div {
+		width: 100%;
```
frassinier: Since its container has `display: flex` by default this `div`, as child, will be `flex: 1` so `width: 100%` is useless I guess.

packages/components/.storybook/head.html
jsomsanith: This is for the list stories.
capture d ecran de 2017-02-07 09-57-44

packages/components/src/List/DisplayTable/DisplayTable.scss
jsomsanith: No it's a flex row so the width is not 100% of it's container

packages/components/.storybook/head.html
frassinier: But what's about `min-height` instead?

packages/components/src/List/DisplayTable/DisplayTable.scss
frassinier: Ok so I prefer explicit `flex: 1;` here.

packages/components/.storybook/head.html
jmfrancois: I'm ok to keep a height so we will be able to tests scroll. The list may be constrained by it's container.
If you put height you will not be able to test scroll

packages/components/src/List/DisplayTable/DisplayTable.component.js
```diff
@@ -1,6 +1,6 @@
-import React, { PropTypes } from 'react';
+import React, {PropTypes} from 'react';
```
jmfrancois: lint please. space after { is required by object-curly-spacing lint rule

packages/components/.storybook/head.html
jsomsanith: Yes, this is only for the story, it demonstrate the scroll in some cases

packages/components/src/List/DisplayTable/DisplayTable.component.js
jsomsanith: oh, sorry for that, will check the linter. Thanks

jmfrancois:
:octocat: [Demo is available here](http://104.talend.surge.sh)

jmfrancois:
:octocat: [Demo is available here](http://104.talend.surge.sh)

packages/components/src/List/DisplayTable/DisplayTable.component.js
jsomsanith: I have issues passing some native props to the Button via Action.
Instead of going inot this dev in this PR, here what I propose : 
* if everything else is ok, approve this PR and let's merge it to be in the next release
* I continue to work on that in another PR right now

What do you think ?

frassinier: I agree

jmfrancois:
:octocat: [Demo is available here](http://104.talend.surge.sh)

PR #105, opened 2017-02-02
chore: move dev dependencies at top

jmfrancois:

If you start containers you will have the famous issue:

> Uncaught Error: Invariant Violation: addComponentAsRefTo(...): Only a ReactOwner can have refs. 

This is because lerna use npm link.

**What is the new behavior?**

No more error. but you have to launch all the things using lerna from top level.

https://github.com/lerna/lerna#common-devdependencies


jmfrancois:
Status: can't be done this way.

> Note that devDependencies providing "binary" executables that are used by npm scripts still need to be installed directly in each package where they're used.

We are using webapack (which is a binary)
And if webpack is in the sub package deps he can't find react in uper packages.

So at the moment I'm stuck with this issue

jmfrancois:
updated with some documentation https://github.com/Talend/ui/wiki/How-to-(monorepo)

PR #115, opened 2017-02-06
fix(containers/list): api call of action creator

jmfrancois:

The action creator on list title is called with only the id of the item

**What is the new behavior?**

it's called with event, data, context as all the other action creator.

- [x] Yes
- [ ] No

If you have used it you just have to update your action creator

```javascript
function myactioncreator(id) {}
```

```javascript
function myactioncreator(event, data, context) {
  const id = data.id;
}
```

AFAIK only TFD is using it

### DIFF
--- a/packages/containers/src/List/List.container.js
+++ b/packages/containers/src/List/List.container.js
@@ -51,6 +51,7 @@ class List extends React.Component {
 	static contextTypes = {
 		store: PropTypes.object,
 		registry: PropTypes.object,
+		router: PropTypes.object,
 	};
 
 	constructor(props) {
--- a/packages/containers/src/List/List.container.js
+++ b/packages/containers/src/List/List.container.js
@@ -97,7 +97,7 @@ class List extends React.Component {
 					api.action.getActionCreatorFunction(
 						this.context,
 						this.props.actions.title,
-					)(p.id),
+					)(event, p, this.context),
 				);
 			};
 		}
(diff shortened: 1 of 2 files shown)


jmfrancois:
:octocat: [Demo is available here](http://115.talend.surge.sh)

jmfrancois:
:octocat: [Demo is available here](http://115.talend.surge.sh)

packages/containers/src/List/List.container.js
```diff
@@ -97,7 +98,7 @@ class List extends React.Component {
 					api.action.getActionCreatorFunction(
 						this.context,
 						this.props.actions.title,
-					)(p.id),
+					)(e, p, this.context),
```
talend-glorieux: `e, p`?

acateland: why passing again the context when the action creator factory was already given it ?

jmfrancois: @acateland the first context is to retrieve the function that will be called.
context.registry.
the second is to call this function which doesnt' know the context.
We have a convention in CMF which said every action provider should accept
event, data, context
as arguments.

To remove this it would means modify the action creator by creating a wrapper around it to inject the context. I don't like this because you don't know the API of the action creator (this is just a convention)

jmfrancois:
:octocat: [Demo is available here](http://115.talend.surge.sh)

jmfrancois:
:octocat: [Demo is available here](http://115.talend.surge.sh)

PR #128, opened 2017-02-08
feat(Action): specify a form to a action

jare-talend:

- [X] The commit(s) message(s) follows our
- [X] Docs have been added / updated (for bug fixes / features)

React bootstrap allows to set a specify form to a button. This is useful if my button is outside of my form for a particular reason.

**What is the new behavior?**
We can set a specify form to a button. All the action (ex: submit) will be associated with the indicated form.

- [ ] Yes
- [X] No

**Other information**:
http://www.w3schools.com/tags/att_button_form.asp

### DIFF
--- a/packages/components/src/Actions/Action/Action.component.js
+++ b/packages/components/src/Actions/Action/Action.component.js
@@ -78,6 +78,7 @@ function Action(props) {
 		model,
 		onClick,
 		tooltipPlacement,
+		form,
 		...rest
 	} = props;
 
@@ -87,6 +88,10 @@ function Action(props) {
 		action: { label, ...rest },
 		model,
 	});
+
+	if (form) {
+		buttonProps.form = form;
+	}
 
 	const buttonContent = getContent(props);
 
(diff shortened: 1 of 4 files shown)


jmfrancois:
:octocat: [Demo is available here](http://128.talend.surge.sh)

talend-glorieux:
I don't see any form props on react-bootstrap button component.

talend-glorieux:
My bad found it ;)

packages/components/src/Actions/Action/Action.component.js
```diff
@@ -88,6 +89,10 @@ function Action(props) {
 		model,
 	});
 
+	if (form) {
+		buttonProps.form = form;
+	}
```
talend-glorieux: I don't think you need that if statement. The property could be passed directly as a prop to Button.

jmfrancois: buttonProps.form = form
if form is undefined will not adding anything. So for sure it's even better.

jare-talend:
The feature is not good, I am going to see that IE, even edge, don't support this attribute :(

packages/components/src/Actions/Action/Action.component.js
jare-talend: Exact, I made this to avoid to pass a "form" prop, undefined.

jare-talend:
I check on Edge 14, this feature isn't implemented.

We can abort this PR.

PR #131, opened 2017-02-08
fix(generator/component): use component did mount

jmfrancois:

If you are using the yo talend:react-component using full component option you will
have a component which use componentWillMount to initialize the state.

Because compoentWillMount will be deprecated in the future release of react it's encouraged to switch to didmount.
There is no real change for us because initState is dispatched to redux which is executed later.
So the rendering will happens twice if initState happens.

**What is the new behavior?**

yo talend:react-component with full option set the state init in componentDidMount

https://github.com/facebook/react/issues/7671


packages/containers/src/state.js
```diff
@@ -31,12 +31,21 @@ export function getStateProps(state, name, id) {
...
 
-export function stateWillMount(props) {
+export function initState(props) {
 	if (!props.state && props.initState) {
 		props.initState(props.initialState);
 	}
 }
 
+// BBB
+export function stateWillMount(props) {
```
acateland: :+1: :godmode:

jmfrancois:
:octocat: [Demo is available here](http://131.talend.surge.sh)

jmfrancois:
:octocat: [Demo is available here](http://131.talend.surge.sh)

PR #158, opened 2017-02-15
fix(forms): onClick error on submit action

talend-glorieux:

JavaScript was throwing an error in case of action without onClick.

**What is the new behavior?**

No more errors.

### DIFF
--- a/packages/forms/src/Form.js
+++ b/packages/forms/src/Form.js
@@ -88,7 +88,10 @@ class Form extends React.Component {
 	}
 
 	handleActionClick(onClick) {
-		return event => onClick(event, this.form.state);
+		if (onClick) {
+			return event => onClick(event, this.form.state);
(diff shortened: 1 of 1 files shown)


jmfrancois:
:octocat: [Demo is available here](http://158.talend.surge.sh)

packages/forms/src/Form.js
```diff
@@ -88,7 +88,10 @@ class Form extends React.Component {
 	}
 
 	handleActionClick(onClick) {
```
jmfrancois: That would not being better to have handleActionClick(onClick=noop) {}
with function noop() {} declared on top ?

jmfrancois: ok the transpilled version of this change proposal is not good at all ... let's keep this implemmentaiton

PR #194, opened 2017-02-24
feat(container): add a ConfirmDialog container

jlame:

No container for confirm dialog component

**What is the new behavior?**
A container was created to use confirm dialog

- [ ] Yes
- [x ] No

**Other information**:


packages/containers/.storybook/config.js
```diff
@@ -21,8 +21,28 @@ function objectView(event, data) {
 	}, data);
 }
 
+function objectView(event, data) {
+	return Object.assign({
```
talend-sromain: spread operator couldn't be nicer here ?

packages/containers/.storybook/config.js
```diff
@@ -21,8 +21,28 @@ function objectView(event, data) {
...
+		type: 'OBJECT_VIEW',
+	}, data);
+}
+
+function hideDialog(event, data) {
+	return Object.assign({
```
talend-sromain: spread operator couldn't be nicer here ?

packages/containers/.storybook/config.js
```diff
@@ -21,8 +21,28 @@ function objectView(event, data) {
...
+		type: 'HIDE_DIALOG',
+	}, data);
+}
+
+function confirmDialog(event, data) {
+	return Object.assign({
```
talend-sromain: spread operator couldn't be nicer here ?

packages/containers/src/ConfirmDialog/ConfirmDialog.container.js
```diff
@@ -0,0 +1,54 @@
+import React, { PropTypes } from 'react';
+import { Map } from 'immutable';
+import { ConfirmDialog as Component } from 'react-talend-components';
+
+import { getActionsProps } from '../actionAPI';
+import { statePropTypes, initState } from '../state';
+
+export const DEFAULT_STATE = new Map({
+	// size: 'small',
```
talend-sromain: This comments shouldn't be removed ?

packages/containers/src/actionAPI.js
```diff
@@ -10,6 +10,14 @@ export function getActionsProps(context, ids, model) {
 		tmpIds = [ids];
 	}
 
+	// in the case of ConfirmDialog, the component is loaded (and hidden) even when not
+	// used at the beginning,
+	// so ids (state.validateAction) is neither a string neither an array in this case
+	// as it takes a default state of the container which is an object.
+	if (!Array.isArray(tmpIds)) {
+		return ids;
+	}
```
jmfrancois: the comment is really not appropriate.
you have passed an object as id. this is not the way to call it.

packages/containers/src/ConfirmDialog/ConfirmDialog.container.js
```diff
@@ -0,0 +1,51 @@
...
+	validateAction: {
+		label: 'Ok',
+		bsStyle: 'primary',
+		onClick: () => {},
+	},
+	cancelAction: {
+		label: 'No',
+		onClick: () => {},
+	},
+});
```
jmfrancois: validateAction and cancelAction are not useful as state and do not reprensent the future action.
it must be a string, so they should not exists as state.

packages/containers/src/ConfirmDialog/ConfirmDialog.container.js
```diff
@@ -0,0 +1,51 @@
...
+	componentDidMount() {
+		initState(this.props);
+	}
+
+	render() {
+		const state = (this.props.state || DEFAULT_STATE).toJS();
+		// looking for the given action setting
+		// and adding onClick event using the retrieved  action creator
+		state.validateAction = getActionsProps(this.context, state.validateAction, state.model);
+		state.cancelAction = getActionsProps(this.context, state.cancelAction, state.model);
```
jmfrancois: the wrong call is here. I would tend to say validateAction and cancelAction should be string props

jlame: yeah in fact they are string props in the end when the host app change the state to update validateAction and cancelAction. The issue now if we remove the added code in ActionAPI and make the props as string is that we'll need to put existing action id in these props, else it will end in error when calling getActionInfo. 
The benefit of having validateAction as object in DEFAULT_STATE is that it won't return propTypes warning and we see clearly what the component needs as an input.

jmfrancois:
:octocat: [Demo is available here](http://194.talend.surge.sh)

PR #213, opened 2017-02-28
feat(containers): add connect function

jmfrancois:

To connect a pure component in CMF a lots of boilerplate is required.
You have to know all the things (how state is persisted into redux and so on).

If tomorrow we want to add a new feature, or refactor one in container, it will be a lots of work.

An other issue we have is the state is persisted through the life of the component. This behavior should be controlled and by default the component state should be dropped with the component unmount.

**What is the new behavior?**

You can connect your component into CMF and containers just by call connect.
The list and the sidepanel contianers has been udpated according to this change.

- [x] No


packages/containers/src/List/List.connect.js
```diff
@@ -20,33 +15,25 @@ function getItems(state, config) {
...
 			['toolbar', 'pagination'],
 			configureGetPagination(state, config)
 		);
 	}
-
 	return props;
 }
 
-export default connect(
+export default connect({
```
acateland: rename it cmfConnect please

packages/containers/src/connect.js
```diff
@@ -0,0 +1,169 @@
...
+ * - props.initState (call it un didMount)
+ *
+ * support for the following props
+ * - initialState (called by props.initState)
+ * - didMountActionCreator (id or array of id)
+ * - willUnMountActionCreator (id or array of id)
+ * - componentId (or will use uuid)
+ * @return {ReactComponent}
+ */
+export default function cmfConnect({
```
acateland: file should be renamed to cmfConnect.js

acateland:
at first look very nice, do you implement all parameter that can be fed to react-redux connect ?

packages/containers/src/cmfConnect.js
```diff
@@ -0,0 +1,165 @@
...
+}) {
+	let props = getStateAccessors(
+		dispatch,
+		getComponentName(WrappedComponent),
+		getComponentId(componentId, ownProps),
+		defaultState,
+	);
+	props.dispatch = dispatch;
+	if (mapDispatchToProps) {
+		props = Object.assign(props, mapDispatchToProps(dispatch, ownProps));
```
jsomsanith: (see comments on `getStateToProps`) Same here.
```
// here cmfProps is your props variable

const userProps = mapDispatchToProps(dispatch, ownProps, cmfProps);
cmfProps = { ...cmfProps, ...userProps };
```

This would be an entry point to customise CMF state accessors functions (not encouraged but possible).
Ex : adding a log before calling the cmf function.

packages/containers/src/cmfConnect.js
```diff
@@ -0,0 +1,165 @@
...
+	}
+	props.dispatchActionCreator = (actionId, event, data, context) => {
```
jsomsanith: This should be before the user's `mapDispatchToProps `.
Internal then user's, even more if the previous comment is relevant.

packages/containers/src/List/index.js
```diff
@@ -1,5 +1,5 @@
 import List from './List.connect';
 
-List.displayName = 'CMF(List)';
+//List.displayName = 'CMF(List)';
```
jsomsanith: Why is it commented ?

packages/containers/src/List/List.connect.js
```diff
@@ -20,33 +15,25 @@ function getItems(state, config) {
...
 export function mapStateToProps(state, ownProps) {
-	const { name, id } = getContainerInfo(ownProps);
-	const props = getStateProps(state, name, id);
+	const props = {};
 	const config = {
 		collectionId: ownProps.collectionId,
 		items: ownProps.items,
 	};
-
 	props.items = getItems(state, config);
```
jsomsanith: Since CMF manages collections, shouldn't the CMF connector resolve the collection items ?
* if `ownProps.collectionId`, CMF resolve it and inject the items (with a defined name, `collectionData` ?)
* in user's `mapStateToProps(state, ownProps, cmfProps)` (see comments in cmf connector), he can add his logics on `collectionData`, like filter

```
props.items = getItems(state, cmfProps.collectionData);
```

packages/containers/src/cmfConnect.js
```diff
@@ -0,0 +1,165 @@
...
+			static contextTypes = {
+				store: PropTypes.object,
+				registry: PropTypes.object,
+				router: PropTypes.object,
+			};
+
+			componentDidMount() {
+				initState(this.props);
+				if (this.props.didMountActionCreator) {
+					this.props.dispatch(
```
jsomsanith: Use `props.dispatchActionCreator` injected from `getDispatchToProps()`

packages/containers/src/cmfConnect.js
```diff
@@ -0,0 +1,165 @@
...
+				...statePropTypes
+			);
+			static contextTypes = {
+				store: PropTypes.object,
+				registry: PropTypes.object,
+				router: PropTypes.object,
+			};
+
+			componentDidMount() {
+				initState(this.props);
```
jsomsanith: Is it interesting to test if the component state is already initialised to avoid too many non useful dispatch ?

packages/containers/src/cmfConnect.js
```diff
@@ -0,0 +1,165 @@
...
+						api.action.getActionCreatorFunction(
+							this.context, this.props.didMountActionCreator
+						)(null, this.props, this.context)
+					);
+				}
+			}
+
+			componentWillUnmount() {
+				if (this.props.willUnmountActionCreator) {
+					this.props.dispatch(
```
jsomsanith: Use props.dispatchActionCreator injected from getDispatchToProps()

packages/containers/src/cmfConnect.js
```diff
@@ -0,0 +1,165 @@
...
+						api.action.getActionCreatorFunction(
+							this.context, this.props.didMountActionCreator
+						)(null, this.props, this.context)
+					);
+				}
+				if (!keepComponentState) {
+					this.props.updateState();
```
jsomsanith: This will merge an empty state into the current state. I think you meant to remove component state instead.

packages/containers/src/List/List.connect.js
jmfrancois: Done with cmfProps.getCollection

packages/containers/src/cmfConnect.js
jmfrancois: let's keep this because it's working in a first version.
I can't remember why but this is the way we delete the component state ... sorry 🙏

packages/cmf/src/componentState.js
```diff
@@ -0,0 +1,61 @@
...
+}
+
+export function initState(props) {
+	if (!props.state && props.initState) {
+		props.initState(props.initialState);
+	}
+}
+
+// BBB
+export function stateWillMount(props) {
```
acateland: we should make it a console.error to make the message more visible and issue a depreciation date prévision.

packages/cmf/src/cmfConnect.js
```diff
@@ -0,0 +1,177 @@
...
+ * - props.
+ *
+ * support for the following props
+ * - initialState (called by props.initState)
+ * - didMountActionCreator (id or array of id)
+ * - willUnMountActionCreator (id or array of id)
+ * - componentId (or will use uuid)
+ * @return {ReactComponent}
+ */
+export default function cmfConnect({
```
acateland: inline documentation should be updated

jmfrancois:
:octocat: [Demo is available here](http://213.talend.surge.sh)

jmfrancois:
:octocat: [Demo is available here](http://213.talend.surge.sh)

PR #224, opened 2017-03-03
feat(http): Dispatch default http error only if onError isn't defined

talend-glorieux:

When handling errors manually with the onError you'll get two dispatch from the http middleware.

**What is the new behavior?**
If an onError is defined the default HTTP error won't be dispatched.


jmfrancois:
please update your branch

jmfrancois:
:octocat: [Demo is available here](http://224.talend.surge.sh)

jare-talend:
Hi, i'm just seeing this PR.
We use the 2 handler on the errors. The first into our application to make something when one error is triggered and the second HTTP_ERRORS which notifies the errors.

PR #234, opened 2017-03-06
fix(components) : fix space between header & table on firefox

talend-sromain:

There is a space of 6 px between header & table content

**What is the new behavior?**

There is no space

- [ ] Yes
- [x] No

**Other information**:


talend-glorieux:
Where is the change? Is it the html element change that fixes the issue? 
https://github.com/Talend/ui/pull/234/files#diff-a24714ed263cd5d3446882f7bb3fa315R154

talend-sromain:
it's not the html element ny himself, actually, it's the display of this html element.
On FF with a span, i got the element at 6px height, when i switch it to display:block, the 'real' height given by css is applied.
So instead of setting a display block on a span, i think it's better to set a div displayed block by default

jmfrancois:
:octocat: [Demo is available here](http://234.talend.surge.sh)

jmfrancois:
:octocat: [Demo is available here](http://234.talend.surge.sh)

jmfrancois:
:octocat: [Demo is available here](http://234.talend.surge.sh)

output/components.eslint.txt
```diff
@@ -16,6 +16,10 @@ Scoping to packages that match 'react-talend-components'
   2:1  error  '@kadira/react-storybook-addon-info' should be listed in the project's dependencies, not devDependencies  import/no-extraneous-dependencies
   7:2  error  Unexpected require()                                                                                      global-require
 
+/home/travis/build/Talend/ui/packages/components/src/List/DisplayTable/DisplayTable.component.js
```
jlame: some lint to fix

jmfrancois:
:octocat: [Demo is available here](http://234.talend.surge.sh)

jmfrancois:
:octocat: [Demo is available here](http://234.talend.surge.sh)

PR #246, opened 2017-03-07
fix(forms): indicate that the field is mandatory

jare-talend:

- [X] The commit(s) message(s) follows our

When we indacate that a field is mandatory, * is missing on the label

**What is the new behavior?**
Add * on the label if the field is required

- [ ] Yes
- [ ] No

**Other information**:


jmfrancois:
:octocat: [Demo is available here](http://246.talend.surge.sh)

jmfrancois:
:octocat: [Demo is available here](http://246.talend.surge.sh)

output/forms.eslint.txt
```diff
@@ -124,7 +128,7 @@ The react/require-extension rule is deprecated. Please use the import/extensions
    30:18  error  'formData' is missing in props validation  react/prop-types
   162:14  error  'onChange' is missing in props validation  react/prop-types
 
-✖ 110 problems (110 errors, 0 warnings)
+✖ 112 problems (112 errors, 0 warnings)
```
jlame: 2 lines lint to fix

packages/forms/src/templates/FieldTemplate.js
```diff
@@ -38,7 +39,7 @@ const CustomFieldTemplate = ({
 			<Label className="form-label" label={label} required={required} id={id} />}
 			{children}
 			{!hasLabelBefore && !isToggle && displayLabel &&
-			<Label label={label} required={required} id={id} className="control-label" />}
+			<Label label={label} required={required || schema.required} id={id} className="control-label" />}
```
jlame: not sure to understand the difference between 
`"fullDatalist": {
       "required": true,
       "$ref": "#/definitions/datalist"
     }`

and `"required": [
      "select1"
    ]`

jlame: Discussed with @jare-talend, the array notation is used in schema-form v4. There's an issue on required field validation, this PR is a workaround for this issue but the required in property form was used in v3.
We should wait for the validation issue to be fixed in v4, so that we don't make unnecessary change in backend.

jlame: https://github.com/mozilla-services/react-jsonschema-form/issues/384

jmfrancois:
:octocat: [Demo is available here](http://246.talend.surge.sh)

jmfrancois:
:octocat: [Demo is available here](http://246.talend.surge.sh)

PR #261, opened 2017-03-09
fix(containers/homelistview): connect to cmf

jmfrancois:

The homelistview is connected by hand so for example the action creator on did mount doesn t get the usual event, data, context args.

**What is the new behavior?**

the homelistview component is connected using cmfconnect.
because it has no handler we don't need the container anymore.


jmfrancois:
:octocat: [Demo is available here](http://261.talend.surge.sh)

jmfrancois:
:octocat: [Demo is available here](http://261.talend.surge.sh)

PR #272, opened 2017-03-13
tests: add coverage step to travis and store results in .static

jmfrancois:
so they will be on surge

We have no way to consult the coverage.
By the way the current coverage include lib folders and snapshots ...

**What is the new behavior?**

The coverage are parts of the tests and the output is put inside .static
The lib folder and snapshots are excluded


jmfrancois:
:octocat: [Demo is available here](http://272.talend.surge.sh)

jmfrancois:
:octocat: [Demo is available here](http://272.talend.surge.sh)

PR #293, opened 2017-03-16
fix(Components): list table double vertical scroll

jsomsanith:

On List Table display with actions, there was a double scroll.
The problem was that the actions are in `position: absolute` aligned on the right. But when the List is big enough to trigger a scroll, the actions remain on the right of the **initial position of the tr**.
So on the last tr, the actions was under the viewport, that enable the second scrollbar.
capture d ecran de 2017-03-17 10-31-46

Hover on the last row : the actions are not on the row --> impossible to click
capture d ecran de 2017-03-17 10-27-54

There is an horizontal scroll du to the long text, but it doesn't appear :/

**What is the new behavior?**
This remains a quick fix as the guideline has been updated with a more complicated actions bar.
The implementation of this version of guideline can take some time, but this bug remains and block the actions click when we scroll.

capture d ecran de 2017-03-17 10-30-36

Hover on the last row
capture d ecran de 2017-03-17 10-30-49

This PR is only to remove this double scroll and actions display out of the tr. They are now back on the right of the title

- [ ] Yes
- [x] No

### DIFF
--- a/packages/components/src/List/DisplayTable/DisplayTable.component.js
+++ b/packages/components/src/List/DisplayTable/DisplayTable.component.js
@@ -63,13 +63,6 @@ function RowRenderer(props) {
 				const displayActions =
 					isTitle &&
 					(!displayModeKey || !item[displayModeKey] || item[displayModeKey] === 'text');
-				const actions = displayActions ?
-					(<Actions
-						actions={item.actions || []}
-						hideLabel
-						link
-					/>) :
-					null;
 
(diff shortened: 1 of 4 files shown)


jmfrancois:
:octocat: [Demo is available here](http://293.talend.surge.sh)

jmfrancois:
:octocat: [Demo is available here](http://293.talend.surge.sh)

PR #294, opened 2017-03-17
feat(cmf/connect): dispatchActionCreator can be called without context

jmfrancois:
it will pass the full context (registry, router, store)

if you want to dispatch an action creator in your connected componetn you must push a context with minimal context to have the registry (to find the actioncretaor)

And we already have a full context in CMFContainer so ...

**What is the new behavior?**

the context is optional and you should not passed it except if you know what you are doing (for example you want to add a new context

- [x] No


jmfrancois:
:octocat: [Demo is available here](http://294.talend.surge.sh)

jmfrancois:
:octocat: [Demo is available here](http://294.talend.surge.sh)

jsomsanith:
Can yo uexplain why do we let the user pass another context instead of always passing the `this.context` ?

jmfrancois:
:octocat: [Demo is available here](http://294.talend.surge.sh)

packages/cmf/src/cmfConnect.js
```diff
@@ -176,5 +185,7 @@ export default function cmfConnect({
 			mergeProps,
 			{ ...rest },
 		)(hoistStatics(CMFContainer, WrappedComponent));
+		Connected.CMFContainer = CMFContainer;
```
jsomsanith: Why is this needed ?

packages/cmf/src/cmfConnect.js
```diff
@@ -121,43 +121,52 @@ export default function cmfConnect({
...
-						this.context
 					);
 				}
 				if (!keepComponentState) {
 					this.props.deleteState();
 				}
 			}
 
+			dispatchActionCreator(actionCreatorId, event, data, context) {
+				const myContext = context || this.context;
```
jsomsanith: As discussed together, this should merge the cmf context with the user context to not break cmf internal process which need the registry in context.

packages/cmf/src/cmfConnect.js
jmfrancois: To being able to tests this class ... it's a bit like a singleton you can't access because there is no link from outside.

jmfrancois:
:octocat: [Demo is available here](http://294.talend.surge.sh)

jmfrancois:
:octocat: [Demo is available here](http://294.talend.surge.sh)

jmfrancois:
:octocat: [Demo is available here](http://294.talend.surge.sh)

PR #299, opened 2017-03-20
feat(components/layout): Add footer option

sdwvit:

no footer option

**What is the new behavior?**
footer option

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/packages/components/src/Layout/Layout.component.js
+++ b/packages/components/src/Layout/Layout.component.js
@@ -25,7 +25,7 @@ body > div {
  * @example
  <Layout mode="TwoColumns" one={one} two={two}></Layout>
  */
-function Layout({ header, mode, drawers, children, ...rest }) {
+function Layout({ header, footer, mode, drawers, children, ...rest }) {
 	const appCSS = classnames(
 		'tc-layout',
 		theme.layout,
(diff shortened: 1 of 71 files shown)


jmfrancois:
:octocat: [Demo is available here](http://299.talend.surge.sh)

jmfrancois:
:octocat: [Demo is available here](http://299.talend.surge.sh)

packages/components/src/Layout/Layout.component.js
```diff
@@ -55,12 +59,16 @@ function Layout({ header, mode, drawers, children, ...rest }) {
 					{children}
 				</Component>
 			) : null}
+			{footer && <div className={footerCSS}>
```
frassinier: You may use `<footer role="contentinfo" />`

packages/components/src/Layout/Layout.scss
```diff
@@ -6,6 +6,7 @@
 /// Layout height of the header
 /// @type Number (Unit)
 $tc-layout-header-height: $navbar-height;
+$tc-layout-footer-height: 2 * $padding-small + $font-size-small;
```
frassinier: This is dangereous since if we change padding or font-size to switch to `em`, `rem` unit, this will fails.
```scss
footer {
  flex-shrink: 0;
  flex-grow: 0;
} 
```
It could be enough?

packages/components/src/Layout/Layout.component.js
sdwvit: oh that's a good idea
дякую

packages/components/src/Layout/Layout.scss
sdwvit: yes, fixed

jmfrancois:
:octocat: [Demo is available here](http://299.talend.surge.sh)

sdwvit:
@jmfrancois @frassinier pls update review

PR #305, opened 2017-03-21
fix(components/List): set max width

jmfrancois:

screen shot 2017-03-21 at 16 11 16
screen shot 2017-03-21 at 16 11 23

You need to set the max width in your project (it was in the story head.html)

**What is the new behavior?**

no more custom set in the head.html, the list is setting the max-width it self.

screen shot 2017-03-21 at 16 11 49

- [x] No

### DIFF
--- a/packages/components/.storybook/head.html
+++ b/packages/components/.storybook/head.html
@@ -13,11 +13,6 @@
         background: #ff69b4;
     }
 
-    .tc-list-fixed-name-column .tc-list-display-table tr > td:first-child + td {
-        width: 400px;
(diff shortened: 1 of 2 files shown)


jmfrancois:
:octocat: [Demo is available here](http://305.talend.surge.sh)

frassinier:
Another solution could be to set `flex-basis: 100%` and add specific class for each columns.
Indeed, each project knows the `max-width` for each column so we can easily delegate to final apps.

frassinier:
Already fixed by #378?

PR #270, opened 2017-03-13
feat(theme): checkbox and radio styles on adjacent sibling element

frassinier:
**Please check if the PR fulfills these requirements**

- [x] The commit(s) message(s) follows our
  [guidelines](https://github.com/talend/tools/blob/master/tools-root-github/CONTRIBUTING.md#commit-message-format)
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**What is the current behavior?** (You can also link to an open issue here)
Firefox could not display same checkboxes and radio buttons as Chrome because it does not support pseudo-elements for `input`

**What is the new behavior?**
A workaround can be found by using `span` to wrap their labels text 

**Does this PR introduce a breaking change?**

- [ ] Yes
- [x] No

**Other information**:
https://github.com/thomaspark/bootswatch/issues/559

PR #306, opened 2017-03-21
fix(components): use Bootstrap markup for list checkboxes

frassinier:

Checkboxes are in the list without a form context

**What is the new behavior?**
Align checkboxes markup to Bootstrap one in order to apply our theme

- [ ] Yes
- [x] No

**Other information**:
http://getbootstrap.com/css/#checkboxes-and-radios


jmfrancois:
:octocat: [Demo is available here](http://306.talend.surge.sh)

packages/components/src/List/DisplayTile/DisplayTile.component.js
```diff
@@ -35,12 +35,17 @@ function Tile({ id, columns, item, itemProps, titleProps }) {
...
-		/>) :
+		(<div className="checkbox">
+			<label htmlFor={id && `${id}-check`}>
+				<input
+					id={id && `${id}-check`}
+					type="checkbox"
+					onChange={e => onToggle(e, item)}
+					checked={isSelected(item)}
+				/>
+				<span><span className="sr-only">Select {item.name}</span></span>
```
sdwvit: span inside of span? why?

frassinier: Because first one is for style with its pseudo-elements (in order to work with Firefox #270) and the second one is for hidden label.

sdwvit: Ok, I see
I usually use <i> for such things :) maybe span inside of span is semantically better

jmfrancois:
:octocat: [Demo is available here](http://306.talend.surge.sh)

packages/components/src/List/DisplayTile/DisplayTile.component.js
frassinier: Yes `<i/>` stands for italic (and is often used for icons) `<span/>` is just an inline wrapper so I prefer its use in this case :)

Issue #90, opened 2017-01-31
[CMF] React router v4 integration


acateland:
React router v4 just hit beta stage and its api is set in stone
https://twitter.com/mjackson/status/826114904914489344

Since V4 is a different beast than router V2/3 we should talk on how we will itegrate it into cmf and our apps.

Here will be collected documentation about our integration plan, links to various ressource etc.

FAQ
Bundle size ?
React Router V4 is lighter than V3/2.
What about the api change ?
React Router have completely different api for route handling and declaration wich is LOT simpler and more powerfull.
V2/3 version are dropped altogether ?
No V3 will be kept and maintened for sometime.

Please contribute in following comments.

// doc entry point
https://github.com/ReactTraining/react-router/tree/v4
// migration guide incomplete
Https://gist.github.com/mjackson/33f51e9d6b9ea18b467c613fbbcf50e1
// codepen playground
https://t.co/EUZkuDfowr

frassinier:
Please, could you produce a **TL;DR** for everyone? :D  
What's your opinion on that upgrade? Should we do it? Now?

jsomsanith:
#324

jmfrancois:
The PR is ready but after some thought react-router will not give us anything that we dont have already.
React-Router add an external component that trigger react rendering.
React-Router-Redux sync history  into redux

I prefer keep sync history into redux + API so we keep only the redux->react notification and source of rendering trigger.

PR #324, opened 2017-03-24
feat(cmf): react-router v4 integration

jmfrancois:

- [ ] Integration has been done in one project (add PR link to it)
- [x] example app provided
- [x] versions script updated
- [ ] wrapchildren patch has been removed

we are using react-router v3.X
#90 

**What is the new behavior?**

We have an app under examples/cmf-app, you can install and start (this is the output from yo talend:react-cmf)

react-router v4.0.0 integration.
The concept has completly changed, so we need to re-think our integration.
Now routing is done using <Router>, <Route>, ... components.
Please check the docs: https://github.com/ReactTraining/react-router

The idea behind this new integration is to keep routes in configuration.

Because it's a WIP we will answer to this question when it will be finished.
- [x] Yes
- [ ] No

**Other information**:
Because CMF support custom routing you already use react-router v4 but you can't use customization through the registry, so this is why this PR is still important.

**Status**
screen shot 2017-03-24 at 21 48 13


packages/cmf/src/UIRouter.js
```diff
@@ -1,37 +1,125 @@
...
+  }
  * @param  {object} props   The waited props (history and routes)
  * @param  {object} context The react context with the registry
  * @return {object} ReactElement
  */
-const UIRouter = (props, context) => {
-	const routes = api.route.getRoutesFromSettings(context, props.routes);
+function CMFRouter(props) {
+	const routes = props.routes;
+	const history = props.history || createHashHistory();
```
jsomsanith: This should not be used explicitly. It is used by App.js which inject the history. We can drop that `createHashHistory()`

packages/cmf/src/UIRouter.js
```diff
@@ -1,37 +1,125 @@
...
-
+import { createHashHistory } from 'history';
 import api from './api';
 
+function CMFRoute(props, context) {
+	let Component = api.route.getComponentFromRegistry(context, props.component);
+	if (props.view && !Component.CMFContainer) {
+		Component = api.route.connectView(context, Component, props.view);
+	}
+	if (props.path === '/') {
```
jsomsanith: Since this is only for the root configuration, it would be better to move this directly in `CMFRouter`

Another thing that we mentionned together was to use the `<Redirect>` component from react-router instead of our custom one for the `/` route. This would introduce a breaking change, since settings.json would change for this root route

packages/cmf/src/App.js
```diff
@@ -3,8 +3,8 @@
  */
 import React from 'react';
 import { Provider } from 'react-redux';
+import { createHashHistory } from 'history';
```
jsomsanith: Done

jmfrancois:
need to be integrated in one project without too much work
The first big problem to solve is the drawer on home page which rely on a `wrapchildren`patch

acateland:
Revival

PR #335, opened 2017-03-30
fix(components): DisplayTable with invisible headers

ncomont:
fix(components/DisplayTable): Allow DisplayTable to render columns with invisble header (screen reader compatible)

DisplayTable is not able to render a column with an invisible header. For example, an "Actions" column is set with an empty label, which is not screen reader compliant.

**What is the new behavior?**
A column accepts a "hideHeader" property. When setted to true, a single <span> will be rendered with the "sr-only" class.

- [ ] Yes
- [x] No

**Other information**:
Can be usefull for table with sort capabilities. Actually, it is possible to focus an empty header by focusing the previous header and pressing "tab".


packages/components/src/List/DisplayTable/DisplayTable.component.js
```diff
@@ -176,7 +178,7 @@ export function ColumnHeader({ index, column, sort }) {
 	return (
 		<th key={index}>
 			{header}
-			<div aria-hidden="true">{column.label}</div>
+			{ !column.hideHeader ? (<div aria-hidden="true">{column.label}</div>) : null }
```
frassinier: You can write `{ column.hideHeader || (<div aria-hidden="true">{column.label}</div>)}`

jmfrancois:
:octocat: [Demo is available here](http://335.talend.surge.sh)

packages/components/src/List/DisplayTable/DisplayTable.snapshot.test.js
```diff
@@ -183,6 +183,22 @@ describe('DisplayTable', () => {
 		expect(wrapper).toMatchSnapshot();
 	});
 
+	it('should render screen reader compatible column actions with invisible header', () => {
+		// given
+		columns.splice(2, 0, { key: 'columnActions', label: 'Actions', hideHeader: true });
```
frassinier: Can you please add a simple comment in order to make it more readable :ok_hand:

jmfrancois:
:octocat: [Demo is available here](http://335.talend.surge.sh)

packages/components/src/List/DisplayTable/DisplayTable.component.js
```diff
@@ -149,7 +149,9 @@ function getNextDirection(isCurrentSortField, currentSort) {
 export function ColumnHeader({ index, column, sort }) {
 	let header;
 
-	if (sort) {
+	if (column.hideHeader) {
```
jsomsanith: Does it make sense to test column.label instead of introducing this property ?
I'm thinking about the use cases and I don't think we would want to pass a label to hide it, or to pass an empty label to display it. So if we base the 2 things on the same data instead of having 2, it will avoid those 2 use cases (user props desynchro)
What do you think ?

jmfrancois:
:octocat: [Demo is available here](http://335.talend.surge.sh)

jmfrancois:
:octocat: [Demo is available here](http://335.talend.surge.sh)

packages/components/src/List/DisplayTable/DisplayTable.component.js
ncomont: The goal of this property is to keep the screen reader capability for the columns with no header. For example, an "Actions" columns as it is in TDP's datasets listing page should have an hidden header but a screen reader must be able to read the content.

In addition, passing an empty label in a table which have sorting capabilities will render an empty button which will be focusable by pressing the 'tab' key and create some issues as described here : https://jira.talendforge.org/browse/TDP-3549

jmfrancois:
:octocat: [Demo is available here](http://335.talend.surge.sh)

jlame:
What about adding an example in story book ?

ncomont:
Thank you @jlame 
I just updated the "Table with column actions" story by adding the "hideLabel" property to its actions column.

jmfrancois:
:octocat: [Demo is available here](http://335.talend.surge.sh)

jmfrancois:
:octocat: [Demo is available here](http://335.talend.surge.sh)

PR #341, opened 2017-03-31
fix(icon/TFD-838): rename DBInput into db-input

jmfrancois:

DBInput is not lower case as all the other icons.

**What is the new behavior?**

db-input


jmfrancois:
:octocat: [Demo is available here](http://341.talend.surge.sh)

PR #347, opened 2017-04-03
fix(forms): handleActionOnClick return noop

jmfrancois:

If a form is defined action without onclick it throw exception.
But for submit action it s a non sens to define onclick because you have defined onSubmit.

**What is the new behavior?**

no need for that


jmfrancois:
:octocat: [Demo is available here](http://347.talend.surge.sh)

jmfrancois:
:octocat: [Demo is available here](http://347.talend.surge.sh)

jmfrancois:
:octocat: [Demo is available here](http://347.talend.surge.sh)

PR #381, opened 2017-04-12
fix(forms): add DisplayName=TalendForm

jmfrancois:

Form has no displayName so in test we have <Unkown />.

**What is the new behavior?**

Form.displayName = 'TalendForm';


jmfrancois:
:octocat: [Demo is available here](http://381.talend.surge.sh)

packages/forms/src/Form.js
```diff
@@ -191,4 +191,6 @@ Form.defaultProps = {
 	buttonBlockClass: 'form-actions',
 };
 
+Form.displayName = 'TalendForm';
```
frassinier: What's the problem with default value `Form`?
https://facebook.github.io/react/docs/react-component.html#displayname

jmfrancois: that default value happens only in dev. in prod the function name is minified.
And in snapshot without displayName I have <unknown />

https://github.com/Talend/data-streams/commit/d2554159919335e29312bf7b2b5e765e7c15f117#diff-8ac575987465edbb6a6b7810963bc2c8R24

jmfrancois:
:octocat: [Demo is available here](http://381.talend.surge.sh)

jmfrancois:
:octocat: [Demo is available here](http://381.talend.surge.sh)

PR #393, opened 2017-04-19
fix(form/enumeration): delete a value when search don't reinit to default state

talend-sromain:

when connected to handler, delete an item reset the list to default state & make some ux issue ( as the searched items are not refresh but the search seems reset )

**What is the new behavior?**

we don't go back to default state when we are in search state

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/output/cmf.eslint.txt
+++ b/output/cmf.eslint.txt
@@ -1,7 +1,7 @@
 Lerna v2.0.0-beta.36
 Scoping to packages that match 'react-cmf'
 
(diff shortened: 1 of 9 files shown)


jmfrancois:
:octocat: [Demo is available here](http://393.talend.surge.sh)

jmfrancois:
:octocat: [Demo is available here](http://393.talend.surge.sh)

PR #394, opened 2017-04-20
fix(components/List): Can't render zeros

talend-glorieux:

If you try and render a list with an item set to 0. The item won't be
displayed.

**What is the new behavior?**
You can now display zeros properly.

### DIFF
--- a/output/components.eslint.txt
+++ b/output/components.eslint.txt
@@ -18,9 +18,6 @@ Scoping to packages that match 'react-talend-components'
   2:1  error  '@kadira/react-storybook-addon-info' should be listed in the project's dependencies, not devDependencies  import/no-extraneous-dependencies
   7:2  error  Unexpected require()                                                                                      global-require
 
-/home/travis/build/Talend/ui/packages/components/src/List/DisplayTable/DisplayTable.component.js
-  26:31  error  Expected '!==' and instead saw '!='  eqeqeq
-
 /home/travis/build/Talend/ui/packages/components/src/ObjectViewer/__snapshots__/config.js
   1:1  error  '@kadira/storybook' should be listed in the project's dependencies, not devDependencies                   import/no-extraneous-dependencies
   2:1  error  '@kadira/react-storybook-addon-info' should be listed in the project's dependencies, not devDependencies  import/no-extraneous-dependencies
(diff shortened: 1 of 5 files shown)


jmfrancois:
:octocat: [Demo is available here](http://394.talend.surge.sh)

packages/components/src/List/DisplayTable/DisplayTable.component.js
```diff
@@ -23,7 +23,7 @@ export function CellContent({ isTitle, item, column, titleProps, id }) {
 			hideLabel
 		/>);
 	}
-	const str = item[column.key] ? item[column.key].toString() : '';
+	const str = item[column.key] !== undefined ? item[column.key].toString() : '';
```
ncomont: `!=` instead of `!==` will return `false` whenever the value is `null` or `undefined` ; which is more appropriate in this case.

jmfrancois:
:octocat: [Demo is available here](http://394.talend.surge.sh)

jmfrancois:
:octocat: [Demo is available here](http://394.talend.surge.sh)

output/components.eslint.txt
```diff
@@ -18,6 +18,9 @@ Scoping to packages that match 'react-talend-components'
   2:1  error  '@kadira/react-storybook-addon-info' should be listed in the project's dependencies, not devDependencies  import/no-extraneous-dependencies
   7:2  error  Unexpected require()                                                                                      global-require
 
+/home/travis/build/Talend/ui/packages/components/src/List/DisplayTable/DisplayTable.component.js
+  26:31  error  Expected '!==' and instead saw '!='  eqeqeq
```
frassinier: Could you consider write more verbose stuff like:
```javascript
const index = item[column.key];
const str = `${(index >= 0 && index) || ''}`;
```
Dunno if it makes sense

talend-glorieux: Yep. I'm also in favor of removing the ternary.
I'll see what I can do. Maybe go back to a good old if statement ;)

jmfrancois:
:octocat: [Demo is available here](http://394.talend.surge.sh)

packages/components/src/List/DisplayTable/DisplayTable.component.js
```diff
@@ -8,6 +8,13 @@ import DisplayPropTypes from '../Display/Display.propTypes';
 
 import theme from './DisplayTable.scss';
 
+function getItemString(item) {
+	if (item === undefined || item === null) {
```
frassinier: What's about `NaN` :sleepy:

PR #413, opened 2017-04-26
fix(forms): manage a list of value with coma separator

jlame:

**What is the new behavior?**
Before we were just using the first attribute of the array as a value, now we send the whole array.

- [x] Yes
- [ ] No

**Other information**:

### DIFF
--- a/output/forms.eslint.txt
+++ b/output/forms.eslint.txt
@@ -130,7 +130,14 @@ The react/require-extension rule is deprecated. Please use the import/extensions
   151:5   error  Value must be omitted for boolean attributes  react/jsx-boolean-value
   172:5   error  Value must be omitted for boolean attributes  react/jsx-boolean-value
 
-✖ 112 problems (112 errors, 0 warnings)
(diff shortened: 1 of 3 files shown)


jmfrancois:
:octocat: [Demo is available here](http://413.talend.surge.sh)

jmfrancois:
:octocat: [Demo is available here](http://413.talend.surge.sh)

jmfrancois:
:octocat: [Demo is available here](http://413.talend.surge.sh)

jlame:
In fact today, there's no UX guide or specific way of managing multiple values on one line. So while waiting for the moment we implement it, it was decided that this would be the current behaviour. It could be possible to remove all of that logic and let the host widget manage this but that would make it impossible to enter multiple value in the "disconnected" mode.

jmfrancois:
:octocat: [Demo is available here](http://413.talend.surge.sh)

jmfrancois:
:octocat: [Demo is available here](http://413.talend.surge.sh)

PR #422, opened 2017-05-04
feat(cmf): add view props support in cmfConnect

jmfrancois:

The view components are connected by the router to support the injection of view.
So only view components support this behavior.

**What is the new behavior?**

Any components connected using cmfConnect support the injection of a props.view which refer to the settings.
So you can re-use settings inside any of your components, not only views.

Bonus: The CMF router detect if the component is connected using cmfConnect.
If yes: do nothing
Else: warn you should connect it using + use the old connection for backward compatibility.

- [x] No


jmfrancois:
:octocat: [Demo is available here](http://422.talend.surge.sh)

jmfrancois:
:octocat: [Demo is available here](http://422.talend.surge.sh)

jmfrancois:
:octocat: [Demo is available here](http://422.talend.surge.sh)

packages/cmf/src/cmfConnect.js
```diff
@@ -41,12 +42,25 @@ export function getStateToProps({
...
 	};
 
+	let viewProps = {};
+	if (ownProps.view) {
+		viewProps = Object.assign(
+			{},
+			state.cmf.settings.views[ownProps.view],
+		);
+		viewProps = attachRef(state, viewProps);
+		Object.keys(viewProps).forEach(
```
jsomsanith: Can you externalize this in an `attachRefs()` function ?
As discussed together, a better solution would be to resolve the dependency recursively only once, before the settings are injected. But this can have an impact and can be tricky. Let's keep that in mind when we have the need.

jmfrancois:
:octocat: [Demo is available here](http://422.talend.surge.sh)

jmfrancois:
:octocat: [Demo is available here](http://422.talend.surge.sh)

jmfrancois:
:octocat: [Demo is available here](http://422.talend.surge.sh)

jmfrancois:
:octocat: [Demo is available here](http://422.talend.surge.sh)

jmfrancois:
:octocat: [Demo is available here](http://422.talend.surge.sh)

PR #430, opened 2017-05-09
fix(cmf): attachRef behavior

acateland:
When attachRef is called with obj as a primitive not being an object, for example a string, said string was convert to an object before being returned.

triggering following problems, if a action creator id was present at the root of the configuration given to attachRefs the id was transformed into an object, being not resolvable by the action resolver.

detect if obj given to attachRef is indeed an object and not another primitive, return passed parameter to caller.

### DIFF
--- a/packages/cmf/src/settings.js
+++ b/packages/cmf/src/settings.js
@@ -10,21 +10,21 @@ import invariant from 'invariant';
  * if an object try to find _ref property and resolve it
  */
 export function attachRef(state, obj) {
-	if (obj !== null && typeof obj === 'object') {
-		let props = Object.assign({}, obj);
-		if (props._ref) {
-			const ref = state.cmf.settings.ref[props._ref];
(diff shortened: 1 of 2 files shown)


packages/cmf/src/settings.js
```diff
@@ -6,19 +6,25 @@
...
+/**
+ * if an object try to find _ref property and resolve it
+ */
 export function attachRef(state, obj) {
-	let props = Object.assign({}, obj);
-	if (props._ref) {
-		const ref = state.cmf.settings.ref[props._ref];
-		invariant(ref, `CMF/Settings: Reference '${props._ref}' not found`);
-		props = Object.assign(
+	if (obj !== null && typeof obj === 'object') {
```
jsomsanith: What about a fast return instead ?
```
if (obj === null || typeof obj !== 'object') {
    return obj;
}
// do your stuff

```

acateland: why not

PR #433, opened 2017-05-11
feat(forms): Adds string completion widget

talend-glorieux:
We are in need of a String field with completion from a remote suggestion list. 

Adds a String completion widget that takes remote suggestion list.

It has the following features:
* Input highlight
* Mouse and keyboard selection
* single or multi sections

Update to the TypeAhead component:
* Maps onFocus property
* Maps Mouse events properties
* Makes multiSection configurable
* Makes autofocus configurable
* Updates component z-index


packages/forms/src/widgets/StringCompletionWidget/__snapshots__/StringCompletionWidget.test.js.snap
```diff
@@ -0,0 +1,343 @@
+exports[`StringCompletionWidget should render 1`] = `
```
frassinier: Does it make sense to check snapshots here?

packages/forms/stories/index.js
```diff
@@ -114,6 +115,43 @@ decoratedStories.add('Multiple actions', () => {
...
+		},
+		uiSchema: {
+			completion: {
+				'ui:widget': 'stringCompletion',
+				'ui:options': {
+					itemsSrc: '/stringCompletionItems.json',
+				},
+			},
+		},
+		properties: {},
```
frassinier: What if I already have properties?

packages/forms/package.json
```diff
@@ -75,9 +75,9 @@
     "lint:es": "eslint --config .eslintrc src",
     "lint:style": "sass-lint -v -q",
     "lint": "npm run lint:es",
-    "storybook": "start-storybook -p 6008",
-    "start": "start-storybook -p 6008",
-    "build-storybook": "build-storybook"
+    "storybook": "start-storybook -s ./stories/static -p 6008",
+    "start": "start-storybook -s ./stories/static -p 6008",
+    "build-storybook": "build-storybook -s ./stories/static"
```
frassinier: Why these 3 lines?

packages/forms/stories/static/stringCompletionItems.json
```diff
@@ -0,0 +1,18 @@
+[{
+	"title": "A",
```
frassinier: I'm wondering, this JSON structure is needed by component but I think we can mask complexity for widget?

```json
[
  { "group": "A", "items": [ "Apple", "Apricot"] },
  { "group": "B", "items": [ "Banana"] },
  { "group": "C", "items": [ "Cherry"] }, 
]
```

packages/forms/package.json
talend-glorieux: Not required anymore. At the beginning I wanted to load the files with a fetch request.
I'll remove it.

packages/forms/src/widgets/StringCompletionWidget/__snapshots__/StringCompletionWidget.test.js.snap
talend-glorieux: Yep I like to unit test the widget.

packages/forms/stories/index.js
talend-glorieux: They'll be put in as the value.

packages/forms/stories/static/stringCompletionItems.json
talend-glorieux: Yes for sure. This is a first implementation to do quick tests on the integration.
It'll be improved later. The widget API isn't fixed.

talend-glorieux:
I might end up merging the remote items loading part of that widget into the DataListWidget.

talend-glorieux:
Closed in favor of #458

PR #434, opened 2017-05-11
fix(CMF): router inject view props if the component is cmfConnected

jmfrancois:

issue related to #422
If the component is connected the `view` props is not injected.
so the cmf settings view is not injected.

If the component is connected using cmfConnect, create a simple factory component `WithView` which inject the view as a props into the connected component.


acateland:
i would like a better PR description for other reviewers.

PR #439, opened 2017-05-12
fix(components/Drawer): stacked drawer with animation

acateland:
No way to disable drawer transition
Stacked drawer did not work

add a props `withoutTransition` on drawer to set if a transtion is not wanted
add a css rule who permit staking effect up to ten drawer 
move translate directive from drawer to transition class


packages/components/src/Drawer/Drawer.component.js
```diff
@@ -23,11 +24,12 @@ DrawerAnimation.propTypes = {
...
 
-function DrawerContainer({ stacked, className, children, ...rest }) {
+function DrawerContainer({ stacked, className, children, withoutTransition = false, ...rest }) {
 	const drawerContainerClasses = classnames(
 		theme['tc-drawer'],
 		className,
 		'tc-drawer',
+		!withoutTransition && theme['tc-drawer-transition'],
 		{
 			[theme['drawer-stacked']]: stacked,
```
jmfrancois: [theme['tc-drawer-transition']]: !withoutTransition, instead of ternary

packages/components/src/Drawer/Drawer.scss
```diff
@@ -50,15 +49,13 @@ $tc-action-bar-background-color: lighten($alto, 5%) !default;
...
-	width: 98%;
-}
-
-:global(.tc-with-drawer-wrapper) > .drawer-stacked +
-:global(.tc-with-drawer-wrapper) > .drawer-stacked +
-:global(.tc-with-drawer-wrapper) > .drawer-stacked {
-	width: 96%;
+:global(.tc-with-drawer-wrapper) {
+	@for $i from 1 through 10 {
+		&:nth-child(#{$i}) > .drawer-stacked {
```
jmfrancois: AFAIK :nth-child is not well supported on all browser.
question: could we have a different selector ?

packages/components/src/Drawer/Drawer.test.js
```diff
@@ -13,6 +13,14 @@ describe('Drawer', () => {
...
 		expect(wrapper).toMatchSnapshot();
 	});
+	it('should render without tc-drawer-transition class', () => {
+		const wrapper = renderer.create(
+			<Drawer withoutTransition>
+				<h1>Hello world</h1>
+			</Drawer>
+		).toJSON();
+		expect(wrapper).toMatchSnapshot();
+	});
```
jmfrancois: please update the test to find the node and node.hasClassName

packages/components/src/Drawer/Drawer.scss
acateland: this version of the selector is pretty well covered by most browsers https://caniuse.com/#feat=css-sel3

packages/components/src/Drawer/Drawer.scss
```diff
@@ -2,7 +2,7 @@
 
 $tc-drawer-initial-offset: 1000px !default;
 $tc-drawer-bgcolor: #FFF !default;
-$tc-drawer-transition-duration: 0.23s !default;
+$tc-drawer-transition-duration: 230ms !default;
 $tc-drawer-transition-easing: cubic-bezier(0.18, 0.89, 0.32, 1.28) !default;
```
acateland: to better match ReactCSSTransitionGroup declaration https://github.com/Talend/ui/pull/439/files#diff-6981c5849d690d2f8cd7b6c4952d054bR14

packages/components/src/Drawer/Drawer.component.js
```diff
@@ -23,11 +24,12 @@ DrawerAnimation.propTypes = {
...
 			stacked,
```
acateland: it isn't a ternary ;p

packages/components/src/Drawer/Drawer.test.js
acateland: will do

acateland: did update test and component

packages/components/stories/Drawer.js
```diff
@@ -164,6 +164,73 @@ storiesOf('Drawer', module)
...
+		/>);
+		return (
+			<Layout
+				header={header}
+				mode="TwoColumns"
+				one={sidePanel}
+				drawers={stackedDrawers}
+			>
+				<span>zone with drawer</span>
+				<p>The content dictate the width</p>
```
talend-glorieux: Can't you make it with a loop ?

packages/components/src/Drawer/Drawer.component.js
```diff
@@ -23,12 +24,16 @@ DrawerAnimation.propTypes = {
...
 };
 
-function DrawerContainer({ stacked, className, children, ...rest }) {
+function DrawerContainer({ stacked, className, children, withoutTransition = false, ...rest }) {
 	const drawerContainerClasses = classnames(
 		theme['tc-drawer'],
 		className,
 		'tc-drawer',
 		{
+			[theme['tc-drawer-transition']]: !withoutTransition,
```
talend-glorieux: It's usually better to try and avoid double negatives. They create confusions on the conditional flow.
Would be better to rename `withoutTransition` to `withTransition` defaulting to `true`.

PR #447, opened 2017-05-18
feat(containers/list): Add localstorage behaviour on List container

talend-sromain:

No persistance of the list state ( on filters / sort )

First proposal : 
- add local storage behaviour on the list only

Another approach to make it would be to use redux storage & change the way the app is initialized

### DIFF
--- a/packages/containers/src/List/List.connect.js
+++ b/packages/containers/src/List/List.connect.js
@@ -1,4 +1,5 @@
 import get from 'lodash/get';
+import { Map } from 'immutable';
 import { cmfConnect } from 'react-cmf';
 import Container, { DEFAULT_STATE } from './List.container';
 import { configureGetFilteredItems, configureGetPagination } from './selector';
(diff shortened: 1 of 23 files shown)


packages/containers/src/List/List.connect.js
```diff
@@ -28,13 +29,22 @@ export function mapStateToProps(state, ownProps, cmfProps) {
...
 	props.items = getItems(state, config);
 	const cmfState = get(cmfProps, 'state');
+
+	if (cmfState && ownProps.localStorageKey) {
+		const itemStored = JSON.parse(localStorage.getItem(ownProps.localStorageKey));
+		let storedState = {};
+		if (itemStored !== null && typeof itemStored === 'object') {
+			storedState = itemStored;
+		}
+		props.state = new Map({ ...cmfProps.state.toJS(), ...storedState });
```
jmfrancois: note: it work only if you have native JS value under the state object

jmfrancois:
This work but just for this container.

If you want to extends this to all containers you will have to copy paste all that code throw all containers.

An other solution could be to put this code inside cmfConnect so all containers got the support for a props to sync the local storage into it.

An other solution is #449 

I would not block that PR because you may want it  in your project but for me it's not the way to go for all containers.

talend-sromain:
I close this to use #449 to store informations on local storage

PR #448, opened 2017-05-18
feat(component/drawer): inject transition status into children

acateland:
If a drawer contain a form, on wich we want an autofocus.
The drawer see itself being pulled by the browser in the viewport, its an accessibility element that many browser implement.
And instead of seeing the drawer being animated properly, we see the page content being thrown to the left then animated to right direction.

inject a property with transtion status, to allow the injection of form with autofocus when transition is done.


packages/components/src/WithDrawer/withDrawer.scss
```diff
@@ -6,7 +6,24 @@
...
+.tc-with-drawer-container > div {
+	height: 100%;
 	position: absolute;
 	top: 0;
 	right: 0;
+	width: 100%;
+}
+
+.tc-with-drawer-container > div {
+	@for $i from 1 through 10 {
```
talend-glorieux: Why 1 through 10?

acateland: arbitrary stuff, i hope no one will try to stack more thant ten drawers.

PR #452, opened 2017-05-18
fix(components/List): add uppercase on sort by dd menu

jmfrancois:

guidelines says the List sort by dropdown menu items must be uppercase.
this is not the case.
on each render we define a function that just doing some rendering of menu item.

this is the case.
the menu item function has been extracted outside (should improve performance ... a bit.

screen shot 2017-05-18 at 14 28 15

### DIFF
--- a/packages/components/src/List/Toolbar/SelectSortBy/SelectSortBy.component.js
+++ b/packages/components/src/List/Toolbar/SelectSortBy/SelectSortBy.component.js
@@ -10,7 +10,6 @@ function SortByItem({ option, index, id }) {
 			id={id && `${id}-by-item-${option.id}`}
 			key={index}
 			eventKey={option}
-			className={theme['sort-by-item']}
 		>
 			{option.name || option.id}
 		</MenuItem>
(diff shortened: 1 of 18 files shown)


packages/components/src/List/Toolbar/SelectSortBy/SelectSortBy.component.js
```diff
@@ -1,9 +1,30 @@
...
 
 import theme from './SelectSortBy.scss';
 
+function SortByItem({ option, index, id }) {
+	return (
+		<MenuItem
+			id={id && `${id}-by-item-${option.id}`}
+			key={index}
+			eventKey={option}
+			className={theme['sort-by-item']}
```
frassinier: Why each item needs that className?
I prefer set 1 className to their parent.

packages/components/src/List/Toolbar/SelectSortBy/__snapshots__/SelectSortBy.test.js.snap
```diff
@@ -37,9 +37,8 @@ exports[`SelectSortBy should render 1`] = `
       >
         <a
           href="#"
-          id={undefined}
+          id={0}
```
frassinier: This `id` should not what you are look for I guess

packages/components/src/List/Toolbar/SelectSortBy/SelectSortBy.component.js
```diff
@@ -16,18 +36,7 @@ function SelectSortBy({ field, id, isDescending, onChange, options }) {
...
-				id={id && `${id}-by-item-${option.id}`}
-				key={index}
-				eventKey={option}
-			>
-				{option.name || option.id}
-			</MenuItem>
-		);
-	}
-
+	const getMenuItem = SortByItem;
```
jmfrancois: note: https://github.com/Talend/ui/pull/452/commits/15ab072bee721de764c80f853ccb4127517a14d0
if you call it using JSX you loose the onKeyDown ...

PR #462, opened 2017-05-22
fix(components/SidePanel): fix propTypes

mtikhinskaya:

SidePanel.selected.propType  is string

SidePanel.selected.propType is object

### DIFF
--- a/packages/components/src/SidePanel/SidePanel.component.js
+++ b/packages/components/src/SidePanel/SidePanel.component.js
@@ -108,7 +108,10 @@ SidePanel.propTypes = {
 	onSelect: React.PropTypes.func,
 	onToggleDock: React.PropTypes.func,
 	docked: React.PropTypes.bool,
-	selected: React.PropTypes.string,
+	selected: React.PropTypes.shape({
+		label: React.PropTypes.string,
+		icon: React.PropTypes.string,
+	}),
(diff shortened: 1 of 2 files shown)


packages/components/src/SidePanel/__snapshots__/SidePanel.snapshot.test.js.snap
```diff
@@ -1238,7 +1238,7 @@ exports[`Storyshots SidePanel default 1`] = `
                     selected
                   </td>
                   <td>
-                    string
+                    other
```
jsomsanith: Why do you have a diff in tests ?

packages/components/src/SidePanel/SidePanel.component.js
```diff
@@ -108,7 +108,10 @@ SidePanel.propTypes = {
 	onSelect: React.PropTypes.func,
 	onToggleDock: React.PropTypes.func,
 	docked: React.PropTypes.bool,
-	selected: React.PropTypes.string,
+	selected: React.PropTypes.shape({
```
jsomsanith: This is not accurate, for what I understand, the selected is the action that is selected.
What I suggest : 
* create a const actionPropType
* `selected : actionPropType` and `actions: PropTypes.arrayOf(actionPropTypes)`

PR #466, opened 2017-05-23
feat(enumeration-widget): add multiple items in a row

talend-sromain:

Currently we can only add one by one items on the list

New action button to keep add another item + default action on enter to reset the field & prepare to add another


packages/components/src/Enumeration/Header/HeaderInput.component.js
```diff
@@ -4,7 +4,7 @@ import classNames from 'classnames';
 import Action from '../../Actions/Action';
 import theme from './Header.scss';
 
-let inputRef;
+let internalInputRef;
```
acateland: if this component get used more than once into another component this value will be shared between them.

output/components.eslint.txt
```diff
@@ -13,6 +13,22 @@ Scoping to packages that match 'react-talend-components'
   2:1  error  '@kadira/react-storybook-addon-info' should be listed in the project's dependencies, not devDependencies  import/no-extraneous-dependencies
   7:2  error  Unexpected require()                                                                                      global-require
 
+/home/travis/build/Talend/ui/packages/components/src/Enumeration/Enumeration.component.js
+   94:2   error  Mixed spaces and tabs                       no-mixed-spaces-and-tabs
```
jare-talend: plz fix the eslint errors

packages/forms/src/widgets/EnumerationWidget/EnumerationWidget.js
```diff
@@ -626,8 +681,12 @@ class EnumerationWidget extends React.Component {
 		return (
 			<input
 				type="file"
-				ref={(element) => { this.inputFile = element; }}
-				onChange={(event) => { this.importFile(event); }}
+				ref={(element) => {
```
jare-talend: it's normal to reformat this. My eslint is good on this part.

packages/forms/src/widgets/EnumerationWidget/EnumerationWidget.js
```diff
@@ -418,6 +430,39 @@ class EnumerationWidget extends React.Component {
...
+				this.addFailHandler.bind(this))
+		) {
+			this.setState({
+				headerInput: this.loadingInputsActions,
+			});
+			this.input.focus();
+		} else if (!this.valueAlreadyExist(value.value)) {
+			this.setState(
+				{
+					items: this.state.items.concat([{
```
acateland: using setState this wayn don't garanties that this.state.items is in intended state. see https://facebook.github.io/react/docs/react-component.html#setstate

packages/forms/src/widgets/EnumerationWidget/EnumerationWidget.js
```diff
@@ -418,6 +430,39 @@ class EnumerationWidget extends React.Component {
...
+						values: this.parseStringValueToArray(value.value),
+						inputValue: '',
+					}]),
+				},
+				this.setFormData.bind(this)
```
acateland: this appear 5 time in the code and this should be bind into the constructor

output/components.eslint.txt
talend-sromain: oh yes, what have done my ide

packages/forms/src/widgets/EnumerationWidget/EnumerationWidget.js
```diff
@@ -450,6 +468,34 @@ class EnumerationWidget extends React.Component {
...
+				this.constructor.parseStringValueToArray(value.value),
+				this.validateAndAddSuccessHandler.bind(this),
+				this.addFailHandler.bind(this))
+		) {
+			this.setState({
+				headerInput: this.loadingInputsActions,
+			});
+			this.input.focus();
+		} else if (!this.valueAlreadyExist(value.value)) {
+			const items = this.state.items.concat([{
```
acateland: same as before this.state my not be in intended shape please use following state update form
this.setState((prevState, props) => { items: prevState...retrieving prevstate data... , inputValue: ''})

Issue #438, opened 2017-05-12
Its impossible to define custom widget without label in forms


sdwvit:
Right now to show custom label you should hack it with custom widget, but then problem with rerendering appears and also should hack css as well..

Developer should have an ability to define schema for custom labels (or even components) to be rendered at any time in any place of the form.

sdwvit:
![screenshot from 2017-05-11 18-34-36](https://cloud.githubusercontent.com/assets/8777372/25999907/8908b19a-372f-11e7-8147-5ba210fcac57.png)
See screenshot

sdwvit:
closing as the red label (as on screenshot) can be done with custom css class and help-label

PR #467, opened 2017-05-23
feat(components): Adds JSONSchemaRenderer

talend-glorieux:
JSONSchema based exchanges between the WebApps and servers are only rendered as forms.
We need a way to display that data in a proper read-only manner by using other HTML tags than the ones available in a form.

Adds a component that renders a read-only version of a given combination
of JSONSchema and data.


packages/components/src/JSONSchemaRenderer/JSONSchemaRenderer.scss
```diff
@@ -0,0 +1,42 @@
...
+		padding-bottom: 10px;
+
+		dt {
+			font-weight: 400;
+			font-size: 1.2rem;
+			margin-bottom: 0;
+		}
+
+		dd {
+			font-size: 14px;
```
frassinier: Please use rem instead

packages/components/src/JSONSchemaRenderer/JSONSchemaRenderer.scss
```diff
@@ -0,0 +1,42 @@
+.json-schema-renderer {
+
+	margin: 20px 0 0 20px;
```
frassinier: Any way to use one margin value instead? 
There are some variables for that in theme package.

packages/components/src/JSONSchemaRenderer/JSONSchemaRenderer.test.js
```diff
@@ -0,0 +1,128 @@
...
+				b: 'test b',
+				c: 'test c',
+			},
+		};
+		const wrapper = shallow(<JSONSchemaRenderer schema={schema} />);
+		expect(wrapper.find('dt').first().text()).toEqual('a');
+		expect(wrapper.find('dt').last().text()).toEqual('c');
+	});
+
+	xit('should handle $ref', () => {});
```
frassinier: Could you add FIXME or TODO? And explain why it's ignored ATM?

output/components.eslint.txt
```diff
@@ -18,6 +18,9 @@ Scoping to packages that match 'react-talend-components'
   2:1  error  '@kadira/react-storybook-addon-info' should be listed in the project's dependencies, not devDependencies  import/no-extraneous-dependencies
   7:2  error  Unexpected require()                                                                                      global-require
 
+/home/travis/build/Talend/ui/packages/components/src/JSONSchemaRenderer/JSONSchemaRenderer.component.js
+  75:41  error  'typeResolver' was used before it was defined  no-use-before-define
```
jmfrancois: please fix it

packages/components/src/JSONSchemaRenderer/JSONSchemaRenderer.component.js
```diff
@@ -0,0 +1,174 @@
...
+import entries from 'lodash/entries';
+
+import css from './JSONSchemaRenderer.scss';
+
+const className = 'json-schema-renderer';
+
+/**
+ * UnkownTypeException
+ *
+ * @param type - The unkown type
```
jmfrancois: missing add param type
@param {string} type

packages/components/src/JSONSchemaRenderer/JSONSchemaRenderer.component.js
```diff
@@ -0,0 +1,174 @@
...
+
+/**
+ * textRenderer
+ *
+ * @param key
+ * @param title
+ * @param text
+ * @returns {string} - HTML markup for a text component
+ */
+function textRenderer(key, title, text) {
```
jmfrancois: why no using a component here ?
Text({key, title, text}) ?
by the way with component the key is not a needed props because it's internal stuff for react.

packages/components/src/JSONSchemaRenderer/JSONSchemaRenderer.component.js
```diff
@@ -0,0 +1,174 @@
...
+ */
+function typeResolver(schema) {
+	return function resolver(e) {
+		if (!schema[e[0]]) {
+			return null;
+		}
+		const type = schema[e[0]].type;
+		const title = schema[e[0]].title;
+
+		const renderer = registry[type];
```
jmfrancois: question: I can't edit the registry before rendering ?

packages/components/src/JSONSchemaRenderer/JSONSchemaRenderer.component.js
talend-glorieux: 1. I don't want to use some props spreading for 3 parameters. Less allocations.
2. The rendering is quite simple and doesn't require the react component's overhead.

packages/components/src/JSONSchemaRenderer/JSONSchemaRenderer.component.js
talend-glorieux: No not yet. But it could make sense to extend it later. The current registry covers most of the JSONSchema cases.

packages/components/src/JSONSchemaRenderer/JSONSchemaRenderer.component.js
jmfrancois: 1. in this case use Text(props) you will have only one allocation
2. what overhead are you taking about ? This is a function that return JSX elements

packages/components/src/JSONSchemaRenderer/JSONSchemaRenderer.component.js
```diff
@@ -0,0 +1,174 @@
...
+	if (!props.schema.jsonSchema || !props.schema.properties) {
+		throw new InvalidSchemaException();
+	}
+	let properties = entries(props.schema.properties);
+	if (props.schema.uiSchema) {
+		properties = orderProperties(props.schema.uiSchema['ui:order'], properties);
+	}
+	const elements = properties.map(typeResolver(props.schema.jsonSchema.properties));
+	return (
+		<dl className={classNames(css[className])}>
```
jmfrancois: please add a global class name so I can it it with CSS and for e2e

sdwvit:
Looks like #438 will be fixed with this one? @glorieux

talend-glorieux:
@sdwvit No that component isn't related to the forms. It's a pure view of JSONSchema formatted data.

PR #472, opened 2017-05-24
fix(containers/SidePanel): Uses getActionsProps

talend-glorieux:
The SidePanel use the old api.action.getActionInfo.

The SidePanel container now uses getActionsProps


packages/containers/src/SidePanel/SidePanel.container.js
```diff
@@ -1,22 +1,25 @@
...
 
+import { getActionsProps } from '../actionAPI';
+
 export const DEFAULT_STATE = new Map({
 	docked: false,
 });
 
-
-export function getActions(actionIds, context) {
+export function getActions(actionIds, context, model) {
```
jsomsanith: This is only used on line 65, passing no model. Why this new arg ?

packages/containers/src/SidePanel/SidePanel.container.js
```diff
@@ -1,22 +1,25 @@
...
 	return actionIds.map((id) => {
-		const info = api.action.getActionInfo(context, id);
-		info.onClick = () => context.store.dispatch(info.payload);
+		const info = getActionsProps(context, id, model);
 		let route = get(info, 'payload.cmf.routerReplace');
 		if (!route) {
 			route = get(info, 'payload.cmf.routerPush');
 		}
+		if (!route) {
+			route = get(info, 'href');
```
jsomsanith: Seems to be a new feature ? Deserve a unit test

packages/containers/src/SidePanel/SidePanel.container.js
talend-glorieux: Because the getActionProps need it.

jsomsanith: This means that `const actions = getActions(actionIds, this.context);` will fail because it doesn't provide it.

talend-glorieux: nope it'll be OK. You just won't have the extra data passed through the model.

PR #478, opened 2017-05-28
fix(components/TooltipTrigger): only bind if needed, performance issue

jmfrancois:

Performance issue caused by the tooltip from react-bootstrap.

Re-work the tooltip trigger


packages/components/src/TooltipTrigger/TooltipTrigger.component.js
```diff
@@ -22,17 +24,75 @@ const props = {
...
+		super(props);
+		this.onMouseOver = this.onMouseOver.bind(this);
+		this.onMouseOut = this.onMouseOut.bind(this);
+		this.state = {
+			hovered: false,
+			id: uuid.v4(),
+		};
+	}
+
+	shouldComponentUpdate(nextProps, nextState) {
```
frassinier: or
```javascript
shouldComponentUpdate(nextProps, nextState) {
  return (
    this.state.hovered !== nextState.hovered || 
    this.props.children !== nextProps.children || 
    this.props.label !== nextProps.label
  );
}
```

packages/components/src/TooltipTrigger/TooltipTrigger.component.js
```diff
@@ -22,17 +24,75 @@ const props = {
...
+			return true;
+		}
+		if (this.props.label !== nextProps.label) {
+			return true;
+		}
+		return false;
+	}
+
+	onMouseOver() {
+		this.setState({ hovered: true });
```
frassinier: What do you think about this line perf?

packages/components/src/TooltipTrigger/TooltipTrigger.component.js
```diff
@@ -22,17 +24,75 @@ const props = {
...
+	render() {
+		if (!this.state.hovered) {
+			const child = React.Children.only(this.props.children);
+			const childProps = child.props;
+			const triggerProps = Object.assign({
+				onMouseOver: this.onMouseOver,
+				onFocus: this.onMouseOver,
+				'aria-describedby': this.state.id,
+				onBlur: noop,
+				onMouseOut: noop,
```
talend-glorieux: Why are those props defined if it's to call a noop function on them?
I would expect them to be define like the `onClick`.

jmfrancois: as mentionned on the commit message they are here to not break 50 snapshots, until we have finished and validate this is the way to go ;)

packages/components/src/TooltipTrigger/TooltipTrigger.component.js
acateland: think this should check actual state, to avoid triggering a render call and even avoid calling shouldComponentUpdate

packages/components/src/TooltipTrigger/TooltipTrigger.component.js
sdwvit: I would have added util function which checks props difference. eg

function checkDiff(oldProps, newProps, [...propsToCheck]) {
    return !!(propsToCheck.filter(el => oldProps[el] !== newProps[el]).length);
}

and then:

shouldComponentUpdate(nextProps, nextState) {
    return checkDiff(this.props, nextProps, ['label', 'children']) 
            || checkDiff(this.state, nextState, ['hovered']);
}

but its a matter of taste

packages/components/src/TooltipTrigger/TooltipTrigger.component.js
sdwvit: debounce + what @acateland said

packages/components/src/TooltipTrigger/TooltipTrigger.component.js
jsomsanith: @sdwvit This make the code harder to read. Here it is clear, if the conditions were more complicated yes that's a good point to externalize it. But here it is simpler to keep this easy checks IMO

PR #484, opened 2017-05-29
feat(components/ConfirmDialog): set max height and overflow option

acateland:
When the content of a modal get to big, the height of the modal get superior than the height of the viewport, making action button unreachable without scrolling

add a max height to the modal and have the content of the modal scroll accordingly.
add a property who remove the overflow visible so user can setup how component will scrolls inside the modal body.
http://guidelines.talend.com/document/92132#/navigation-layout/dialog-box


packages/components/stories/ConfirmDialog.js
```diff
@@ -109,4 +132,276 @@ storiesOf('ConfirmDialog', module)
...
+					{
+						id: 10,
+						name: 'Title with long long long long long long long long long long long text',
+						created: '2016-09-22',
+						modified: '2016-09-22',
+						author: 'Jean-Pierre DUPONT',
+						actions,
+						icon: 'talend-file-xls-o',
+						display: 'text',
+						className: 'item-3-class',
```
talend-glorieux: Could you generate all that content with a loop?

packages/components/src/ConfirmDialog/ConfirmDialog.component.js
```diff
@@ -39,24 +41,25 @@ function ConfirmDialog({
...
 	progressValue,
+	bodyOverflow = true,
 }) {
 	return (
 		<Modal
 			bsSize={size}
 			show={show}
 			keyboard
-			className={classNames(theme['tc-confirm-dialog'])}
+			className={classNames(theme['tc-confirm-dialog'], bodyOverflow && theme['modal-body-overflow'])}
```
jmfrancois: please use classNames({ foo: true }, { bar: true }); // => 'foo bar'

packages/components/src/ConfirmDialog/ConfirmDialog.scss
```diff
@@ -6,8 +6,17 @@ $tc-dialog-body-padding: 30px !default;
...
 
+	&.modal-body-overflow {
+		:global(.modal-body) {
+			overflow: auto;
+		}
+	}
+
 	:global(.modal-content) {
 		background-color: tint($tc-dialog-header-color, 90%);
+		max-height: 80vh;
```
frassinier: Be careful with IE11... `max-height` *could* not work

packages/theme/src/theme/_modals.scss
```diff
@@ -1,5 +1,12 @@
 
+.modal-dialog {
+	max-height: 80vh;
```
frassinier: Same here

packages/components/src/ConfirmDialog/ConfirmDialog.component.js
acateland: those are not ternary expression, this is a simple js expression returning je last element if bodyOverflow is truthy, it may even be faster than creating a new object literal, checking if first key is true and then extracting said key.

packages/components/src/ConfirmDialog/ConfirmDialog.scss
acateland: added a workaround, IE sigh

PR #492, opened 2017-05-31
feat(forms/component): add choices in the EnumerationWidget

jare-talend:
Add some options on the action on the component EnumerationWidget

The actions in the header can be either a Action or a DropDown


packages/forms/src/widgets/EnumerationWidget/EnumerationWidget.js
```diff
@@ -463,6 +494,18 @@ class EnumerationWidget extends React.Component {
...
+
+	onImportButtonClick() {
+		this.callActionHandler(
+			ENUMERATION_IMPORT_FILE_CLICK,
+			{
+				simulateClickInputFile: this.simulateClickInputFile.bind(this),
+			},
+			this.importFileHandler.bind(this),
+			this.importFileHandler.bind(this)
+		);
```
talend-glorieux: That's a lot of binding. Won't it hurt performances?

packages/forms/src/widgets/EnumerationWidget/EnumerationWidget.test.js
```diff
@@ -208,8 +208,136 @@ describe('EnumerationWidget', () => {
...
+				<EnumerationWidget
+					registry={registry}
+					schema={{
+						allowImport: true,
+					}}
+				/>);
+
+			// when
+			wrapper
+				.find('.tc-enumeration-header div.btn-group-link button').at(0)
```
talend-glorieux: please aggregate the wrapper find method call into a single line.

packages/forms/src/widgets/EnumerationWidget/EnumerationWidget.test.js
```diff
@@ -208,8 +208,136 @@ describe('EnumerationWidget', () => {
...
+				'ENUMERATION_IMPORT_FILE_CLICK',
+				{
+					simulateClickInputFile: jasmine.any(Function),
+				},
+				jasmine.any(Function),
+				jasmine.any(Function),
+			);
+		});
+
+		it('should send a event with the choice done by the user', () => {
```
jlame: instead of something generic like this, can you change it so the expect is more specific ? (we need to read the test to know what it does)
in that case the aim is to send an import file event in append mode.

packages/forms/src/widgets/EnumerationWidget/EnumerationWidget.test.js
```diff
@@ -208,8 +208,136 @@ describe('EnumerationWidget', () => {
...
+			expect(registry.formContext.handleAction).toBeCalledWith(
+				undefined,
+				'ENUMERATION_IMPORT_FILE_APPEND_MODE',
+				null,
+				jasmine.any(Function),
+				jasmine.any(Function),
+			);
+		});
+
+		it('should send a event with the choice done by the user', () => {
```
jlame: instead of something generic like this, can you change it so the expect is more specific ? (we need to read the test to know what it does)
in that case the aim is to send an import file event in overwrite mode.

PR #497, opened 2017-06-01
feat(forms): Adds children to Forms

talend-glorieux:
We can't pass a children to the Forms in order to inject extra content on it. 
Adds children to Forms.


jsomsanith:
Can you tell why this is needed ?

acateland:
trying to solve this https://jira.talendforge.org/browse/TFD-622

as you see the children table should sit between input field rendered by the form component and the action block also rendered by the form component, not a big fan of this but...

PR #499, opened 2017-06-01
fix(cmf): remove all internal props already used by the container

jmfrancois:

cmfConnect create a container which wrap the given component to add some behavior.
For example didMountActionCreator props make the container dispatch an action after mount.
those props are at the moment spreaded to the wrapped component.

So the component must be aware of those props which is not acceptable.
For example if a subcomponent get a didMountActionCreator because of a simple spread, this is bad.

remove internal props already used by the container.


packages/cmf/src/cmfConnect.js
```diff
@@ -169,6 +169,11 @@ export default function cmfConnect({
 					this.props,
 					{ dispatchActionCreator: this.dispatchActionCreator },
 				);
+				// remove all internal props already used by the container
```
acateland: i think it should be a delete instead of setting the props to undefined ?
if the props is still present any function checking property existence will return true.

jsomsanith: I would suggest to set a `CMF_PROPS` array + a function that delete all those props.
So adding a new cmf props will automatically remove it here. Maintaining an array of const is easier than maintaining code

packages/cmf/__tests__/cmfConnect.test.js
```diff
@@ -337,4 +337,28 @@ describe('cmfConnect', () => {
 		expect(context.store.dispatch.mock.calls[0][0]).toEqual(firstCall);
 		expect(context.store.dispatch.mock.calls[1][0]).toEqual(secondCall);
 	});
+	it('should remove internal props', () => {
+		const iProps = {
+			didMountActionCreator: 'myactionCreator',
```
jsomsanith: Please add the other internal props and assert their deletion

PR #500, opened 2017-06-01
feat(components): Progress - style changes

Sylvain-Talend:
- Increase progress rapidity to conform to http://guidelines.talend.com/document/92132#/ui-controls/loading-feedback => set to 1s as discussed with UX team
- Decrease length of the space between green lines

Change CSS style


packages/components/src/Progress/Progress.scss
```diff
@@ -40,7 +40,7 @@ $gradient-part-width: 100% / 6;
 		&.infinite {
 			background: transparent;
 			width: 100%;
-			animation: infinite-progression-keyframes 5s infinite;
+			animation: infinite-progression-keyframes 2s infinite;
```
acateland: The duration is not defined on linked frontify specification, can you ask UX/UI team to update spec

ncomont: I'd like to add that you're not increasing the progress bar speed, but the time it took to the bare to achieve a complete tour.

The speed will depends of the progress bar width. Maybe this value should be part of the progress bar API ?

Sylvain-Talend: Thanx for these returns,  I asked @maximenicoul to clarify those points

Sylvain-Talend: @ncomont I just set the value in the progress bar API. Please tell me if it sounds good to you

Sylvain-Talend: As seen with @maximenicoul frontify documentation will be updated to set the new speed time. It will be hard-coded, no need to parametrize it.

packages/components/src/Progress/Progress.scss
```diff
@@ -1,5 +1,6 @@
 $tc-progress-color: $brand-success !default;
-$gradient-part-width: 100% / 6;
+$solid-gradient-part-width: 22%;
```
ncomont: Where does these values come from ? I think you should write the calculation you did from readability / maintenance reasons  (`$gradient-part-width: 100% / 6;`). It is preprocessed by SASS, so it doesnt affect performances.

Sylvain-Talend: done

PR #501, opened 2017-06-01
fix: lost state of the children when the tooltip disparears

jare-talend:
When the tooltip disappears, the children come back to the previous state.
To reproduce the bug: http://talend.surge.sh/components/?selectedKind=ActionDropdown&selectedStory=default&full=0&down=1&left=1&panelRight=0&downPanel=kadirahq%2Fstorybook-addon-actions%2Factions-panel

On the last dropdown, click on the dropdown to expand it. Move the mouse. The toolip disappears and the dropdown collapse it.

Don't remove the TooltipTrigger after his creating.


acateland:
DON'T MERGE THIS AS IT REQUIRE DISCUSSION
see https://github.com/Talend/ui/pull/478

jare-talend:
Except this bug is very blocking for us.

packages/components/src/TooltipTrigger/TooltipTrigger.component.js
```diff
@@ -78,7 +78,6 @@ class TooltipTrigger extends React.Component {
 				placement={this.props.tooltipPlacement}
 				overlay={tooltip}
 				delayShow={400}
-				onExited={this.onMouseOut}
```
acateland: the bind of this.onMouseOut id removed but the function attached to the component remain and is unused.

jare-talend: exact

PR #504, opened 2017-06-02
feat(component/List): add virtualized

jmfrancois:

The list is slow when lot's of items in it.

The list accept a new props 'virtualized' which trigger the use of VirtualizedList
with a compatibility mode.

**state**

This PR is ready to be used in some project like TFD. Some missing binding are left to other teams:

- support for selection
- any other ?


packages/components/src/List/ListToVirtualizedList/ListToVirtualizedList.component.js
```diff
@@ -0,0 +1,59 @@
...
+function ListToVirtualizedList(props) {
+	const { id, items, columns, titleProps } = props;
+	if (!titleProps.actionsKey) {
+		titleProps.actionsKey = 'actions';
+	}
+	// Backward compatibility: find array in object attr:
+	const supposedActions = {};
+	if (props.items.length > 0) {
+		const item = props.items[0];
+		Object.keys(item).forEach((key) => {
```
talend-glorieux: I'd rather see a reduce used here.

packages/containers/src/List/List.container.js
```diff
@@ -102,6 +102,7 @@ class List extends React.Component {
 				items,
 				columns: get(this.props, 'list.columns', []),
 			},
+			virtualized: this.props.virtualized || false,
```
jsomsanith: Remove `|| false`, not needed (+ update snapshots)

packages/components/src/List/ListToVirtualizedList/ListToVirtualizedList.component.js
```diff
@@ -0,0 +1,59 @@
...
+					<VirtualizedList.Content key={index} {...cProps} />
+				);
+			})}
+		</VirtualizedList>
+	);
+}
+
+ListToVirtualizedList.propTypes = {
+	id: PropTypes.string,
+	displayMode: PropTypes.oneOf(['LARGE', 'TABLE']),
```
jsomsanith: Warning on the storybook : 
```
warning.js:36 Warning: Failed prop type: Invalid prop `displayMode` of value `large` supplied to `ListToVirtualizedList`, expected one of ["LARGE","TABLE"].
```

jlame:
Why not add some rows (like a lot)  to the story book as the list is now virtualize ?

packages/components/src/List/ListToVirtualizedList/ListToVirtualizedList.component.js
jsomsanith: Reduce in this case is not really improving the readability. But we still can improve that : 
```
Object.keys(item)
	.filter(key => Array.isArray(item[key]))
	.forEach((key) => { supposedActions[key] = true; });
``

PR #505, opened 2017-06-05
fix(components): Adds ellipsis to Drawers title

talend-glorieux:
The Drawers header can contain text that overflows the container.

Add an elipsis when the header text is too long.


talend-glorieux:
@mtikhinskaya Yes sure this can be a future requirement. For now, even the ellipsis isn't clearly defined on the guidelines :(

PR #508, opened 2017-06-05
fix(components): Manage middle-click (mousewheel) on item title

Sylvain-Talend:
Manage middle click (mouse wheel) on item title

Add a mouseDown event bound to onClick function

### DIFF
--- a/packages/components/src/List/ItemTitle/ItemTitle.component.js
+++ b/packages/components/src/List/ItemTitle/ItemTitle.component.js
@@ -123,7 +123,6 @@ function ItemTitle({ id, className, item, titleProps }) {
 		iconKey,
 		displayModeKey,
 		onClick,
-		onMouseDown,
 		onEditSubmit,
 		onEditCancel,
 	} = titleProps;
@@ -141,7 +140,6 @@ function ItemTitle({ id, className, item, titleProps }) {
 				className={className}
 				item={item}
 				onClick={onClick}
-			    onMouseDown={onMouseDown}
 			/>);
 		} else {
 			titleElement = renderText({ id, value, className });
@@ -173,7 +171,6 @@ ItemTitle.propTypes = {
 		iconKey: PropTypes.string,
 		displayModeKey: PropTypes.string,
 		onClick: PropTypes.func,
-		onMouseDown: PropTypes.func,
 		onEditSubmit: PropTypes.func,
 		onEditCancel: PropTypes.func,
 	}).isRequired,
--- a/packages/components/src/List/ItemTitle/ItemTitle.component.js
+++ b/packages/components/src/List/ItemTitle/ItemTitle.component.js
@@ -12,14 +12,19 @@ const TITLE_MODE_INPUT = 'input';
 const ESC_KEY = 27;
 const ENTER_KEY = 13;
 
-export function RenderButton({ id, value, className, item, onClick }) {
+export function RenderButton({ id, value, className, item, onClick, onMouseDown }) {
 	function click(event) {
 		event.stopPropagation();
 		onClick(event, item);
 	}
 
+	function mouseDown(event) {
+		event.stopPropagation();
+		onMouseDown(event, item);
+	}
+
 	return (
-		<Button id={id} className={className} onClick={click} role="link" bsStyle="link">
+		<Button id={id} className={className} onMouseDown={mouseDown} onClick={click} role="link" bsStyle="link">
 			{value}
 		</Button>
 	);
@@ -123,6 +128,7 @@ function ItemTitle({ id, className, item, titleProps }) {
 		iconKey,
 		displayModeKey,
 		onClick,
+		onMouseDown,
 		onEditSubmit,
 		onEditCancel,
 	} = titleProps;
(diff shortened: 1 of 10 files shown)


Sylvain-Talend:
I close this PR as it can be managed directly by the onClick event

talend-glorieux:
Isn't it automatically handled by the browser if we use a link instead of a button?

frassinier:
Yes but please avoid using anchor because of our different router impl

Sylvain-Talend:
Indeed would be great to manage Ctrl-Click and Middle-Click natively. Will do some POC to test that but I'm afraid of huge impact in transforming Button into Link.

packages/components/src/List/ItemTitle/ItemTitle.component.js
```diff
@@ -14,12 +14,23 @@ const ENTER_KEY = 13;
 
 export function RenderButton({ id, value, className, item, onClick }) {
 	function click(event) {
-		event.stopPropagation();
-		onClick(event, item);
+		// only fires onclick when left-click
+		if (event.button === 0) {
```
jsomsanith: Why do we need this test ?

packages/components/src/List/ItemTitle/ItemTitle.test.js
```diff
@@ -122,7 +122,30 @@ describe('ItemTitle', () => {
 
 		// when
 		const wrapper = mount(titleInstance);
-		wrapper.find(Button).simulate('click', { stopPropagation: () => {} });
+		wrapper.find(Button).simulate('click', { button: 0, stopPropagation: () => {} });
+
+		// then
+		expect(props.titleProps.onClick).toBeCalled();
+		const callArgs = props.titleProps.onClick.mock.calls[0];
+		expect(callArgs[1]).toBe(item);
```
jsomsanith: Replace the 3 lines above by `expect(props.titleProps.onClick).toBeCalledWith(item)`

packages/components/src/List/ItemTitle/ItemTitle.test.js
```diff
@@ -122,7 +122,30 @@ describe('ItemTitle', () => {
...
+			},
+		};
+		const titleInstance = <ItemTitle {...props} />;
+
+		// when
+		const wrapper = mount(titleInstance);
+		wrapper.find(Button).simulate('mouseDown', { button: 1, stopPropagation: () => {} });
 
 		// then
 		expect(props.titleProps.onClick).toBeCalled();
```
jsomsanith: `expect(props.titleProps.onClick).toBeCalledWith(item)`

packages/components/src/List/ItemTitle/ItemTitle.component.js
```diff
@@ -14,12 +14,23 @@ const ENTER_KEY = 13;
...
+			event.stopPropagation();
+			onClick(event, item);
+		}
+	}
+
+	function mouseDown(event) {
+		// only fires onclick when middle-click
+		if (event.button === 1) {
+			event.stopPropagation();
```
jsomsanith: If you remove the if() in click() function, here just call `click(event)` when you have a middle click. The stopPropagation will be handled.
This would represent basically what you want : `in case of middle click, execute the same thing as click callback.`

packages/components/src/List/ItemTitle/ItemTitle.component.js
Sylvain-Talend: In TDS-side, in callback `onClick` we do a redirection on a new page (`target = _blank`)
In that case, Chrome and FF behave differently when we do a middle-click (`mouseDown`) event . Chrome will not trigger a mouse click event after `mouseDown` event, but FF will. 
With this test, I prevent FF (and surely other browsers) to trigger a `onClick` event when there was a `mouseDown` event triggered before.

packages/components/src/List/ItemTitle/ItemTitle.component.js
Sylvain-Talend: I did not remove the if() as explained above, but there was however a little code duplication I removed, thanx

packages/components/src/List/ItemTitle/ItemTitle.test.js
Sylvain-Talend: indeed, done

packages/components/src/List/ItemTitle/ItemTitle.test.js
Sylvain-Talend: indeed, done

packages/components/src/List/ItemTitle/ItemTitle.component.js
```diff
@@ -13,13 +13,27 @@ const ESC_KEY = 27;
...
 export function RenderButton({ id, value, className, item, onClick }) {
-	function click(event) {
+	function executeOnClickAction(event) {
 		event.stopPropagation();
 		onClick(event, item);
 	}
 
+	function click(event) {
+		// only fires onclick when left-click
+		if (event.button === 0) {
```
jsomsanith: Why this test ?

Sylvain-Talend: Please see above

PR #509, opened 2017-06-06
feat(forms): deal with arrays

frassinier:
**What is the problem this PR is trying to solve?**
Arrays are not supported

**What is the chosen solution to this problem?**
Impl Select widgets for them

**Please check if the PR fulfills these requirements**
- [x] The PR commit message follows our [guidelines](https://github.com/talend/tools/blob/master/tools-root-github/CONTRIBUTING.md)
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)
- [ ] Related design / discussions / pages (not in jira), if any, are all linked or available in the PR

**[ ] This PR introduces a breaking change**

PR #517, opened 2017-06-08
feat(UIForm): select

jsomsanith:
**Warning: ** this is only available for the next version of forms
Arrays are not supported

Started from cherry-pick of #509 commit.
Impl Select widgets for them


packages/forms/src/UIForm/utils/properties.test.js
```diff
@@ -1,7 +1,7 @@
-import { getValue, omit } from './properties';
+import { getValue, omit, convertValue } from './properties';
 
 describe('Properties utils', () => {
-	describe('getValue', () => {
+	describe('#getValue', () => {
```
jmfrancois: why do you put # on test case name ?

packages/forms/src/UIForm/utils/properties.js
```diff
@@ -31,3 +31,15 @@ export function omit(properties, key) {
 		});
 	return result;
 }
+
+/**
+ * Convert a string value to the wanted type
+ * @param type The string type
+ * @param value The value to convert
+ */
+export function convertValue(type, value) {
```
jmfrancois: this doesn't deal with properties
utils/converValue ?

packages/forms/src/UIForm/fields/Select.js
```diff
@@ -0,0 +1,83 @@
+import React, { PropTypes } from 'react';
+import FieldTemplate from './FieldTemplate';
+
+function getSelectedOptions(select, multiple) {
+	if (multiple) {
+		return Array.from(select.options)
```
jmfrancois: note: Array.from is not supported by IE it needs the babel polyfill

packages/forms/src/UIForm/fields/Select.js
```diff
@@ -0,0 +1,83 @@
...
+			labelAfter
+		>
+			<select
+				id={id}
+				multiple={multiple}
+				autoFocus={autoFocus}
+				className="form-control"
+				disabled={disabled}
+				onChange={
+					event => onChange(event, schema, getSelectedOptions(event.target, multiple))
```
jmfrancois: I'm used to event, data
here it's event, schema, data
not sure about the order, should we either doing event, data, schema ?
it depends if the schema is mandatory or not for the usage.

packages/forms/src/UIForm/fields/Select.js
```diff
@@ -0,0 +1,83 @@
...
+		return Array.from(select.options)
+			.filter(option => option.selected)
+			.map(option => option.value);
+	}
+
+	return select.value;
+}
+
+export default function Select(props) {
+	const { id, isValid, errorMessage, onChange, schema, value } = props;
```
talend-sromain: destructuring props in component definition fonction ?

packages/forms/src/UIForm/utils/properties.test.js
```diff
@@ -55,4 +55,28 @@ describe('Properties utils', () => {
...
 		});
 	});
+
+	describe('#convertValue', () => {
+		it('should return the string value', () => {
+			// given
+			const value = '3';
+
+			// when
+			const convertedValue = convertValue('string', value);
```
talend-sromain: As number is the only type converted, i'm not sure this test is not misleading as it's now

packages/forms/src/UIForm/utils/properties.test.js
jsomsanith: So in the description we can see `Properties utils #getValue should ...` 
It's a way to mark/highlight the function that is tested

packages/forms/src/UIForm/utils/properties.js
jsomsanith: I put it here because it convert it to be stored in the properties. It's like a properties cleaning/preparation.
If you insist on putting it out after my explanation, I can do it.

packages/forms/src/UIForm/fields/Select.js
jsomsanith: Yes this is mandatory.
All the fields here will call onChange(event, schema, value). What you call payload would be `{ schema, value }`.
This can change, but we have to change it in all fields and onChange implem.
We  can do that afterward in another PR to align every field. I add the point in the [form TODO list](https://github.com/Talend/ui/issues/538#issue-236131440)

packages/forms/src/UIForm/fields/Select.js
jsomsanith: Replaced with Array.prototype.slice.call(select.options)

packages/forms/src/UIForm/utils/properties.test.js
jsomsanith: Changed it to 'unknown'

packages/forms/src/UIForm/fields/Select.js
```diff
@@ -0,0 +1,82 @@
...
+	);
+}
+
+if (process.env.NODE_ENV !== 'production') {
+	Select.propTypes = {
+		id: PropTypes.string,
+		isValid: PropTypes.bool,
+		errorMessage: PropTypes.string,
+		onChange: PropTypes.func,
+		schema: PropTypes.shape({
```
acateland: titleMap minimal element shape is not specified here

packages/forms/src/UIForm/fields/Select.js
```diff
@@ -0,0 +1,82 @@
...
+						);
+					})
+				}
+			</select>
+		</FieldTemplate>
+	);
+}
+
+if (process.env.NODE_ENV !== 'production') {
+	Select.propTypes = {
```
acateland: PropTypes not marked as required, without default value could lead in actual implementation to runtime error

PR #519, opened 2017-06-08
feat(UIForm): widget property in uiSchema instead of type

jsomsanith:
**WARNING** : this is for the new implementation that is not available yet
To stick a little more with previous api of uiSchema (and to make @frassinier and TCOMP happy), we should use `widget`property instead of `type`.

Use `widget`instead of `type` in uiSchema
Before
```
{
      "type": "fieldset",
      "title": "My awesome USER form",
      "items": []
}
```
Now
```
{
      "widget": "fieldset",
      "title": "My awesome USER form",
      "items": []
}
```


talend-glorieux:
Why did you take the fields title out of the jsonSchema?

packages/forms/stories-core/json/core-fieldset.json
```diff
@@ -60,7 +60,7 @@
       ]
     },
     {
-      "type": "fieldset",
+      "widget": "fieldset",
```
charlesnguyen: line 79:  "type": "textarea" => "widget": "textarea"

jsomsanith:
@talend-glorieux They are in the uiSchema now since this is tied to the form and not the model definition. But using jsfc, they merge the 2 and either we put them in jsonSchema or uiSchema, we will have them

packages/forms/stories-core/json/core-fieldset.json
jsomsanith: Changed

PR #522, opened 2017-06-08
fix(forms): Unrestricted datalist handles arbitrary input

talend-glorieux:
The DataList widget doesn't set the value in case it's not on the enum or remote list.
This renders the formData null.

The DataList is now calling the onChange callback in case the value isn't on the enum or remote list and the field isn't restricted.


packages/forms/src/widgets/DatalistWidget/DatalistWidget.test.js
```diff
@@ -185,4 +185,26 @@ describe('DatalistWidget', () => {
...
+		);
+		const input = wrapper.find('input').at(0);
+		input.simulate('change', { target: { value: 'unknown' } });
+
+		// when
+		input.simulate('blur');
+
+		// then
+		expect(onChange).toBeCalled();
+		expect(toJson(wrapper)).toMatchSnapshot();
```
jmfrancois: remove unecessary snapshots here.
if you have things to check use wrapper.find(AutoWhatever).props() ...

packages/forms/src/widgets/DatalistWidget/DatalistWidget.test.js
```diff
@@ -185,4 +185,26 @@ describe('DatalistWidget', () => {
...
+			/>
+		);
+		const input = wrapper.find('input').at(0);
+		input.simulate('change', { target: { value: 'unknown' } });
+
+		// when
+		input.simulate('blur');
+
+		// then
+		expect(onChange).toBeCalled();
```
jsomsanith: Can you ensure that it is called with the expected value ? `toBeCalledWith('unknown')`

PR #529, opened 2017-06-13
fix(theme): Fixes forms style according to guidelines

talend-glorieux:
The forms style aren't following the guidelines.

Fix the forms style according to the guidelines.


talend-glorieux:
I'm not convinced by the lines colors.
Please tell me what you think.

talend-glorieux:
I've updated the forms styles.
The only thing missing is the lines colors transitions on input focus. This is not currently feasible with the forms library. I'll look into ways to implement it into the new forms library implementation.

PR #531, opened 2017-06-14
fix(Container/HomeListView): crash when there was no drawer

jsomsanith:
HomeListView container crash when there is no drawer.
The problem is that when the component has no children, `props.children` is undefined. But a piece of code in the container test that `props.children === null`, which resulted with `drawers = [undefined]`.
The component then tries to access to `drawers[i].props` --> boom

Change the drawers adaptation in container.
Bonus : fix warning and clean the HomeListView story


packages/containers/examples/ExampleHomeListView.js
```diff
@@ -184,11 +184,17 @@ const ExampleHomeListView = {
...
 				list={listProps}
 			>
-				<Drawer stacked title="Im stacked drawer 1">
+				<Drawer
+					stacked title="Im stacked drawer 1"
+					footerActions={Object.assign({}, basicProps, { selected: 0 })}
+				>
 					<h1>Hello drawer 1</h1>
-					<p>You should not being able to read this because I'm first</p>
+					<p>You should not being able to read this because I&#39;m first</p>
```
frassinier: Why did you escape it?

jsomsanith: linter on IDE
capture d ecran de 2017-06-14 17-51-19

PR #532, opened 2017-06-14
chore(Component/VirtualizedList) : improve CellActions styling

jsomsanith:
VirtualizedList only : Custom css on cell actions on Table and Large displays
capture d ecran de 2017-06-14 11-55-39
capture d ecran de 2017-06-14 11-55-27

Rollback to classical button on Table display
Set custom style on Large display to fit the layout but more icon-centered
capture d ecran de 2017-06-14 11-56-56
capture d ecran de 2017-06-14 11-57-08


packages/components/src/VirtualizedList/CellActions/CellActions.scss
```diff
@@ -1,17 +1,32 @@
...
-
 :global(.tc-list-large-row) .tc-list-actions :global(.tc-actions) > button,
 :global(.ReactVirtualized__Table__row) .tc-list-actions :global(.tc-actions) > button:focus,
 :global(.ReactVirtualized__Table__row):hover .tc-list-actions :global(.tc-actions) > button {
 	opacity: 1;
 	transition: opacity 0.15s ease-in;
 }
+
+// Large display specificities
+:global(.tc-list-large-row) .tc-list-actions :global(.tc-actions) > button {
```
jmfrancois: I don't like selector which start with global, this break the css module protection

jsomsanith: In fact we don't have a choice because I want to style the cell within the cell's css.
This selector is to scope to a specific display.

Why : 
* an add/modification on cell won't require to touch to other parts
* the style stays just next to the cell that is styled, resier to search (same concept as the tests)

packages/components/src/VirtualizedList/CellActions/CellActions.scss
```diff
@@ -1,17 +1,32 @@
...
+	height: 20px;
```
frassinier: Where does it come from?

packages/components/src/VirtualizedList/CellActions/CellActions.scss
frassinier: I agree with @jmfrancois, this is dangerous, we have to add more classes to our produced HTML to avoid this kind of selectors.

Why did you set it here and not in list style?

jsomsanith:
After the 2 reviews, I reworked the thing.

To keep the original idea to have the cell style int the cell folder. Now we have 1 scss file per RowRenderer that we want to customize. Those styles are injected in the cell style (css module).

With that, we have theme aggregator for Table and Large. Those theme aggregator import the `.row` class from the scss and gather them in an array. This array is injected in the Table or Large renderer (css modules).

This allows to have the classname from css modules on the row and on the cell. This avoids most of the `:global()`

packages/components/src/VirtualizedList/ListTable/ListTable.component.js
```diff
@@ -41,7 +43,7 @@ function ListTable(props) {
 			headerHeight={35}
 			height={height}
 			id={id}
-			rowClassName={theme.row}
+			rowClassName={rowThemes}
```
frassinier: Please use classnames here

packages/components/src/VirtualizedList/CellActions/CellActions.component.js
```diff
@@ -1,18 +1,20 @@
 import React, { PropTypes } from 'react';
-import classnames from 'classnames';
```
frassinier: What's the problem with `classnames`?

packages/components/src/VirtualizedList/RowLarge/RowLarge.component.js
```diff
@@ -32,7 +34,7 @@ function RowLarge({ className, index, key, parent, style }) {
 
 	return (
 		<div
-			className={`tc-list-large ${theme['tc-list-large']}`}
+			className={classNames('tc-list-large', rowThemes)}
```
frassinier: You could factorize and put `tc-list-large` const in RowTheme.js

packages/components/src/VirtualizedList/CellActions/CellActions.component.js
```diff
@@ -1,18 +1,20 @@
...
  */
 function CellActions({ cellData }) {
 	return (
 		<div
-			className={classnames(
+			className={classNames(
 				'tc-list-actions',
-				theme['tc-list-actions'],
+				tableTheme['tc-list-actions'],
+				largeTheme['tc-list-actions'],
```
frassinier: You could factorize and put `tc-list-actions`value into const

packages/components/src/VirtualizedList/CellActions/CellActions.component.js
jsomsanith: Nothing, just the IDE that points me out that there is typo in the word `classnames` :)

PR #535, opened 2017-06-14
feat(Component/List): map List sort to VirtualizedList

jsomsanith:
Sort where not supported by ListToVistualizedList, so no Table header sort.

Adaptation done


packages/components/src/List/ListToVirtualizedList/ListToVirtualizedList.component.js
```diff
@@ -1,10 +1,23 @@
...
 import CellTitle from '../../VirtualizedList/CellTitle';
 import CellActions from '../../VirtualizedList/CellActions';
 
+function adaptOnSort(onChange) {
+	if (!onChange) {
+		return null;
+	}
+	return function onSortChange({ sortBy, sortDirection }) {
+		return onChange(
+			null,
```
jmfrancois: question: no way to get the event here 🙏 ?

jmfrancois: please down the event here 🙏

jsomsanith: :/ we don't have it. It's react-virtualized sort on click callback that provides only those infos

PR #541, opened 2017-06-15
fix(forms/datalist): do not trigger onchange if value does not change

frassinier:
Datalist trigger onchange even if value does not change.

Check value before trigger onchange

### DIFF
--- a/packages/forms/src/widgets/DatalistWidget/DatalistWidget.js
+++ b/packages/forms/src/widgets/DatalistWidget/DatalistWidget.js
@@ -36,7 +36,7 @@ function renderDatalistInput(props) {
 		<div className={theme['typeahead-input-icon']}>
 			<FormControl {...props} />
 			<div className={theme['dropdown-toggle']}>
-				<span className="caret"/>
+				<span className="caret" />
 			</div>
 		</div>);
 }
--- a/packages/forms/src/widgets/DatalistWidget/DatalistWidget.js
+++ b/packages/forms/src/widgets/DatalistWidget/DatalistWidget.js
@@ -36,7 +36,7 @@ function renderDatalistInput(props) {
 		<div className={theme['typeahead-input-icon']}>
 			<FormControl {...props} />
 			<div className={theme['dropdown-toggle']}>
-				<span className="caret" />
+				<span className="caret"/>
 			</div>
 		</div>);
 }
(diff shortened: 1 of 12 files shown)


packages/forms/src/widgets/DatalistWidget/DatalistWidget.js
```diff
@@ -151,12 +151,16 @@ class DatalistWidget extends React.Component {
...
+	onBlur(event) {
 		if (this.props.options &&
-			this.props.options.restricted && this.state.initalItems.indexOf(this.state.value) === -1) {
+			this.props.options.restricted &&
+			this.state.initalItems.indexOf(this.state.value) === -1) {
 			this.resetValue();
 		} else {
-			this.props.onChange(this.state.value);
+			const { value } = event.target;
+			if (value !== this.state.value) {
```
ncomont: This assignation seems usesless assignation. Why didn't you do that like that : `if (event.target.value !== this.state.value)` ?

packages/forms/src/widgets/DatalistWidget/DatalistWidget.js
```diff
@@ -151,12 +151,16 @@ class DatalistWidget extends React.Component {
 		};
 	}
 
-	onBlur() {
+	onBlur(event) {
 		if (this.props.options &&
-			this.props.options.restricted && this.state.initalItems.indexOf(this.state.value) === -1) {
+			this.props.options.restricted &&
```
ncomont: Accessing nested properties has a performance cost. I think it could be better like that : 

```
const { options } = this.props;
[...]
if(options && options.restricted
[...]
```

packages/forms/src/widgets/DatalistWidget/DatalistWidget.js
```diff
@@ -151,12 +151,16 @@ class DatalistWidget extends React.Component {
...
+			this.state.initalItems.indexOf(this.state.value) === -1) {
```
ncomont: `Array.prototype.includes()` can be used here :

Instead of :
`this.state.initalItems.indexOf(this.state.value) === -1`

You can write :
`!this.state.initalItems.includes(this.state.value)`

ncomont:
Can you please add a unit test for this feature ?

packages/forms/src/widgets/DatalistWidget/DatalistWidget.js
frassinier: Right, I will do it :+1:

packages/forms/src/widgets/DatalistWidget/DatalistWidget.js
frassinier: Yes, I will change it.

packages/forms/src/widgets/DatalistWidget/DatalistWidget.js
frassinier: Yes

PR #542, opened 2017-06-15
feat(forms): Adds column support to fields

talend-glorieux:
We can't set two fields in columns. Only JsonSchema objects.

Add column support to fields.


packages/forms/src/widgets/ColumnsWidget/__snapshots__/ColumnsWidget.test.js.snap
```diff
@@ -11,6 +11,10 @@ ShallowWrapper {
   "node": <div
     className="tf-widget-columns undefined"
 >
+    <TitleField
```
jsomsanith: Can you snapshot `wrapper.node` instead ? The snapshot would be less big and more readable

packages/forms/src/widgets/ColumnsWidget/ColumnsWidget.scss
```diff
@@ -1,11 +1,16 @@
...
+	flex-wrap: wrap;
 
 	.column {
 		flex: 1;
 
 		fieldset {
 			margin-right: auto;
+
+		}
+		input {
```
jsomsanith: Can you fix the indent ?

packages/forms/src/widgets/ColumnsWidget/ColumnsWidget.scss
```diff
@@ -1,11 +1,16 @@
...
+			width: 20rem;
```
jsomsanith: can you set the min-widget instead ?

PR #543, opened 2017-06-16
fix(components/dialog): backport ConfirmDialog enhancement

acateland:
see https://github.com/Talend/ui/pull/484
see https://github.com/Talend/ui/pull/484

All specific css for this feature put into ConfirmDialog css module was ported to the bootstrap theme and both ConfirmDialog and Dialog use it


output/theme.sasslint.txt
```diff
@@ -23,13 +23,22 @@ src/theme/_forms.scss
   316:9   error    Qualifying elements are not allowed for class selectors  no-qualifying-elements
   457:8   warning  Vendor prefixes should not be used                       no-vendor-prefixes
 
+src/theme/_modals.scss
+  10:3  warning  Expected indentation of 1 tab but found 2   indentation
```
talend-sromain: Some lint here

packages/theme/src/theme/_modals.scss
```diff
@@ -5,12 +5,42 @@
 	flex-direction: column;
 }
 
+/** IE-FIX */
+.modal-dialog {
```
jsomsanith: Can you merge with the previous class ? Those 2 blocks are for `.modal-dialog` and the second one overrides some properties of the first one.

packages/theme/src/theme/_modals.scss
```diff
@@ -5,12 +5,42 @@
...
+	display: flex;
+	flex-direction: row;
+	flex-wrap: nowrap;
+
+	>.modal-content {
```
jsomsanith: This is supposed to style the same elements as the `.modal-content` below. Can you merge them ?

packages/theme/src/theme/_modals.scss
```diff
@@ -25,10 +55,17 @@
 	}
 
 	&-body {
+		flex-basis: auto;
 		flex-grow: 1;
+		flex-shrink: 1;
+		display: flex;
+		flex-direction: column;
+		overflow: hidden;
```
jsomsanith: I don't understand why we have those 3 properties above. Why do we style the content that can be whatever we want ?

PR #544, opened 2017-06-16
feat(UIForm): FieldTemplate to gather common field markup

jsomsanith:
**WARNING** : this is on next version of forms only
Currently, fields have to add the form-group, label, and message markup themself. There all have a lot in common, this implies some markup duplications upon fields.

Create a FieldTemplate to add all those extra infos

When implementing a new UI widget you should wrap it inside a FieldTemplate.component
example : https://github.com/Talend/ui/pull/544/files#diff-27eca92122e6896405f8b853923b92bc

*Still need to do*
If #517 is merged before this one : use FieldTemplate in Select before merging this one.
If #547 is merged before this one : use FieldTemplate in TextArea before merging this one.

### DIFF
--- a/packages/forms/src/UIForm/fields/FieldTemplate/FieldTemplate.component.js
+++ b/packages/forms/src/UIForm/fields/FieldTemplate/FieldTemplate.component.js
@@ -1,0 +1,44 @@
+import React, { PropTypes } from 'react';
+import classNames from 'classnames';
+
+import Message from '../../Message';
+
+function FieldTemplate(props) {
+	const groupsClassNames = classNames(
+		'form-group',
+		{ 'has-error': !props.isValid },
+	);
+
+	return (
+		<div className={groupsClassNames}>
+			{
+				props.labelBefore &&
+				<label htmlFor={props.id} className="control-label">{props.labelBefore}</label>
+			}
+			{props.children}
+			{
+				props.labelAfter &&
+				<label htmlFor={props.id} className="control-label">{props.labelAfter}</label>
+			}
+			<Message
+				errorMessage={props.errorMessage}
+				description={props.description}
+				isValid={props.isValid}
+			/>
+		</div>
+	);
+}
+
+if (process.env.NODE_ENV !== 'production') {
+	FieldTemplate.propTypes = {
+		children: PropTypes.element,
+		description: PropTypes.element,
+		errorMessage: PropTypes.string,
+		id: PropTypes.string,
+		isValid: PropTypes.bool,
+		labelAfter: PropTypes.string,
+		labelBefore: PropTypes.string,
+	};
+}
+
+export default FieldTemplate;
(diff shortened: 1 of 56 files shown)


packages/forms/src/UIForm/fields/FieldTemplate/FieldTemplate.component.js
```diff
@@ -0,0 +1,44 @@
...
+	return (
+		<div className={groupsClassNames}>
+			{
+				props.labelBefore &&
+				<label htmlFor={props.id} className="control-label">{props.labelBefore}</label>
+			}
+			{props.children}
+			{
+				props.labelAfter &&
+				<label htmlFor={props.id} className="control-label">{props.labelAfter}</label>
```
frassinier: This is repeated, you should use another Label render fn?

packages/forms/src/UIForm/fields/FieldTemplate/FieldTemplate.component.js
```diff
@@ -0,0 +1,44 @@
...
+			}
+			<Message
```
acateland: Im not sure since Message handle Description or Error that its wise to use Message for description, since on Error it may hide some helpfull info for end user

packages/forms/src/UIForm/fields/FieldTemplate/FieldTemplate.component.js
```diff
@@ -0,0 +1,44 @@
...
+}
+
+if (process.env.NODE_ENV !== 'production') {
+	FieldTemplate.propTypes = {
+		children: PropTypes.element,
+		description: PropTypes.element,
+		errorMessage: PropTypes.string,
+		id: PropTypes.string,
+		isValid: PropTypes.bool,
+		labelAfter: PropTypes.string,
```
acateland: is there an use case where we want to display a label both before and after ?

jsomsanith: No in practice it's only one or the other. Hmm are you telling me that this is possible to precise via PropTypes ?

packages/forms/src/UIForm/fields/FieldTemplate/FieldTemplate.component.js
jsomsanith: It's out of the PR scope. I'll add the point in the forms TODO list to see with the UX and implement the wanted behavior.

packages/forms/src/UIForm/fields/FieldTemplate/FieldTemplate.component.js
jsomsanith: Done thanks

packages/forms/src/UIForm/fields/FieldTemplate/FieldTemplate.component.js
acateland: i would prefer as props
label
labelAfter = false

jsomsanith: :+1: done

PR #547, opened 2017-06-19
feat(UIForm): textarea

jsomsanith:
**Warning: ** this is only available for the next version of forms
TextArea are not supported

Impl TextArea widgets


packages/forms/src/UIForm/fields/TextArea.js
```diff
@@ -0,0 +1,54 @@
+import React, { PropTypes } from 'react';
+import FieldTemplate from './FieldTemplate';
+
+export default function TextArea(props) {
+	const { id, isValid, errorMessage, onChange, schema, value } = props;
```
talend-sromain: Move the spread to the function call ?

packages/forms/src/UIForm/fields/TextArea.js
```diff
@@ -0,0 +1,54 @@
...
+			readOnly: PropTypes.bool,
+			title: PropTypes.string,
+			type: PropTypes.string,
+		}),
+		value: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
+	};
+}
+TextArea.defaultProps = {
+	isValid: true,
+	value: '',
```
frassinier: Can we have default for `rows` attr?

packages/forms/src/UIForm/fields/TextArea.test.js
```diff
@@ -0,0 +1,99 @@
...
+describe('Text field', () => {
+	const schema = {
+		autoFocus: true,
+		description: 'my text input hint',
+		placeholder: 'Type something here',
+		title: 'My input title',
+		type: 'text',
+	};
+
+	it('should render input', () => {
```
frassinier: ~input~ should be textarea

packages/forms/src/UIForm/fields/TextArea.test.js
```diff
@@ -0,0 +1,99 @@
...
+				value={'toto'}
+			/>
+		);
+		const event = { target: { value: 'totoa' } };
+
+		// when
+		wrapper.find('textarea').simulate('change', event);
+
+		// then
+		expect(onChange).toBeCalledWith(event, schema, 'totoa');
```
frassinier: DRY, use const for that dummy value

packages/forms/src/UIForm/fields/__snapshots__/TextArea.test.js.snap
```diff
@@ -0,0 +1,73 @@
...
+  <textarea
+    autoFocus={true}
+    className="form-control"
+    disabled={true}
+    id="myForm"
+    label="My input title"
+    onChange={[Function]}
+    placeholder="Type something here"
+    readOnly={undefined}
+    type="text"
```
frassinier: Where does the type come from? It's an error :/

packages/forms/src/UIForm/fields/TextArea.js
```diff
@@ -0,0 +1,54 @@
...
+		errorMessage: PropTypes.string,
+		onChange: PropTypes.func,
+		schema: PropTypes.shape({
+			autoFocus: PropTypes.bool,
+			description: PropTypes.string,
+			disabled: PropTypes.bool,
+			placeholder: PropTypes.string,
+			readOnly: PropTypes.bool,
+			title: PropTypes.string,
+			type: PropTypes.string,
```
frassinier: `type` should not be here

packages/forms/src/UIForm/fields/TextArea.js
```diff
@@ -0,0 +1,54 @@
...
+			<textarea
+				id={id}
+				autoFocus={autoFocus}
+				className="form-control"
+				disabled={disabled}
+				label={title}
+				placeholder={placeholder}
+				onChange={event => onChange(event, schema, event.target.value)}
+				readOnly={readOnly}
+				type={type}
```
frassinier: Textarea shouldn't have `type` attr

packages/forms/src/UIForm/fields/TextArea.js
```diff
@@ -0,0 +1,54 @@
...
+			isValid={isValid}
+			label={title}
+			labelAfter
+		>
+			<textarea
+				id={id}
+				autoFocus={autoFocus}
+				className="form-control"
+				disabled={disabled}
+				label={title}
```
frassinier: What? Please remove this! In other hand, we need a `name` instead!

packages/forms/src/UIForm/fields/TextArea.js
```diff
@@ -0,0 +1,54 @@
...
+				placeholder={placeholder}
+				onChange={event => onChange(event, schema, event.target.value)}
+				readOnly={readOnly}
+				type={type}
+				value={value}
+			/>
```
frassinier: There is no way to render it required with HTML5 validation?

packages/forms/src/UIForm/fields/TextArea.js
```diff
@@ -0,0 +1,54 @@
...
+			labelAfter
+		>
+			<textarea
+				id={id}
+				autoFocus={autoFocus}
+				className="form-control"
+				disabled={disabled}
+				label={title}
+				placeholder={placeholder}
+				onChange={event => onChange(event, schema, event.target.value)}
```
frassinier: `onchange` looks like overkill, what do you think about `onblur`?

jsomsanith: It's a controlled textarea. So we need to trigger the onChange to update the value in state and inject it back as props.value.
What could be overkill is the validation on each change. But this is another subject that I need to talk about with UX?

packages/forms/src/UIForm/fields/TextArea.js
jsomsanith: Sorry for that :(

packages/forms/src/UIForm/fields/TextArea.js
jsomsanith: Sorry for that, wrong copy/paste

packages/forms/src/UIForm/fields/TextArea.js
jsomsanith: The idea is to validate it through the same validation process as other validation

packages/forms/src/UIForm/fields/__snapshots__/TextArea.test.js.snap
jsomsanith: It's the type in jsfc merged schema that transform the string to text
But this will go after the `type` removal in the textarea

packages/forms/src/UIForm/fields/TextArea.js
jsomsanith: done

PR #549, opened 2017-06-19
fix(component/list/TFD-1033): keep columns with empty list

jmfrancois:

keeping the display of columns even if no items to display

move the message inside the rendering of Table display

### DIFF
--- a/packages/components/src/List/DisplayTable/DisplayTable.component.js
+++ b/packages/components/src/List/DisplayTable/DisplayTable.component.js
@@ -306,7 +306,7 @@ function DisplayTable(props) {
 	);
 	let message = null;
 	if (!items || items.length === 0) {
-		message = (<span className={theme['no-result']}>No result found</span>);
+		message = <span className={theme['no-result']}>No result found</span>;
 	}
 
 	return (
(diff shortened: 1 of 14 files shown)


packages/components/src/List/DisplayTable/DisplayTable.component.js
```diff
@@ -304,6 +304,11 @@ function DisplayTable(props) {
 		'tc-list-display-table',
 		theme.table,
 	);
+	let message = null;
+	if (!items || items.length === 0) {
```
ncomont: `items.length === 0` is useless here.

packages/components/src/List/DisplayTable/DisplayTable.component.js
```diff
@@ -304,6 +304,11 @@ function DisplayTable(props) {
...
+		message = (<span className={theme['no-result']}>No result found</span>);
```
ncomont: Parenthesis are not needed for single line JSX.

packages/components/src/List/DisplayTable/DisplayTable.component.js
jmfrancois:
```
> !![]
true
> !![].length
false
```
oh you mean === 0 is useless , thanks

jmfrancois:
```![]
false
[].length === 0 
true```

ncomont: Oh, sorry, I had read `!items.length || items.length === 0` :)

PR #550, opened 2017-06-19
chore(components/VirtualizedList): increase dataset in the stories

jare-talend:

increase the dataset to see in the stories the behavior when we scroll the table

### DIFF
--- a/packages/components/stories/VirtualizedList.js
+++ b/packages/components/stories/VirtualizedList.js
@@ -7,8 +7,6 @@ import VirtualizedList, { listTypes } from '../src/VirtualizedList';
 import CellActions from '../src/VirtualizedList/CellActions';
 import CellTitle from '../src/VirtualizedList/CellTitle';
 
-const NUMBER_ELEMENTS_HUGE_DATASET = 100;
-
 const icons = {
 	'talend-badge': talendIcons['talend-badge'],
 	'talend-caret-down': talendIcons['talend-caret-down'],
@@ -81,7 +79,7 @@ const actions = [
 	},
 ];
 
-let collection = [
+const collection = [
 	{
 		id: 0,
 		name: 'Title with icon and actions',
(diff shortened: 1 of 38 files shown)


packages/components/stories/VirtualizedList.js
```diff
@@ -170,6 +172,24 @@ const collection = [
...
+	description: 'Simple row with icon and actions',
+	author: 'Jean-Pierre DUPONT',
+	icon: 'talend-file-xls-o',
+	display: 'text',
+	className: 'item-0-class',
+	actions,
+	titleActions,
+}));
+
+collection = collection.concat(hugeDataSet);
```
ncomont: Do you really still need `collection` ?

packages/components/stories/VirtualizedList.js
```diff
@@ -7,6 +7,8 @@ import VirtualizedList, { listTypes } from '../src/VirtualizedList';
 import CellActions from '../src/VirtualizedList/CellActions';
 import CellTitle from '../src/VirtualizedList/CellTitle';
 
+const NUMBER_ELEMENTS_HUGE_DATASET = 100;
```
jsomsanith: Constant that doesn't need to be a constant

packages/components/stories/VirtualizedList.js
```diff
@@ -170,6 +172,24 @@ const collection = [
 	},
 ];
 
+let hugeDataSet = Array.from(new Array(NUMBER_ELEMENTS_HUGE_DATASET));
```
jsomsanith: Why not just new Array()

packages/components/stories/VirtualizedList.js
```diff
@@ -79,7 +81,7 @@ const actions = [
 	},
 ];
 
-const collection = [
+let collection = [
```
jsomsanith: This can stay a const

packages/components/stories/VirtualizedList.js
```diff
@@ -170,6 +172,24 @@ const collection = [
 	},
 ];
 
+let hugeDataSet = Array.from(new Array(NUMBER_ELEMENTS_HUGE_DATASET));
+
+hugeDataSet = hugeDataSet.map((item, index) => ({
```
jsomsanith: You allocate an array of 100 elements just to use `.map()` instead of a for loop

packages/components/stories/VirtualizedList.js
jare-talend: new Array(100).map(() => 10) do nothing. We need to pass by Array.from

packages/components/stories/VirtualizedList.js
jare-talend: I already use a map

packages/components/stories/VirtualizedList.js
jare-talend: I reallocate collection with collection.concat(randomArray). I can't use const

packages/components/stories/VirtualizedList.js
jsomsanith: Yes ... that's the problem --> `You allocate an array of 100 elements just to use .map() instead of a for loop`

packages/components/stories/VirtualizedList.js
jsomsanith: This was a comment to go with my solution

PR #552, opened 2017-06-20
fix(components/typeahead): don't fail on undefined text

ncomont:
If the title or the description is undefined: `Cannot read property 'split' of undefined`

`emphasise()` returns an empty string if trying to highlight an undefined string.

### DIFF
--- a/packages/components/src/Typeahead/Typeahead.component.renderers.js
+++ b/packages/components/src/Typeahead/Typeahead.component.renderers.js
@@ -131,11 +131,11 @@ export const renderSectionTitle = (section) => {
 };
 
 function emphasise(text, value) {
-	if (!value) {
-		return [text];
-	}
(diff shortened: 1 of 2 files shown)


packages/components/src/Typeahead/Typeahead.component.renderers.js
```diff
@@ -134,6 +134,9 @@ function emphasise(text, value) {
 	if (!value) {
 		return [text];
 	}
+	if (!text) {
```
jsomsanith: Can you put that before the test on `value` ?
Or else if there is no text and no value, it will return `[undefined]`. This is well supported in our use but much cleaner if it returns empty string.

ncomont: Oops, sorry, fixed

PR #553, opened 2017-06-20
chore(Components/Containers/Cmf): Open in new tab

Sylvain-Talend:
Now, for CMF Apps, the middle-click (mousewheel) and CTRL-Click (Command key for MAC) is not opening a new tab when route changes.

Manage ctrl-click and middle-click in CMF routing when route change. Now, for whatever route you define, if you do a ctrl-click or a middle-click, it will open the route in a new tab.

No but major changes

:warning: When doing  a middle-click, Firefox will under particular circumstances trigger `onClick` event after `onMouseDown` (see comments below to have example of what happens)


packages/containers/src/actionAPI.js
```diff
@@ -20,8 +20,8 @@ export function getActionsProps(context, ids, model) {
 				);
 			} else {
 				context.store.dispatch(Object.assign({
-					model,
-				}, info.payload));
+					model
+				}, info.payload, { openInNewTab: event.ctrlKey || event.button === 1 }));
```
jmfrancois: why not `{ event: { ctrlKey: event.ctrlKey, button: event.button } }` ?
so the logic is on the onClick

Sylvain-Talend: :+1: Done

packages/cmf/__tests__/middlewares/cmf.test.js
```diff
@@ -82,6 +91,9 @@ describe('CMF middleware', () => {
 				},
 			},
 			response: { id: 28 },
+			event: {
```
frassinier: DRY, use a constant for this event on top of the file with a consistent name

packages/cmf/__tests__/middlewares/cmf.test.js
```diff
@@ -90,4 +102,38 @@ describe('CMF middleware', () => {
...
+					return `/route/${data.response.id}`;
+				},
+			},
+			response: { id: 28 },
+			event: {
+				button: 1,
+			},
+		};
+		const res = middleware(action);
+		console.log(res)
```
frassinier: Remove this console.log statement

packages/cmf/__tests__/middlewares/cmf.test.js
```diff
@@ -90,4 +102,38 @@ describe('CMF middleware', () => {
...
+			},
+		};
+		middleware(action);
+		expect(global.open).toHaveBeenCalled();
+
+	});
+	it('should open route in new tab when middle-click (mousewheel)', () => {
+		global.open = jest.fn();
+		const action = {
+			cmf: {
```
frassinier: DRY, use same action

packages/cmf/__tests__/middlewares/cmf.test.js
```diff
@@ -90,4 +102,38 @@ describe('CMF middleware', () => {
...
+			event: {
+				ctrlKey: true,
+			},
+		};
+		middleware(action);
+		expect(global.open).toHaveBeenCalled();
+
+	});
+	it('should open route in new tab when middle-click (mousewheel)', () => {
+		global.open = jest.fn();
```
frassinier: DRY, use beforeEach to define spy

packages/cmf/src/middlewares/cmf/index.js
```diff
@@ -17,13 +17,18 @@ const cmfMiddleware = store => next => (action) => {
...
 		}
-		store.dispatch({
-			type: '@@router/CALL_HISTORY_METHOD',
-			payload: {
-				method: config.routerPush ? 'push' : 'replace',
-				args: [route],
-			},
-		});
+
+		if (action.event && (action.event.ctrlKey || action.event.button === 1)) {
```
frassinier: You need to manage Mac users by dealing with meta key. I recommend you to use
```
const { event } = action;
if (event && ((event.button === 0 && (event.ctrlKey || event.metaKey)) || event.button === 1)) {
```

packages/cmf/__tests__/middlewares/cmf.test.js
Sylvain-Talend: :+1: done

packages/cmf/__tests__/middlewares/cmf.test.js
Sylvain-Talend: :+1: done

packages/cmf/__tests__/middlewares/cmf.test.js
Sylvain-Talend: :+1: done

packages/cmf/__tests__/middlewares/cmf.test.js
Sylvain-Talend: :+1: done

packages/cmf/src/middlewares/cmf/index.js
Sylvain-Talend: :+1: done

packages/components/src/Actions/Action/Action.component.js
```diff
@@ -90,14 +90,29 @@ function Action(props) {
...
+				model,
+			});
+		}
+	});
 
-	const rMouseDown = event => onMouseDown(event, {
-		action: { label, ...rest },
-		model,
+	const rMouseDown = ((event) => {
+		// middle-click: call onClick event
```
frassinier: If you deal with bug please add link to source

packages/components/src/Actions/Action/Action.component.js
```diff
@@ -90,14 +90,29 @@ function Action(props) {
 
 	const buttonProps = getPropsFrom(Button, rest);
 	const style = link ? 'link' : bsStyle;
-	const rClick = onClick && (event => onClick(event, {
-		action: { label, ...rest },
-		model,
-	}));
+	const rClick = onClick && ((event) => {
+		// to prevent odd bug with FF
```
frassinier: If you deal with bug please add link to source

Sylvain-Talend: Yes to better understand what happen on Chrome and FF see here https://jsfiddle.net/5zbvw0g8/10/
In the console, when you you middle-click : 
- FF will trigger mouseDown AND mouseClick
- Chrome will trigger only mouseClick
Strangely, it behaves the same with native JS https://jsfiddle.net/sylvainmaes/pfmnz1tf/

packages/components/src/Actions/Action/Action.component.js
Sylvain-Talend: Please see above

Sylvain-Talend:
I did the necessary reworks  and documented in a Jsfiddle what happens in FF and Chrome (https://jsfiddle.net/sylvainmaes/5zbvw0g8/), @frassinier if it's fine for you, could you approve?

Sylvain-Talend:
I put the PR in WIP because in fact it is a workaround for using `<button>` instead of `<a> `for links.  With `<a>`, the behavior is supported natively, and it would be ideal.

jsomsanith:
This PR has not evolved since a while, which means that this is not needed yet.
The way to do it should be reconsider when the need comes

jmfrancois:
I think this was the best way to a partial support of the standard.
Which is rendering anchor.
We need to be able to render anchor in react and make them work well in angular ...

One idea is `<Action href="/foo" />` to render `<a href="/foo" />` (only if href props exists)
So for example in full rect APP I could add this props for actions in the SidePanel

PR #559, opened 2017-06-21
feat(virtualized): POC improve scroll performance

jare-talend:

After some research to try to improve the performance on the list on IE11, I found some recommandations on the scroll:

- Avoid the hover during the scroll. react-virtualized do it with adding the css pointer-events: none during the scroll by removing all event on the container

- Avoid many operations on the DOM during the scroll. During the scroll, react-virtualized removes or adds some element in the viewport. All react components created to each scroll, if the component is high, this can take many times and slows down the frame-rate. The scroll can be jerky.

Create a skeleton during the scroll. This is a prototype, it misses the style on the skeleton.

On Firefox:
Before:
ff_scroll
After:
ff_skeleton

Unfortunately, on IE11, i can see  a improvement.


jsomsanith:
I'm not a big fan of that in a user point of vue. It introduce a delay before I can read the items.
Let's see what the others think before I do a proper review

packages/components/src/VirtualizedList/Row/Row.component.js
```diff
@@ -0,0 +1,51 @@
...
+ */
+function getRowRenderer({ selectionToggle, isSelected }) {
+	const RowTableRender = selectionToggle ?
+		RowSelectionRenderer( // eslint-disable-line new-cap
+			DefaultTableRowRenderer,
+			{
+				isSelected,
+				getRowData: rowProps => rowProps.rowData,
+			}) :
+		DefaultTableRowRenderer;
```
talend-glorieux: Could you please use a function instead of a multi-line ternary?

packages/components/src/VirtualizedList/Row/Row.component.js
```diff
@@ -0,0 +1,51 @@
...
+
+	function Row(props) {
+		const { isScrolling } = props;
+
+		if (isScrolling) {
```
talend-glorieux: Could you please use `props.isScrolling` instead of spreading the props?
That allocation is useless.

packages/components/src/VirtualizedList/Row/Row.component.js
```diff
@@ -0,0 +1,51 @@
...
+			// improve the performance on the browsers
+			// waiting guidelines from UX
+			return (
+				<div
+					{...props}
+				>
+					<div
+						className={theme['row-placeholder']}
+						style={{
+							width: `${random(15, 50)}%`,
```
talend-glorieux: Why do you need a random on the width?

packages/components/src/VirtualizedList/Row/RowPlaceholder.scss
```diff
@@ -0,0 +1,6 @@
+.row-placeholder {
+	background-color: #DDD;
```
talend-glorieux: Please use one of the themes colors.

jsomsanith:
After the discussion on Slack, there are multiple points to change in the behavior : 
* have the skeleton match the columns, not only 1 block
* have the skeleton on new rendered element, not all because we loose info on scrolling
* consider a scroll velocity to show the skeleton only if we scroll fast

For the implem, I haven't done a proper review but it would be better to create a Higher Order Component (HOC) and compose them instead of defining a component that holds our selection HOC composition logic.

jsomsanith:
lazy

jare-talend:
@jsomsanith the issue that you have referenced, lag after the scroll, is a know issue on react-virtualized: https://github.com/bvaughn/react-virtualized/issues/722. I'm waiting a fix on this.

PR #565, opened 2017-06-22
fix(Virtualized): fix display on IE on the display large

jare-talend:
fix display on IE
add min-height to 0 on the content. By default, when display:flex is enable, min-height set to auto instead 0. With column-count, it tries to fill all the height before to fill the other columns.


jare-talend:
improve on Edge, not on IE11

packages/components/src/VirtualizedList/RowLarge/RowLarge.scss
```diff
@@ -31,7 +31,8 @@ $tc-list-large-field-value-color: $dove-gray !default;
 
 	.content {
 		column-count: 3;
-		flex-grow: 1;
+		flex: 1;
+		min-height: 0; // need on Edge
```
talend-glorieux: Please remove comment.

PR #567, opened 2017-06-23
feat(cmf): http middleware support action.cmf.http structure

jmfrancois:

if you are doing an http request the TYPE must be in 'GET', 'POST', ...
you can't provide custom type.
We want to make our type more explicit (like 'MyComponent.didMount' or 'MyComponent.click.addButton', ...) so the logs are more readable

add support for the structure in the middleware


packages/cmf/src/middlewares/http/middleware.js
```diff
@@ -122,40 +122,47 @@ export const httpMiddleware = ({ dispatch }) => next => (action) => {
...
+		const errorObject = {
+			name: error.name,
+			message: error.description || error.message,
+			number: error.number,
+			stack: error.stack,
+		};
+		const clone = get(error, 'stack.response.clone');
+		if (!clone) {
+			dispatch({
+				type: 'HTTP_REDUCE_ERROR',
```
jlame: what is this used for ?

packages/cmf/src/middlewares/http/middleware.js
```diff
@@ -122,40 +122,47 @@ export const httpMiddleware = ({ dispatch }) => next => (action) => {
...
+				error: errorObject,
+				action: httpAction,
+			});
+		} else {
+			// clone the response object else the next call to text or json
+			// triggers an exception Already use
```
sdwvit: sorry I'm having really hard times understanding what do you mean by this comment, could you rephrase?

jsomsanith:
2 Approvals but we'll wait for @jmfrancois to answer the questions and merge if everything is ok

packages/cmf/src/middlewares/http/middleware.js
```diff
@@ -12,7 +12,7 @@ export const DEFAULT_HTTP_HEADERS = {
 };
 
 export function isHTTPRequest(action) {
-	return action.type in HTTP_METHODS;
+	return action.type in HTTP_METHODS || !!get(action, 'cmf.http');
```
talend-glorieux: I'd rather see a regular if conditional than a weird double negative statement.

packages/cmf/src/middlewares/http/middleware.js
jmfrancois: at the moment when we have a JS error in a middleware it's hidden by can't read clone attribute of undefined.
So let's dispatch that error so we may display the error 'to the end user or not'

May I just throw the error ?

packages/cmf/src/middlewares/http/middleware.js
jmfrancois: not sure to understand. will catch you today

packages/cmf/src/middlewares/http/middleware.js
jmfrancois: it was here before I must admit me too

PR #568, opened 2017-06-23
fix(TUX-262): show pagination navigation only when necessary

mtikhinskaya:

When there is no items the pages count is 1/0

Don't show pagination navigation when there is less than 2 pages

### DIFF
--- a/packages/components/src/List/Toolbar/Pagination/Pagination.component.js
+++ b/packages/components/src/List/Toolbar/Pagination/Pagination.component.js
@@ -34,6 +34,7 @@ function Pagination({ id, startIndex, itemsPerPage, totalResults, onChange, ...o
 	} = paginationIconProps;
 	const currentPage = Math.ceil(startIndex / itemsPerPage);
 	const pagesLength = Math.ceil(totalResults / itemsPerPage);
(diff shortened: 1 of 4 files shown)


packages/components/src/List/Toolbar/Pagination/Pagination.component.js
```diff
@@ -73,32 +74,32 @@ function Pagination({ id, startIndex, itemsPerPage, totalResults, onChange, ...o
 			>
 				{itemsPerPageOptions.map((option, index) => getMenuItem(option, index))}
 			</NavDropdown>
-			{itemsPerPage > 0 && (
+			{isNavigationShown && (
```
jsomsanith: Can you extract the navigation items to a function `getNavigationItems()` so you can write 
```
{isNavigationShown && getNavigationsItems()}
```
This would avoid the `sNavigationShown && ` on each items (5 times)

jsomsanith:
BTW, the branch name has evolved 2 months ago. can you read the contribution guideline ?
The problem here is the branch name

mtikhinskaya:
Add fixes and change branch name in https://github.com/Talend/ui/pull/571

PR #573, opened 2017-06-26
fix(components/Virtualized): adapt the toolbar stories

jare-talend:
Autosizer takes the height from her parent. https://github.com/bvaughn/react-virtualized/blob/master/docs/usingAutoSizer.md

With the toolbar is enable, the list height is the toolbar height + display list height instead of only the list height. 
To avoid to get a height of the list higher, we need to set a wrapper around the autosize to determinate the height.


packages/components/src/List/List.scss
```diff
@@ -19,5 +19,7 @@
 		box-shadow: $shadow-default;
 	}
 
-
+	:global(.tc-list-display-virtualized) {
+		overflow-y: hidden;
```
jsomsanith: This is not a good solution. This hides a scrolling issue du to css placement.

packages/components/src/List/List.component.js
```diff
@@ -54,12 +54,18 @@ function DisplayModeComponent({ id, useContent, displayMode, list, virtualized }
 		);
 	}
 	if (virtualized) {
+		const classnames = classNames(
+			'tc-list-display',
+			'tc-list-display-virtualized',
```
jsomsanith: This is not needed, only `tc-list-display` is

packages/components/stories/List.js
```diff
@@ -600,4 +600,34 @@ storiesOf('List', module)
...
+		inputDebounceProps.toolbar.filter.debounceTimeout = 300;
+
+		return (<div style={{ height: '60vh' }} className="virtualized-list">
+			<IconsProvider />
+
+			<h1>List</h1>
+			<h2>Definition</h2>
+			<p>Toolbar Filter</p>
+			<h2>Docked</h2>
+			<List {...dockedProps} virtualized />
```
jsomsanith: To avoid the list to extends its size, let's wrap each list in a controlled size div

packages/components/stories/VirtualizedList.js
```diff
@@ -188,7 +188,7 @@ for (let i = collection.length; i < 100; i += 1) {
 
 storiesOf('Virtualized List', module)
 	.add('List > Table', () => (
-		<div style={{ height: '60vh' }} className="virtualized-list">
+		<div className="virtualized-list">
```
jsomsanith: Why the changes in this file ?

packages/components/src/List/List.component.js
jare-talend: I have added tc-list-display-virtualized to avoid to override overflow-y: hidden; on tc-list-display during the migration

packages/components/stories/VirtualizedList.js
jare-talend: I wrapped the list into <section style={{ height: '300px'}}></section>. It is unused to define a height.

packages/components/src/List/List.component.js
jsomsanith: Yes but `overflow-y: hidden;` is not a good solution (see above)

packages/components/stories/VirtualizedList.js
jsomsanith: To avoid List to take the h1 and p size :+1:

PR #574, opened 2017-06-26
feat(components): i18n

jsomsanith:
No i18n

* i18n via i18next
* storybook serve / (containing locales/)
* storybook copy locales/ on build

To use the next version, you will have to configure i18next on your app


packages/components/stories/config/i18n.js
```diff
@@ -0,0 +1,26 @@
+import i18n from 'i18next';
+import XHR from 'i18next-xhr-backend';
+
+i18n
+	.use(XHR)
```
jmfrancois: it should work by default without the xhr, we must provide default traduction without traduction. So even bad english will be fixed by the translations :)

packages/components/src/HeaderBar/HeaderBar.component.js
```diff
@@ -179,27 +182,28 @@ User.propTypes = {
...
 }
-
-Products.propTypes = {
+PureProducts.propTypes = {
 	renderers: React.PropTypes.shape({
 		ActionDropdown: React.PropTypes.func,
 	}),
+	t: React.PropTypes.func.isRequired,
 };
+const Products = translate('tui-components-headerBar')(PureProducts);
```
jmfrancois: we should be able to only translate the higher components and spread t props in it

packages/components/src/HeaderBar/HeaderBar.component.js
```diff
@@ -41,21 +41,22 @@ function Logo({ isFull, renderers, ...props }) {
 				bsStyle="link"
 				className={actionClassName}
 				hideLabel
-				label="Go to Portal"
+				label={t('GO_PORTAL')}
```
jmfrancois: t('GO_PORTAL', 'Go to Portal') with a defaultTranslate function which support this second argument

jsomsanith:
Close this PR as it was a POC

PR #583, opened 2017-06-29
feat(TFD-1459): add available expression to action

jmfrancois:

Actions are always displayed in most components. We need to handle security and let the user only access to an action if the current user can do it.

This kind of check need multiple concern:
* does the current user has a permission ? (action context less like 'add XX')
* does the resource is a state that make this action possible ? (action with context 'edit XX')

So we need to express this kind of stuff in the CMF way (understand configurable)

example for a contextual action. Note the model is the action model, so the resource object.
Here the backend add metadata to infor client what are the possible actions using a simple string: 'XUD' for execute, update, delete

`settings.json`
```
    "dataset:edit": {
      "id": "dataset:edit",
      "icon": "talend-pencil",
      "label": "Edit this dataset",
      "actionCreator": "redirect",
      "path": "/datasets/$id/edit",
      "available": {
        "id": "hasMode",
        "args": ["U"]
      }
    },
```

`configure.js`
```
import { api } from 'react-cmf';

api.expression.register('hasMode', (context, right) => context.model.__talend_modes.indexOf(right) !== -1);


api.expression.register('hasPermission', (context, permission) => {
	return context.store.getState()
		.cmf.collections.getIn(['user', 'permissions'])
		.find(item => item.name === permission)
		.size === 1;
});
```

note this PR touch to all the stack: cmf, containers, component


packages/containers/src/actionAPI.js
```diff
@@ -1,5 +1,22 @@
 import { api } from 'react-cmf';
 
+/**
+ * add support for expression in actions.
+ * it change the action props by their expression value
+ */
+function evalExpressions(action, model, context) {
+	const copyContext = Object.assign({}, context, { model });
```
acateland: its a bit more thant a copied context since model was merged in

packages/cmf/src/mock/settings.js
```diff
@@ -35,7 +35,6 @@ const settings = {
 			id: 'menu:article',
 			name: 'My article',
 			icon: 'icon-article',
-			available: 'model.actions.delete',
```
jmfrancois: a relic of what it was intended to achieve but we don't have angular expression ^^

packages/components/src/Actions/Action/Action.component.js
```diff
@@ -101,7 +103,11 @@ function Action(props) {
 	});
 
 	const buttonContent = getContent(props);
-
+	if (available === false) {
+		buttonProps.className = classnames({
+			hidden: available === false,
+		}, buttonProps.className || '');
```
jsomsanith: `available === false` --> `!available` ?
`buttonProps.className || ''` --> `buttonProps.className`, the falsy value is managed by `classNames()`

jsomsanith: I see that as a possible problem, because the button is rendered in the DOM and it is clickable (programmatically). Or just by disabling the css property.
IMO, this responsibility should be managed by the container, rendering the Action or not, instead of letting the component do it with a flag.

packages/containers/examples/ExampleAction.js
```diff
@@ -12,15 +12,50 @@ const myAction = {
...
 
+export default function ExampleAction() {
 	return (
 		<div>
 			<IconsProvider />
 			<p>Using names</p>
 			<Action name="menu:first" />
 			<p>Using actions</p>
 			<Action {...myAction} />
+			<p>Using available expression (4 Actions 2 hidden)</p>
```
jsomsanith: This story is hard to understand. Can you add some description ?
(And change the 2nd `is True expression` by perhaps `should not be displayed: false expression`

packages/containers/src/Action/Action.component.js
```diff
@@ -11,11 +11,13 @@ function Action({ name, ...rest }, context) {
...
+	let action;
 	if (name) {
-		const action = api.action.getActionInfo(context, name);
-		return (<PureAction {...action} onClick={onClick} />);
+		action = actions.getProps(context, name, rest.model);
+	} else {
+		action = actions.evalExpressions(rest, context);
 	}
-	return (<PureAction {...rest} onClick={onClick} />);
+	return (<PureAction {...action} onClick={onClick} />);
```
jsomsanith: To complete my comment on Action component, here we should evaluate expression if present, and if it is falsy, return null.

packages/components/src/Actions/Action/Action.component.js
jmfrancois: lets remove the change on component

packages/containers/src/Action/Action.component.js
jmfrancois: and render null il action.available === false

packages/containers/src/Action/Action.component.js
```diff
@@ -11,11 +11,16 @@ function Action({ name, ...rest }, context) {
 	const onClick = (event, payload) => {
```
jsomsanith: Please move this onClick creation after the `if (action.avaialable === false)`

PR #589, opened 2017-06-30
chore: improve travis run

jsomsanith:
We decided to disable the double CI check (branch + PR). The PR run was merging master, then checking the result. But this was not really useful, moreover, we must merge master on branch before the PR is mergeable. And this triggers the CI again.
So now we only have the branch run.

Another problem is that we trigger tests, and after success we want to get the code coverage, we trigger the tests again with coverage flag.

STILL ISSUES ON SNAPSHOTS

Simplify the travis  scripts
Run test with coverage from the beginning


.travis/after_success_storybook.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
```
frassinier: There is another tasks than storybooks, you could talk about "after_success_demo.sh"

.travis/after_success_static.sh
```diff
@@ -5,38 +5,16 @@ cd "$TRAVIS_BUILD_DIR"
 
 rm -rf .static
 mkdir .static
+mkdir .static/cmf
 mkdir .static/components
 mkdir .static/containers
 mkdir .static/forms
 mkdir .static/icons
+mkdir .static/logging
```
frassinier: What's the goal?

.travis.yml
```diff
@@ -12,20 +12,26 @@ addons:
...
   - .travis/after_success_coverage.sh
+  - .travis/after_success_deploy.sh
+  - .travis/after_success_lint.sh
   - .travis/after_success_screenshots.sh
   - .travis/after_success_git.sh
-  - .travis/after_success_deploy.sh
 branches:
-  except:
-    - /^v\d+\.\d+\.\d+$/
+  only:
```
frassinier: ORLY This is working for PR too?

.travis.yml
```diff
@@ -12,20 +12,26 @@ addons:
 node_js:
   - '6'
 before_install:
-  - if [[ "$TRAVIS_COMMIT_MESSAGE" =~ 'test(ci)' ]]; then travis_terminate 0; fi
+#   Fail fast if ci is triggered by ci commit
+  - export TALEND_COMMIT_MSG=$TRAVIS_COMMIT_MESSAGE
+  - if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then export TALEND_COMMIT_MSG="$(git log --format=%B -n 1 $TRAVIS_PULL_REQUEST_SHA)"; fi
+  - echo $TALEND_COMMIT_MSG
```
frassinier: Why do you need it?

.travis/after_success_static.sh
jsomsanith: For coverage

.travis.yml
jsomsanith: You're right, it was for debug. I remove it

.travis.yml
jsomsanith: For the PR that want to merge into master yes

.travis/after_success_static.sh
jmfrancois: so lets call it coverage ?

jsomsanith: No it's because the coverage folder goes in `.static/<package-name>`
So the coverage of logging will be moved to `.static/logging`, that's why we create this folder

PR #591, opened 2017-06-30
feat(form/datalist): on restricted mode handle blur with valid value

acateland:

When typping in the search zone a value exactly corresponding to an element in the list of value available, and leaving the focus, the component did not a trigger an onChange event, thus preventif hook to trigger further actions

onBlur if restricted mode is activated and the typped info match the list of available value, a change event is triggered and list of suggested element retracted.


packages/forms/src/widgets/DatalistWidget/DatalistWidget.test.js
```diff
@@ -207,6 +207,30 @@ describe('DatalistWidget', () => {
...
+		const input = wrapper.find('input').at(0);
+
+		// when
+		input.simulate('focus'); // to display suggestions
+		input.simulate('change', { target: { value: 'banane' } });
+		input.simulate('blur');
+
+		// then
+		expect(onChange).toBeCalled();
+		expect(toJson(wrapper)).toMatchSnapshot();
```
jmfrancois: useless snapshot here.
find the render of the selected value to ensure the props is 'banane'

PR #592, opened 2017-07-03
DO NOT MERGE feat(cmf): add Registry as higher order component

jmfrancois:

the registry is a singleton created at 'bootstrap' time.
we have no tools to inspect it and it makes it less reactish

Plug the registry as a react component to declare it
so the configuration happens during first rendering.

It will introduce the following possibility:

before 
```js
import { App, api } from 'react-cmf';

api.route.registerComponent('MyComponent', MyComponent);
render(
	<App ... />,
	document.getElementById('app'),
);
```

after
```js
import { App } from 'react-cmf';

('MyComponent', MyComponent);
render(
	<App ... >
		<App.Register id="MyComponent" component={MyComponent} />
		<App.Register id="myActionCreator" actionCreator={myActionCreator} />
		<App.Register id="myExpression" expression={myExpression} />
                <App.Router />
	</App>,
	document.getElementById('app'),
);
```

Then you also may use #611


packages/cmf/src/RegistryProvider.js
```diff
@@ -27,6 +35,35 @@ export default class RegistryProvider extends React.Component {
 	}
 }
 
+function Register(props, context) {
+	let id;
+	let item;
```
jsomsanith: Remove those declaration and declare `const` in the 1st `if` bloc

packages/cmf/src/RegistryProvider.js
```diff
@@ -27,6 +35,35 @@ export default class RegistryProvider extends React.Component {
...
+		action.registerActionCreator(props.id, props.actionCreator, context);
+	} else if (props.expression) {
+		expression.register(props.id, props.expression, context);
+	}
+	return null;
+}
+
+Register.contextTypes = {
+	registry: PropTypes.object,
+};
```
jsomsanith: There is no propTypes on `Register`

packages/cmf/src/RegistryProvider.js
jmfrancois: my bad, was for a previous implementation, thanks 👍 
I will update that

PR #593, opened 2017-07-03
fix(forms/datalist): remove difference check between user input and internal state on select

acateland:
When typing the exact sentance on one of the slectable element available
Then navigating to it with up/down key.
Then hitting enter, the selection and change does not trigger

remove an equality check onSelect for the restricted mode. 


acateland:
waiting for a better depiction of what this component should do from UX team

PR #597, opened 2017-07-03
fix(TFD-1459/components/Action): available=false add hidden class

jmfrancois:

Currently available = false can be applied by CMF but not all component use / can use Container(Action) but they directly use Action.

So the props available is set to false but the action is displayed to the end user.

let's add hidden bootstrap class to the action if props.available === false.

Note: `available === false` is not `!available` because most of the time available === undefined


packages/components/src/Actions/Action/Action.component.js
```diff
@@ -114,8 +116,9 @@ function Action(props) {
 			{buttonContent}
 		</Button>
 	);
-
-	if (hideLabel || tooltip || tooltipLabel) {
+	if (!available) {
```
jsomsanith: Adopt the fail fast principle by putting this at the top of the function please

PR #599, opened 2017-07-04
feat(forms): Adds read-only field columns support

talend-glorieux:
Columns fields can't be set as read-only.

Add read-only support to column fields.


packages/forms/src/widgets/ColumnsWidget/ColumnsWidget.js
```diff
@@ -13,16 +13,17 @@ function getColumnRenderer(type) {
 	return ObjectField;
 }
 
-function Column({ className, schema, formData, onChange, onBlur, registry }) {
-	const Renderer = getColumnRenderer(schema.type);
+function Column(props) {
```
talend-sromain: why go back on props instead of destructuring ?
too many params ?

talend-sromain: ( as we spoke, this params are not used in other case than passing to another component, so we skip variable assignements )

acateland: you were right on this one

PR #608, opened 2017-07-11
feat(components/Enumeration): added customizable label property in He…

vyshkov:
…ader.component

In order to finish new TMC module 'Projects' we need to implement a drawer with a list of users authorized to work with selected project. This list based on Enumeration component, but default label 'Values*' is not acceptable. We need to have possibility to define custom label (like 'Users') 

Add custom property.

### DIFF
--- a/packages/components/src/Enumeration/Header/Header.component.js
+++ b/packages/components/src/Enumeration/Header/Header.component.js
@@ -49,10 +49,10 @@ function getAction(action, index) {
 	);
 }
 
-function Header({ headerDefault, required, label }) {
+function Header({ headerDefault, required, label = 'Values' }) {
 	return (
 		<header className={headerClasses()}>
-			<span>{label || 'Values'}{required && '*'}</span>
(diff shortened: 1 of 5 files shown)


packages/components/src/Enumeration/Header/Header.component.js
```diff
@@ -49,10 +49,10 @@ function getAction(action, index) {
 	);
 }
 
-function Header({ headerDefault, required }) {
+function Header({ headerDefault, required, label }) {
 	return (
 		<header className={headerClasses()}>
-			<span>Values{required && '*'}</span>
+			<span>{label || 'Values'}{required && '*'}</span>
```
talend-sromain: May we use default proptypes instead of this || ?

jare-talend: Replace this by defaultProps?

vyshkov: Ok :)

PR #610, opened 2017-07-11
feat(forms): fix for form lib + custom elements

acateland:
Forked ArrayField position handler aren't bind to this, thus could not work and trigger an error.
Forked ArrayField template customisation was invoqued only as function instead as an React Element.
Form does not accept customFields trought the host application.

Add a constructor to ArrayField and bind function to this
Updated how ArrayFieldTemplate are called matching what latest version of the form lib does
Updated Forms so customFields can be merged with default one, and be given trought a props.


acateland:
@jmfrancois yeah may change in the future if tcomp and us agree on some way to predetermine the shape of the title

output/forms.eslint.txt
```diff
@@ -5,31 +5,31 @@
...
   644:3  error  Prop type `object` is forbidden  react/forbid-prop-types
+  645:3  error  Prop type `object` is forbidden  react/forbid-prop-types
 
 /home/travis/build/Talend/ui/packages/forms/src/fields/BooleanField.js
   71:3  error  Prop type `object` is forbidden  react/forbid-prop-types
   72:3  error  Prop type `object` is forbidden  react/forbid-prop-types
   73:3  error  Prop type `object` is forbidden  react/forbid-prop-types
 
+/home/travis/build/Talend/ui/packages/forms/src/fields/CustomObjectField.js
+  81:8  error  Visible, non-interactive elements should not have mouse or keyboard event listeners  jsx-a11y/no-static-element-interactions
```
jsomsanith: Is it possible to do something with that ? Replace divs by buttons ?

packages/forms/src/fields/CustomObjectField.js
```diff
@@ -0,0 +1,151 @@
...
+					</div>
+					{(uiSchema['ui:description'] || schema.description) &&
+						<DescriptionField
+							id={`${idSchema.$id}__description`}
+							description={uiSchema['ui:description'] || schema.description}
+							formContext={formContext}
+						/>}
+					{!formData.isClosed &&
+						orderedProperties.map((fieldName, index) => {
+							if (fieldName !== 'isClosed') {
```
jsomsanith: I would prefer a fast fail 
```
if(isClosed) { return null; }
return <SchemaField ...
```

packages/forms/src/templates/ArrayFieldTemplate.scss
```diff
@@ -0,0 +1,105 @@
...
+		transition: box-shadow 0.3s cubic-bezier(.25, .8, .25, 1);
+		margin-bottom: $padding-normal;
+
+		&:hover {
+			background: $white;
+			box-shadow: 0 7px 14px rgba(0, 0, 0, 0.25), 0 5px 5px rgba(0, 0, 0, 0.22);
+			border: 1px solid $gallery;
+
+			.delete {
+				color: $chestnut-rose;
```
jsomsanith: Use theme var

packages/forms/src/templates/ArrayFieldTemplate.scss
```diff
@@ -0,0 +1,105 @@
...
+			}
+
+			button {
+				background: transparent;
+				border: 0;
+				margin: 0 0 1em 0;
+				color: $dark-silver;
+
+				&:hover {
+					color: $dove-gray;
```
jsomsanith: Can you externalize the common colors to component sass var ? Here for example you repeat `$dove-gray` multiple time for hover

packages/forms/src/fields/CustomObjectField.js
```diff
@@ -0,0 +1,151 @@
+import React, { PropTypes } from 'react';
+import { Icon } from 'react-talend-components';
+import { orderProperties, retrieveSchema } from 'react-jsonschema-form/lib/utils';
+
+function createCustomObjectField(title) {
+	class CustomObjectField extends React.Component {
```
jsomsanith: I don't like this name :)
Something more descriptive would be better, CollapsibleFieldset ?

acateland: good one

packages/forms/src/templates/ArrayFieldTemplate.scss
acateland: its for two different elements wich may be styled in different fashion later

packages/forms/src/templates/ArrayFieldTemplate.scss
jsomsanith: Sorry, not clear, I meant $brand-danger

packages/forms/src/fields/CollapsibleFieldset.js
```diff
@@ -0,0 +1,151 @@
...
+						</pre>
+					</div>
+				);
+			}
+			const iconTransform = !formData.isClosed ? 'flip-vertical' : '';
+			return (
+				<fieldset>
+					<div onDoubleClick={this.toggle} role="button">
+						{title &&
+							<div onClick={this.toggle} role="button">
```
frassinier: Maybe this could a `button.btn.btn-link`?

PR #611, opened 2017-07-12
feat(cmf): add Inject component

jmfrancois:

When writing import Action from '../Action' we are creating hard dependency between component.
There is no way to unplug this component to replace it by an other (for example Action)

```js
import { Inject } from 'react-cmf';

function MyComponent(props) {
  return <Inject component="Action" name="home:return" />
}
```


packages/cmf/src/Inject.component.js
```diff
@@ -0,0 +1,34 @@
+import React, { PropTypes } from 'react';
+import api from './api';
+
+const getComponentFrom = api.route.getComponentFromRegistry;
+
+function NotFoundComponent({ children, error, ...props }) {
+	console.error(e);
```
ncomont: `e` is not defined

packages/cmf/src/Inject.component.js
```diff
@@ -0,0 +1,34 @@
...
+	return (
+		<div {...props}>
+			<div className="alert alert-danger">{error.message}</div>
+			{children}
```
talend-sromain: Should we render the chidren element here in case of error ? 
We might not want to render freely that be meant to be render in injected component

packages/cmf/src/Inject.component.js
```diff
@@ -0,0 +1,34 @@
...
+	);
+}
+
+function Inject({ component, ...props }, context) {
+	try {
+		const Component = getComponentFrom(context, component);
+		return (
+			<Component {...props} />
+		);
+	} catch(error) {
```
ncomont: lint error : missing space beetween `catch` and `(error)`

packages/cmf/src/Inject.component.js
```diff
@@ -0,0 +1,34 @@
...
+		</div>
+	);
+}
+
+function Inject({ component, ...props }, context) {
+	try {
+		const Component = getComponentFrom(context, component);
+		return (
+			<Component {...props} />
+		);
```
talend-glorieux: No need for multilines.

packages/cmf/src/Inject.component.js
```diff
@@ -0,0 +1,34 @@
...
+			<div className="alert alert-danger">{error.message}</div>
+			{children}
+		</div>
+	);
+}
+
+function Inject({ component, ...props }, context) {
+	try {
+		const Component = getComponentFrom(context, component);
+		return (
```
ncomont: could be `return <Component {...props} />;`

packages/cmf/src/Inject.component.js
```diff
@@ -0,0 +1,34 @@
+import React, { PropTypes } from 'react';
+import api from './api';
+
+const getComponentFrom = api.route.getComponentFromRegistry;
+
+function NotFoundComponent({ children, error, ...props }) {
```
jsomsanith: Linter: missing propTypes

packages/cmf/src/Inject.component.js
```diff
@@ -0,0 +1,34 @@
...
+}
+
+function Inject({ component, ...props }, context) {
+	try {
+		const Component = getComponentFrom(context, component);
+		return (
+			<Component {...props} />
+		);
+	} catch(error) {
+		return <NotFoundComponent error={error.message} {...props} />
```
jsomsanith: Linter : missing semicolon

PR #624, opened 2017-07-13
fix(forms/listview): use enumNames instead of enum

frassinier:
`ListView` component does not use `enumNames` if they are provided.

Return `enum` values and display `enumNames` ones as usual.


packages/forms/stories/json/listView.json
```diff
@@ -10,6 +10,213 @@
...
+            "country196",
+            "country197",
+            "country198",
+            "country199",
+            "country200",
+            "country201",
+            "country202",
+            "country203",
+            "country204",
+            "country205"
```
talend-glorieux: Please use a loop to generate that content.

frassinier: Oh it's dummy json

PR #625, opened 2017-07-13
feat(List): map selection to virtualized list

jsomsanith:
Old List to VirtualizedList does not pass selection props

Map List selection props to VirtualizedList selection props

### DIFF
--- a/packages/components/src/List/ListToVirtualizedList/ListToVirtualizedList.component.js
+++ b/packages/components/src/List/ListToVirtualizedList/ListToVirtualizedList.component.js
@@ -16,7 +16,7 @@ function adaptOnSort(onChange) {
 }
 
 function ListToVirtualizedList(props) {
-	const { id, items, columns, titleProps } = props;
+	const { id, items, columns, itemProps, sort, titleProps } = props;
 
 	if (!titleProps.actionsKey) {
 		titleProps.actionsKey = 'actions';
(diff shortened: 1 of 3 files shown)


packages/components/src/List/ListToVirtualizedList/ListToVirtualizedList.component.js
```diff
@@ -16,7 +16,7 @@ function adaptOnSort(onChange) {
 }
 
 function ListToVirtualizedList(props) {
-	const { id, items, columns, titleProps } = props;
+	const { id, items, columns, itemProps, sort, titleProps } = props;
```
talend-glorieux: Why not also add display mode to the spread and put it in the argument definition?

packages/components/stories/List.js
```diff
@@ -428,19 +397,59 @@ storiesOf('List', module)
...
 				<h1>List</h1>
-				<p>Display a list with columns containing actions.</p>
+				<p>A column can contains only actions that appear on mouseover.</p>
 				<IconsProvider defaultIcons={icons} />
 				<List {...columnActionsProps} virtualized />
 			</div>
 		);
 	})
+	.add('Virtualized - selection', () => {
+		const selectedItemsProps = Immutable.fromJS(props).toJS();
```
talend-glorieux: `fromJS` `toJS` ?

talend-glorieux: Why not use something like lodash's cloneDeep. It avoid's moving from a mutable to an immutable data structure and back.

Issue #43, opened 2017-01-25
[containers] bug: Action or Button with Icon inside, when disabled, can still click on Icon and trigger eventlistener


talend-glorieux:
_From @Peiren on November 23, 2016 9:2_

_Copied from original issue: Talend/react-talend-components#110_

PR #627, opened 2017-07-17
 fix(components/Action): tooltip stays on disabled button after timeout

sdwvit:
If you have a disabled button with tooltip trigger, button event listeners are ignored by browser. However, when hovering over button's child elements, child triggers onMouseOver and propagates it on button, this way tooltip is displayed. However, somehow onMouseOut is not triggered and Tooltip component's flag this.hover stays true forever. 

If child element covers whole button, then onMouseOver and onMouseOut are always triggered and tooltip is destroyed. Decided to wrap children in div with display: inline-box style to assure no side-effects.

Related to #43


acateland:
im not sure this is the purpose of storybook, better add a non regression integration test.

jsomsanith:
+1, the storybook is more a showcase of what we do, not what are the bugs.
You can fix it in this PR or perhaps create a bug with the code to reproduce it in the description.

sdwvit:
sure, I'll remove story

sdwvit:
@jsomsanith this pr actually introduces patch to fix bug, story was added as a showcase, but I'll remove story

PR #628, opened 2017-07-17
fix(components/datalist): onblur on same value

ncomont:
Datalist onblur is triggered even if value is the same.

onblur is only triggered when necessary.


packages/forms/src/widgets/DatalistWidget/DatalistWidget.js
```diff
@@ -159,7 +160,20 @@ class DatalistWidget extends React.Component {
 		if (hasItems && Object.prototype.hasOwnProperty.call(itemsMap, value)) {
 			return itemsMap[value];
 		}
-		return value;
+		return value != null ? value : '';
```
talend-glorieux: Why not use a default parameter in the function definition?

packages/forms/src/widgets/DatalistWidget/DatalistWidget.js
```diff
@@ -117,14 +108,24 @@ class DatalistWidget extends React.Component {
...
-	onBlur() {
-		const { options } = this.props;
-		const included = this.state.initalItems.includes(this.state.value);
+	onBlur(event) {
+		const inputLabel = event.target.value;
+		const { options, onChange } = this.props;
+		const { initialItems, itemsMap, value, lastKnownValue } = this.state;
+		const inputValue = this.getValue(inputLabel);
+		const isIncluded = initialItems.includes(value)
+			|| Object.keys(itemsMap).some(k => itemsMap[k] === value);
```
talend-glorieux: Do you think this could be extracted into a function?

packages/forms/src/widgets/DatalistWidget/DatalistWidget.js
talend-glorieux: Why not use a default parameter in the function definition?

ncomont: Unfortunately, I can't. If `value` is `null`, the default parameter is not taken in account : 

```javascript
➜  ui git:(master) ✗ node
> function yo(test = 'lo') { console.log('yo', test); };

> yo();
yolo

> yo(null);
yonull

> 
```

But I learned something !

PR #630, opened 2017-07-17
chore(cmf): add flowtype, for TAB

jmfrancois:

detect errors at compile time.
proptypes is a tool that is used at runtime and consume a lot of resource in non production environment.

exposing flow type which is the typing system used at facebook.
we try to cover:

* Dispatcher
* App
* actions


packages/cmf/flow-typed/npm/redux_v3.x.x.js
```diff
@@ -0,0 +1,107 @@
+// flow-typed signature: 76c3dfb5a40ce169d5c08fe80dbd029a
+// flow-typed version: 37b1c6d953/redux_v3.x.x/flow_>=v0.33.x
```
jmfrancois: I believe this file should be removed and installed using flow-typed CLI

PR #631, opened 2017-07-18
fix(tabs): move tabs style to theme

mtikhinskaya:

theme tabs style differ from TabBar component style

TabBar component style moved to theme nav-tabs


packages/theme/src/theme/_navs.scss
```diff
@@ -73,41 +73,42 @@
 	}
 
 	&-tabs {
+		background-color: tint($brand-primary, 90);
```
frassinier: Is there a way to create some global values for these tint in variables.scss?
Like https://github.com/Talend/ui/blob/master/packages/theme/src/theme/_variables.scss#L21

mtikhinskaya: I think it should be a separate PR . We need to search occurrences and create a variable if it really needed. And not only for this color but for other ones

jsomsanith: This means we need 2 new shades of $brand-primary : 
* tab background
* tab active background
We already have $brand-primary-light, $brand-primary-lighter. If we come up with 2 new names that express that they are more light, ok for me.

mtikhinskaya: I don't understand what for we need this variable. It's used only in one place. It shouldn't be rewritten. It's not rewrite bootstrap variable. Could you please explain?

jsomsanith:
Close for no common solution and non activity (means this is not needed)

PR #634, opened 2017-07-18
feat(VirtualizedList): adapt and enhance component objects

jsomsanith:
With virtualizedList, the component object List is no longer up to date.

Adapt it, but not only : 
* enhance its use to have a more fluent api
* add missing selectors

The List api and POJO have completely changed.
See breaking change log


packages/components/e2e/component-objects/pom.xml
```diff
@@ -5,7 +5,7 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>com.talend</groupId>
     <artifactId>component-objects</artifactId>
-    <version>1.0.0</version>
+    <version>0.95.0-SNAPSHOT</version>
```
frassinier: Are you sure that it will be easily to bump?

jsomsanith: I discussed with @kmouzouri and @atriki from TFD team, they are the only ones using the lib, and they don't maintain any (release) branch for now. So it's the time to align version and override the 1.0.0.

packages/components/e2e/component-objects/src/main/java/org/talend/component/Component.java
```diff
@@ -58,6 +75,11 @@ public Component(WebDriver driver, String name, String selector) {
      */
     public WebElement getElement() throws NotFoundException {
         LOGGER.info(this.name + ".getElement " + this.selector);
+
+        if(this.root != null) {
```
talend-glorieux: Missing space before if.

packages/components/e2e/component-objects/src/main/java/org/talend/component/Component.java
```diff
@@ -41,6 +43,21 @@ public Component(WebDriver driver, String name, String selector) {
...
+     * @param driver Selenium WebDriver
+     * @param root Component root element
+     */
+    public Component(WebDriver driver, String name, WebElement root) {
+        LOGGER.info("Component " + name + " from provided root");
+        this.driver = driver;
+        this.name = name;
+        this.root = root;
+
+        this.driver.manage().timeouts().implicitlyWait(10, TimeUnit.SECONDS);
```
talend-glorieux: Why the timeout?

packages/components/e2e/component-objects/src/main/java/org/talend/component/list/table/Table.java
```diff
@@ -0,0 +1,117 @@
...
+        this(driver, null);
+    }
+
+    /**
+     * Constructor with table id
+     *
+     * @param driver Selenium WebDriver
+     */
+    public Table(final WebDriver driver, final String id) {
+        super(driver, NAME, String.format(TABLE_SELECTOR, id != null ? "#" + id : ""));
```
talend-glorieux: Please set a default value for id instead of using a ternary.

packages/components/e2e/component-objects/src/main/java/org/talend/component/list/table/Table.java
```diff
@@ -0,0 +1,117 @@
...
+    /**
+     * Get all rendered items, represented by a row in the table.
+     *
+     * @return The list of Items
+     */
+    public List<Item> getItems() {
+        return this.getElement() //
+                .findElements(By.cssSelector(TABLE_ITEM_SELECTOR)) //
+                .stream() //
+                .map(webElement -> new Item(driver, webElement)) //
```
talend-glorieux: Those comments slashes are not required.

packages/components/e2e/component-objects/src/main/java/org/talend/component/list/table/Item.java
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * A List is used to easy access to WebElements of the react-talend-component List component - Table - Item.
+ */
+public class Item extends Component {
+
+    private static final String NAME = "Item";
+
+    private static final String TABLE_ITEM_TITLE_SELECTOR = ".tc-list-title > button";
+
+    private static final String TABLE_ITEM_ACTIONS_SELECTOR = ".tc-list-title > .tc-actions button";
```
frassinier: DRY, create var for `.tc-list-title`

packages/components/e2e/component-objects/src/test/java/org/talend/component/StorybookTest.java
```diff
@@ -0,0 +1,56 @@
+package org.talend.component;
+
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.openqa.selenium.By;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+import org.openqa.selenium.chrome.ChromeDriver;
+
+public class StorybookTest {
```
frassinier: What's about run this test into Travis CI?

packages/components/e2e/component-objects/src/main/java/org/talend/component/Component.java
jsomsanith: This was set by the TFD QA, I just kept it.
This is their settings. But now that you point it, it should not be there, it overrides the app test config.
I'll remove it.

packages/components/e2e/component-objects/src/main/java/org/talend/component/list/table/Table.java
jsomsanith: I can't, no default value in methode prototype in java.
Plus, super() must be the first instruction. So no choice :(

packages/components/e2e/component-objects/src/main/java/org/talend/component/list/table/Table.java
jsomsanith: If I don't put them, and format it with Talend config, it is less readable.
That's why I choose to put them.

packages/components/e2e/component-objects/src/test/java/org/talend/component/StorybookTest.java
jsomsanith: :+1: let's do that in another PR

packages/components/e2e/component-objects/src/main/java/org/talend/component/list/table/Table.java
talend-glorieux: :roll_eyes:

PR #638, opened 2017-07-19
fix(components/datalist): datalist in modal

ncomont:

For a strange reason, when a datalist is contained in a modal, user can't scroll using the items-container scrollbar. `onBlur` is called when the **input** loose the focus, and then resets the suggestions. (I don't understand why it's working fine without modal, but that's a fact).

nofix

A `mousedown` event listener on the items container prevent click actions to be propagated, so the `onBlur` method is not called.

fix

### DIFF
--- a/packages/forms/stories/index.js
+++ b/packages/forms/stories/index.js
@@ -8,11 +8,31 @@ import { withKnobs, object } from '@kadira/storybook-addon-knobs';
 import Well from 'react-bootstrap/lib/Well';
 import IconsProvider from 'react-talend-components/lib/IconsProvider';
 import { Action } from 'react-talend-components';
+import { Dialog } from 'react-talend-components';
 
 import Form from '../src/Form';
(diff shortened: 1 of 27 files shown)


packages/forms/stories/index.js
```diff
@@ -8,6 +8,7 @@ import { withKnobs, object } from '@kadira/storybook-addon-knobs';
 import Well from 'react-bootstrap/lib/Well';
 import IconsProvider from 'react-talend-components/lib/IconsProvider';
 import { Action } from 'react-talend-components';
+import { Dialog } from 'react-talend-components';
```
frassinier: You could use the previous line

talend-glorieux:
Does it work when using the keyboard?

ncomont:
@glorieux this problem occurs when the user **click** on the scrollbar. The current behaviour is not affected by this change.

PR #639, opened 2017-07-20
feat(containers/Action): add expr and actionCreator support

jmfrancois:

if you provide an action id to Action which resolve on actionCreator it will not be used.

inProgress props is really often something dynamic, so let's make it support expression.

label may also change depending on the app state (Run -> Running, ...)

let's all that being supported !


output/containers.eslint.txt
```diff
@@ -7,6 +7,11 @@ The react/require-extension rule is deprecated. Please use the import/extensions
 /home/travis/build/Talend/ui/packages/containers/src/Action/Action.test.js
   2:1  error  'react-test-renderer' should be listed in the project's dependencies, not devDependencies  import/no-extraneous-dependencies
 
+/home/travis/build/Talend/ui/packages/containers/src/actionAPI.test.js
```
acateland: fix lint

PR #646, opened 2017-07-24
feat(forms/TFD-1660): add widget code

jmfrancois:

tcomp have / will have 'code' widget which is /was a textarea

bind to react-ace except if possible / fallback to textarea if not installed in the webapp
Note: you have to add 'react-ace' deps in your client if you want it.

screen shot 2017-08-10 at 15 57 44


packages/forms/src/widgets/CodeWidget/CodeWidget.component.js
```diff
@@ -0,0 +1,124 @@
+/* eslint-disable no-console */
+/* eslint-disable import/no-extraneous-dependencies */
+/* eslint-disable import/no-mutable-exports */
+/* eslint-disable global-require */
+
+import React, { PropTypes } from 'react';
```
acateland: please import Proptypes from 'prop-type' package

packages/forms/src/widgets/CodeWidget/CodeWidget.component.js
```diff
@@ -0,0 +1,124 @@
...
+			const {
+				id,
+				options,
+				value,
+				disabled,
+				readonly,
+				autofocus,
+			} = this.props;
+			if (options && options.language && DEFAULT_LANGUAGES.indexOf(options.language) === -1) {
+				console.error(`${options.language} language not supported`);
```
acateland: this error is fine, but maybe we could add a visual cue when the component render ?

PR #651, opened 2017-07-25
feat(UIForms): arrays of objects

jsomsanith:
*WARNING*: this is for the next version of forms only
No array of objects widget

Implement it with the syntaxe of jsfc array
This widget manage arrays of objects or simple fields like strings. It basically manage indexes :
* Add control to remove/reorder
* Add the index on the schema-key of its children
* On remove : shift the errors indexes after the removed index
* On reorder : shift the errors indexes between the old and new indexes

Next step would be to implement DnD (nice to have).


packages/forms/src/UIForm/utils/validation.js
```diff
@@ -19,28 +19,131 @@ export function validateValue(schema, value, properties, customValidationFn) {
...
-			Object.assign(results, subResults);
-		}
+		const { key } = schema;
+		const value = key && getValue(properties, key);
+		const subResults = validateSingle(
+			schema,
+			value,
+			properties,
+			customValidationFn,
+			true /* deep validation */
```
talend-glorieux: Please use a line comment instead of a block one.

packages/forms/src/UIForm/utils/validation.js
```diff
@@ -50,9 +153,9 @@ export function validateAll(mergedSchema, properties, customValidationFn) {
  * It is invalid if :
  * - the schema is an invalid field (errors[key] is falsy)
  * - the schema has items (ex: fieldset, tabs, ...), and at least one of them is invalid
- * @param schema The schema
- * @param errors The errors
- * @returns {boolean} true if it is invalid, false otherwise
+ * @param schema The schema.
+ * @param errors The errors.
```
talend-glorieux: Please remove periods on `@param` definitions.

packages/forms/src/UIForm/fieldsets/Array/Array.component.js
```diff
@@ -0,0 +1,186 @@
...
+		super(props);
+
+		this.onAdd = this.onAdd.bind(this);
+		this.onRemove = this.onRemove.bind(this);
+		this.onReorder = this.onReorder.bind(this);
+	}
+
+	onAdd(event) {
+		const schema = this.props.schema;
+		const defaultValue = schema.schema.items.type === 'object' ? {} : '';
```
jsomsanith: Yes, `const schema = this.props.schema;` refers the the array schema from json-schema-form-core. But in it we have a property `schema` with the original schema value from json input.

I'll change the names to be a little clearer.
```
const arrayMergedSchema = this.props.schema;
const defaultValue = arrayMergedSchema.schema. ...
```

packages/forms/src/UIForm/UIForm.component.js
```diff
@@ -40,23 +40,38 @@ export default class UIForm extends React.Component {
...
+			error,
+		};
+		this.props.onChange(event, payload);
+
+		if (schema.triggers && schema.triggers.includes(TRIGGER_AFTER)) {
 			this.onTrigger(event, { type: TRIGGER_AFTER, ...payload });
 		}
+
+		if (widgetChangeErrors) {
+			const errors = Object.assign(widgetChangeErrors(this.props.errors));
```
ncomont: Is it here to perform a copy of the object ? If it is, this should be 

`const errors = Object.assign({}, widgetChangeErrors(this.props.errors));`

packages/forms/src/UIForm/fieldsets/Array/Array.component.js
```diff
@@ -0,0 +1,192 @@
...
+	onRemove(event, indexToRemove) {
+		const schema = this.props.schema;
+		const value = this.props.value.slice(0);
+		value.splice(indexToRemove, 1);
+
+		return this.props.onChange(
+			event,
+			{ schema, value },
+			{
+				widgetChangeErrors(errors) {
```
ncomont: I think this could be clearer if `widgetChangeErrors` were defined before the `return`

packages/forms/src/UIForm/fieldsets/Array/Array.component.js
```diff
@@ -0,0 +1,192 @@
...
+					// shift up the items errors after the one we remove
+					return shiftArrayErrorsKeys(
+						errors,
+						{
+							arrayKey: schema.key,
+							minIndex: indexToRemove,
+							shouldRemoveIndex(index) {
```
ncomont: Why didn't you use arrow function ? 

```javascript
{
	arrayKey: schema.key,
	minIndex: indexToRemove,
	shouldRemoveIndex: index => index === indexToRemove,
	getNextIndex: index => index - 1,
},
```

packages/forms/src/UIForm/fieldsets/Array/Array.component.js
```diff
@@ -0,0 +1,192 @@
...
+		const schema = this.props.schema;
+		const value = this.props.value.slice(0);
+		const [item] = value.splice(previousIndex, 1);
+		value.splice(nextIndex, 0, item);
+
+		return this.props.onChange(
+			event,
+			{ schema, value },
+			{
+				widgetChangeErrors(errors) {
```
ncomont: Same here

packages/forms/src/UIForm/fieldsets/Array/Array.component.js
```diff
@@ -0,0 +1,192 @@
...
+		);
+	}
+
+	render() {
+		const { errorMessage, id, isValid, schema, value, ...restProps } = this.props;
+
+		return (
+			<div>
+				<ol id={id} className={theme['tf-array']}>
+					{value.map((itemValue, index) => {
```
ncomont: What do yo think about adding a new function to render the `li` ? `{value.map(renderItem)}`

packages/forms/src/UIForm/fieldsets/Array/ArrayItem.component.js
```diff
@@ -0,0 +1,66 @@
...
+		hasMoveDown,
+		hasMoveUp,
+		id,
+		index,
+		onRemove,
+		onReorder,
+		value,
+	} = props;
+	return (
+		<div className={theme['tf-array-item']} >
```
ncomont: Could you add a CSS class in addition to the CSS module one ? 

`<div className={'tf-array-item', theme['tf-array-item']} >`

packages/forms/src/UIForm/fieldsets/Array/ArrayItem.component.js
```diff
@@ -0,0 +1,66 @@
...
+		children,
+		hasMoveDown,
+		hasMoveUp,
+		id,
+		index,
+		onRemove,
+		onReorder,
+		value,
+	} = props;
+	return (
```
ncomont: (missing new line)

packages/forms/src/UIForm/fieldsets/Array/ArrayItem.component.js
```diff
@@ -0,0 +1,66 @@
...
+					>
+						<Icon name="talend-caret-down" />
+					</button>
+				</div>
+			}
+			{children}
+		</div>
+	);
+}
+if (process.env.NODE_ENV !== 'production') {
```
ncomont: (missing new line)

packages/forms/src/UIForm/utils/validation.js
```diff
@@ -19,28 +19,131 @@ export function validateValue(schema, value, properties, customValidationFn) {
...
+	if (error) {
+		results[key] = error;
+	}
+
+	// validate each value of the array
+	if (deepValidation && value) {
+		for (let valueIndex = 0; valueIndex < value.length; valueIndex += 1) {
+			// adapt items schema with value index
+			const indexedItems = items.map((item) => {
+				const indexedKey = [...item.key];
```
ncomont: Could you declare `indexedItems` and `indexedKey` as variables outside of this double loop (O(n2)) ?

packages/forms/src/UIForm/utils/validation.js
```diff
@@ -19,28 +19,131 @@ export function validateValue(schema, value, properties, customValidationFn) {
...
-			if (error) {
-				results[key] = error;
-			}
-		}
-		if (items) {
-			const subResults = validateAll(items, properties);
-			Object.assign(results, subResults);
-		}
+		const { key } = schema;
+		const value = key && getValue(properties, key);
```
ncomont: Please move `const value = getValue(properties, key);` outside of the loop

packages/forms/src/UIForm/utils/validation.js
```diff
@@ -70,3 +173,68 @@ export function isValid(schema, errors) {
...
+ * @param maxIndex The max item index (EXCLUDED)
+ */
+export function filterArrayErrorsKeys(errors, arrayKey, minIndex, maxIndex) {
+	const minArrayIndexKey = Number.isInteger(minIndex) && arrayKey.concat(minIndex).toString();
+	const maxArrayIndexKey = Number.isInteger(maxIndex) && arrayKey.concat(maxIndex).toString();
+
+	return Object.keys(errors)
+		.filter(errorKey => errorKey.startsWith(arrayKey))
+		.filter(errorKey => !minArrayIndexKey || errorKey >= minArrayIndexKey)
+		.filter(errorKey => !maxArrayIndexKey || errorKey < maxArrayIndexKey);
```
ncomont: A single filter could do the job here

packages/forms/src/UIForm/UIForm.component.js
jsomsanith: Oh, thanks, in fact the next line replace what I wanted to do with Object.assign.
I'll remove that

packages/forms/src/UIForm/fieldsets/Array/Array.component.js
jsomsanith: Why not :)
Ok I'll change it

packages/forms/src/UIForm/utils/validation.js
jsomsanith: No I can't. the purpose of this code is to create the new schema key for the item. The new key contains the index, so it must be in the loop.

packages/forms/src/UIForm/utils/validation.js
jsomsanith: No it depends on the schema key, so the current item in the loop.

packages/forms/src/UIForm/utils/validation.js
jsomsanith: It seemed clearer to me like that, we have a filter for each single condition, easier to understand.
We could write 
```
return Object.keys(errors)
		.filter(isOnTargetArray)
		.filter(isAfterMinIndex)
		.filter(isBeforeMaxIndex);
```

packages/forms/src/UIForm/utils/validation.js
ncomont: Oh, sorry, I didn't see `const { key } = schema;`

packages/forms/src/UIForm/utils/validation.js
ncomont: With you example, you'll potentially loop `3 * errors.length` times.

I think that the following code is not less clear, but I let you decide :

```javascript
With you example, you'll potentially loop `3 * errors.length` times.

I think that the following code is not less clear, but I let you decide :

```javascript
return Object.keys(errors)
		.filter((e) => {
			const isOnTargetArray = ...;
			const isAfterMinIndex = ...;
			const isBeforeMaxIndex = ...;
			
			return isOnTargetArray && isAfterMinIndex && isBeforeMaxIndex;
		});
```

packages/forms/src/UIForm/utils/validation.js
ncomont: :+1:

packages/forms/src/UIForm/fieldsets/Array/Array.component.js
jsomsanith: I'm always a little reluctent to separate ol/li because they are tied :/
Here the code is not too big IMO, and we have a clear view of the markup without scrolling.

packages/forms/src/UIForm/utils/validation.js
jsomsanith: `you'll potentially loop 3 * errors.length times` --> I always see it like a reactive pipeline but it's not. You're right, better perf. I'll change it

PR #652, opened 2017-07-25
chore(cmf&containers): refactor actionApi

jmfrancois:
actionAPI support well the case with an id as input to plug onclick and so on
but when you action info already exists you have to create the onclick yourself
this lead to code duplication

cmf action API support only id as string

Action container use actionAPI
cmf action support id with string or object


packages/cmf/src/action.js
```diff
@@ -65,12 +65,17 @@ function getActionInfo(context, id) {
  * Return the action object ready to be dispatched
  * This is supposed to be used outside of content type
  * @param  {object} context
- * @param  {String} id
+ * @param  {String|Object} id or the action
  * @param  {object} event event which have trigger this action
  * @param  {object} data data attached to the action
  */
 function getActionObject(context, id, event, data) {
```
jsomsanith: `id` is not necessarily an id anymore. Perhaps we should rename this param.
Proposal : name the param action and the `let` actionInfo

packages/cmf/src/action.js
```diff
@@ -65,12 +65,17 @@ function getActionInfo(context, id) {
...
-	const action = getActionInfo(context, id);
+	let action;
+	if (typeof id === 'string') {
+		action = getActionInfo(context, id);
+	} else {
+		action = id;
+	}
```
jsomsanith: I would prefer a ternary here but @glorieux won't be happy :P

jmfrancois: this is why I have used an if ;)

PR #657, opened 2017-07-27
feat(ConfirmDialog/TMC-10855): add ability to define more actions in the dialog

vyshkov:

In order to implement this task
https://jira.talendforge.org/browse/TMC-10855
we need ability to define more than one confirm action in the dialog

Image of mockup

Custom property with array of actions which is not required.


packages/components/src/ConfirmDialog/ConfirmDialog.component.js
```diff
@@ -64,7 +65,10 @@ function ConfirmDialog({
 			</Modal.Body>
 			<Modal.Footer>
 				<Action {...cancelAction} />
-				<Action {...validateAction} />
+				<div>
```
talend-glorieux: I'd rather see a class added to that div in order to style it.

packages/components/stories/ConfirmDialog.js
```diff
@@ -302,4 +302,38 @@ storiesOf('ConfirmDialog', module)
...
+			],
+		};
+		const content = (
+			<div>
+				<p>1 item will be removed PERMANENTLY</p>
+				<ul>
+					<li>Demo-Project</li>
+				</ul>
+				<p>This operation cannot be undone.</p>
+				<p>You can keep project(s) on Github or delete them there as well. What do you prefer?</p>
```
talend-glorieux: Please use a non project specific sentence.

PR #658, opened 2017-07-27
feat(components): i18n

jare-talend:
No i18n

* i18n via i18next
* implement test with i18n like https://react.i18next.com/misc/testing.html

**[] This PR introduces a breaking change**


packages/components/stories/HeaderBar.js
```diff
@@ -140,11 +143,13 @@ const props = {
 
 const decoratedStories = storiesOf('HeaderBar', module)
 	.addDecorator(story => (
-		<div>
-			{story()}
-			<div className="container" style={{ paddingTop: 40 }} />
-			<IconsProvider defaultIcons={icons} />
-		</div>
+		<I18nextProvider i18n={i18n}>
+			<div>
```
Sylvain-Talend: Since there is a wrapper declared `TranslateWrapper.component.js` above, why explicitely declare `I18nextProvider` here

packages/components/stories/config/i18n.js
```diff
@@ -0,0 +1,21 @@
...
+		// defaultNS: 'common',
+
+		// Fallback language
+		// fallbackLng: 'en',
+		debug: false,
+		wait: true, // globally set to wait for loaded translations in translate hoc
+	});
+
+
+export default i18n;
```
Sylvain-Talend: Maybe add a comment to indicate that in the future, it is in this file where we will get JSON translation file with https://github.com/i18next/i18next-xhr-backend

packages/forms/stories/config/i18n.js
```diff
@@ -0,0 +1,21 @@
...
+		// defaultNS: 'common',
+
+		// Fallback language
+		// fallbackLng: 'en',
+		debug: false,
+		wait: true, // globally set to wait for loaded translations in translate hoc
+	});
+
+
+export default i18n;
```
Sylvain-Talend: Would it be possible to declare i18 configuration once, in order to prevent this file duplication?

packages/components/src/utils/nameSpaceI18n.js
```diff
@@ -0,0 +1 @@
+export default 'tui-components';
```
jmfrancois: I prefer src/constants.js with I18N_DOMAIN please

I18n.md
```diff
@@ -0,0 +1,124 @@
...
+
+### Use the HOC TranslateWrapper
+
+```javascript
+import { TranslateWrapper } from 'react-talend-component';
+import i18n from ...
+
+import MyComponentToTranslate from ...
+
+const MyComponentTranslated = TranslateWrapper(MyComponentToTranslate, { i18n });
```
jmfrancois: Question: TranslateWrapper is for ngReact only right ?

packages/components/test/translate.js
```diff
@@ -0,0 +1 @@
+export default (key, options) => options.defaultValue;
```
jmfrancois: I would love to see this function as a the default 't' value props on all components;
src/translate.js ?

I18n.md
jare-talend: nope It is a HOC to integrate the i18n provider to the translatable components.

package.json
```diff
@@ -24,7 +25,10 @@
     "publish": "lerna publish --force-publish=*",
     "build-storybook": "export TRAVIS_BRANCH='master' && export TRAVIS_BUILD_DIR=`pwd` && .travis/after_success_static.sh && .travis/after_success_demo.sh && .travis/after_success_coverage.sh",
     "publish-storybook": "npm run build-storybook && cd .static && surge",
-    "changelog": "git log --pretty=\"format:%C(bold green)%ad%C(reset) %<(70,trunc)%s\" --date=short --color"
+    "changelog": "git log --pretty=\"format:%C(bold green)%ad%C(reset) %<(70,trunc)%s\" --date=short --color",
+    "extract-i18n": "npm run extract-i18n-components && npm run extract-i18n-containers && npm run extract-i18n-forms",
+    "extract-i18n-components": "i18next packages/components/src -r -o i18n/components -n tui-components -l en --write-old false",
```
Sylvain-Talend: `extract-i18n-containers` is not defined , to delete

packages/components/src/HeaderBar/__snapshots__/HeaderBar.snapshot.test.js.snap
```diff
@@ -1,5803 +1,10184 @@
...
-        version="1.1"
-        viewBox="0 0 16 16"
-        xmlns="http://www.w3.org/2000/svg"
-      >
-        <polygon
-          points="8,0 10.5,5 16,5.8 12,9.7 12.9,15.2 8,12.6 3.1,15.2 4,9.7 0,5.8 5.5,5 "
-        />
-      </svg>
-    </symbol>
-    <
```
jmfrancois: 10 000 lignes of snashots is way to much.
either revert to storyshots or review all the tests to make only interesting ones

packages/forms/src/widgets/EnumerationWidget/EnumerationWidget.js
```diff
@@ -772,14 +779,22 @@ class EnumerationWidget extends React.Component {
...
 }
 
-export default EnumerationWidget;
+EnumerationForm.defaultProps = {
+	t: defaultTranslateFn,
+};
+
+export {
+	EnumerationForm,
+};
```
jsomsanith: Why do you rename/export the non translated component ? The translated one without translation should work with the default values.

packages/components/stories/List.js
```diff
@@ -338,7 +344,11 @@ storiesOf('List', module)
 		);
 	})
 	.add('Tile empty list', () => {
-		const emptyListProps = cloneDeep(props);
+		const emptyListProps =
+
+
+
+          (props);
```
jsomsanith: I think this is not wanted

packages/components/stories/List.js
```diff
@@ -790,5 +800,16 @@ storiesOf('List', module)
 			<IconsProvider defaultIcons={icons} />
 			<List {...getPropsFor('tile')} />
 		</div>
+	))
+	.add('DEPRECATED - list with i18n', () => (
```
jsomsanith: Please turn that into a non deprecated story please. The translation is only on toolbar, you can make the same story with Virtualized story

packages/forms/src/Form.js
```diff
@@ -211,4 +212,7 @@ Form.defaultProps = {
 
 Form.displayName = 'TalendForm';
 
+export {
+	I18N_DOMAIN,
+};
```
jsomsanith: Why are you importing/exporting the i18n domain ?

packages/components/src/List/List.component.js
```diff
@@ -131,7 +138,7 @@ ListDisplay.propTypes = DisplayModeComponent.propTypes;
 }
  <List {...props}></List>
  */
-function List({ displayMode, id, list, toolbar, useContent, virtualized }) {
+export function List({ displayMode, id, list, toolbar, useContent, virtualized, t }) {
```
jsomsanith: Why are you exporting the non translated component ?
The translated one should behave the same as the non translated one if no configuration is provided.

packages/components/src/HeaderBar/index.js
```diff
@@ -1,3 +1,3 @@
-import HeaderBar from './HeaderBar.component';
+import Component from './HeaderBar.component';
 
-export default HeaderBar;
+export default Component;
```
jsomsanith: wat

packages/components/src/List/index.js
```diff
@@ -1,3 +1,3 @@
-import List from './List.component';
+import Component from './List.component';
 
-export default List;
+export default Component;
```
jsomsanith: wat

packages/components/src/HeaderBar/HeaderBar.component.js
```diff
@@ -260,22 +218,85 @@ HeaderBar.Help = Help;
...
+			Search: PropTypes.func,
+			User: PropTypes.func,
+			Products: PropTypes.func,
+		}),
+		t: PropTypes.func,
+	};
+}
+
+export {
+	HeaderBar,
```
jsomsanith: Still the same question : why export the non translated component as it should behave the same as translated one without conf

packages/components/src/HeaderBar/HeaderBar.snapshot.test.js
```diff
@@ -1,5 +1,245 @@
...
+		onSelect: jest.fn(),
+	},
+};
+
+describe('HeaderBar', () => {
+	it('should render default HeaderBar', () => {
+		// given
+		const headerProps = Immutable.fromJS(props).toJS();
+		// when
+		const wrapper = mount(<HeaderBar {...headerProps} />);
```
jsomsanith: You can shallow instead of mount. And snapshot wrapper.node
(same for the other tests)

packages/components/src/Badge/Badge.component.js
```diff
@@ -54,10 +58,16 @@ Badge.propTypes = {
...
 
 Badge.defaultProps = {
 	tcStyle: 'solid',
+	t: defaultTranslateFn,
 };
 
-export default Badge;
+export {
+	Badge,
+};
```
jsomsanith: Same question, why this export ?

packages/forms/src/widgets/EnumerationWidget/EnumerationWidget.js
Sylvain-Talend: It seems that for unit tests, you need to test the non-translated component (https://react.i18next.com/misc/testing.html)
Indeed, the component will not have i18n configuration set by the wrapper : `<I18nextProvider i18n={i18n}>`

packages/forms/src/Form.js
jare-talend: I exporting the i18n domain to be able to reuse it when i create the i18n config in my application and, so, to avoid this to hard-coded the domain.
e.g:

```
.init({
		resources: {
			en: {
				[I18N_DOMAIN_COMPONENTS]: TRANSLATIONS_COMPONENTS_EN,
				[I18N_DOMAIN_FORMS]: TRANSLATIONS_FORMS_EN,
			},
			fr: {
				[I18N_DOMAIN_COMPONENTS]: TRANSLATIONS_COMPONENTS_FR,
				[I18N_DOMAIN_FORMS]: TRANSLATIONS_FORMS_FR,
			},
		},
	});
```

packages/components/src/HeaderBar/HeaderBar.component.js
```diff
@@ -260,22 +217,77 @@ HeaderBar.Help = Help;
...
+			Search: PropTypes.func,
+			User: PropTypes.func,
+			Products: PropTypes.func,
+		}),
+		t: PropTypes.func,
+	};
+}
 
-export default HeaderBar;
+export default translate(I18N_DOMAIN_COMPONENTS, { i18n: i18n.init() })(HeaderBar);
```
jare-talend: We have to init the i18n else we haven't no translations if any provider is given.
https://github.com/i18next/i18next/blob/master/src/i18next.js#L247

There is not issue with the init on the provider because i18next extends the options with the existing options: https://github.com/i18next/i18next/blob/master/src/i18next.js#L37

PR #659, opened 2017-07-31
fix(TFD-1709/containers): wrapchildren for drawers support

jmfrancois:

children may have props.children which are not arrays, so drawers will fails

recursive call to wrap the children in array.

### DIFF
--- a/output/containers.eslint.txt
+++ b/output/containers.eslint.txt
@@ -1,5 +1,5 @@
 
-> react-talend-containers@0.97.0 lint:es /home/travis/build/Talend/ui/packages/containers
+> react-talend-containers@0.98.0 lint:es /home/travis/build/Talend/ui/packages/containers
 > eslint --config .eslintrc src
 
 The react/require-extension rule is deprecated. Please use the import/extensions rule from eslint-plugin-import instead.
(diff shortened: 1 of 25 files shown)


output/containers.eslint.txt
```diff
@@ -17,7 +17,8 @@ The react/require-extension rule is deprecated. Please use the import/extensions
   2:1  error  'react-test-renderer' should be listed in the project's dependencies, not devDependencies  import/no-extraneous-dependencies
 
 /home/travis/build/Talend/ui/packages/containers/src/HomeListView/HomeListView.test.js
-  2:1  error  'enzyme' should be listed in the project's dependencies, not devDependencies  import/no-extraneous-dependencies
+  2:1   error  'enzyme' should be listed in the project's dependencies, not devDependencies  import/no-extraneous-dependencies
+  5:21  error  'wrapChildren' is defined but never used                                      no-unused-vars
```
jmfrancois: it's used, what that false positive error ?

packages/containers/src/HomeListView/HomeListView.component.js
```diff
@@ -12,13 +12,22 @@ function getContent(Component, props) {
 	return (<Component {...props} />);
 }
 
+function wrapChildren(children) {
+	if (children && children.props && children.props.children) {
+		return [children, ...wrapChildren(children.props.children)];
+	} else if (children && !children.props) {
+		return [];  // this happens ony in tests with enzyme's mount
```
talend-glorieux: Please put comment on top of return.

PR #662, opened 2017-07-31
feat(UIForms): toggle widget

jsomsanith:
No Toggle widget in UIForms

Migrate Toggle widget


packages/forms/src/UIForm/fields/Toggle/Toggle.component.js
```diff
@@ -0,0 +1,43 @@
+import React, { PropTypes } from 'react';
+import Toggle from 'react-talend-components/lib/Toggle';
+import FieldTemplate from '../FieldTemplate';
+
+const ToggleWidget = (props) => {
```
talend-glorieux: Please use a regular function instead of an arrow one.

jsomsanith: Oh sorry :( copy/paste daemon has been here

packages/components/src/Toggle/__snapshots__/Toggle.snapshot.test.js.snap
```diff
@@ -8,6 +8,7 @@ exports[`Toggle should render a Checkbox 1`] = `
     htmlFor="id"
   >
     <input
+      autoFocus={undefined}
```
jmfrancois: use default value to get false here

jsomsanith: What is the benefit ? In DOM using undefine will just not add the attribute. With `false`, we will always have the attribute, most of the time with `false` value.

PR #665, opened 2017-07-31
feat(UIForms): radio or select widget

jsomsanith:
No RadioOrSelectWidget in UIForms.

Migrate it


packages/forms/src/UIForm/fields/RadioOrSelect/RadioOrSelect.component.js
```diff
@@ -0,0 +1,20 @@
+import React, { PropTypes } from 'react';
+import Radio from '../Radios';
+import Select from '../Select';
+
+const MAX_TO_RADIO = 2;
+
+export default function RadioOrSelect(props) {
+	if (props.schema.titleMap && props.schema.titleMap.length <= MAX_TO_RADIO) {
+		return (<Radio {...props} />);
```
ncomont: Parentheses are useless for single line JSX

packages/forms/src/UIForm/fields/RadioOrSelect/RadioOrSelect.component.js
```diff
@@ -0,0 +1,20 @@
...
+	}
+	return (<Select {...props} />);
```
ncomont: Parentheses are useless for single line JSX

packages/forms/src/UIForm/fields/RadioOrSelect/RadioOrSelect.component.js
```diff
@@ -0,0 +1,20 @@
+import React, { PropTypes } from 'react';
+import Radio from '../Radios';
+import Select from '../Select';
+
+const MAX_TO_RADIO = 2;
```
ncomont: Could this property be part of the component API ?

packages/forms/src/UIForm/fields/RadioOrSelect/RadioOrSelect.component.js
jsomsanith: It's a habit to wrap the jsx within parentheses, but I'll remove them :)

packages/forms/src/UIForm/fields/RadioOrSelect/RadioOrSelect.component.js
jsomsanith: In fact I asked on the Slack forms channel why this component existed, because it is based on static enum.
I don't see the benefit the set a 'RadioOrSelect' with static enum, I would have set Radio or Select widget directly.
Adding this to the api is even worst, it means the dev knows the number of possible value, et set a threshold to use one of the other, instead of using it directly.

I added it to ensure compatibility.

PR #679, opened 2017-08-08
feat(cmf-cqrs): add package to process with CQRS

talend-sromain:
When working with a backend in cqrs with a websocket, there is no components to help us.
This is an extract of a currently working stuff on data-streams

 This package contains actions / compoents / reducer & middleware to work with websockets.


packages/cmf-cqrs/.editorconfig
```diff
@@ -0,0 +1,13 @@
+# http://editorconfig.org
```
acateland: not sure editor config should be commited to the repo

packages/cmf-cqrs/src/components/ACKDispatcher/ACKDispatcher.container.js
```diff
@@ -0,0 +1,73 @@
...
+		data: { foo: 'bar' },
+		actionCreator: 'dataset:fetchAll'
+	}
+	{
+		type: 'ACK_RECEIVE_MESSAGE',
+		requestId: '123',
+	}
+
+ */
+class ACKDispatcher extends React.Component {
```
acateland: a big fat comment should be put in this section, since i hope this component will disapear fast in favor of a a clean solution living in middleware realm.
this solution mean that many useless rerender may be trigger, and lot of other issue that could happen around race conditions.

packages/cmf-cqrs/.editorconfig
jmfrancois: true we have one at the root of the UI repository

PR #683, opened 2017-08-17
feat(components): Enumeration - empty list

yyanwang:
https://jira.talendforge.org/browse/TMC-11003
The Enumeration component should display the text "The list is empty", if the list is empty. 
font style is the hint style (placeholder) on the forms;
Also after search/ filter done, if there is no result there should be a text "No results"

render EmptyListPlaceholder when there's no item in list.

### DIFF
--- a/packages/components/src/Enumeration/Enumeration.component.js
+++ b/packages/components/src/Enumeration/Enumeration.component.js
@@ -78,12 +78,14 @@ function EmptyListPlaceholder({ displayMode }) {
 	const EMPTY_LIST_PLACEHOLDER_SEARCH = 'No results';
 
 	return (<p className="help-block">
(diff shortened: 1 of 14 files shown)


packages/components/src/Enumeration/Enumeration.component.js
```diff
@@ -73,7 +73,29 @@ Enumeration.propTypes = {
...
-function ItemsEnumeration({ items, itemsProp, searchCriteria, currentEdit }) {
+function hintClasses() {
+	return classNames({
+		[theme['tc-enumeration-hint']]: true,
+		'tc-enumeration-hint': true,
+	});
+}
+
+function EmptyListPlaceholder({ displayMode }) {
+	const EMPTY_LIST_PLACEHOLDER_DEFAULT = 'The list is empty';
```
acateland: those two const could be moved outside the function declaration, so to avoid allocation and garbage collection for each executions

yyanwang: +1. fix is pushed.

PR #694, opened 2017-08-30
feat(UIForms): password input

jsomsanith:
No password in UIForms.

Migrate it


packages/forms/src/UIForm/fields/Text/__snapshots__/Text.component.test.js.snap
```diff
@@ -48,6 +48,30 @@ exports[`Text field should render input 1`] = `
...
+  errorMessage="My error message"
+  id="myForm"
+  isValid={true}
+  label="My input title"
+  labelAfter={true}
+>
+  <input
+    autoFocus={true}
+    className="form-control"
+    disabled={undefined}
```
acateland: i rather have true or false than undefined here

jsomsanith: This won't output the disabled attribute in DOM, while passing false as default will add it in DOM.
That's basically the same result, so Is there a reason ?

acateland: yes i think a false convey a better meaning than an undefined.
I set this to false instead of i may have forgotten it.

PR #695, opened 2017-08-31
feat(forms): Collapse previous array items on add

talend-glorieux:
Adding elements to an array forces the user to scroll to fill the newly added item.

Collapse all previous items when adding a new one.

### DIFF
--- a/packages/forms/src/fields/ArrayField.js
+++ b/packages/forms/src/fields/ArrayField.js
@@ -198,11 +198,7 @@ class ArrayField extends Component {
 		if (isFixedItems(schema) && allowAdditionalItems(schema)) {
 			itemSchema = schema.additionalItems;
 		}
-		formData.map((item) => {
-			const i = item || {};
-			i.isClosed = true;
-			return i;
-		});
+		formData.forEach(item => Object.assign(item, { isClosed: true }));
 		this.props.onChange([
 			...formData,
 			getDefaultFormState(itemSchema, undefined, definitions),
--- a/packages/forms/src/fields/ArrayField.js
+++ b/packages/forms/src/fields/ArrayField.js
@@ -150,7 +150,6 @@ function DefaultNormalArrayFieldTemplate(props) {
 					description={props.schema.description}
 				/>
 			) : null}
-
 			<div
 				className="row array-item-list"
 				key={`array-item-list-${props.idSchema.$id}`}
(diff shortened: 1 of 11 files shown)


packages/forms/src/fields/ArrayField.js
```diff
@@ -199,6 +198,11 @@ class ArrayField extends Component {
 		if (isFixedItems(schema) && allowAdditionalItems(schema)) {
 			itemSchema = schema.additionalItems;
 		}
+		formData.map((item) => {
+			const i = item || {};
+			i.isClosed = true;
+			return i;
+		});
```
jmfrancois: please use forEach instead of map

packages/forms/src/fields/ArrayField.js
```diff
@@ -199,6 +198,11 @@ class ArrayField extends Component {
 		if (isFixedItems(schema) && allowAdditionalItems(schema)) {
 			itemSchema = schema.additionalItems;
 		}
+		formData.map((item) => {
+			const i = item || {};
+			i.isClosed = true;
```
jmfrancois: create a const just to assign a value if exists ?
i && i.isClosed = true 
_.set(i, 'isClosed', true)

packages/forms/src/templates/ArrayFieldTemplate.js
```diff
@@ -57,23 +57,24 @@ if (process.env.NODE_ENV !== 'production') {
...
 					name="btn-new-element"
 					disabled={items.length >= maxItems}
 					onClick={onAddClick}
 				>
 					{`NEW ${props.type}`}
 				</button>}
+			{items &&
+				items.map(element =>
+					<FieldTemplate element={element} cantDelete={items.length <= minItems} />
+				)}
```
jmfrancois: map need key

packages/forms/src/templates/ArrayFieldTemplate.scss
```diff
@@ -1,6 +1,11 @@
 $tf-filterwidget-shadow-color: $shadow !default;
 
 .ArrayFieldTemplate {
+
```
jmfrancois: usless return, please remove

packages/forms/src/templates/ArrayFieldTemplate.js
talend-glorieux: eslint doesn't require it.

jmfrancois: but you know we should do it, may be a false positive.

PR #696, opened 2017-08-31
feat(UIForms): validation and triggers on blur

jsomsanith:
The triggers were only triggers explicitly within the widget, or on value change for trigger `after`.
For automatic triggers, there are some issues : 
* it is tied to the change, so the trigger is performed at each change (when user type)
* it triggers only `after` but others are usually triggered on change

The validation should be performed when user finishes his change (so basically with the triggers)

Separate triggers and validation from value change. 
* The changes are taken into account at each user actions
* Triggers are done on a `finish` event. This event is triggered depending on the widget. On Text input it is on blur. On array add/remove it is when action is performed. etc

Each widget controls when to trigger the `change` and `finish` events.


output/forms.eslint.txt
```diff
@@ -37,6 +37,9 @@ The react/require-extension rule is deprecated. Please use the import/extensions
 /home/travis/build/Talend/ui/packages/forms/src/UIForm/fields/CheckBox/SimpleCheckBox.component.js
   8:4  error  Form controls using a label to identify them must be programmatically associated with the control using htmlFor  jsx-a11y/label-has-for
 
+/home/travis/build/Talend/ui/packages/forms/src/UIForm/fields/Datalist/Datalist.component.js
+  135:29  error  A space is required after ','  comma-spacing
```
talend-glorieux: Please fix that error.

packages/forms/src/UIForm/UIForm.component.js
```diff
@@ -37,51 +38,81 @@ export default class UIForm extends React.Component {
...
+	 * This allows to perform triggers/validation while changing a value
+	 * (ex: add an element in array)
+	 * @param widgetChangeErrors Change errors hook, allows any widget to manipulate the errors map
+	 * (ex: shift the errors in array elements on remove)
+	 */
+	onFinish(event, { schema, value }, { deepValidation = false, widgetChangeErrors } = {}) {
+		// validate current field
+		const newValue = value !== undefined ?
+			value :
+			getValue(this.props.properties, schema.key);
```
talend-glorieux: Please avoid using multi-line ternaries.

packages/forms/src/UIForm/Widget/Widget.component.js
```diff
@@ -6,7 +6,7 @@ import defaultWidgets from '../utils/widgets';
 import { getValue } from '../utils/properties';
 
 export default function Widget(props) {
-	const { errors, formName, onChange, onTrigger, properties, schema, widgets } = props;
+	const { errors, formName, onChange, onFinish, onTrigger, properties, schema, widgets } = props;
```
talend-glorieux: A lot of spreading for values that are only used once.

packages/forms/src/UIForm/fields/CheckBox/SimpleCheckBox.component.js
```diff
@@ -11,6 +11,7 @@ export default function SimpleCheckBox({ id, label, onChange, schema, value }) {
 					autoFocus={autoFocus}
 					disabled={disabled}
 					label={label}
+					onBlur={event => onFinish(event, { schema })}
```
talend-glorieux: Why did the onBlur got renamed to onFinish?

jsomsanith: Because it's a widget props. When user finish to change value, this is triggered. In checkbox, it is on blur, so if the checkbox is changed multiple time, the onFinish will be triggered once, on blur

talend-glorieux: OK. So there is a clear difference between onBlur and onFinish. 
Sounds good.

PR #699, opened 2017-09-01
chore(UIForms): disabled and readonly set to false by default

jsomsanith:
@acateland 's happiness is one of my biggest concern. That's the reason of this PR.
Currently, the disabled/readonly props of UIForms fields are undefined by default.

Set them to false by default


packages/forms/src/UIForm/fields/Button/SingleButton.component.js
```diff
@@ -6,7 +6,7 @@ import classNames from 'classnames';
 export default function SingleButton({ className, id, onTrigger, schema }) {
 	const {
 		bsStyle,
-		disabled,
+		disabled = false,
```
frassinier: Any way to do this with defaultProps instead?

acateland: we should not defaultProps is lot slower mecanism and is framework bound when this way is faster and manguage bound

jsomsanith: No we can't it's not a props but a property of a props. If `schema` is provided without `disabled` or `readOnly` property, they will be undefined, even with a default props defined

frassinier: :+1:

PR #706, opened 2017-09-04
chore(versions): add addons and update all packages

jmfrancois:

versions.js has a stable set of versions.
It doesn't support common react addons that may used accross talend project.

let's add addons used in TFD and update all current packages in UI


packages/components/src/VirtualizedList/ListGrid/__snapshots__/ListGrid.test.js.snap
```diff
@@ -48,6 +48,7 @@ exports[`ListGrid should render no-rows component 1`] = `
       noRowsRenderer={[Function]}
       onRowsRendered={[Function]}
       onScroll={[Function]}
+      onScrollbarPresenceChange={[Function]}
```
jmfrancois: this is due to react-virtualized update

packages/components/src/VirtualizedList/ListTable/__snapshots__/ListTable.test.js.snap
```diff
@@ -109,6 +109,7 @@ exports[`ListGrid should render no-rows component 1`] = `
         noRowsRenderer={[Function]}
         onRowsRendered={[Function]}
         onScroll={[Function]}
+        onScrollbarPresenceChange={[Function]}
```
jmfrancois: this is due to react-virtualized update

version.js
```diff
@@ -17,6 +17,27 @@ program
 const REACT_VERSION = '15.5.4';
 const JEST_VERSION = '20.0.3';
 
+const ADDONS = {
```
jsomsanith: This list doesn't seem to be used

version.js
```diff
@@ -241,12 +241,12 @@ files.forEach((ppath) => {
...
-			}
-			//fs.unlink(yarnLock);
-		}
+		// const yarnLock = path.join(path.dirname(ppath), 'yarn.lock');
+		// if (fs.existsSync(yarnLock)) {
+		// 	if (!program.quite) {
+		// 		console.log(`delete ${yarnLock}`);
+		// 	}
+		// 	fs.unlink(yarnLock);
+		// }
```
jsomsanith: Please remove that, you have git history to get it back if needed

package.json
```diff
@@ -8,7 +8,7 @@
   },
   "scripts": {
     "postinstall": "lerna clean --yes && lerna exec -- yarn --pure-lockfile --no-progress --mutex network && ./copylibs.js",
-    "update-versions": "./version.js",
+    "update-versions": "./version.js && lerna exec -- yarn && ./copylibs.js",
```
jmfrancois: this is to update the yarn.lock

jmfrancois:
seems to be a screenshot issue, when you look at the live demo we still have italic + light grey

jmfrancois:
Now I'm sure: This is because we set the 2 last browser and firefox 50 is not part of this.
So the :placeholder is supported by firefox and we have no -moz prefix in the generated css

PR #707, opened 2017-09-04
fix(List): check for an optional props definition

jsomsanith:
When there is no titleProps passed to List, the mapping from List to VirtualizedList crashes.

Test for titleProps to be defined.


jmfrancois:
a unit test  would be nice

PR #712, opened 2017-09-07
feat(cmf): add saga router to handle side effects

talend-sromain:
To handle some side effects ( for example route change ) we don't have some elegant solution to catch actions with some context to make side effect behaviours

Add saga router to define in your application an association of routes & actions to yield. as it use generator. The generator would be canceled on route change. more documentation about redux saga here : https://redux-saga.js.org/


packages/cmf/package.json
```diff
@@ -1,109 +1,106 @@
...
-    "redux-storage-decorator-filter": "^1.1.8",
-    "redux-storage-decorator-immutablejs": "^1.0.4",
-    "redux-storage-engine-localstorage": "^1.1.4",
-    "redux-thunk": "2.2.0"
-  },
-  "version": "0.107.0",
-  "dependencies": {
-    "hoist-non-react-statics": "^1.2.0"
-  }
+	"name": "react-cmf",
```
jmfrancois: the style for json is 
```js
[*.json]
indent_style = space
indent_size = 2
```
from the editorconfig.
So please use 2 space here

packages/cmf/src/sagaRouter/router.js
```diff
@@ -0,0 +1,65 @@
+import 'babel-polyfill';
```
jmfrancois: this should not be done here

packages/cmf/src/sagaRouter/router.js
```diff
@@ -0,0 +1,65 @@
...
+			const routeFragment = routeFragments[index];
+			const match = matchPath(currentLocation.pathname, { path: routeFragment });
+			const maybeSaga = sagas[routeFragment];
+			if (shouldStartSaga(maybeSaga, match)) {
+				sagas[routeFragment] = {
+					saga: yield spawn(routes[routeFragment], match.params),
+					match,
+				};
+			} else if (shouldCancelSaga(maybeSaga, match)) {
+				if (maybeSaga) {
```
jare-talend: useful? we have already a test on maybeSaga in shouldCancelSaga to enter in this block

packages/cmf/src/sagaRouter/router.js
```diff
@@ -0,0 +1,65 @@
...
+					yield cancel(maybeSaga.saga);
+				}
+			} else if (shouldRestartSaga(maybeSaga, match)) {
+				if (maybeSaga) {
```
jare-talend: same thing

packages/cmf/src/sagaRouter/router.js
talend-sromain: not that usefull, i fix that

packages/cmf/src/sagaRouter/router.js
talend-sromain: not that usefull, i fix that

packages/cmf/src/sagaRouter/matchPath.js
```diff
@@ -0,0 +1,58 @@
...
+	const [url, ...values] = match;
+	const isExact = pathname === url;
+
+	if (exact && !isExact) return null;
+
+	return {
+		path, // the path pattern used to match
+		url: path === '/' && url === '' ? '/' : url, // the matched portion of the URL
+		isExact, // whether or not we matched exactly
+		params: keys.reduce((memo, key, index) => {
```
jare-talend: you can use this:

params: keys.reduce((memo, key, index) => ({
...memo,
[key.name] = values[index],
});

talend-sromain: it's code from react-router v4, i don't think it's usefull to replace it. As it's now, we have this file, but when migration on v4 would be done, this file would be replace by the one from the dependency

jare-talend: ok

PR #715, opened 2017-09-10
doc: update badges

jmfrancois:

we can't see at first if there is a problem with our dependencies

add badges to watch them all


talend-glorieux:
That's a lot of badges.
Is there a way to put them inside a table?

jmfrancois:
david-dm doesn't support lerna mono repository :'(

PR #718, opened 2017-09-11
feat(Drawer): update stacked Drawer layout

vyshkov:

Frontify definition for the Drawer component updated http://guidelines.talend.com/document/92132#/navigation-layout/drawer-br

We need to center drawer content like on the picture:
image

Updated styles and layout

# Breaking change
If you have wide drawer content (like multi-column forms) you will need to update them.

Here an example of TMC task https://jira.talendforge.org/browse/TMC-11183


output/components.sasslint.txt
```diff
@@ -3,11 +3,14 @@
 > sass-lint -v -q
 
 
+src/Drawer/Drawer.scss
+  92:3  warning  Mixed tabs and spaces  indentation
```
talend-glorieux: Please fix that issue.

packages/components/src/Drawer/Drawer.scss
```diff
@@ -46,15 +48,36 @@ $tc-action-bar-background-color: $concrete !default;
...
+}
+
 .tc-drawer-header {
-	padding: $padding-normal;
 	background-color: $brand-primary-lighter;
 
-	h1 {
+	.tc-drawer-header-title {
+		padding: $padding-normal;
+		position: relative;
```
frassinier: Why is this relative?

packages/components/src/Drawer/Drawer.scss
```diff
@@ -46,15 +48,36 @@ $tc-action-bar-background-color: $concrete !default;
...
+		}
+	}
+}
+
 .tc-drawer-header {
-	padding: $padding-normal;
 	background-color: $brand-primary-lighter;
 
-	h1 {
+	.tc-drawer-header-title {
```
frassinier: Please, use nesting, DRY
```scss
.tc-drawer-header {
  &-title {
    /* ... */
  }
}
```

packages/components/src/Drawer/Drawer.scss
```diff
@@ -46,15 +48,36 @@ $tc-action-bar-background-color: $concrete !default;
...
+		padding: $padding-normal;
+		position: relative;
+	}
+
+	.tc-drawer-header-title h1 {
```
frassinier: Please, use nesting, DRY
```scss 
.tc-drawer-header {
  &-title {
    h1 {
    }
  }
}
```

packages/components/src/Drawer/Drawer.component.js
```diff
@@ -93,8 +93,8 @@ function DrawerTitle({ title, children, onCancelAction }) {
 		return null;
 	}
 	return (
-		<div>
-			<div className={theme['tc-drawer-header']}>
+		<div className={theme['tc-drawer-header']}>
+			<div className={theme['tc-drawer-header-title']}>
```
frassinier: If we can have ellipsis, I suggest to deal with `title` attribute to display full content if needed.

packages/components/src/Drawer/Drawer.scss
frassinier: This is optional

packages/components/src/Drawer/Drawer.scss
frassinier: This is optional

packages/components/src/Drawer/Drawer.scss
vyshkov: @frassinier because 'cancel' cross button has position absolute.

PR #720, opened 2017-09-12
fix(i18n): createInstance instead of init

jare-talend:
Every time that we load a component, we lose the configuration done on the application

To pass a i18n by default to the component, we clone the i18n instance.
- If a previous instance exists, it will be ignored. translate take the i18n in the context
- If any instance exists, so he creates an new instance and uses it. We need to pass some values to allow to init the internal services (https://github.com/i18next/i18next/issues/936#issuecomment-307550677)


output/components.eslint.txt
```diff
@@ -12,21 +12,24 @@ The react/require-extension rule is deprecated. Please use the import/extensions
...
   4:2  error  Unexpected require()                                                                    global-require
 
 /home/travis/build/Talend/ui/packages/components/src/HeaderBar/HeaderBar.component.js
-  293:58  error  Caution: `i18n` also has a named export `init`. Check if you meant to write `import {init} from 'i18next'` instead  import/no-named-as-default-member
+  293:1   error  Line 293 exceeds the maximum line length of 100                                                                                         max-len
+  293:58  error  Caution: `i18n` also has a named export `createInstance`. Check if you meant to write `import {createInstance} from 'i18next'` instead  import/no-named-as-default-member
 
 /home/travis/build/Talend/ui/packages/components/src/List/List.component.js
-  177:58  error  Caution: `i18n` also has a named export `init`. Check if you meant to write `import {init} from 'i18next'` instead  import/no-named-as-default-member
+  177:58  error  Caution: `i18n` also has a named export `createInstance`. Check if you meant to write `import {createInstance} from 'i18next'` instead  import/no-named-as-default-member
```
talend-glorieux: Please fix those errors

output/forms.eslint.txt
```diff
@@ -54,7 +54,8 @@ The react/require-extension rule is deprecated. Please use the import/extensions
   79:3  error  Prop type `object` is forbidden  react/forbid-prop-types
 
 /home/travis/build/Talend/ui/packages/forms/src/widgets/EnumerationWidget/EnumerationWidget.js
-  808:53  error  Caution: `i18n` also has a named export `init`. Check if you meant to write `import {init} from 'i18next'` instead  import/no-named-as-default-member
+  808:1   error  Line 808 exceeds the maximum line length of 100                                                                                         max-len
+  808:53  error  Caution: `i18n` also has a named export `createInstance`. Check if you meant to write `import {createInstance} from 'i18next'` instead  import/no-named-as-default-member
```
talend-glorieux: Please fix those errors.

I18n.md
```diff
@@ -17,7 +17,7 @@ Each packages has their own namespace:
 
 Each components who has to be internationalizable are exported by default with translate from 'react-i18next' with its namespace.
 
-**The component can be to use with or without a internationalizable application. For this, in the creation of the component, we pass it a default i18n. Translate will check if a i18n context exists and uses it else it uses the default i18n. We need to initialised the i18n passed to the component to translate the key.**
+**The component can be to use with or without a internationalizable application. For this, in the creation of the component, we pass it a default i18n. Translate will check if a i18n context exists and uses it else it uses the default i18n. We need to initialised the i18n passed to the component to translate the key. We use createInstance (with empty options and function https://github.com/i18next/i18next/issues/936#issuecomment-307550677)  instead of init to avoid to lose the previous init that the application done on the application. **
```
jsomsanith: The component can be use within and internationalized application or not. We pass a default i18n configuration that will be ignored if one is provided in the context by `react-i18next` provider.

`We use createInstance (with empty options and function https://github.com/i18next/i18next/issues/936#issuecomment-307550677)  instead of init to avoid to lose the previous init that the application done on the application.` --> this is an implementation comment. this should not be here but next to the code. (This won't be duplicated, see next comments)

I18n.md
```diff
@@ -30,7 +30,7 @@ const HelloWorld = ({t}) => (
  <div>{`Hi ${t('HELLO', { defaultValue: 'Hello world' })}`}</div>
 );
 
-export default translate(I18N_DOMAIN, { i18n: i18n.init() })(HelloWorld);
+export default translate(I18N_DOMAIN, { i18n: i18n.createInstance({}, () => {}) })(HelloWorld);
```
jsomsanith: Please externalize the only i18n default configuration.
So each translated component import it and set it instead of creating a default configuration on each component.
+ that will fix the linter error on line of code size

output/components.eslint.txt
jsomsanith: +1
Importing createInstance won't have the same context issue than importing init.

jsomsanith: +1
Importing createInstance won't have the same context issue than importing init.

packages/components/src/translate.js
```diff
@@ -1 +1,12 @@
...
+import { createInstance } from 'i18next';
+
+// https://github.com/i18next/i18next/issues/936#issuecomment-307550677
+const DEFAULT_I18N = createInstance({}, () => {});
+
+export function getDefaultTranslate(key, options) {
+	return options.defaultValue;
+}
+
+export function getDefaultI18n() {
```
jare-talend: `export const DEFAULT_I18N = createInstance({}, () => {});,` 
then 

`import { DEFAULT_I18N } from ...
translate(I18N_DOMAIN_FORMS, { i18n: DEFAULT_I18N })(EnumerationForm);
`

don't work

jsomsanith: Why ? What is the error ?

I18n.md
```diff
@@ -17,20 +17,21 @@ Each packages has their own namespace:
 
 Each components who has to be internationalizable are exported by default with translate from 'react-i18next' with its namespace.
 
-**The component can be to use with or without a internationalizable application. For this, in the creation of the component, we pass it a default i18n. Translate will check if a i18n context exists and uses it else it uses the default i18n. We need to initialised the i18n passed to the component to translate the key.**
+**The component can be to use with or without a internationalizable application. For this, in the creation of the component, we pass it a new instance of I18n. Translate will check if a i18n context exists and uses it else it uses the default i18n. We need to initialised the i18n passed to the component to translate the key. **
 
 ```javascript
 import React from 'react';
-import i18n from 'i18next';
+import { createInstance } from 'i18next';
```
jsomsanith: You can remove that

PR #733, opened 2017-09-18
fix(cmf/saga-router): fix cancel/spawn order

acateland:
unmatched sagas could be cancel after a new matching sagas where started.
Making finally block containing concurrent state change command to be executed on the wrong unnatural order.

Make sure all route/sagas match are canceled if needed, before checking all route for spawn.
Since cancellation is a blocking command all finally are executed before new sagas are spawned.


packages/cmf/src/sagaRouter/router.js
```diff
@@ -30,25 +30,47 @@ function shouldRestartSaga(maybeSaga, match) {
...
 					match,
 				};
-			} else if (shouldCancelSaga(maybeSaga, match)) {
-				yield cancel(maybeSaga.saga);
-			} else if (shouldRestartSaga(maybeSaga, match)) {
-				yield cancel(maybeSaga.saga);
+			}
+			index += 1;
+		}
+		for (let index = 0; index < routeFragments.length;) {
```
jsomsanith: Instead of running the loop twice, can you populate a sagaQueue in the first loop where you store the saga to start (routes/params/match) ?
Then after the loop, you only loop on this sagaQueue to start the sagas, avoiding `parseSagaState` computation on each element twice.

packages/cmf/src/sagaRouter/router.js
```diff
@@ -30,31 +30,55 @@ function shouldRestartSaga(maybeSaga, match) {
...
+	return {
+		routeFragment,
+		match: matchPath(currentLocation.pathname, { path: routeFragment }),
+		maybeSaga: sagas[routeFragment],
+	};
+}
+
 export default function* routerSaga(history, routes) {
 	const sagas = {};
+	let shouldStart = [];
```
jsomsanith: Remove that and declare le `const` in the While loop

packages/cmf/src/sagaRouter/router.js
```diff
@@ -30,31 +30,55 @@ function shouldRestartSaga(maybeSaga, match) {
...
 					saga: yield spawn(routes[routeFragment], match.params),
 					match,
 				};
+			} else if (shouldStartSaga(maybeSaga, match)) {
+				shouldStart.push(routeFragment);
 			}
 			index += 1;
 		}
+		for (let index = 0; index < shouldStart.length;) {
+			const { routeFragment, match } = parseSagaState(
```
jsomsanith: Can you store the result of parseSagaState() in shouldStart instead of calling it again ?

packages/cmf/src/sagaRouter/router.js
```diff
@@ -30,31 +30,49 @@ function shouldRestartSaga(maybeSaga, match) {
...
+				index,
+			);
+			if (shouldCancelSaga(maybeSaga, match)) {
 				yield cancel(maybeSaga.saga);
 			} else if (shouldRestartSaga(maybeSaga, match)) {
 				yield cancel(maybeSaga.saga);
 				sagas[routeFragment] = {
 					saga: yield spawn(routes[routeFragment], match.params),
 					match,
 				};
```
jmfrancois: please use shouldStart.push here

PR #734, opened 2017-09-18
feat(SlidePanel/TSCH-52): new toggle button

vyshkov:

Frontify definition for SlidePanel updated http://guidelines.talend.com/document/92132#/navigation-layout/left-side-panel.

In this PR i changed only the button. Hover color will be updated with all colors in separate PR.


packages/components/stories/SidePanel.js
```diff
@@ -9,6 +9,7 @@ const icons = {
 	'talend-dataprep': talendIcons['talend-dataprep'],
 	'talend-download': talendIcons['talend-download'],
 	'talend-star': talendIcons['talend-star'],
+	'talend-opener': talendIcons['talend-opener'],
```
talend-sromain: Can you also add this icon to containers stories ( like homelist view )

packages/components/src/SidePanel/SidePanel.scss
```diff
@@ -6,6 +6,8 @@
 /// Side panel toggle button color
 /// @type Color
 $tc-side-panel-toggle-color: tint($brand-primary, 75);
+$list-item-hover-color: darken($brand-primary, 2%);
```
talend-sromain: It's not the background color of the guideline ( maybe in another pr ? )
Also, the selected background color is not the same in the guideline
Would we do another pr for both background color & this one is only about toggle button ?

vyshkov: @talend-sromain i'll update all sidebar colors in separate PR. Because if i update only toggle button (to use white highlight) it will look bad.

mloussot: @vyshkov indeed, it would... can you add me as reviewer also on this other PR ? Thanks

vyshkov: @mloussot, ok, i'll add you

mloussot: Thanks

PR #747, opened 2017-09-25
	feat(UIForms): ListView widget

jsomsanith:
*WARNING*: This is only on UIForms, not the current forms
1. No ListView widget
2. ListView component doesn't support multiple instance in the DOM because of ideas on checkboxes.

1. Migrate it
2. Introduce optional id on component props to alter the checkboxes ids. With no id, the component acts like today to ensure having the same ids to not break QA.

Missing feature (?) :
* formContext.handleAction. Seems not to work and not to be used
* disabled --> on chexkboxes. Not present in current form widget, needed ?


packages/forms/src/UIForm/fields/ListView/ListView.component.test.js
```diff
@@ -0,0 +1,285 @@
...
+		};
+	});
+
+	describe('render', () => {
+		it('should render ListView', () => {
+			// when
+			const wrapper = shallow(<ListView {...props} />);
+
+			// then
+			expect(wrapper.node).toMatchSnapshot();
```
jmfrancois: please use wrapper.getNode

packages/forms/src/UIForm/fields/ListView/ListView.component.test.js
```diff
@@ -0,0 +1,285 @@
...
+			const instance = ReactDOM.render(<ListView {...props} />, node);
+			expect(instance.state).toMatchSnapshot();
+
+			const allValues = props.schema.titleMap.map(option => option.value);
+
+			// when : trigger a props update
+			ReactDOM.render(<ListView {...props} value={allValues} />, node);
+
+			// then
+			expect(instance.state).toMatchSnapshot();
```
jmfrancois: This too hard to check if this tests is working.
Please find a better way to make this assertion.
I don't see why instance should have changed here, I must read both snapshots.
May be you should check previous state and next state ?

packages/forms/src/UIForm/fields/ListView/ListView.component.test.js
```diff
@@ -0,0 +1,285 @@
...
+
+			// when : trigger a props update
+			ReactDOM.render(<ListView
+				{...props}
+				schema={alternativeSchema}
+				value={allValues}
+			/>, node);
+
+			// then
+			expect(instance.state).toMatchSnapshot();
```
jmfrancois: same here

packages/forms/src/UIForm/fields/ListView/ListView.component.test.js
```diff
@@ -0,0 +1,285 @@
...
+			// given
+			const wrapper = mount(<ListView {...props} />);
+			const initialHeader = wrapper.find('.tc-listview-header').at(0);
+			expect(initialHeader.node).toMatchSnapshot();
+
+			// when
+			initialHeader.find('button').at(0).simulate('click');
+
+			// then
+			expect(wrapper.find('.tc-listview-header').at(0).node).toMatchSnapshot();
```
jmfrancois: snapshots doesn't test anything here.
please try to select the item which is the search button and make it equal 'found'

packages/forms/src/UIForm/fields/ListView/ListView.component.test.js
```diff
@@ -0,0 +1,285 @@
...
+
+		it('should filter items on input change', () => {
+			// given
+			const wrapper = mount(<ListView {...props} />);
+
+			// when
+			filter(wrapper, 'ia');
+
+			// then
+			expect(wrapper.find('.tc-listview-items').at(0).node).toMatchSnapshot();
```
jmfrancois: snapshots doesn't test anything here.

packages/forms/src/UIForm/fields/ListView/ListView.component.test.js
```diff
@@ -0,0 +1,285 @@
...
+
+		it('should show no result message', () => {
+			// given
+			const wrapper = mount(<ListView {...props} />);
+
+			// when
+			filter(wrapper, 'lol');
+
+			// then
+			expect(wrapper.find('.tc-listview').at(0).node).toMatchSnapshot();
```
jmfrancois: same, find the no result message

packages/forms/src/UIForm/fields/ListView/ListView.component.test.js
jsomsanith: I disagree
* The first snapshot set the initial state with a label
* The second one set the next state with the input

The 2 snapshots define the test `should switch to input mode on search button click`

packages/forms/src/UIForm/fields/ListView/ListView.component.test.js
jsomsanith: Snapshots test that the suggestions in DOM are filtered. If I break the filter, the snapshot will add 2 more values in items list

packages/forms/src/UIForm/fields/ListView/ListView.component.js
```diff
@@ -0,0 +1,222 @@
...
+		this.props.onFinish(event, payload);
+	}
+
+	/**
+	 * Search input change
+	 * @param { Object } event The change event
+	 * @param { Array } value
+	 */
+	onInputChange(event, { value }) {
+		if (this.timerSearch) {
```
ncomont: You could delete this test. `clearTimeout(undefined);` is not a problem, so this method could be simplified : 

```javascript
clearTimeout(this.timerSearch);
this.timerSearch = setTimeout(() => {
	this.setState(getItemsProps(this.state.items, value));
}, 400);
```

packages/forms/src/UIForm/fields/ListView/ListView.component.js
```diff
@@ -0,0 +1,222 @@
...
+			this.timerSearch = null;
+		}, 400);
+	}
+
+	/**
+	 * Search input ENTER/ESC management
+	 * @param { Object } event The keydown event
+	 */
+	onInputKeyDown(event) {
+		if (event.keyCode === keycode('enter')) {
```
ncomont: A `switch/case` or an `else if` could be more efficient here.

packages/forms/src/UIForm/fields/ListView/ListView.component.js
```diff
@@ -0,0 +1,222 @@
...
+			);
+		} else if (this.state.toggleAllChecked) {
+			// User uncheck all items
+			checkedItems = [];
+		} else {
+			// User check all items
+			checkedItems = this.state.items;
+		}
+
+		const value = checkedItems.map(item => item.value);
```
ncomont: Useless assignation

packages/components/src/ListView/Items/Items.component.js
```diff
@@ -99,7 +99,10 @@ class Items extends React.PureComponent {
 	}
 
 	renderItem(item, index) {
-		const computedIndex = this.hasToggleAll ? index + 1 : index;
+		let computedIndex = this.hasToggleAll ? index + 1 : index;
```
acateland: i think this is really hight on id collision chance, even with `item` appended, i would not allow to render an id if the component user didn't provide an prefix or postfix id fragment

packages/forms/src/UIForm/fields/ListView/ListView.utils.js
```diff
@@ -0,0 +1,71 @@
...
+/**
+ * Update the check status of the list items
+ * @param { Array } items The listView items
+ * @param { Array } value The listView value
+ * @param { string } searchCriteria The filter criteria
+ * @returns { Object } The items related state
+ */
+export function updateItems(items, value, searchCriteria) {
+	function updateChecked(item) {
+		const checked = value.indexOf(item.value) !== -1;
```
ncomont: You could use `includes()` here.

packages/forms/src/UIForm/utils/widgets.js
```diff
@@ -24,21 +25,24 @@ const widgets = {
...
 	select: Select,
 	submit: Button,
 	text: Text,
 	textarea: TextArea,
+
+	// widgets
+	buttons: Buttons,
+	checkboxes: CheckBoxes,
+	datalist: Datalist,
+  keyValue: KeyValue,
```
ncomont: oops

packages/forms/src/UIForm/fields/ListView/ListView.component.js
```diff
@@ -0,0 +1,222 @@
...
+
+		this.defaultHeaderActions = [
+			{
+				icon: 'talend-search',
+				id: `${props.id}-search`,
+				label: 'Search for specific values',
+				onClick: this.switchToSearchMode.bind(this),
+			},
+		];
+		this.searchHeaderActions = [
```
acateland: defaultSearchHeaderActions maybe ?

packages/forms/src/UIForm/fields/ListView/ListView.component.js
```diff
@@ -0,0 +1,222 @@
...
+	}
+
+	/**
+	 * Toggle an item
+	 * @param { Object } event The toggle event
+	 * @param { Object } changedItem The item to toggle
+	 */
+	onToggleItem(event, changedItem) {
+		const value = this.state.items
+			.filter((item) => {
```
acateland: if im not mistaken only element is affected each time, a find will be faster in this case, since its stop iterrating once the value is found.

packages/forms/src/UIForm/fields/ListView/ListView.component.js
jsomsanith: For only 2 cases without any default, I'll use `else if`

packages/forms/src/UIForm/fields/ListView/ListView.component.js
jsomsanith:
```
this.onChange(event, checkedItems.map(item => item.value));
```
is much more hard to read than
```
const value = checkedItems.map(item => item.value);
this.onChange(event, value);
```

packages/forms/src/UIForm/utils/widgets.js
jsomsanith: oops :)

packages/forms/src/UIForm/fields/ListView/ListView.component.js
jsomsanith: No this rebuild the new value, so we have to iterate over all values
It's just that if we toggle item X, we need to test the reverse checked value to know if it is in the values

PR #749, opened 2017-09-25
ObjectViewer style improvements

sgendre:
Fixes reviews from https://jira.talendforge.org/projects/TFD/issues/TFD-2086?filter=allopenissues

Take some points directly and forward some others to UX validation


packages/components/src/Icon/Icon.scss
```diff
@@ -15,6 +15,10 @@ $tc-icon-size: $svg-md-size !default;
 	}
 }
 
+.link {
```
acateland: what is the point of this modification since the icon can be already wrapper in a link where it is used ? 
https://github.com/Talend/ui/blob/master/packages/components/src/Icon/Icon.component.js#L75

sgendre: a tag with xlinkHref attribute does not provide the right pointer, so I enforced it.

output/components.sasslint.txt
```diff
@@ -3,11 +3,14 @@
 > sass-lint -v -q
 
 
+src/ObjectViewer/JSONLike/JSONLike.scss
+  65:21  warning  Hex notation should all be upper case  hex-notation
```
jmfrancois: prettier put it in lowercase ... what should we do ?

acateland: desactivate prettier for scss or change the lint rule.

PR #753, opened 2017-09-26
fix(list): remove the onclick on the title if no action

jare-talend:
There is always an click event on each title items

If action defines on title, we set this action. 
Fix some style errors when the title is not wrapped in a button


packages/components/src/List/ItemTitle/ItemTitle.scss
```diff
@@ -4,6 +4,14 @@ $tc-list-item-title-icon-size: $svg-lg-size !default;
...
 	display: flex;
 	min-width: 0;
+	line-height: $btn-line-height;
+
+	& > span {
+		padding: 0 $padding-normal;
+		overflow: hidden;
+		text-overflow: ellipsis;
+		white-space: nowrap;
+	}
```
jsomsanith: The not virtualized list will die, next time, don't bother fixing that.

packages/components/stories/List.js
```diff
@@ -354,6 +354,25 @@ storiesOf('List', module)
 		);
 	})
 
+	.add('Virtualized - table display without click on title', () => {
```
jsomsanith: Can you move that to special cases below (around empty) and rename it more generically ?
`Virtualized - title without click`

PR #755, opened 2017-09-27
fix(list) - conform to guidelines

Sylvain-Talend:
- Virtualized list is not fully conform to guidelines (border color, etc)

Virtualized list is conform to guidelines: 
- CSS fixing
- add a story to display 4 columns instead of 3 by default


packages/components/stories/VirtualizedList.js
```diff
@@ -457,6 +486,65 @@ storiesOf('Virtualized List', module)
 			</section>
 		</div>
 	))
+	.add('List > Large with mor ethan 6 items', () => (
```
jare-talend: ... more than ...

packages/components/.storybook/preview-head.html
```diff
@@ -44,4 +44,9 @@
     .virtualized-list div.tc-list-cell-modified {
         flex: 0 0 90px;
     }
+
```
jlame: can make sense to show how to switch to different number of column, but as its not done using props i guess this shouldn't be here.

Sylvain-Talend: Ok I removed it

PR #776, opened 2017-10-05
feat(components): add inProgress feature to Virtualized List

Innes333:
inProgress feature is added to Virtualized List

List props are extended to have bool prop inProgress

https://jira.talendforge.org/browse/TPSVC-3014 - related bug


packages/components/stories/List.js
```diff
@@ -393,6 +393,21 @@ storiesOf('List', module)
...
+		loadingListProps.list.inProgress = true;
+		return (
+			<div style={{ height: '60vh' }}>
+				<h1>List</h1>
+				<p>When the list is loading, a CircularProgress is displayed instead of the rows.</p>
+				<IconsProvider defaultIcons={icons} />
+				<h2>Table</h2>
+				<List {...loadingListProps} virtualized />
+				<h2>Large</h2>
+				<List {...loadingListProps} displayMode="large" virtualized />
```
jmfrancois: I think there is a problem in your example:
Table -> virtualized 
Large -> same without virtualized ?
I think just one is enough (virtualized will becomes the default soon)

output/components.sasslint.txt
```diff
@@ -3,6 +3,9 @@
 > sass-lint -v -q
 
 
+src/List/List.scss
+  35:18  warning  No unit allowed for values of 0  zero-unit
```
jmfrancois: please fix lint issue

packages/components/src/List/List.scss
```diff
@@ -24,5 +24,14 @@
 		flex-grow: 1;
 		flex-shrink: 1;
 		box-shadow: $shadow-default;
+		height: 100%;
+	}
+
+	:global(.tc-list-progress) {
```
jsomsanith: This should not be defined here. It is something in VirtualizedList, put it in its style please.
Plus, we can use css modules instead of global class

packages/components/src/List/List.scss
```diff
@@ -24,5 +24,14 @@
...
+		display: flex;
+		align-items: center;
+		justify-content: center;
+		height: 100%;
+		margin-bottom: 0px;
```
jsomsanith: Margin-bottom seems not necessary

packages/components/stories/List.js
jsomsanith: @jmfrancois  The second one is large with virtualized
But agree, you can remove the large example

packages/components/src/VirtualizedList/VirtualizedList.test.js
```diff
@@ -83,4 +83,28 @@ describe('VirtualizedList', () => {
 		// then
 		expect(rendererInstance).toMatchSnapshot();
 	});
+	it('should render CircularProgress', () => {
```
jsomsanith: Add a new line before please

packages/components/src/List/List.scss
Innes333: Which way is better from your point of you, add as css module?

packages/components/src/VirtualizedList/VirtualizedList.component.js
```diff
@@ -29,6 +32,16 @@ function VirtualizedList(props) {
 		selectionToggle,
 	});
 
+	if (inProgress) {
+		return (
+			<div className="tc-list-progress">
+				<Status
+					status="inProgress"
+					label={t('LIST_LOADING_LABEL', { defaultValue: 'Loading...' })}
```
frassinier: Avoid using three dots, please change to ellipsis char instead

packages/components/src/VirtualizedList/VirtualizedList.component.js
```diff
@@ -29,6 +32,16 @@ function VirtualizedList(props) {
 		selectionToggle,
 	});
 
+	if (inProgress) {
+		return (
+			<div className="tc-list-progress">
```
frassinier: You could add  `aria-atomic="true" aria-busy="true"` here, for fun!

packages/components/src/VirtualizedList/VirtualizedList.component.js
Innes333: This component will be replaced with CircularProgress

packages/components/src/List/List.scss
jsomsanith: add as css module in vrituazedList.scss

jsomsanith:
In the screenshot, 
* the list has a green border
* the loading panel has a red border

This is du to your `height:100%` on `tc-list-virtualized`. Without it le loading panel doesn not take 100% height, you need to find another way

capture d ecran de 2017-10-06 09-31-56

packages/components/src/VirtualizedList/VirtualizedList.component.js
```diff
@@ -29,6 +33,13 @@ function VirtualizedList(props) {
 		selectionToggle,
 	});
 
+	if (inProgress) {
+		return (
+			<div className={theme['tc-list-progress']}>
```
frassinier: What's about my comment about ARIA attributes? :(

Innes333: I will add them :)

PR #777, opened 2017-10-05
feat(cmf): add api to register many components

jmfrancois:

At the moment in configure.js of cmf app we register component one by one
We can't provide a set of component in an object and we also need to register their actions and expressions one by one.

add API to register a component which also register actions and expressions.
add API to regster many components (using for example src/app/component/index.js)

```javascript
import components from './components';

api.component.registerMany(components);  // and voila
```


output/cmf.eslint.txt
```diff
@@ -10,11 +10,8 @@ The react/require-extension rule is deprecated. Please use the import/extensions
...
   8:2  warning  Unexpected console statement  no-console
 
-/home/travis/build/Talend/ui/packages/cmf/src/RegistryProvider.js
-  17:17  error  Expected 'this' to be used by class method 'getChildContext'  class-methods-use-this
-
 /home/travis/build/Talend/ui/packages/cmf/src/sagaRouter/router.js
   45:2  warning  Unexpected constant condition  no-constant-condition
 
-✖ 4 problems (1 error, 3 warnings)
+✖ 3 problems (0 errors, 3 warnings)
```
jsomsanith: Nice

packages/cmf/src/component.js
```diff
@@ -0,0 +1,70 @@
...
+ * @module react-cmf/lib/component
+ */
+
+/**
+ * return a component from the registry
+ * @param  {object} context
+ * @param  {string} id
+ * @return {function} the react component
+ */
+function get(contextOrId, idOrUndefined) {
```
jsomsanith: This makes the following code had to understand.
As I understand, you introduced anew way to use the get, providing only the id : `get(id)` and the old way to use it is still ok `get(context, id)`
A better idea would be to change the function signature : `get(id, [context])`
This means a breaking change, but if the apps are not using this low level function, the refacto is easy (only in route.js)

@jmfrancois @talend-sromain can you evaluate the difficulty to switch the old calls `route.getComponentFromRegistry(context, id)` to `route.getComponentFromRegistry(id, context)` ?

packages/cmf/src/component.js
```diff
@@ -0,0 +1,70 @@
...
+		register(key, components[key], context);
+	});
+}
+
+
+export default {
+	get,
+	register,
+	registerMany,
+};
```
jsomsanith: Can you migrate `registerComponent` unit tests from `routes.test.js` to `component.test.js` ?
+ add missing unit tests please.

packages/cmf/src/component.js
sdwvit: Agree with @jsomsanith, api can and should be simplified

jmfrancois: ok I will do it !

PR #778, opened 2017-10-05
enhancement(cmf/view settings): memoize the setting retrieval

acateland:
On each container render, setting are fetched from the store then cloned to avoid store reference mutation.
Problem this prevent cmfConnect shallow testing of computed props to prevent children render if props haven't changed.

Memoize settings retrieved by this function with a cache storage on key composed of (viewName, component name, component id)

### DIFF
--- a/packages/cmf/src/settings.js
+++ b/packages/cmf/src/settings.js
@@ -32,6 +32,21 @@ export function attachRefs(state, props) {
 	return attachedProps;
 }
 
+export function nonMemoizedMapStateToViewProps(state, ownProps, componentName, componentId) {
+	let viewProps = {};
+	let viewId = ownProps.view;
+	if (!ownProps.view && componentName && !componentId) {
+		viewId = componentName;
+	} else if (!ownProps.view && componentName && componentId) {
+		viewId = `${componentName}:${componentId}`;
+	}
+	if (viewId && state.cmf.settings.views[viewId]) {
+		viewProps = Object.assign({}, state.cmf.settings.views[viewId]);
+		viewProps = attachRefs(state, viewProps);
+	}
+	return viewProps;
+}
+
 /**
  * return props for a given view with reference and override support
  * this function is memoized and the map key is computed using
@@ -74,26 +89,11 @@ export function attachRefs(state, props) {
  * @param  {Object} ownProps   the props passed to the component. may have a view attribute
  * @return {Object}           React props for the component injected from the settings
  */
-export const mapStateToViewProps = memoize((
-	state,
-	ownProps,
-	componentName,
-	componentId,
-) => {
-	let viewProps = {};
-	let viewId = ownProps.view;
-	if (!ownProps.view && componentName && !componentId) {
-		viewId = componentName;
-	} else if (!ownProps.view && componentName && componentId) {
-		viewId = `${componentName}:${componentId}`;
-	}
-	if (viewId && state.cmf.settings.views[viewId]) {
-		viewProps = Object.assign({}, state.cmf.settings.views[viewId]);
-		viewProps = attachRefs(state, viewProps);
-	}
-	return viewProps;
-},
-(state, ownProps, componentName, componentId) => `${ownProps.view}-${componentName}-${componentId}`);
+export const mapStateToViewProps = memoize(
+	nonMemoizedMapStateToViewProps,
+	(state, ownProps, componentName, componentId) =>
+		`${ownProps.view}-${componentName}-${componentId}`,
+);
 
 export default {
 	attachRef,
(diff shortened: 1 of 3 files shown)


output/cmf.eslint.txt
```diff
@@ -16,5 +16,8 @@ The react/require-extension rule is deprecated. Please use the import/extensions
 /home/travis/build/Talend/ui/packages/cmf/src/sagaRouter/router.js
   45:2  warning  Unexpected constant condition  no-constant-condition
 
-✖ 4 problems (1 error, 3 warnings)
+/home/travis/build/Talend/ui/packages/cmf/src/settings.js
+  29:37  error  Expected parentheses around arrow function argument having a body with curly braces  arrow-parens
```
jsomsanith: We really should do something about this prettier/eslint issue

packages/cmf/src/settings.js
```diff
@@ -17,28 +18,41 @@ export function attachRef(state, obj) {
...
+		viewId = `${componentName}:${componentId}`;
+	}
+	if (viewId && state.cmf.settings.views[viewId]) {
+		viewProps = Object.assign({}, state.cmf.settings.views[viewId]);
+		viewProps = attachRefs(state, viewProps);
+	}
+	return viewProps;
+}
+
+memoize.Cache = Map;
```
jsomsanith: Why do you replace the default cache implementation ?

acateland: So we can clear it when testing, so one test cache doesn't leak on another one.

output/cmf.eslint.txt
acateland: yes sorry :/

packages/cmf/src/settings.js
jsomsanith: https://github.com/lodash/lodash/blob/4.17.4/lodash.js#L2047

output/cmf.eslint.txt
jsomsanith: not your fault

packages/cmf/src/settings.js
```diff
@@ -70,30 +84,18 @@ export function attachRefs(state, props) {
 	listview: {
 	  ...
 	}
-  }
+	}
```
jmfrancois: wrong indent in JSDoc

jmfrancois:
Object.freeze() seems a really good candidate
- load settings, and on load ok, freeze state.cmf.settings
- remove Object.assign, so we don't need to memoize it

packages/cmf/src/settings.js
```diff
@@ -17,28 +18,39 @@ export function attachRef(state, obj) {
...
+	Object.keys(attachedProps).forEach(key => {
+		attachedProps[key] = attachRef(state, attachedProps[key]);
+	});
 	return attachedProps;
 }
 
+export function nonMemoizedMapStateToViewProps(state, ownProps, componentName, componentId) {
+	let viewProps = {};
+	let viewId = ownProps.view;
+	if (!ownProps.view && componentName && !componentId) {
```
sdwvit: pls extract condition into meaningful variable
also, it can be simplified to:
if (!ownProps.view && componentName) {
   if (componentId) { ... } else { ... }
}

packages/cmf/src/settings.js
```diff
@@ -17,28 +18,39 @@ export function attachRef(state, obj) {
...
+		viewId = componentName;
+	} else if (!ownProps.view && componentName && componentId) {
+		viewId = `${componentName}:${componentId}`;
+	}
+	if (viewId && state.cmf.settings.views[viewId]) {
```
sdwvit: this seems fishy as well, could you pls explain what this check should do?
maybe it can be simplified to (state.cmf.settings.views[viewId]) ?

acateland: it is done this way because of
```
const data = {
  "undefined": "value"
};

console.log(data[undefined]);
```

packages/cmf/src/settings.js
```diff
@@ -2,9 +2,48 @@
...
+}
+
+/**
+ * try to retrieve view settings for a cmfconnected component
+ * @param {Object} state the application state
+ * @param {*} ownProps props given to the cmfConnected component
+ * @param {*} componentName name of the cmfConnect component
+ * @param {*} componentId componentId, can be undefined
+ */
+export function nonMemoizedMapStateToViewProps(state, ownProps, componentName, componentId) {
```
talend-glorieux: No need to call it `nonMemoized`.

packages/cmf/src/settings.js
```diff
@@ -2,9 +2,48 @@
...
+	let viewProps = {};
+	let viewId = ownProps.view;
+
+	viewId = generateDefaultViewId(viewId, componentName, componentId);
+
+	if (viewId && state.cmf.settings.views[viewId]) {
+		viewProps = state.cmf.settings.views[viewId] || {};
```
talend-glorieux: I don't think the `||` is required since you're already initialise `viewProps`.

PR #779, opened 2017-10-05
feat(objectviewer): add openFirst, remove root label, perf improvement

jmfrancois:

ObjectViewer display a 'root' label which should not be required
To open the first item we have to push an initialState
Sub objects are rendered even if not opened. This push thousand of components for nothing ('hidden')

Add openFirst props
remove 'root' default label
do not render sub components if not opened


packages/containers/src/ObjectViewer/ObjectViewer.container.js
```diff
@@ -95,6 +96,9 @@ class ObjectViewer extends React.Component {
 
 	render() {
 		const state = (this.props.state || DEFAULT_STATE).toJS();
+		if (this.props.state === DEFAULT_STATE && this.props.openFirst) {
```
ncomont: Did you mean `(state === DEFAULT_STATE && this.props.openFirst)` ?

packages/components/src/ObjectViewer/JSONLike/JSONLike.component.js
```diff
@@ -257,24 +257,26 @@ export function ComplexItem({ data, name, opened, edited, jsonpath, info, onSele
...
-										jsonpath={jpath}
-										opened={opened}
-										edited={edited}
-										onSelect={onSelect}
-									/>
-								</li>
-							);
-						})}
-					</ul>
+					{isOpened ? (
```
ncomont: 60+ lines for this `return`. Could you split it ?

packages/containers/src/ObjectViewer/ObjectViewer.container.js
jmfrancois: no beacuse state is not immutable it's the result of toJS()

jmfrancois: here this is a kind of pristine test

PR #782, opened 2017-10-07
fix(container/virtualized-list) : Allow sorting with header column in display table mode

Sylvain-Talend:
Currently, it is not possible to sort via connected List when clicking column header

It is now possible to sort with column header


packages/containers/src/List/List.container.js
```diff
@@ -102,6 +102,13 @@ class List extends React.Component {
 			list: {
 				items,
 				columns: get(this.props, 'list.columns', []),
+				sort: {
+					field: state && state.sortOn,
+					isDescending: state && !state.sortAsc,
```
jsomsanith: Test if `state` is truthy is useless, it is never falsy.

packages/containers/src/List/List.container.js
```diff
@@ -102,6 +102,13 @@ class List extends React.Component {
...
+					onChange: (event, data) => {
```
jsomsanith: `onChange: this.onSelectSortBy` should be sufficient, no need to create a new function

packages/containers/src/List/List.container.js
Sylvain-Talend: :+1:

packages/containers/src/List/List.container.js
Sylvain-Talend: :+1:

PR #783, opened 2017-10-10
chore: update all webpack plugins

jmfrancois:

Following webpack3 update in # we have some warning like this one:
```
warning "extract-text-webpack-plugin@2.1.0" has incorrect peer dependency "webpack@^2.2.0".
```

update all webpack plugins to their last version
remove ^


packages/containers/yarn.lock
```diff
@@ -5165,6 +5152,10 @@ pify@^2.0.0, pify@^2.3.0:
   version "2.3.0"
   resolved "https://registry.yarnpkg.com/pify/-/pify-2.3.0.tgz#ed141a6ac043a849ea588498e7dca8b15330e90c"
 
+pify@^3.0.0:
```
jmfrancois: :disapointed:
├─┬ sass-loader@6.0.6
│ └── pify@3.0.0
├─┬ jest-cli@20.0.3
│ └── pify@2.3.0

So we got it twice :/ but only for build tools.

jmfrancois: If we want to update jest we have to update babel and this should be into another PR

PR #795, opened 2017-10-18
chore(cmf): remove object assign to increase performance

jmfrancois:

if you are using view object to inject props the component will be rendered on every frame.
This is because mapStateToViewProps is doing an Object.assign,
so every call is providing new object, so the shouldComponentUpdate of react-redux will return true.

To remove that object.assign we have preprocess the references in the reducer.
So settings are ready to be used.
We have also added a call to Object.freeze so we are trying to protect you from bad code behavior.
You should never mutate those properties.


packages/cmf/src/reducers/settingsReducers.js
```diff
@@ -10,10 +12,58 @@ export const defaultState = {
...
+	Object.keys(attachedProps).forEach(
+		(key) => {
+			attachedProps[key] = attachRef(refs, attachedProps[key]);
+		}
+	);
+	return attachedProps;
+}
+
+/**
+ * attache reference to produce a ready to use freezed object
```
talend-glorieux: `attache`?

packages/cmf/src/reducers/settingsReducers.js
```diff
@@ -10,10 +12,58 @@ export const defaultState = {
 	actions: {},
 	views: {},
 	routes: {},
-	ref: {},
 };
 
 /**
+ * if an object try to find _ref property and resolve it
+ */
+export function attachRef(refs, obj) {
```
talend-glorieux: Could you add some JsDoc?

packages/cmf/src/reducers/settingsReducers.js
```diff
@@ -10,10 +12,58 @@ export const defaultState = {
...
+	if (obj === null || typeof obj !== 'object' || Array.isArray(obj)) {
+		return obj;
+	}
+	let props = Object.assign({}, obj);
+	if (props._ref) {
+		const ref = refs[props._ref];
+		invariant(ref, `CMF/Settings: Reference '${props._ref}' not found`);
```
talend-glorieux: ref is only used once.

PR #796, opened 2017-10-19
feat(ObjectViewer): Adds value selection

talend-glorieux:
ObjectViewer values can't be selected easily.

Change value from button to span.
@frassinier Please don't hate me. I haven't found a better way to make the button text easily selectable. 
I'm open to any suggestion if you have a better way to handle this.


jmfrancois:
the onClick do not do anything.
so we may render a span without onClick

sgendre:
it would be nice to select all value with a double click but I can't find the trick either on

https://www.w3.org/TR/DOM-Level-2-Traversal-Range/ranges.html

packages/components/src/ObjectViewer/JSONLike/JSONLike.component.js
```diff
@@ -105,12 +102,8 @@ export function LineItem({
...
 
 	return (
-	// eslint-disable-next-line jsx-a11y/no-static-element-interactions
-		<span
-			className={classes}
-			onClick={e => stopAndSelect(e, { onSelect, jsonpath })}
-			{...props}
-		>
+		// eslint-disable-next-line jsx-a11y/no-static-element-interactions
+		<span className={classes} onClick={e => stopAndSelect(e, { onSelect, jsonpath })} {...props}>
```
frassinier: Don't miss on key events and focus management here too

packages/components/src/ObjectViewer/JSONLike/JSONLike.component.js
```diff
@@ -35,20 +35,17 @@ export function NativeValue({ data, edit, onSelect, onChange, jsonpath }) {
...
-	);
+	const lineValueClasses = classNames(theme.native, theme[type], theme['line-value']);
 
 	return (
-		<button
+		<span
 			className={lineValueClasses}
-			type="button"
+			role="button"
+			tabIndex="0"
```
acateland: shouldn't tabIndex be provided, by the iterator key ?

frassinier: Could you add `onKeyUp` event management too? `onKeyUp={e => stopAndSelectWithEnterOrSpace(e, { onSelect, jsonpath })}`

frassinier: It doesn't matter since It will be a relative order if they are all with value `0`

PR #800, opened 2017-10-23
chore(containers/HomeListView): use Inject to set renderers

jmfrancois:

We need to switch component to better support futur refactor.
HomeListView is tied coupled to AppHeaderBar and we have now HeaderBar which is connected.

Let's use Inject as component where it's possible.


packages/containers/src/HomeListView/HomeListView.component.js
```diff
@@ -1,16 +1,13 @@
...
 import React from 'react';
 import { Layout } from '@talend/react-components';
+import { Inject } from '@talend/react-cmf';
 
-import AppHeaderBar from '../AppHeaderBar';
-import List from '../List';
-import SidePanel from '../SidePanel';
-
-function getContent(Component, props) {
+function getContent(component, props) {
```
acateland: please add jsdoc, wich describe
for example component can be either : a component, a string, null, undefined
Should we add a deprecation warning on `if (React.isValidElement(props)) {` block

jsomsanith:
Something bothers me in this PR. Before, the HomeListView worked out of the box. Now we need a pre-step : register components with `AppHeaderBar`, `SidePanel`, and `List` id. Otherwise you end up with a `NotFound` component.
What would be cool, is that you register the default needed components somewhere (automatically ? or via an exposed component like a `DefaultRegistration` thing that the apps put like `IconProvider`), and user's config can override the default one.

So that would ensure furthur changes : 
* if there is a new component needed in the containers, we add them in the `DefaultRegistration` component, so any upgrade will automatically register them
* if there is a refacto like here, same thing

Otherwise everytime you do a refactor to use Inject, it will potentially break apps. What do you think ?

jsomsanith:
(another idea)
Or better : if `getComponent()` returns a NotFoundComponent, we use the default ones, so no registration required.

jsomsanith:
Done in Layout component

PR #801, opened 2017-10-23
feat(components/badge): add ids

sdwvit:
QA want ids for badges to complete testing

add id props for onselect and ondelete badge

### DIFF
--- a/packages/components/src/Badge/Badge.component.js
+++ b/packages/components/src/Badge/Badge.component.js
@@ -8,8 +8,9 @@ import { DEFAULT_I18N } from '../translate';
 import Action from '../Actions/Action';
 import theme from './Badge.scss';
 
-function renderDeleteIcon(onClick, deleteId, disabled, t) {
+function renderDeleteIcon(onClick, id, disabled, t) {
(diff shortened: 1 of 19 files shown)


jmfrancois:
should we prefix id; sth like tc-badge-${id} ?

packages/components/src/Badge/Badge.component.js
```diff
@@ -54,7 +66,9 @@ Badge.propTypes = {
 	label: PropTypes.string,
 	category: PropTypes.string,
 	onDelete: PropTypes.func,
+	deleteId: PropTypes.string,
```
jsomsanith: Please avoid having multiple ids as input. You can align it with other components we have : 
* get a unique id from props
* if present, generate a delete id and a select id from it, and set it

jsomsanith: + add a unit test please

sdwvit: @jsomsanith thanks, updating..

sdwvit:
@jmfrancois @jsomsanith @acateland updated

PR #804, opened 2017-10-24
feat(forms): update styles of error msg

Innes333:
 Removing of the bullet points, and left align the text displayed. 
If more than one error message needs to be displayed, this will be done as today, on a separate line/ under each other.
Changes connected to:
     -  http://guidelines.talend.com/document/92132#/ui-controls/input-field
     -  https://jira.talendforge.org/browse/TPSVC-3545

Styling adjustment

### DIFF
--- a/packages/forms/stories/index.js
+++ b/packages/forms/stories/index.js
@@ -496,3 +496,44 @@ decoratedStories.add('Form Children', () => {
 		</Form>
 	);
 });
+
+
+decoratedStories.add('Form with live validation', () => {
+	const schema = {
+		jsonSchema: {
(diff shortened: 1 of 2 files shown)


Innes333:
@jsomsanith the guideline says that errors should have font-size 12px, and about classes - it's come from react-bootstrap-forms ul>li

jsomsanith:
Yes, it should be 12px, but part of the error text has 14px font-size.
So we have 12 and 14, and that's quite ugly :)

`about classes - it's come from react-bootstrap-forms ul>li` --> yes but you define a new css class. To include the label to have the same color, you should set it in `.has-error` instead

PR #806, opened 2017-10-24
feat(components/ObjectViewer): fix errors when attributes are null and add schema (type and dqtype) to header (Table.component)

talend-sromain:
This component doesn't render the children when there is null value in the children.
We can't switch to grid view when we have null value.
Need to add type or dqtype to column header.

Add some conditions
Also add a schema attributes in the incoming data of the ObjectViewer. 
The schema holds the type of the header column. 


packages/components/src/ObjectViewer/List/List.component.js
```diff
@@ -5,12 +5,14 @@ import JSONLike from '../JSONLike';
 import theme from './List.scss';
 
 function List({ data, ...props }) {
-	if (!Array.isArray(data)) {
+	if (!data || (!Array.isArray(data) && !Array.isArray(data.datas))) {
```
jare-talend: remove parenthesis (!Array.isArray(data) && !Array.isArray(data.datas). AND is priority.

packages/components/src/ObjectViewer/List/List.component.js
```diff
@@ -5,12 +5,14 @@ import JSONLike from '../JSONLike';
...
 		return null;
 	}
+	// The datas can be an array or an array in an object. We assign the value correctly here.
+	const datas = Array.isArray(data) ? data : data.datas;
 	return (
 		<ul className={`${theme.container} tc-object-viewer list-unstyled`}>
-			{data.map((obj, index) => (
+			{datas.map((obj, index) => (
```
jare-talend: change datas by dataset

packages/components/src/ObjectViewer/Table/Table.component.js
```diff
@@ -23,23 +23,49 @@ export function getAbsolutePath(index, key, flat) {
...
+ * If there is a type we add it.
+ * @param {array} headers
+ */
+export function buildContentHeaders(headers, schema) {
+	return headers.map((key, index) => {
+		let type;
+		if (schema) {
+			type = schema.get(key);
+		}
+		return type ? (
```
jare-talend: prefer to use
if(type) {
 return <JSX>
}

return JSX

PR #807, opened 2017-10-25
enhancement(component/Headerbar): small change

acateland:
HeaderBar/Brand use name property to render as label wich prevent to use CMFAction wich use name to resolve CMFAction from the registry.
Target property could not be passed down to an Action component

Add target to the whitelist.
Change headerBar Brand descriptor to label.
Name now can be used to resolve CMFAction from the registry. 

- [X] This PR introduces a breaking change
The brand element of the HeaderBar don't use name as label wich is conflicting with CMFAction name used for action resolution.

### DIFF
--- a/packages/components/src/utils/getPropsFrom.js
+++ b/packages/components/src/utils/getPropsFrom.js
@@ -4,6 +4,7 @@ const NATIVE_PROPS = [
 	'className',
 	'id',
 	'name',
+	'target',
 ];
 
 function extractComponentProps(Component, props) {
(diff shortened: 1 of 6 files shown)


jsomsanith:
I'll approve, but note that you don't follow the guideline in branch/commit naming.
Next time I'll have to punish you

packages/components/src/utils/getPropsFrom.js
```diff
@@ -4,6 +4,7 @@ const NATIVE_PROPS = [
 	'className',
 	'id',
 	'name',
+	'target',
```
jare-talend: could you update test getPropsFrom.test.js with this props to avoid regression later

PR #808, opened 2017-10-25
chore(tests): Using faker to generate test data

talend-glorieux:
Subtitle: How to get rid of Jean-Pierre DUPONT and toto.

We have in many of our test files random string and sentences. This
shouldn't be up the the developers to figure out those random values.

I've added faker in order to generate those values. Even though faker is
usually used to generate "trully random" values a seed can be set in
order to have repeatable results.

The only thing missing is moving the seed to the root of the tests so it doesn't need to be repeated.


jsomsanith:
I like the initiative, and understand the concern, but there are 2 things that I don't like, all linked the the thing you've done on it to generate determinist data:
* it's kind of a hack, and we must reinit the state on top of each test file
* it generate a sequence of data. This means the data you want to use is now ordered, if you move/remove/refactor a test, it impacts all the following tests in the file.

I don't have any idea for now to fix those, but the second one is a blocker for me.

PR #811, opened 2017-10-26
fix(sagas): Export pending constants

sgendre:
Pending constants were not exported.
Change that forgetting

### DIFF
--- a/packages/sagas/package.json
+++ b/packages/sagas/package.json
@@ -56,6 +56,7 @@
     "redux-saga": "0.15.4"
   },
   "devDependencies": {
+    "babel": "^6.23.0",
     "babel-cli": "6.24.1",
     "babel-core": "6.24.1",
     "babel-eslint": "7.2.3",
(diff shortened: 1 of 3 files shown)


packages/sagas/package.json
```diff
@@ -56,6 +56,7 @@
     "redux-saga": "0.15.4"
   },
   "devDependencies": {
+    "babel": "^6.23.0",
```
talend-glorieux: Why was babel added?

sgendre: because without it local package build can't be achieved

packages/sagas/yarn.lock
```diff
@@ -815,6 +815,10 @@ babel-types@^6.18.0, babel-types@^6.19.0, babel-types@^6.23.0, babel-types@^6.24
     lodash "^4.17.4"
     to-fast-properties "^1.0.3"
 
+babel@^6.23.0:
```
talend-glorieux: Please update your warn lock.

PR #812, opened 2017-10-26
feat(cmf&container): auto evaluate expression based on props convention

acateland:
only a few selected action properties key could be evaluated.
Meaning each time we need one more we have to touch this package

Each action properties ending by `Expression` are automaticaly evaluated,  and result put in action propertie missing the `Expression` ending.
`action.stuffExpression` value get evaluated and result put in `action.stuff` and  `action.stuffExpression` get deleted

### DIFF
--- a/packages/containers/src/actionAPI.js
+++ b/packages/containers/src/actionAPI.js
@@ -1,4 +1,7 @@
 import { api } from '@talend/react-cmf';
+import forIn from 'lodash/forIn';
+
+const regexExpression = /(.*)Expression/g;
 
 /**
  * add support for expression in actions.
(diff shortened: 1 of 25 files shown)


packages/containers/src/actionAPI.js
```diff
@@ -16,14 +19,13 @@ function evalExpressions(action, context, payload = {}) {
...
-		delete newAction.labelExpression;
-		newAction.label = api.expression.call(action.labelExpression, context, newAction);
-	}
-	if (action.iconExpression) {
-		delete newAction.iconExpression;
-		newAction.icon = api.expression.call(action.iconExpression, context, newAction);
-	}
+	forIn(action, (value, key) => {
+		if (key.match(regexExpression)) {
+			const newKey = key.slice(0, -10);
```
talend-glorieux: Guess you could use a regexp with capture to avoid slicing based on `Expression` length.

acateland: its more safe to

packages/containers/src/actionAPI.js
```diff
@@ -16,14 +19,13 @@ function evalExpressions(action, context, payload = {}) {
...
 	);
-	if (action.labelExpression) {
-		delete newAction.labelExpression;
-		newAction.label = api.expression.call(action.labelExpression, context, newAction);
-	}
-	if (action.iconExpression) {
-		delete newAction.iconExpression;
-		newAction.icon = api.expression.call(action.iconExpression, context, newAction);
-	}
+	forIn(action, (value, key) => {
```
jmfrancois: I prefer Object.keys().forEach which do not need lodash

acateland: we already have loadash on our codebase so.

PR #821, opened 2017-10-28
feat(cmf): add expression to cmfConnect

jmfrancois:

To support expression you have to create higher order component.

All cmfConnected components will have support for first level props keys.
Following #812 


packages/cmf/package.json
```diff
@@ -108,8 +108,7 @@
     ],
     "roots": [
       "src",
-      "__tests__",
-      "__mocks__"
```
jmfrancois: it makes jest --watch fails because it doesn't exists

packages/cmf/src/cmfConnect.js
```diff
@@ -134,6 +141,32 @@ export function getDispatchToProps({
 }
 
 /**
+ * Interal: you should not have to use this
```
ncomont: typo ?

PR #823, opened 2017-10-30
feat(Action): add displayMode support

jare-talend:

No way to resolve ActionDropdown and ActionSplitDropdown using just Action
So if I want to display an ActionDropdown I need to use Actions ...

Action become a component selector
The old Action is in ActionButton component


packages/components/src/index.js
```diff
@@ -68,7 +68,7 @@ import {
 
 import I18N_DOMAIN_COMPONENTS from './constants';
 
-import { Action, ActionDropdown, ActionSplitDropdown, Actions } from './Actions';
+import { Action, ActionDropdown, ActionButton, ActionSplitDropdown, Actions } from './Actions';
```
acateland: should ActionDropdown, ActionButton, ActionSplitDropdown  still be exported ? Since Action will render to those given the right props ?

packages/components/src/Actions/Actions.component.js
```diff
@@ -108,15 +86,12 @@ function Actions(props) {
 Actions.propTypes = {
 	actions: PropTypes.arrayOf(
 		PropTypes.oneOfType([
-			PropTypes.shape(Action.propTypes),
-			PropTypes.shape({
-				displayMode: PropTypes.oneOf([TYPE_DROPDOWN, TYPE_SPLIT_DROPDOWN]),
-			}),
+			PropTypes.shape(ActionButton.propTypes),
```
acateland: it is weird  to render abstract Action and use ActionButton as PropType provider

packages/components/src/Actions/Actions.component.js
```diff
@@ -108,15 +86,12 @@ function Actions(props) {
...
 		]),
 	),
 	className: PropTypes.string,
 	hideLabel: PropTypes.bool,
-	tooltipPlacement: Action.propTypes.tooltipPlacement,
+	tooltipPlacement: ActionButton.propTypes.tooltipPlacement,
```
acateland: it is weird  to render abstract Action and use ActionButton as PropType provider

packages/components/src/Actions/Action/Action.test.js
```diff
@@ -1,142 +1,59 @@
...
-		const wrapper = renderer.create(
-			<Action
-				className="navbar-btn"
-				inProgress
-				{...myAction}
-			/>).toJSON();
-
-		// then
-		expect(wrapper).toMatchSnapshot();
+	it('should return MyActionButton without displayMode', () => {
```
acateland: test label doesn't match result

packages/components/src/Actions/Action/Action.test.js
```diff
@@ -1,142 +1,59 @@
...
-		const wrapper = renderer.create(
-			<Action
-				className="navbar-btn"
-				disabled
-				{...myAction}
-			/>).toJSON();
-
-		// then
-		expect(wrapper).toMatchSnapshot();
+	it('should return MyActionSplitDropdown if displayMode = splitDropdown', () => {
```
acateland: test label doesn't match result

packages/components/src/Actions/Action/Action.test.js
```diff
@@ -1,142 +1,59 @@
...
-		wrapper.props.onMouseDown();
-
-		// then
-		expect(mouseDownAction.onMouseDown).toHaveBeenCalled();
-		expect(mouseDownAction.onMouseDown.mock.calls.length).toBe(1);
-		const args = mouseDownAction.onMouseDown.mock.calls[0];
-		expect(args.length).toBe(2);
-		expect(args[0]).toBe();
-		expect(args[1].action.extra).toBe('extra');
+	it('should return MyActionDropdown if displayMode = dropdown', () => {
```
acateland: test label doesn't match result

packages/components/src/Actions/Actions.component.js
jare-talend: exact

packages/components/src/Actions/Actions.component.js
jare-talend: exact

packages/components/src/index.js
jare-talend: if no dependencies on this and anyone use it alone

PR #845, opened 2017-11-13
feat(containers): dynamic items for ActionSplitDropDown & ActionDropDown

jare-talend:
provide dynamic items on the ActionSplitDropdown & ActionDropdown

We can provide, on the container DropDown, the items by settings, expressions or straightly by props (to avoid to broke compatibility)


packages/containers/src/ActionSplitDropdown/ActionSplitDropdown.connect.js
```diff
@@ -15,7 +15,10 @@ export function mapStateToProps(state, { actionId, actionIds } = {}) {
 		props = api.action.getActionInfo(context, actionId);
 	}
 	if (actionIds) {
-		props.items = actionIds.map(itemId => api.action.getActionInfo(context, itemId));
+		props.actionIds = actionIds;
```
jmfrancois: actionIds comes from external props, why do you want to add it from mapStateToProps result ?

packages/containers/src/ActionSplitDropdown/ActionSplitDropdown.connect.js
```diff
@@ -44,38 +47,34 @@ export function ContainerActionSplitDropdown(props) {
...
+							},
+							item.payload,
+						),
+					);
+				}
+			},
+			...item,
+		}));
 	}
-	delete newProps.actionIds;
```
jmfrancois: This is a double of props.actionIds.
Why ?

packages/containers/src/ActionSplitDropdown/ActionSplitDropdown.connect.js
```diff
@@ -26,56 +31,31 @@ export function mergeProps(stateProps, dispatchProps, ownProps) {
...
+	if (newProps.items) {
+		newProps.items = props.items.map(item => ({
+			...getOnClick(item, props),
+			...item,
+		}));
+
+		delete newProps.actionIds;
 	}
-	delete newProps.actionIds;
+
```
jmfrancois: I think this delete should always been done here not only if items

packages/containers/src/ActionDropdown/ActionDropdown.connect.js
```diff
@@ -30,31 +35,22 @@ export function mergeProps(stateProps, dispatchProps, ownProps) {
...
-			},
-		}, item));
+
+	if (newProps.items) {
+		newProps.items = props.items.map(item => ({
+			...getOnClick(item, props),
+			...item,
+		}));
+
+		delete newProps.actionIds;
```
jmfrancois: same here the delete can be moved just before line 33

packages/containers/src/ActionSplitDropdown/ActionSplitDropdown.connect.js
jmfrancois: So it's to resolve it if the actionIds is in the first actionid resolution

packages/containers/src/ActionSplitDropdown/ActionSplitDropdown.connect.js
```diff
@@ -26,56 +31,31 @@ export function mergeProps(stateProps, dispatchProps, ownProps) {
...
-			),
-		);
+
+	if (newProps.items) {
+		newProps.items = props.items.map(item => ({
+			...getOnClick(item, props),
+			...item,
+		}));
+
+		delete newProps.actionIds;
```
jmfrancois: this delete should be in mergeProps

PR #846, opened 2017-11-13
feat(components/sidepanel): accept custom action ids

ncomont:

Actions ids are base on the action's label. It cause problems with i18n (for example TDP onboarding uses theses IDs to target elements).

Actions can have a new `id` property.

Before this, action ids were ignored. Now, if an id is provided per action, it will be used instead of the label ; so it could break some QA tests.

### DIFF
--- a/packages/components/src/SidePanel/SidePanel.component.js
+++ b/packages/components/src/SidePanel/SidePanel.component.js
@@ -10,7 +10,6 @@ import Action from '../Actions/Action';
 
 import theme from './SidePanel.scss';
 
-
 /**
  * return the formatted action id
  * if there is no action id, it is generated from the action label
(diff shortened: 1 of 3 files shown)


packages/components/src/SidePanel/SidePanel.component.js
```diff
@@ -62,12 +62,19 @@ function SidePanel({ id, selected, onSelect, actions, docked, onToggleDock, t, r
...
 				{actions.map(action => {
-					const isSelected = isActionSelected(action);
 					const a11y = {};
+					const extra = {};
+					const isSelected = isActionSelected(action);
+					const actionId =
+						action.id ||
+						action.label
+							.toLowerCase()
+							.split(' ')
```
acateland: a regexp could be used here but i guess it is a lot more readable the way you did it

packages/components/src/SidePanel/SidePanel.component.js
```diff
@@ -62,12 +62,19 @@ function SidePanel({ id, selected, onSelect, actions, docked, onToggleDock, t, r
...
+							.join('-');
```
talend-glorieux: Please move is to an actionId generator.

PR #853, opened 2017-11-14
feat(cmf-cqrs): add saga needed to do ack

jmfrancois:

we are using saga and cqrs so we need to do ack inside saga

add API to do so


packages/cmf-cqrs/src/sagas/ack.js
```diff
@@ -0,0 +1,49 @@
...
+		yield put(chan, action.requestId);
+	}
+}
+
+/**
+ * this function will wait for the ack to be filled.
+ * @param {Object} channel redux-saga channel instance
+ * @param {string} requestId the requestId to check in the channel
+ */
+function* check(chan, requestId) {
```
ncomont: What do you think about : 

```javascript
function* check(chan, requestId) {
	let ackRequestId;
	do {
		ackRequestId = yield take(chan);
	} while (ackRequestId !== requestId);

	return true;
}
```

Or something like that : 

```javascript
function* check(chan, requestId) {
	for (;;) {
		const ackRequestId = yield take(chan);
		if (ackRequestId === requestId) {
			return true;
		}
	}
}
```

Or even something like :

```javascript
function* check(chan, requestId) {
	for (let ack; ack !== requestId; ack = yield take(chan));
	return true;
}
```

These three version are eslint compliant.

acateland: i vote in favor of the first proposition or what is actually available in code, for lisibility and understandability purpose

packages/cmf-cqrs/src/sagas/ack.js
```diff
@@ -0,0 +1,47 @@
...
+}
+ */
+
+/**
+ * this function let you listen a channel ACK_RECEIVE_MESSAGE
+ * @param {Object} channel redux-saga channel
+ */
+function* listen(chan) {
+	// eslint-disable-next-line no-constant-condition
+	while (true) {
```
ncomont: `for(;;) {}` could avoidn using this `eslint-disable-next-line`

jmfrancois: no, it exactly the same thing except a normal brain take more time to understand what happens.

PR #854, opened 2017-11-14
feat(containers): add TreeView

jmfrancois:

no treeView in containers but in components
no way to hide the header
cmf do not expose a way to omit the props it has. so it's hard to omit them to spread the rest.
the component wait for `itemSelectCallback` and `itemToggleCallback` props to handle some click events. This doesn't let enzyme doing simulate which wait for onSomething.

add it in containers
add a noHeader props support in the components
export cmfConnect.INJECTED_PROPS constant so you can do the following to get only the props to spread:

```
const props = omit(this.props, cmfConnect.INJECTED_PROPS);
```

Also we deprecate `itemSelectCallback` and `itemToggleCallback` and add onClick and onSelect to make the component more testable following the event handler convention


packages/components/src/TreeView/TreeView.component.js
```diff
@@ -44,51 +48,72 @@ import theme from './TreeView.scss';
...
+			</nav>
+		</div>
+	);
+}
 
 TreeView.propTypes = {
 	id: PropTypes.string,
 	headerText: PropTypes.string,
-	structure: PropTypes.arrayOf(TreeViewItem.propTypes.item),
+	structure: PropTypes.arrayOf(PropTypes.object),
```
smouillour: Why change this type?

acateland:
Can you describe the change on the component itself ?

packages/components/src/TreeView/TreeView.component.js
jmfrancois: because TreeViewItem import ../../index.js so it make circular deps and tests fails on a can't read propTypes of undefined.

jmfrancois:
done !

packages/components/src/TreeView/TreeView.component.js
acateland: can you move proptypes in a place where it does not cause circular deps ?

packages/components/src/TreeView/TreeView.component.js
```diff
@@ -44,51 +48,72 @@ import theme from './TreeView.scss';
...
 	addAction: PropTypes.func,
 	addActionLabel: PropTypes.string,
 	itemSelectCallback: PropTypes.func.isRequired,
```
acateland: can you remove the .isRequired it is misleading

PR #857, opened 2017-11-15
fix(component/ActionFile): Input clear after onChange

acateland:
When selecting the same file twice in the row, the second time the onChange props did not get called.

After onChange props get called, the input got its value cleared, so the second time you select the same file, onChange get trigered as well.


packages/components/src/Actions/ActionFile/ActionFile.component.js
```diff
@@ -61,82 +61,95 @@ function getButtonContent(props) {
...
+		}
+		const localId = id || uuid.v4();
+		const buttonContent = getButtonContent(this.props);
+		const labelClasses = classNames(
+			'btn',
+			theme['btn-file'],
+			(disabled || inProgress) && 'disabled',
+		);
+		const btn = (
+			<form
```
jare-talend: I know that @frassinier prefer that all input need to nest in a form excect that if the action is integrate in a form, this form is ignored.

try this https://jsfiddle.net/4j6dravd/ and look the DOM. The second form nested in a form is removed:

nested form

PR #859, opened 2017-11-16
fix(VirtualizedList) titleProps get overwrite when column type is specified

yyanwang:
there's a problem when column config is
```
{
  key: 'name',
  label: 'Name',
  type: 'title'
}
```
column is rendered not clickable because titleProps gets overwrite when resolve cell renderer type.
and it's set as
change the order of code snippet, assign `titleProps` to columnData after resolving renderer type

### DIFF
--- a/packages/components/src/List/ListToVirtualizedList/ListToVirtualizedList.component.js
+++ b/packages/components/src/List/ListToVirtualizedList/ListToVirtualizedList.component.js
@@ -49,17 +49,17 @@ function ListToVirtualizedList(props) {
 					label: column.label,
 					dataKey: column.key,
 				};
-				if (titleProps && column.key === titleProps.key) {
-					Object.assign(cProps, CellTitle, {
-						columnData: titleProps,
-					});
-				}
 				if (supposedActions[column.key]) {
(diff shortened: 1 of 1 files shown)


packages/components/src/List/ListToVirtualizedList/ListToVirtualizedList.component.js
```diff
@@ -62,6 +57,11 @@ function ListToVirtualizedList(props) {
 						columnData: column.data,
 					});
 				}
+				if (titleProps && column.key === titleProps.key) {
```
jsomsanith: We have 3 cases here : 
* supposed action (for retro compatibility) --> `CellActions`
* explicit type --> the specified cell type
* titleProps defined --> `CellTitle` with props

Those are not compatible with each other. So we have to define an order and use `if/else` instead of `if` x3
I would suggest : titleProps - actions - type (as before). Type is the generic case, so it falls at the end.

yyanwang: Hi jimmy,
thanks for explaining! I've updated according your comment.

PR #862, opened 2017-11-17
fix(components/sidepanel): accept empty ids

ncomont:

image

Add a check on `action.id` and `action.label` in the component
Add action.id and action.label in SidePanel mapStateToProps so the component has it.

### DIFF
--- a/packages/components/src/SidePanel/SidePanel.component.js
+++ b/packages/components/src/SidePanel/SidePanel.component.js
@@ -18,13 +18,16 @@ import theme from './SidePanel.scss';
  * @return {string}        	formatted id
  */
 function getActionId(id, action) {
-	const actionId =
-		action.id ||
-		action.label
-			.toLowerCase()
-			.split(' ')
-			.join('-');
-	return id && `${id}-nav-${actionId}`;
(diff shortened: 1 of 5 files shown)


packages/containers/examples/index.js
```diff
@@ -32,5 +31,4 @@ export default {
 	ObjectViewerExample,
 	SidePanelExample,
 	TreeViewExample,
-	DeleteResource,
```
jmfrancois: is twice in the index.js

packages/components/src/SidePanel/SidePanel.component.js
```diff
@@ -18,13 +18,16 @@ import theme from './SidePanel.scss';
...
-		action.id ||
-		action.label
-			.toLowerCase()
-			.split(' ')
-			.join('-');
-	return id && `${id}-nav-${actionId}`;
+	if (action.id || action.label) {
+		const actionId =
+			action.id ||
+			action.label
```
jsomsanith: I don't think we should force an id. If the user wants an id, it provides it.
Any reason or other opinion ?

PR #864, opened 2017-11-17
fix(containers/components): uniformise the action on dropdown

jare-talend:

fix the event on main button on SplitDropdown
uniform the onClick event to be like the ActionButton


packages/components/src/Actions/utils.js
```diff
@@ -0,0 +1,27 @@
...
+
+	return (
+		onClick &&
+		(event =>
+			onClick(event, {
+				action: { label, ...rest },
+				model,
+			}))
+	);
+}
```
talend-glorieux: Please use an actual file name rather than utils.

packages/components/src/Actions/utils.js
```diff
@@ -0,0 +1,27 @@
...
+	const { label, model, onClick, ...rest } = omit(action, IGNORED_PROPS);
+
+	return (
+		onClick &&
+		(event =>
+			onClick(event, {
+				action: { label, ...rest },
+				model,
+			}))
+	);
```
talend-glorieux: Please a regular if statement rather than the `&&` construct.

packages/components/src/Actions/utils.js
```diff
@@ -0,0 +1,27 @@
...
+	'hideLabel',
+	'link',
+	'onMouseDown',
+	'tooltipPlacement',
+	'tooltip',
+	'tooltipLabel',
+	'available',
+];
+
+export default function getRClick(action) {
```
talend-glorieux: Why is it called RClick if it doesn't contain any specific logic related to right click?

packages/components/src/Actions/ActionDropdown/ActionDropdown.test.js
```diff
@@ -6,27 +6,46 @@ describe('ActionDropdown', () => {
...
+		});
+		expect(onItemClick.mock.calls[0][0].type).toBe('click');
 
 		// when
-		actionDropdownInstance.find('MenuItem').at(1).find('SafeAnchor').simulate('click');
+		actionDropdownInstance
+			.find('MenuItem')
+			.at(1)
+			.find('SafeAnchor')
+			.simulate('click');
```
talend-glorieux: Could you please avoid text repetitions?

packages/components/src/Actions/utils.js
```diff
@@ -0,0 +1,27 @@
...
+	const { label, model, onClick, ...rest } = omit(action, IGNORED_PROPS);
```
talend-glorieux: Why don't you do the props restriction prior to calling `getRClick`?
It look like a lot of work to do it on every user click.

packages/components/src/Actions/Action/Action.component.js
```diff
@@ -1,5 +1,6 @@
 import PropTypes from 'prop-types';
 import React from 'react';
+import omit from 'lodash/omit';
```
talend-glorieux: Looks like `omit` isn't used.

packages/components/src/Actions/Action/Action.test.js
```diff
@@ -74,3 +74,57 @@ describe('Action', () => {
...
+					inProgress: 'inProgress',
+					label: 'label',
+					link: 'link',
+					onMouseDown: 'onMouseDown',
+					otherProperty: 'otherProperty',
+					tooltip: 'tooltip',
+					tooltipLabel: 'tooltipLabel',
+					tooltipPlacement: 'tooltipPlacement',
+				},
+				model: { id: '#model' },
```
talend-glorieux: The entire action object is duplicated. Could you please put it all into a single object?

PR #878, opened 2017-11-27
feat(components): add reversed theme for SidePanel

romainseb:
We have to get some Side panel in reversed theme in some detail view in data catalog
-> guideline here : http://guidelines.talend.com/document/92132#/navigation-layout/detail-view-mode/sub-side-left-panel
reverse the theme from sidepanel


packages/components/src/SidePanel/SidePanel.component.js
```diff
@@ -114,9 +133,12 @@ function SidePanel({ id, selected, onSelect, actions, docked, onToggleDock, t, r
...
 							key={action.key || action.label}
-							className={classNames('tc-side-panel-list-item', {
-								active: isSelected,
-							})}
+							className={classNames(
+								theme['tc-side-panel-list-item'],
+								'tc-side-panel-list-item',
+								{ active: isSelected },
+								{ [theme.active]: isSelected },
+							)}
```
ncomont: I think that you could use a single object here

packages/components/src/SidePanel/SidePanel.scss
```diff
@@ -22,10 +27,94 @@ $toggle-button-padding: $padding-small - 1;
 		width: $tc-side-panel-width;
 	}
 
-	:global(.btn) svg:global(.tc-svg-icon) {
-		width: $tc-side-panel-icons-size;
-		height: $tc-side-panel-icons-size;
-		margin: 0;
+	&-list {
+		border-right: 2px solid $alto;
```
ncomont: Use variable, please

packages/components/src/SidePanel/SidePanel.component.js
```diff
@@ -49,13 +49,32 @@ function getActionId(id, action) {
...
+		'nav',
+		'nav-pills',
+		theme['nav-pills'],
+		'nav-stacked',
+		theme['nav-stacked'],
 		theme['action-list'],
+		'tc-side-panel-list',
+		theme['tc-side-panel-list'],
+		{ 'nav-inverse': !reverse },
+		{ [`${theme['nav-inverse']}`]: !reverse },
```
ncomont: I think that you could use a single object here

ncomont: I don't understand why you're using template string here instead of just doing something like that: 

```
	{
		'nav-inverse': !reverse,
		[theme['nav-inverse']]: !reverse,
	},
```

Am I missing something ?

packages/components/src/SidePanel/SidePanel.scss
```diff
@@ -22,10 +27,94 @@ $toggle-button-padding: $padding-small - 1;
...
+			}
+
+			.nav-pills {
+				.link {
+					padding: $padding-normal $tc-side-panel-large-padding;
+				}
+			}
+
+			.toggle-btn {
+				padding-left: 4px;
```
ncomont: Where does this value come from ?

romainseb: it's for centered align when docked & large display ^^'
I will make a variable :)

packages/components/src/SidePanel/SidePanel.component.js
romainseb: You're not missing something 👍

PR #880, opened 2017-11-28
feat(SelectObject): display results as ListGroupItem

jmfrancois:
**What is the problem this PR is trying to solve?**

Search on tree is not really good. 

**What is the chosen solution to this problem?**

When starting typing in the FilterBar let's switch to list of results

**Please check if the PR fulfills these requirements**
- [x] The PR commit message follows our [guidelines](https://github.com/talend/tools/blob/master/tools-root-github/CONTRIBUTING.md)
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)
- [ ] Related design / discussions / pages (not in jira), if any, are all linked or available in the PR

PR #882, opened 2017-11-28
fix(components/TreeView): add keyboard navigation

jmfrancois:

TreeView component hide the outlines everywhere.
So you can't see where you are.
We think there is no difference between open and select action
If 'id' props is not set -> undefined-add and other values like this in snapshots

Merge select and click so the toggle icon is not an action anymore
Get the outline back
add default value id='tc-treeview'


acateland:
> Merge select and click
has this been check with UX can @maximenicoul approve on this PR ?

packages/components/src/TreeView/TreeView.component.js
```diff
@@ -92,7 +92,7 @@ function TreeView({
 				<ul className={theme['tc-treeview-ul']}>
 					{structure.map((item, i) => (
 						<TreeViewItem
-							id={id && `${id}-${i}`}
+							id={`${id}-${i}`}
```
acateland: this change is not listed on the pr description, and potentially break QA tests, it need to be listed as Breaking change

packages/components/src/TreeView/TreeView.test.js
```diff
@@ -36,8 +37,38 @@ const defaultProps = {
...
 
 describe('TreeView', () => {
+	it('should be exported', () => {
+		expect(TreeView).toBe(ExportedTreeView);
+	});
 	it('should render normally with all buttons and custom labels', () => {
 		const wrapper = shallow(<TreeView {...defaultProps} />);
 		expect(wrapper.getNode()).toMatchSnapshot();
 	});
+	it('should call add action', () => {
```
acateland: should be rephrased example 'user click on XXX should call `function/props` on XXX'

packages/components/src/TreeView/TreeView.test.js
```diff
@@ -36,8 +37,38 @@ const defaultProps = {
...
+		const props = {
+			...defaultProps,
+			addAction: jest.fn(),
+		};
+		const wrapper = shallow(<TreeView {...props} />);
+		wrapper.find('Action').simulate('click');
+		expect(props.addAction).toHaveBeenCalled();
+	});
+	it('should call onClick', () => {
```
acateland: same as above

packages/components/src/TreeView/TreeView.test.js
```diff
@@ -36,8 +37,38 @@ const defaultProps = {
...
+		const props = {
+			...defaultProps,
+			onClick: jest.fn(),
+		};
+		const wrapper = shallow(<TreeView {...props} />);
+		wrapper.find('TreeViewItem').simulate('click');
+		expect(props.onClick).toHaveBeenCalled();
+	});
+	it('should call onSelect', () => {
```
acateland: same as above

packages/components/src/TreeView/TreeViewItem/TreeViewItem.component.js
```diff
@@ -34,122 +37,139 @@ function getActionHandler(func, item) {
...
 
-	function getTreeViewItem(child, i) {
+	onKeyDown(event) {
+		if (event.keyCode === keycode.codes.enter) {
+			this.onSelect();
+		}
+	}
+
+	onSelect() {
+		this.elementRef.focus();
```
acateland: this variable could be a lot more explicit

packages/components/src/TreeView/TreeViewItem/TreeViewItem.test.js
```diff
@@ -29,9 +30,46 @@ const defaultProps = {
 	onClick: () => null,
 };
 
+describe('getItemIcon', () => {
+	it('should return icon', () => {
```
acateland: should be split in more test, for more explicitness.

packages/components/src/TreeView/TreeViewItem/TreeViewItem.test.js
```diff
@@ -29,9 +30,46 @@ const defaultProps = {
...
+		expect(getItemIcon(undefined, false)).toBe('talend-folder-closed');
+	});
+});
+
 describe('TreeView', () => {
 	it('should render normally with all buttons and custom labels', () => {
 		const wrapper = shallow(<TreeViewItem {...defaultProps} />);
 		expect(wrapper.getNode()).toMatchSnapshot();
 	});
+	it('should select call onClick and onSelect', () => {
```
acateland: the simulated action is user click, do not match description

packages/components/src/TreeView/TreeViewItem/TreeViewItem.test.js
```diff
@@ -29,9 +30,46 @@ const defaultProps = {
...
+		};
+		const focus = jest.fn();
+		const wrapper = shallow(<TreeViewItem {...props} />);
+		wrapper.instance().elementRef = { focus };
+		wrapper.find('.tc-treeview-item').simulate('click');
+		expect(props.onClick).toHaveBeenCalled();
+		expect(props.onSelect).toHaveBeenCalled();
+		expect(focus).toHaveBeenCalled();
+	});
+	it('should enter call onClick and onSelect', () => {
```
acateland: need to rephrased, looks like a focus thing should be called to

packages/components/src/TreeView/TreeView.component.js
jmfrancois: id has been added as not required to help QA.
I prefer add a default value so I don't need to check
Description added to the PR

jsomsanith: We should align how ids work within our components. If no id in props, then no generated ids, same for add (all) button

packages/components/src/TreeView/TreeViewItem/TreeViewItem.component.js
```diff
@@ -10,14 +11,16 @@ import css from './TreeViewItem.scss';
...
-		func(item);
-	};
+/**
+ * return the default open or closed folder icon if non is specified on item
+ * or if it is specified return the specified icon with `-closed` append if not toggled
+ * @param {String} icon - icon name
+ * @param {Boolean} toggled - state of the item
+ * @return {String}
+ */
+export function getItemIcon(itemIcon = 'talend-folder', itemStateToggled) {
```
jsomsanith: The jsDoc params names are not correct

packages/components/src/TreeView/TreeViewItem/TreeViewItem.component.js
```diff
@@ -34,122 +37,139 @@ function getActionHandler(func, item) {
...
+	constructor(props) {
+		super(props);
+		this.onSelect = this.onSelect.bind(this);
+		this.onKeyDown = this.onKeyDown.bind(this);
+		this.renderTreeViewItem = this.renderTreeViewItem.bind(this);
 	}
 
-	function getTreeViewItem(child, i) {
+	onKeyDown(event) {
+		if (event.keyCode === keycode.codes.enter) {
```
jsomsanith: Shouldn't we support `space` key too ? (Just an idea)

packages/components/src/TreeView/TreeViewItem/TreeViewItem.component.js
```diff
@@ -34,122 +37,139 @@ function getActionHandler(func, item) {
...
+			this.onSelect();
+		}
+	}
+
+	onSelect() {
+		this.containerRef.focus();
+		this.props.onClick(this.props.item);
+		return this.props.onSelect(this.props.item);
```
jsomsanith: I don't understand the purpose of click and select. Why do we have both ?

packages/components/src/TreeView/TreeViewItem/TreeViewItem.component.js
```diff
@@ -34,122 +37,139 @@ function getActionHandler(func, item) {
...
-					onClick,
-					depth: depth + 1,
-					key: i,
-				}}
+				id={`${this.props.id}-${i}`}
+				item={child}
+				onSelect={this.props.onSelect}
+				onClick={this.props.onClick}
+				depth={this.props.depth + 1}
+				key={i}
```
jsomsanith: OMG the old code. Thank you for this change

packages/components/src/TreeView/TreeViewItem/TreeViewItem.component.js
```diff
@@ -34,122 +37,139 @@ function getActionHandler(func, item) {
...
-				onClick={getActionHandler(action, item)}
+				onClick={event => {
+					event.stopPropagation();
+					action(this.props.item);
+				}}
 				tooltipPlacement="right"
 				hideLabel
 				key={label}
-				id={id_ || `${id}-${icon_}`}
+				id={id_ || `${this.props.id}-${icon_}`}
```
jsomsanith: We should not have id if no `this.props.id`

packages/components/src/TreeView/TreeViewItem/TreeViewItem.component.js
jmfrancois: Because they represent two actions: select and toggle open/close the tree

packages/components/src/TreeView/TreeViewItem/TreeViewItem.component.js
jmfrancois: done in #880

packages/components/src/TreeView/TreeViewItem/TreeViewItem.component.js
jmfrancois: done in #880

PR #883, opened 2017-11-29
 feat(UIForms): close closeable items on array add

jsomsanith:
*WARNING: this is only on the next version of forms, not current one*
This is an implementation if #695 : Adding elements to an array forces the user to scroll to fill the newly added item.

Arrays with closeable elements will close all items on new element insertion.


packages/forms/src/UIForm/fieldsets/Array/Array.component.test.js
```diff
@@ -94,14 +98,14 @@ describe('Array component', () => {
...
-			/>
+			/>,
 		);
 
 		// then
 		expect(wrapper.node).toMatchSnapshot();
 	});
 
-	it('should render array that can\'t be reordered', () => {
+	it("should render array that can't be reordered", () => {
```
jmfrancois: why " ? it's simple quote

packages/forms/src/UIForm/fieldsets/Array/Array.component.js
```diff
@@ -61,7 +62,13 @@ export default class ArrayWidget extends React.Component {
...
 		const arrayMergedSchema = this.props.schema;
 		const defaultValue = arrayMergedSchema.schema.items.type === 'object' ? {} : '';
-		const value = [...this.props.value, defaultValue];
+
+		const itemWidget = widgets[this.props.schema.itemWidget];
+		const currentValue =
+			itemWidget && itemWidget.isCloseable
+				? this.props.value.map(item => ({ ...item, isClosed: true }))
+				: this.props.value;
+		const value = currentValue.concat(defaultValue);
```
jmfrancois: What do you think of
```javascript
let value = [...this.props.value, defaultValue];
if (get(widgets, `${this.props.schema.itemWidget}.isCloseable`, false)) {
  value = value.map(item => ({ ...item, isClosed: true });
}
```

packages/forms/src/UIForm/fieldsets/Array/Array.component.test.js
ncomont: Prettier do that to avoid having antislash in strings

packages/forms/src/UIForm/fieldsets/Array/Array.component.js
jsomsanith: hmm I prefer to add some vars with explicit names to be clearer.
Plus, your code close the new inserted item, to have the same behavior, you have to add a test in your `.map()` to skip the last item.

If my code is a little hard to read it's because of the ternary I think. I'll change it to an `if` (@glorieux will be happy). What do you think ?

packages/forms/src/UIForm/fieldsets/Array/Array.component.test.js
jsomsanith: :+1: it's prettier, not me :)

PR #884, opened 2017-11-29
feat(VirtualizedList): row height per list display

jlame:
Needed to change the default rowHeight on virtualized list when there is too much data to display, but the rowHeight wasn't managed by the list container.

Added the missing props to list container and some other in components. The rowHeight props is used for all the display which causes an issue as our different display modes have in fact different height. So the list container now takes a rowHeight props which is a object containing the wanted height value for each display (if they are not given it uses the default one define in the component).


output/components.eslint.txt
```diff
@@ -15,6 +15,9 @@ The react/require-extension rule is deprecated. Please use the import/extensions
   1:1  error  '@storybook/react' should be listed in the project's dependencies, not devDependencies  import/no-extraneous-dependencies
   4:2  error  Unexpected require()                                                                    global-require
 
+/home/travis/build/Talend/ui/packages/components/src/List/ListToVirtualizedList/ListToVirtualizedList.component.js
+  40:21  error  'rowHeight' is missing in props validation  react/prop-types
```
jsomsanith: Fix this one please

packages/components/src/List/List.component.js
```diff
@@ -54,15 +54,20 @@ ListToolbar.propTypes = {
...
 		return <Content id={id && `${id}-content`} displayMode={displayMode} {...translatedList} />;
 	}
 	if (virtualized) {
 		return (
 			<div className={'tc-list-display-virtualized'}>
-				<ListToVirtualizedList id={id} displayMode={displayMode} {...translatedList} />
+				<ListToVirtualizedList
+					id={id}
+					displayMode={displayMode}
+					rowHeight={rowHeight}
```
jsomsanith: Can you add or modify the tests to ensure that the info is passed ?

packages/containers/examples/ExampleList.js
```diff
@@ -44,6 +44,16 @@ const props = {
 	toolbar,
 };
 
+const customHeight = {
+	tile: undefined,
+	large: 200,
+	table: 200,
```
jsomsanith: Change the value between large and table to demonstrate that the height is per type

packages/containers/src/List/List.container.js
```diff
@@ -110,6 +110,9 @@ class List extends React.Component {
 			virtualized: this.props.virtualized,
 			renderers: this.props.renderers,
 		};
+		if (this.props.rowHeight) {
+			props.rowHeight = this.props.rowHeight[props.displayMode];
```
jsomsanith: If there is no `props.rowHeight`, the app will crash

packages/containers/src/List/List.test.js
```diff
@@ -211,6 +206,21 @@ describe('Container List', () => {
...
+			large: 2,
+			tile: 1,
+		};
+		const wrapper = shallow(
+			<Container {...cloneDeep(settings)} items={items} rowHeight={rowHeight} />,
+			{ lifecycleExperimental: true },
+		);
+		const props = wrapper.props();
+		expect(props.displayMode).toBe('table');
+		expect(props).toMatchSnapshot();
```
jsomsanith: It's a been over-engineered, you just want to check the `rowHeight`, just check it instead of the whole props.

packages/containers/src/List/List.container.js
jsomsanith: As you said in private, yeah, I missed the `if` :)
Sorry for that

packages/components/src/List/List.component.js
jlame: added test on ListToVirtualizedList component

packages/components/src/List/ListToVirtualizedList/ListToVirtualizedList.test.js
```diff
@@ -53,6 +54,12 @@ describe('ListToVirtualizedList', () => {
 		expect(large.type).toBe('LARGE');
 	});
 
+	it('should support rowHeight', () => {
+		const rProps = { ...props, rowHeight: 200 };
+		const table = shallow(<ListToVirtualizedList {...rProps} displayMode="table" />);
+		expect(toJson(table)).toMatchSnapshot();
```
jsomsanith: Please check just the prop you want instead of snapshot

PR #888, opened 2017-12-04
chore(forms): add uiform props on Form to use UIForm (migration)

jmfrancois:

We have two uiSchema form and two libs

Let our `Form` lib render `UIForm` with the props `uiform`.
add `moz` props on UIForm to align API with `Form`

Override merge feature to compute props and migrate uiSchema of Form to UIForm.

Tests are done using some of the old schema available for the story and match them with the corresponding results.

We also support custom widgets by wrapping them in a HOC.

Todo on Storybbok:

- [x] Form / Widgets broken (join of undefined)
- [x] UIForm / all actions boken
- [x] UIForm / columns (impossible to set name input)
- [x] Use internal Form state for the container implementation
- [x] Remove redux implementation which is done in container


packages/forms/src/UIForm/fields/Datalist/Datalist.component.js
```diff
@@ -130,7 +127,8 @@ class Datalist extends Component {
 		const previousValue = persist ? value : this.state.previousValue;
 		this.setState({ value, previousValue });
 		if (persist) {
-			const payload = { schema: this.props.schema, value };
+			const enumValue = this.props.schema.titleMap.find(item => item.name);
```
jmfrancois: what is this find is supposed to do ?

output/containers.eslint.txt
```diff
@@ -1,12 +1,15 @@
...
+> @talend/react-containers@0.134.0 lint:es /home/travis/build/Talend/ui/packages/containers
 > eslint --config ../../.eslintrc src
 
 The react/require-extension rule is deprecated. Please use the import/extensions rule from eslint-plugin-import instead.
 
 /home/travis/build/Talend/ui/packages/containers/src/ActionSplitDropdown/ActionSplitDropdown.test.js
   4:10  error  'ActionSplitDropdown' is defined but never used  no-unused-vars
 
+/home/travis/build/Talend/ui/packages/containers/src/Form/Form.container.js
+  90:2  error  formActions should be placed after setErrors  react/sort-comp
```
jmfrancois: 👍

packages/forms/src/UIForm/UIForm.component.js
```diff
@@ -46,11 +44,31 @@ export default class UIForm extends React.Component {
 	onChange(event, { schema, value }) {
 		const payload = {
 			formName: this.props.formName,
-			properties: this.props.properties,
+			properties: this.props.properties || {},
```
jsomsanith: Default props please

packages/forms/src/UIForm/UIForm.component.js
```diff
@@ -46,11 +44,31 @@ export default class UIForm extends React.Component {
...
 			schema,
 			value,
 		};
-		this.props.onChange(event, payload);
+		payload.formData = mutateValue(payload.properties, schema.key, value);
```
jsomsanith: If the purpose is to have the new data eveytime (moz of not), you can add it in payload declaration just up there.
Or else, move it in the `if(moz)`

packages/forms/src/UIForm/UIForm.component.js
```diff
@@ -46,11 +44,31 @@ export default class UIForm extends React.Component {
...
+		if (this.props.moz) {
+			this.props.onChange(payload);
+		} else {
+			this.props.onChange(event, payload);
+		}
+		// trigger if value is correct
```
jsomsanith: moving trigger in `onChange` implies that trigger will be triggered even if value is not valid
onChange/onFinish separation was to allow widgets and fields to manage when those are called. Some widgets call onChange then onFinish on the same event.
We can for example on selection widget, to trigger both, but on user typing widgets, call onFinish on blur

packages/forms/src/UIForm/UIForm.component.js
```diff
@@ -46,11 +44,31 @@ export default class UIForm extends React.Component {
...
+		if (schema.triggers && schema.triggers.length) {
+			let formData = payload.formData;
+			let propertyName = schema.key.join('.');
+			if (this.props.moz) {
+				schema.key.forEach((key, index) => {
+					if (index !== schema.key.length - 1) {
+						formData = formData[key];
```
jsomsanith: What is that ? on old forms, the trigger get the parent object value instead of the value ?

packages/forms/src/UIForm/UIForm.component.js
```diff
@@ -119,28 +124,13 @@ export default class UIForm extends React.Component {
...
-		)
-			.then(newForm => updateForm(
-				formName,
-				newForm.jsonSchema,
-				newForm.uiSchema,
-				newForm.properties,
-				newForm.errors)
-			)
-			.catch(({ errors }) => setError(formName, errors));
+		return onTrigger(formData, formId, propertyName, propertyValue);
```
jsomsanith: Where is the part that update the form if we receive a new form from trigger in connector ?
I see it in container that manage internal state, but not in the connected one anymore. Does it mean that user has to manage it himself ?

packages/forms/src/UIForm/UIForm.component.js
```diff
@@ -178,26 +180,31 @@ export default class UIForm extends React.Component {
...
+			},
+		];
 
 		return (
 			<form
 				acceptCharset={this.props.acceptCharset}
 				action={this.props.action}
 				autoComplete={this.props.autoComplete}
-				className={this.props.className}
+				className={`tf-uiform ${this.props.className || ''}`}
```
jsomsanith: `classnames()` to manage `this.props.className || ''` ?

packages/forms/src/UIForm/UIForm.container.js
```diff
@@ -24,6 +24,19 @@ export default class UIForm extends React.Component {
 		this.setErrors = this.setErrors.bind(this);
 	}
 
+	/**
+	 * Update the state with the new schema.
+	 * @param jsonSchema
+	 * @param uiSchema
+	 */
+	componentWillReceiveProps(nextProps) {
```
jsomsanith: Can you add a test ?

packages/forms/src/UIForm/UIForm.component.js
jmfrancois: I don't think we have this use case.
All our triggers are on controlled field (like select). So result can't be invalid and happens on change.
I guess trigger the validation onChange will slow down the rendering which is already quite slow.

packages/forms/src/UIForm/UIForm.component.js
jmfrancois: yes :(
Sorry for that it's for compatibility and change that need to update all forms.
Worse we have also backend change to support this use case.

packages/forms/src/UIForm/UIForm.component.js
jmfrancois: it's react so one directional flow
trigger is an action managed by the parent.
And schema is passed by the parent.
So it's the parent that must handle the update of the schema

packages/forms/stories-core/json/fields/core-datalist.json
```diff
@@ -10,6 +10,11 @@
       "restrictedDatalist": {
         "type": "string",
         "enum": [ "Apple", "Pine[apple]", "Banana", "Cher[ry", "Lemo}n", "Grapefruit" ]
+      },
+      "keyValue": {
+        "type": "string",
+        "enum": [ "a", "p", "b", "c", "l", "g" ],
+        "enumNames": [ "Apple", "Pine[apple]", "Banana", "Cher[ry", "Lemo}n", "Grapefruit" ]
```
jsomsanith: Are we really supposed to support enumNames as exposed feature ?
IMO this is supposed to be in an old version use case, mapped with merge.js, so no example in the core storybooks

packages/forms/src/UIForm/Widget/Widget.component.js
```diff
@@ -7,12 +7,17 @@ import { getValue } from '../utils/properties';
 
 export default function Widget(props) {
 	const { errors, formName, properties, schema } = props;
-	const { key, type, validationMessage, widget } = schema;
+	const { key, type, validationMessage, widget, options } = schema;
 	const widgetId = widget || type;
+
+	if (widgetId === 'hidden') {
+		return null;
```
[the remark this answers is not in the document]
jmfrancois: I have copy / paste the way it's done in `Form` to not break our app

packages/forms/stories-core/json/fields/core-text.json
```diff
@@ -11,6 +11,12 @@
       },
       "password": {
         "type": "string"
+      },
+      "enum": {
+        "type": "string",
+        "enum": [
+          "one", "two", "three"
```
jsomsanith: This is a select, use case covered in core-select.json, not an input text

packages/forms/src/UIForm/merge.js
```diff
@@ -0,0 +1,207 @@
...
+				ui.items = updateWidgets(ui.items, config, widgets, `${ui.key}.`);
+			}
+		}
+		return ui;
+	});
+}
+
+/**
+ * migrate from react-jsonschema-form to UIschema
+ * @param {Object} mergedSchema
```
jsomsanith: Outdated jsDoc

packages/forms/src/UIForm/merge.js
```diff
@@ -0,0 +1,207 @@
...
+ */
+export function migrate(jsonSchema, uiSchema) {
+	const safeUISchema = parseProperties(jsonSchema, true);
+	const widgets = {};
+	if (!uiSchema || uiSchema === {}) {
```
jsomsanith: `uiSchema === {}` is always false
BTW this test should be on top of the function. Calling parseProperties is useless, and should return uiSchema as an empty array.

packages/forms/src/UIForm/fields/Button/SingleButton.component.js
```diff
@@ -3,20 +3,23 @@ import React from 'react';
...
+export default function SingleButton({ className, id, onTrigger, onClick, schema }) {
 	const {
 		bsStyle,
 		disabled = false,
 		inProgress,
+		label,
 		name,
-		title,
 		triggers,
+		model,
```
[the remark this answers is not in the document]
jmfrancois: it comes from cmf, but I have replace all this by rest spread operation in props

packages/forms/src/UIForm/merge.js
```diff
@@ -0,0 +1,207 @@
...
+		return { jsonSchema, uiSchema: safeUISchema };
+	} else if (!Array.isArray(uiSchema) && typeof uiSchema === 'object') {
+		if (uiSchema['ui:order']) {
+			safeUISchema[0].items.sort((left, right) => {
+				const leftOrderIndex = uiSchema['ui:order'].indexOf(left.key);
+				const righttOrderIndex = uiSchema['ui:order'].indexOf(right.key);
```
jsomsanith: type `righttOrderIndex` --> `rightOrderIndex`

packages/forms/src/UIForm/merge.js
```diff
@@ -0,0 +1,207 @@
...
+				if (leftOrderIndex < righttOrderIndex) {
+					return -1;
+				}
+				if (leftOrderIndex > righttOrderIndex) {
+					return 1;
+				}
+				return 0;
```
jsomsanith: if/if/else --> `return leftOrderIndex - rightOrderIndex` ?

packages/forms/src/UIForm/merge.js
```diff
@@ -0,0 +1,207 @@
...
+
+function parseArray(items, key) {
+	if (items.type === 'object') {
+		return parseProperties(items.properties, false, `${key}[]`);
+	}
+	return [getUISchemaFromObject(items, key)];
+}
+
+function getUISchemaFromObject(schema, key) {
+	const ui = { key };
```
jsomsanith: This method is called with `key = undefined` when node to parse is root, this will produce `{ undefined: undefined }`

packages/forms/src/UIForm/merge.js
```diff
@@ -0,0 +1,207 @@
...
+	if (schema.enum && schema.enumNames) {
+		ui.titleMap = schema.enum.map((value, index) => ({ value, name: schema.enumNames[index] }));
+		delete schema.enumNames;
+	}
+	if (schema.type === 'object') {
+		ui.widget = 'fieldset';
+		ui.items = parseProperties(schema.properties, false, key);
+	} else if (schema.type === 'string') {
+		if (!schema.enum) {
+			ui.widget = 'text';
```
jsomsanith: I think this is not necessary, by default, widget is `text`. And if we have an enum in jsonSchema, by default it is a select.
You can see that in action in `core-text.js` and `core-select.js`

packages/forms/src/UIForm/merge.js
```diff
@@ -0,0 +1,207 @@
...
+			if (schema.format === 'data-url') {
+				ui.type = 'file';
+			}
+		}
+	} else if (schema.type === 'array') {
+		ui.widget = 'array';
+		ui.items = parseArray(schema.items, key);
+	} else if (schema.type === 'number') {
+		ui.widget = 'text';
+		ui.type = 'number';
```
jsomsanith: In uiSchema, type and widget are the same, but for number, `type` is used to set the input type. You can remove ui.widget = 'text' that is only used to get the `<Text>` component, which is also mapped to `number`

packages/forms/src/UIForm/merge.js
```diff
@@ -0,0 +1,207 @@
+import React, { PropTypes } from 'react';
+import { merge } from 'talend-json-schema-form-core';
+import get from 'lodash/get';
+
+import FieldTemplate from './fields/FieldTemplate';
+
+/* eslint-disable no-param-reassign */
+
+function parseArray(items, key) {
```
jsomsanith: Can you add jsDoc ?
And on the other functions please

output/forms.eslint.txt
```diff
@@ -13,5 +13,10 @@ The react/require-extension rule is deprecated. Please use the import/extensions
 /home/travis/build/Talend/ui/packages/forms/src/UIForm/fields/Radios/Radios.component.js
   23:7  error  Form controls using a label to identify them must be programmatically associated with the control using htmlFor  jsx-a11y/label-has-for
 
-✖ 3 problems (3 errors, 0 warnings)
+/home/travis/build/Talend/ui/packages/forms/src/UIForm/merge.js
+  11:10  error  'parseProperties' was used before it was defined        no-use-before-define
+  13:10  error  'getUISchemaFromObject' was used before it was defined  no-use-before-define
+  30:14  error  'parseProperties' was used before it was defined        no-use-before-define
```
jsomsanith: Please fix those errors

packages/forms/src/UIForm/utils/widgets.js
```diff
@@ -17,6 +17,11 @@ import Text from '../fields/Text';
 import TextArea from '../fields/TextArea';
 import Toggle from '../fields/Toggle';
 
+import Code from '../../widgets/CodeWidget';
+import { wrapCustomWidget } from '../merge';
+
+const WrappedCode = wrapCustomWidget(Code);
```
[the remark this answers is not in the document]
jmfrancois: I think that can be done later. without impact any project

packages/forms/src/UIForm/merge.test/ordering.json
```diff
@@ -0,0 +1,35 @@
+[
```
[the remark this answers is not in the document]
jmfrancois: moved in __snapshots__

packages/forms/src/UIForm/UIForm.component.js
jsomsanith: wow this is a weird choice

packages/forms/src/UIForm/UIForm.component.js
jsomsanith: yes, that's something I was torn appart, but what I meant here is that you moved that on UIForms.container (the react state one), but not in UIForms.connect.js (the redux one)

packages/forms/src/UIForm/Widget/Widget.component.js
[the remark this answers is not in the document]
jsomsanith: But with this, we can't add a hidden input anymore

packages/forms/src/UIForm/merge.js
jmfrancois: uiSchema === {} is useless true but
we need a uiSchema in all case to handle enumNames which is not part of the spec

packages/forms/stories-core/json/fields/core-datalist.json
jmfrancois: But it's used in our backend. so yes, but until backend has moved to the new uiSchema.

packages/forms/src/UIForm/merge.test.js
```diff
@@ -0,0 +1,162 @@
...
+		wrapper.find(component).simulate('change', { foo: 'bar' });
+		expect(onChange).toHaveBeenCalledWith(expect.anything(), {
+			schema: { triggers: ['after'] },
+			value: { foo: 'bar' },
+		});
+	});
+});
+
+describe('default merge', () => {
+	it('should return props', () => {
```
jsomsanith: --> `should return default uiSpecs props`

packages/forms/src/UIForm/merge.test.js
```diff
@@ -0,0 +1,162 @@
...
+		expect(merge({}, [])).toEqual({
+			jsonSchema: {},
+			uiSchema: [],
+			mergedSchema: [],
+		});
+	});
+	it('should return props', () => {
```
jsomsanith: --> `should adapt old format and merge them`

packages/forms/src/UIForm/UIForm.component.js
```diff
@@ -44,13 +43,37 @@ export default class UIForm extends React.Component {
...
+		if (this.props.moz) {
+			this.props.onChange(payload);
+		} else {
+			this.props.onChange(event, payload);
+		}
+		// moving trigger in onChange implies that trigger will be triggered even if value is not valid
+		// onChange/onFinish separation was to allow widgets and fields to manage when those are called. Some widgets call onChange then onFinish on the same event.
+		// We can for example on selection widget, to trigger both, but on user typing widgets, call onFinish on blur
+		// trigger if value is correct
+		// if (schema.triggers && schema.triggers.length) {
```
jsomsanith: µYou can keep the comment explaining why we don't have trigger here, but remove the commented code please

packages/forms/src/UIForm/UIForm.component.js
```diff
@@ -44,13 +43,37 @@ export default class UIForm extends React.Component {
 	 * @param value The payload new value
 	 */
 	onChange(event, { schema, value }) {
+		console.log('##########@');
```
jsomsanith: Remove that please

packages/forms/src/UIForm/fields/Button/Buttons.component.test.js
```diff
@@ -32,7 +32,14 @@ describe('Buttons field', () => {
 		};
 
 		// when
-		const wrapper = shallow(<Buttons id={'myForm'} onTrigger={jest.fn()} schema={schema} />);
+		const wrapper = shallow(
+			<Buttons
+				id={'myForm'}
+				onTrigger={jest.fn()}
+				onClick={jest.fn(onClick => onClick && onClick())}
```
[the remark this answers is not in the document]
jsomsanith: Why is this onClick so complicated ? 
You don't seem to test the call in this snapshot test, it can only be `jest.fn()`

packages/forms/src/UIForm/fields/FieldTemplate/FieldTemplate.component.js
```diff
@@ -39,21 +42,20 @@ function FieldTemplate(props) {
 
 if (process.env.NODE_ENV !== 'production') {
 	FieldTemplate.propTypes = {
-		children: PropTypes.oneOf(
-			PropTypes.element,
-			PropTypes.arrayOf(PropTypes.element)
-		),
+		children: PropTypes.oneOf(PropTypes.element, PropTypes.arrayOf(PropTypes.element)),
```
jsomsanith: This is not du to your changes but `PropTypes.oneOf` accepts an array

packages/forms/src/UIForm/fields/Select/Select.component.js
```diff
@@ -39,32 +33,22 @@ export default function Select({ id, isValid, errorMessage, onChange, onFinish,
...
-				onBlur={event => onFinish(event, { schema })}
-				onChange={
-					event => onChange(
-						event,
-						{ schema, value: getSelectedOptions(event.target, multiple) }
-					)
-				}
+				onChange={event => {
+					onChange(event, { schema, value: getSelectedOptions(event.target, multiple) });
+					onFinish(event, { schema, value: getSelectedOptions(event.target, multiple) });
```
jsomsanith: Can you init only 1 time the payload ? Store it in a local var and pass it to onChange and onFinish

packages/forms/src/UIForm/fields/Select/Select.component.test.js
```diff
@@ -179,7 +179,7 @@ describe('Select field', () => {
 		expect(onChange).toBeCalledWith(event, { schema: multipleSchema, value: ['foo', 'lol'] });
 	});
 
-	it('should trigger onFinish on select blur', () => {
+	it('should trigger onFinish on select change', () => {
```
jsomsanith: I think you can merge it with the previous test, adding the onFinish assertion

packages/forms/src/UIForm/fields/Text/Text.component.js
```diff
@@ -24,6 +24,7 @@ export default function Text(props) {
 			isValid={isValid}
 			label={title}
 			labelAfter
+			required={schema.required || false}
```
jsomsanith: `|| false` is unnecessary. FieldTemplate test the truthiness with `classnames()` to add a css class

packages/forms/src/UIForm/Widget/Widget.component.js
```diff
@@ -6,32 +6,31 @@ import defaultWidgets from '../utils/widgets';
 import { getValue } from '../utils/properties';
 
 export default function Widget(props) {
-	const { errors, formName, properties, schema } = props;
-	const { key, type, validationMessage, widget } = schema;
+	const { key, type, validationMessage, widget, options } = props.schema;
 	const widgetId = widget || type;
+
+	if (widgetId === 'hidden') {
```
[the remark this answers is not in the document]
jmfrancois: We have type hidden every where in our schema :/

packages/forms/src/UIForm/fields/FieldTemplate/FieldTemplate.component.js
jmfrancois: and it's oneOfType

packages/theme/src/theme/_forms.scss
```diff
@@ -288,9 +288,6 @@ form {
 	}
 
 	.has-error {
-		font-size: $font-size-small;
-		color: $chestnut-rose;
```
jmfrancois: this one make the datalist content red

output/forms.eslint.txt
```diff
@@ -8,10 +8,13 @@ The react/require-extension rule is deprecated. Please use the import/extensions
...
-  8:4  error  Form controls using a label to identify them must be programmatically associated with the control using htmlFor  jsx-a11y/label-has-for
+  9:4  error  Form controls using a label to identify them must be programmatically associated with the control using htmlFor  jsx-a11y/label-has-for
 
 /home/travis/build/Talend/ui/packages/forms/src/UIForm/fields/Radios/Radios.component.js
-  23:7  error  Form controls using a label to identify them must be programmatically associated with the control using htmlFor  jsx-a11y/label-has-for
+  24:7  error  Form controls using a label to identify them must be programmatically associated with the control using htmlFor  jsx-a11y/label-has-for
 
-✖ 3 problems (3 errors, 0 warnings)
+/home/travis/build/Talend/ui/packages/forms/src/UIForm/UIForm.component.js
+  4:1  error  'tv4' should be listed in the project's dependencies. Run 'npm i -S tv4' to add it  import/no-extraneous-dependencies
```
jmfrancois: fixed, waiting for CI

PR #895, opened 2017-12-08
fix(components): disable the overlay when inProgress

jare-talend:

When we set the props InProgress with an overlay, the overlay shows

### DIFF
--- a/packages/components/src/Actions/ActionButton/ActionButton.test.js
+++ b/packages/components/src/Actions/ActionButton/ActionButton.test.js
@@ -126,6 +126,7 @@ describe('Action', () => {
 		const wrapper = shallow(<ActionButton {...props} />);
 
 		// then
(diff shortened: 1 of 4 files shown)


packages/components/src/Actions/ActionButton/ActionButton.test.js
```diff
@@ -110,6 +110,25 @@ describe('Action', () => {
...
+			inProgress: true,
+			overlayComponent: OverlayComponent,
+			overlayPlacement: 'bottom',
+		};
+
+		// when
+		const wrapper = shallow(<ActionButton {...props} />);
+
+		// then
+		expect(wrapper.getNode()).toMatchSnapshot();
```
jmfrancois: snapshot are not read today because we have too much.  instead expect you try to find it but not found.

PR #897, opened 2017-12-11
fix(forms/Enumeration): fix error on search then edit

Sylvain-Talend:
When search and edit, the item is not the correct one. Please see animated GIF attached

err-enum-edit

When we search for items, the component state was not refreshed, hence the item to edit was not the correct one.
Call setState() when searching fixes the problem

:warning: I had to change some code to make PRETTIER and ESLINT work well ; Prettier was changing ternary in a weird format that result in an ESLINT error on _EnumerationWidget.js_


packages/forms/src/widgets/EnumerationWidget/EnumerationWidget.js
```diff
@@ -448,6 +448,7 @@ class EnumerationForm extends React.Component {
 			headerInput: this.searchInputsActions,
 			searchCriteria: prevState.loadingSearchCriteria,
 			loadingSearchCriteria: '',
+			items: this.searchItems(prevState.loadingSearchCriteria),
```
jare-talend: It seems that we loose the old items after a search

Sylvain-Talend: :+1:  in non-connected  mode only (hence not this line), it has been fixed

PR #898, opened 2017-12-11
fix(components): defaultHeight for VirtualizedList

frassinier:
VirtualizedList leads on its parent element for defining its height.
If its parent has height set to 0px (ie. inline element), it can not render any list item.
ngReact wrap React pure components into Angular component (which they are inline) and unit tests fail.

Introduce `defaultHeight` for VirtualizedList in order to not render it at 0px


packages/components/src/VirtualizedList/VirtualizedList.component.js
```diff
@@ -73,6 +74,7 @@ function VirtualizedList(props) {
 VirtualizedList.displayName = 'VirtualizedList';
 VirtualizedList.propTypes = propTypes;
 VirtualizedList.defaultProps = {
+	defaultHeight: 250,
```
jmfrancois: question: why 250 ?

packages/components/src/VirtualizedList/VirtualizedList.test.js
```diff
@@ -43,6 +43,34 @@ describe('VirtualizedList', () => {
...
+			</VirtualizedList>,
+		);
+
+		const renderer = wrapper.node.props.children;
+
+		// when
+		const rendererInstance = renderer({ height: 0, width: 300 });
+
+		// then
+		expect(rendererInstance).toMatchSnapshot();
```
jmfrancois: I don't understand this tests. sorry. What do you really expect ?

frassinier: I'm expecting `height: <DEFAULT_HEIGHT>` if AutoSizer returns `height: 0`

packages/components/src/VirtualizedList/VirtualizedList.component.js
frassinier: `$> ./randomValue.sh`, did you prefer another one?

frassinier: I don't care, I don't want `height: 0` :)

packages/components/src/VirtualizedList/VirtualizedList.test.js
jmfrancois: In that case please remove snapshots and do
```javascript
expect(rendererInstance.props().height).toBe(600)
```

PR #904, opened 2017-12-12
feat(components/drawer): fix withAnimation, fix working with router

acateland:
`withAnimation` doesn't work properly
fixing the way `withAnimation` work, adding a way for the router to manage if a component should work with or without animation

withDrawer now is able to use a children drawer props withTransition={true|false}
withDrawer now is able to use a children of any type withTransition={true|false}

and if you use cmf and that withDrawer is injected by react router you can rely on props.route.state.withTransition to manage if or if not the drawer or children containing the drawer should be injected with an animation.


packages/components/src/Drawer/Drawer.component.js
```diff
@@ -21,14 +23,15 @@ class DrawerAnimation extends React.Component {
...
+		const { children, withTransition, ...rest } = this.props;
+		const transitionDuration = withTransition ? DEFAULT_TRANSITION_DURATION : 0;
 		return (
 			<CSSTransition
 				{...rest}
 				onTransitionComplete={this.handleTransitionComplete}
 				defaultStyle={{ transform: 'translateX(100%)' }}
-				enterStyle={{ transform: transit('translateX(0%)', 350, 'ease-in-out') }}
-				leaveStyle={{ transform: transit('translateX(100%)', 350, 'ease-in-out') }}
+				enterStyle={{ transform: transit('translateX(0%)', transitionDuration, 'ease-in-out') }}
```
frassinier: What's about `enterStyle={withTransition && { transform: transit('translateX(0%)', DEFAULT_TRANSITION_DURATION, 'ease-in-out') }}`

packages/components/src/Drawer/Drawer.component.js
```diff
@@ -21,14 +23,15 @@ class DrawerAnimation extends React.Component {
...
+				leaveStyle={{ transform: transit('translateX(100%)', transitionDuration, 'ease-in-out') }}
```
frassinier: Can we create a transition object as var?

packages/components/src/WithDrawer/WithDrawer.component.js
```diff
@@ -27,14 +28,19 @@ function WithDrawer({ drawers, children }) {
...
-					>
-						{drawer}
-					</Drawer.Animation>
-				))}
+				{drawers &&
+					drawers.map((drawer, key) => (
+						<Drawer.Animation
+							withTransition={
+								get(drawer, 'props.withTransition', true) &&
+								get(drawer, 'props.route.state.withTransition')
```
frassinier: nice :+1:

packages/components/src/Drawer/Drawer.component.js
jsomsanith: Isn't it better not to return `<CSSTransition>` if `withTransition` is false ?

acateland: sadly not really, many forms with autorfocus rely on the onTransitionComplete hook, so some visual glitch don't happen.

PR #917, opened 2017-12-15
chore(Forms/EnumerationWidget) - Refacto in connected and non connected mode

Sylvain-Talend:
The PR https://github.com/Talend/ui/pull/897 is a workaround to fix the problem when searching and then editing in the dictionary. 
In **connected mode**, there is no need to filter on items : there are all up to date (as they have been just fetched from server)

I removed the modification on the old PR, and removed filtering items in the _render()_ method when connected mode.
Moreover, in non-connected mode, I have to recompute the correct item when actions (edit, search, submit) are triggered.


jlame:
issue with the disconnected remove in search mode, the same changes needs to be add (calling searchItem).

PR #919, opened 2017-12-15
chore: use yarn workspace and centralize babelrc

jsomsanith:

We have an issue with copyLibs, the process is that it compiles each modules, then copy the libs.

Let's say that we upgrade a dependency `MyAwesomeLib` in components. During build :
* containers fetch previous version of components, so previous version of package.json, it will resolve the old version of `MyAwesomeLib`
* after all resolution/compilation, copyLibs runs. It will replace component's code in containers/node_modules/@talend/react-components

But in the end, we have old version of `MyAwesomeLib` in containers/node_modules/MyAwesomeLib

I have the pb in react 16 branch, where components depends on react-autowhatever 7.0.0, upgraded to 10.1.0, and in containers, I have the 7.0.0 version

Use yarn workspace.
* it resolves common dependencies and put them in `<root_dir>/node_modules`. So contrary to npm link, there is no dependency duplication (like React)
* modules only contains their own dependencies
* create a link for all Talend/ui packages in `<root_dir>/node_modules/<package_name>`. So each package is directly impacted by other packages changes. No copyLibs needed.

Thank you copyLib.js, you served us well, RIP.

**WARNING**: you may need to update yarn


package.json
```diff
@@ -3,22 +3,13 @@
...
   },
   "scripts": {
-    "postinstall": "lerna clean --yes && lerna exec -- yarn --pure-lockfile --no-progress --mutex network && node copylibs.js",
-    "update-versions": "node version.js && lerna exec -- yarn && node copylibs.js",
-    "build": "lerna exec --ignore=generator-talend -- yarn run prepublish && node copylibs.js",
-    "build-cmf": "lerna exec --scope=*cmf -- yarn run prepublish && node copylibs.js --scope=cmf",
-    "build-components": "lerna exec --scope=*components -- yarn run prepublish && node copylibs.js --scope=components",
-    "build-containers": "lerna exec --scope=*containers -- yarn run prepublish && node copylibs.js --scope=containers",
-    "build-forms": "lerna exec --scope=*forms -- yarn run prepublish && node copylibs.js --scope=forms",
-    "build-icons": "lerna exec --scope=@talend/icons -- yarn run prepublish && node copylibs.js --scope=icons && lerna exec --scope=@talend/react-components -- yarn test -- -u",
```
frassinier: I used this one from Travis CI before script

jsomsanith: I'll revert this line

PR #923, opened 2017-12-17
chore(cmf): rename settings.view in settings.props

jmfrancois:

Naming is important.
the settings view doesn't means a lot to users.
In the past it was created for the router to associate some props to the component.
Now every cmfConnected component try to fetch those props from the settings.

Rename while keeping compatibility at the reducer level.
No one should use the settings in the mapStateToProps so removing it from the store.
The reducer warn you only in development mode if you fetch settings with 'view';


acateland:
im not sure props have already a general meaning in react context, user of the lib may misundrstood wich props those are, same apply to this.setState and this.props.setState !

packages/cmf/src/reducers/settingsReducers.js
```diff
@@ -40,14 +40,23 @@ export function attachRefs(refs, props) {
 
 /**
  * attach reference to produce a ready to use freezed object
- * @param {object} originalSettings the full settings with `views` and `ref` attribute
+ * @param {object} originalSettings the full settings with `` and `ref` attribute
```
talend-glorieux: Looks like that comment wasn't updated.

PR #924, opened 2017-12-19
feat(component/layout): Add the subheader

nicomaligne:
The new component [SubHeaderBar ](https://github.com/Talend/ui/blob/master/packages/components/src/SubHeaderBar/SubHeaderBar.component.js)is missing in the [Layout.component](https://github.com/Talend/ui/blob/master/packages/components/src/Layout/Layout.component.js).
Add it to the layout. We will used a subHeader props, it is an element so we can inject the component either the container.


packages/components/src/Layout/Layout.component.js
```diff
@@ -53,22 +40,26 @@ function Layout({ header, footer, mode, drawers, tabs, children, ...rest }) {
...
-				<div className={headerCSS}>{header}</div>
-			}
-			{Component &&
-				<Component drawers={drawers} tabs={tabs} {...rest}>{children}</Component>
-			}
-			{footer &&
-				<footer role="contentinfo" className={footerCSS}>{footer}</footer>
-			}
+			{header && <div className={headerCSS}>{header}</div>}
+			{subheader && subheader}
```
talend-glorieux: ?

nicomaligne: Haha yeah already removed in local :)

PR #932, opened 2017-12-22
fix: remove Slider from the global export

jmfrancois:

Each time we add a new component with a dependency it becomes hard for every one to update.
Think to add this deps into the project.
It's also embed in the build because all components are exposed by the global index.js.

This is the case with rc-slider.

Let's remove Slider from index.js (breaking change)

This will be done for every components so we will decrease our bundle size.
The main idea is to import component from it's folder when it's possible.
Here we have update forms package to do so.


packages/forms/src/Form.js
```diff
@@ -3,7 +3,7 @@ import React from 'react';
 
 import RJSForm from 'react-jsonschema-form/lib/index';
 
-import { Action } from '@talend/react-components';
+import Action from '@talend/react-components/lib/Actions/Action';
```
jmfrancois: this one is weird, we should put Action at the root (I think)

PR #935, opened 2017-12-22
fix(List): enforce use of selectionToggle with isSelected props

jsomsanith:
Layout storybook is broken because `itemProps.selectionToggle` is defined, which will render a selection column (checkboxes) but not `itemProps.isSelected` which is required when we have a selection column.

Fix storybook
Add a warning when one is defined and not the other, ignoring the selection


packages/components/src/VirtualizedList/utils/tablerow.js
```diff
@@ -28,6 +28,8 @@ export function insertSelectionConfiguration({ children, isSelected, selectionTo
 				{...CellCheckboxRenderer}
 			/>);
 		contentsConfiguration = [toggleColumn].concat(contentsConfiguration);
+	} else if (selectionToggle || isSelected) {
+		console.warn('VirtualizedList : props.selectionToggle goes with props.isSelected. You defined only one of those props, selection is ignored.');
```
jmfrancois: I suspect this to be printed on each render. If this is the case -> avoid this behavior and add a marker to trigger the print only once.

packages/components/src/VirtualizedList/utils/tablerow.js
```diff
@@ -26,8 +26,13 @@ export function insertSelectionConfiguration({ children, isSelected, selectionTo
 				cellDataGetter={({ rowData }) => isSelected(rowData)}
 				columnData={{ label: 'Select this element', onChange: selectionToggle }}
 				{...CellCheckboxRenderer}
-			/>);
+			/>
+		);
 		contentsConfiguration = [toggleColumn].concat(contentsConfiguration);
+	} else if (selectionToggle || isSelected) {
+		console.warn(
```
jmfrancois: I suspect this to be printed on each render. If this is the case -> avoid this behavior and add a marker to trigger the print only once.

packages/components/src/VirtualizedList/utils/tablerow.js
jsomsanith: I don't know : 
* the check should be here (in this function) because it manages the selection feature
* it is not a component but a function. This means that the flag will be managed by the instance of List. This breaks the 1st point
* having it on each render, is not a problem because it is a dev issue. But in prod mode, yes it is a problem

So what I can do is not print it if not in production mode

packages/components/src/VirtualizedList/utils/tablerow.js
jsomsanith: I don't know : 
* the check should be here (in this function) because it manages the selection feature
* it is not a component but a function. This means that the flag will be managed by the instance of List. This breaks the 1st point
* having it on each render, is not a problem because it is a dev issue. But in prod mode, yes it is a problem

So what I can do is not print it if not in production mode

PR #938, opened 2017-12-22
fix(Slider): icon should be inactive when value is null

romainseb:
Currently, when the slider value is null, we have the first icon active ( when we use icons of course )
Add a if to do so


packages/components/src/Slider/Slider.component.js
```diff
@@ -26,6 +26,9 @@ function isIconsAvailables(icons) {
  * @param {number} max - maximum value of the slider
  */
 export function getSelectedIconPosition(icons, value, min, max) {
+	if (value === null) {
+		return -1;
```
acateland: why not throwing and forcing the client of this function to handle the error case cleanly ?

romainseb: It's not an "error", it allow us to reset the slider to "no value" state.

packages/components/src/Slider/Slider.component.js
```diff
@@ -26,6 +26,9 @@ function isIconsAvailables(icons) {
  * @param {number} max - maximum value of the slider
  */
 export function getSelectedIconPosition(icons, value, min, max) {
+	if (value === null) {
```
ncomont: Why didn't you check for undefined values ?  With a simple `value == null` you could handle it at the same time.

packages/components/src/Slider/Slider.snap.test.js
```diff
@@ -63,5 +63,14 @@ describe('Slider component', () => {
 			// then
 			expect(result).toBe(2);
 		});
+
+		it('should return -1 when value is null', () => {
```
ncomont: I think that this kind of tests has nothing to do in `snapshot` tests file. `FilterBar`, for example, has a `FilterBar.snapshot.test.js` for snapshot tests and a `FilterBar.test.js` for unit tests. Could you move it ?

PR #939, opened 2017-12-26
feat(components/list): custom noRowRenderer when no rows

jare-talend:
I have two list on one same component with 2 different labels each others when the list is empty. I can't do with the i18n because the same label 'll be applied in the 2 list.

Add a property noRowsRenderer on the List components.
go up the noRowRender to get an uniq method to generate the NoRows


packages/components/src/VirtualizedList/NoRows/NoRows.component.js
```diff
@@ -6,20 +6,30 @@ import { getDefaultTranslate } from '../../translate';
 
 import theme from './NoRows.scss';
 
-function NoRows({ t }) {
+function getLabel(props) {
+	if (props.noRowsLabel !== undefined) {
```
ncomont: I think that overriding the i18n key when necessary could be more "i18n compliant"

jare-talend: I have two list on one same component with 2 different labels each others when the list is empty. I can't do with the i18n because the same label 'll be applied in the 2 list.

jare-talend: I prefer use a label to custom this by the application. This is more flexible if we have use more features from i18n (interpolation, ...)

packages/components/src/VirtualizedList/RendererSelector.test.js
```diff
@@ -93,4 +97,100 @@ describe('RendererSelector', () => {
...
+			>
+				{contentFields}
+			</RendererSelector>,
+		);
+
+		// then
+		expect(wrapper.props().noRowsRenderer()).toMatchSnapshot();
+	});
+
+	it('should render the grid with the default NoRows', () => {
```
jsomsanith: Same test than previous

packages/components/src/VirtualizedList/RendererSelector.test.js
```diff
@@ -93,4 +97,100 @@ describe('RendererSelector', () => {
...
+			>
+				{contentFields}
+			</RendererSelector>,
+		);
+
+		// then
+		expect(wrapper.props().noRowsRenderer()).toMatchSnapshot();
+	});
+
+	it('should render the table with the noRowsRenderer', () => {
```
jsomsanith: Same test than previous.
You meant to mix custom/default with table/large, but you ended up with table-custom x2 and large-default x2 
:)

packages/components/src/VirtualizedList/RendererSelector.component.js
```diff
@@ -24,6 +25,13 @@ function getRowRenderer(type) {
 	return rowRenderer;
 }
 
+export function getNoRowsRenderer(t, NoRowsRenderer) {
+	if (NoRowsRenderer) {
+		return () => <NoRowsRenderer />;
+	}
+	return () => <NoRows t={t} />;
```
jsomsanith: Instead of creating this function, I think it would be better to 
* wrap the default NoRows with a translate instead of passing `t` like the old code
* use default props to set NoRows (without `t` to pass :point_up:)

packages/components/src/List/ListToVirtualizedList/ListToVirtualizedList.component.js
```diff
@@ -53,6 +53,7 @@ export function ListToVirtualizedList(props) {
 			sortBy={sort && sort.field}
 			sortDirection={sort && sort.isDescending ? SORT_BY.DESC : SORT_BY.ASC}
 			type={props.displayMode.toUpperCase()}
+			noRowsRenderer={props.noRowsRenderer}
 			t={props.t}
```
jsomsanith: You can remove that, it no longer exists in VirtualizedList props.
Don't forget variable declaration and propTypes

packages/components/src/List/List.component.js
```diff
@@ -118,7 +113,6 @@ function List({ displayMode, id, list, toolbar, defaultHeight, t, renderers, row
 					displayMode={displayMode}
 					defaultHeight={defaultHeight}
 					rowHeight={rowHeight}
-					t={t}
```
jsomsanith: You removed this one, but you forgot the one you pass to <ListToolbar> that is outdated.
Don't forget to remove the variable declaration.

packages/components/src/List/List.component.js
```diff
@@ -129,11 +123,10 @@ function List({ displayMode, id, list, toolbar, defaultHeight, t, renderers, row
 List.propTypes = {
 	...omit(ListToolbar.propTypes, 't'),
```
jsomsanith: You can remove the `omit()` now.
Don't forget the import

packages/components/src/VirtualizedList/RendererSelector.test.js
```diff
@@ -93,4 +97,100 @@ describe('RendererSelector', () => {
...
+				sortBy={'name'}
+				sortDirection={'DESC'}
+				width={1024}
+			>
+				{contentFields}
+			</RendererSelector>,
+		);
+
+		// then
+		expect(wrapper.props().noRowsRenderer()).toMatchSnapshot();
```
jsomsanith: Your snapshots are useless in the 4 tests. That only snapshot what you pass as props.
If the purpose is to make sure that the renderer is passed through the row renderers, you should mount them with empty collection and snapshot the grid containing the mounted no row renderer

packages/components/src/List/Toolbar/Toolbar.component.js
```diff
@@ -2,6 +2,7 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import Navbar from 'react-bootstrap/lib/Navbar';
 import omit from 'lodash/omit';
```
jsomsanith: You forgot to remove this import that you no longer use

packages/components/src/VirtualizedList/RendererSelector.component.js
```diff
@@ -56,14 +56,14 @@ function RendererSelector(props) {
 				id={id}
 				isActive={isActive}
 				isSelected={isSelected}
+				noRowsRenderer={() => <props.noRowsRenderer />}
```
ncomont: If I refer to the Proptypes, this could be an element instead of a function. So could it be written like `noRowsRenderer={<props.noRowsRenderer />}`

packages/containers/examples/ExampleList.js
```diff
@@ -104,7 +104,13 @@ const ExampleList = {
 		<div>
 			<IconsProvider />
 			<div className="list-container">
-				<List {...props} items={items} rowHeight={customHeight} initialState={defaultListState} virtualized />
+				<List
+					{...props}
+					items={items}
+					rowHeight={customHeight}
+					initialState={defaultListState}
+					virtualized
```
ncomont: Is it still useful ?

packages/components/src/VirtualizedList/ListTable/ListTable.component.js
```diff
@@ -86,19 +84,18 @@ ListTable.propTypes = {
 	id: PropTypes.string,
 	isActive: PropTypes.func,
 	isSelected: PropTypes.func,
+	noRowsRenderer: PropTypes.func,
```
ncomont: Should it be `noRowsRenderer: PropTypes.oneOf(PropTypes.element, PropTypes.func),` as it is in `PropTypes.js` ?

packages/components/src/VirtualizedList/RendererSelector.component.js
jsomsanith: I tried to pass only the component `noRowsRenderer={props.noRowsRenderer}` but it failed saying that a class cannot be called like a function.
I haven't looked into react-virtualized to spot the line of code, but my assumption is that they just call the noRowsRenderer as a function, but our default NoRows is wrapped in a translate HOC that returns a class.

packages/components/src/VirtualizedList/ListTable/ListTable.component.js
jsomsanith: No it is passed by RendererSelector that inject a function no matter what (see your next comment)

packages/containers/examples/ExampleList.js
jsomsanith: Nope :)

packages/components/src/VirtualizedList/RendererSelector.component.js
jare-talend: noRowsRender need a function which 'll be called by react-virtualized

PR #941, opened 2017-12-27
feat(log): ng/react/redux error logger

jsomsanith:
To uniformise the way we log frontend errors in our apps, we need a unified way to register our logger.
This logger should be triggered on each error :
* on angularJS process
* on react process
* on redux process

**AngularJS** : decorate $exceptionHandler + configure logger in module.run
**React** : hook in BatchingStrategy to catch all errors within react
**Redux** : middleware

Apps must provide a state getter in logger configuration and a state transformer to obfuscate sensitive data.
During configuration, we configure TraceKit.
Inside each implementation, we catch the errors and call Tracekit to report the error to a backend endpoint.


packages/logging/src/redux/README.md
```diff
@@ -0,0 +1,32 @@
...
+* react logger [documentation](../react/README.md)
+
+## Configuration
+
+You need to configure the logger in order to fit your backend logging service and add your app state.
+
+```javascript
+import { createStore, applyMiddleware } from 'redux';
+import initReactLogger from '@talend/log/lib/react';
+import tLoggerMiddleware from '@talend/log/lib/redux';
```
acateland: why *t*LoggerMiddleware ?

packages/logging/src/api/transport.js
```diff
@@ -22,13 +22,6 @@ export function sendReport(payload, transportOpts, attempt = 0) {
...
-function defaultPayloadMiddleware(payload) {
-	return {
-		time: new Date(),
-		...payload,
-	};
-}
-
 const defaultHandlers = {
 	success: response => {
 		console.info('Logging: reported', response);
```
acateland: do we need to console info outside the dev environnement ?

packages/logging/src/api/payloadMiddleware.js
```diff
@@ -0,0 +1,10 @@
+export default function getStatePayloadMiddleware(getState) {
+	return function payloadMiddleware(payload) {
+		return {
+			time: new Date(),
```
acateland: would change the name to clearly state that is a browser time and as such dependent on user localisation and os preferences.
Maybe use toJSON on the Date object instance to retrieve clean date format ? see https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Objets_globaux/Date/toJSON

packages/logging/src/api/transport.js
jsomsanith: No, and in fact I don't think it's really useful even in dev mode.
I just didn't see that, let's remove it.

jsomsanith:
@acateland what do you mean by `and also a middleware for the api` ?
Is it the framework specific piece of code ?

packages/logging/src/redux/README.md
jsomsanith: `t` for Talend. But it's a default import you can name it whatever you want, it's just an example

PR #953, opened 2018-01-03
TPSVC-4079: The "can't delete system role" tip still display even no focus on it

andrii-holovko-dev:
https://jira.talendforge.org/projects/TPSVC/issues/TPSVC-4079
This PR solves problem of handling mouse events for disabled action buttons. By default disabled elements don't handle mouse events and block their further propagation (bubbling). We have to handle these events to show/hide tooltips.

The solution is to wrap the disabled element with another element which obviously is not disabled and properly handles all the events. To allow the event to be bubbled to parent element (wrapper) the solution is adding "pointer-events: none" CSS rule for disabled element (action button in our case). Eventually OverlayTrigger component could work as expected.


packages/components/src/Actions/ActionButton/ActionButton.component.js
```diff
@@ -110,13 +112,15 @@ function ActionButton(props) {
 	}
 
 	const buttonContent = getContent(props);
+	const btnIsDisabled = inProgress || disabled;
 
 	let btn = (
 		<Button
+			className={classNames({ [theme['btn-disabled']]: btnIsDisabled })}
```
jmfrancois: buttonProps may contains a className. please try to update buttonProps with a className if exists, else add it ?
```javascript
buttonProps.className = classNames(buttonProps.className, { [theme['btn-disabled']]: btnIsDisabled });
```

PR #954, opened 2018-01-03
fix(ObjectViewer): Fixes object viewer style

talend-glorieux:
Object viewer has styles issues in grid mode since changes on the tree view mode were not isolated. 

* Isolate each view style. 
* Isolate long text case into it's own story. 


packages/components/src/ObjectViewer/Table/Table.component.js
```diff
@@ -87,7 +87,12 @@ function Table({ flat, data, ...props }) {
 								const path = getAbsolutePath(index, key, flat);
 								return (
 									<td key={j}>
-										<JSONLike data={flattenRow[key]} {...props} jsonpath={path} />
+										<JSONLike
+											data={flattenRow[key]}
+											{...props}
+											jsonpath={path}
+											nativeValueClassName={'native-value-table'}
```
jmfrancois: `nativeValueClassName="native-value-table"`

packages/components/src/ObjectViewer/JSONLike/JSONLike.component.js
```diff
@@ -36,7 +36,7 @@ function stopAndSelectWithEnterOrSpace(event, { onSelect, jsonpath }) {
 	}
 }
 
-export function NativeValue({ data, edit, onSelect, onChange, jsonpath }) {
+export function NativeValue({ data, edit, className, onSelect, onChange, jsonpath }) {
```
jmfrancois: rename as `themeAttr`

jmfrancois:
The demo on containers seems broken.
Could you take a look ?

packages/components/src/ObjectViewer/JSONLike/JSONLike.scss
```diff
@@ -91,6 +91,13 @@
 		color: $dove-gray;
 	}
 
+	.native-value-table {
```
jsomsanith: This is used in Table component, but it means that this component has to know the implementation of JSONLike component.
IMO 
* this should be defined in Table.scss
* JSONLike takes a native value className and doesn't resolve it from the theme, but set it directly
* Table component resolves from its theme and pass it to JSONLike

PR #956, opened 2018-01-04
feat(cmf/component): add Inject to react-components

nicomaligne:

CMF has a `Inject` component but we can't add cmf as a dependency of our components lib.
So we have tried in the past to use renderers which is in the end hard to maintains and do not fit well with the component registry.
 
Add Inject component and two new props to replace the renderer approche:

* getComponent function which take a componentId
* component the componentId to render

The main idea is to accept this kind of props:

```javascript
[{
  component: 'ActionButton',
  componentId: 'add:dataset',
}];
```

So we do not need super complex component and we can compose in a really easy way.

This is the `children` approche on **steroid** (aka no code, just configuration)

We have take the SidePanel as an example in this PR.


packages/cmf/src/cmfConnect.js
```diff
@@ -51,6 +51,7 @@ const CMF_PROPS = [
 export const INJECTED_PROPS = [
 	'setState',
 	'state',
+	'getComponent',
```
jmfrancois: please remove it from here or the component will never have it.

packages/components/src/Inject/Inject.test.js
```diff
@@ -0,0 +1,68 @@
...
+
+import Inject, { NotFoundComponent } from './Inject.component';
+
+const error = { message: 'MyError' };
+
+describe('Inject', () => {
+	it('should render', () => {
+		const wrapper = shallow(<Inject />);
+		expect(wrapper.getElement()).toEqual(null);
+		expect(wrapper.getElement()).toMatchSnapshot();
```
jmfrancois: if it's null why do you snapshot it ?

packages/components/src/SidePanel/SidePanel.component.js
```diff
@@ -132,7 +140,7 @@ function SidePanel({
 							id: getActionId(id, action),
 							bsStyle: 'link',
 							role: 'link',
-							className: classNames(theme.link, action.className),
+							className: action.className,
```
jsomsanith: Since you removed `theme.link`, you don't need that anymore.
`action.className` is already in the resulting object with the `Object.assign({}, action)`
You can remove this line

jsomsanith:
But talked on Slack, IMO we should turn SidePanel into a component with a behavior and accept children as content.
And the menu would be a SidePanelMenu that hold the wanted style from guideline, to control the homogenous display accross app

packages/components/src/SidePanel/SidePanel.component.js
```diff
@@ -91,73 +93,75 @@ function SidePanel({
 	const expandLabel = t('SIDEPANEL_EXPAND', { defaultValue: 'Expand' });
 	const collapseTitle = t('SIDEPANEL_COLLAPSE', { defaultValue: 'Collapse' });
 	const toggleButtonTitle = docked ? expandLabel : collapseTitle;
-
+	const Components = Inject.getAll(getComponent, { Action });
 	return (
 		<nav className={navCSS} role="navigation">
+			{injected('preul')}
```
jsomsanith: This is not explicit except if
* you read it right ^^ (pre-ul and not preul that doesn't mean anything :D)
* you know the implementation

I think we should name them something like top/beforeActions/actions/afterActions/bottom
We should document it. Add a readme on sidepanel with an example of all insertion points

packages/components/src/SidePanel/SidePanel.component.js
```diff
@@ -91,73 +93,75 @@ function SidePanel({
...
 							icon="talend-opener"
 							label=""
 						/>
 					</li>
 				)}
-				{actions.map(action => {
-					const a11y = {};
-					const extra = {};
-					const isSelected = isActionSelected(action);
+				{injected('beforeactions')}
```
jsomsanith: beforeActions and actions imply that we know the implementation, to inject li tags.
We should reshape this, by adding a div with toggle button (outside of ul), then ul/li are rendered only if we have props.actions. Else, we render the injected actions, completely independent from ul/li.

packages/components/src/SidePanel/SidePanel.component.js
```diff
@@ -91,73 +93,75 @@ function SidePanel({
...
+				{actions &&
+					actions.map(action => {
```
jsomsanith: It would be good to externalize the menu in a SidePanelMenu component to have only 
```
{actions ?
    (<SidePanelMenu actions= {actions} selected={selected} getComponent={getComponent} />) :
    injected('actions')
}
```
in SidePanel render

packages/containers/examples/ExampleSidePanel.js
```diff
@@ -2,11 +2,30 @@ import React from 'react';
...
 			<SidePanel actionIds={['menu:first', 'menu:second', 'menu:third']} />
 		</div>
-	);
-}
+	),
+	'injected settings': () => (
+		<div>
+			<IconsProvider />
+			<SidePanel
+				actionsComponents={[
```
jsomsanith: With the last changes, this seems outdated

packages/components/src/Inject/Inject.component.js
```diff
@@ -0,0 +1,108 @@
...
+	};
+};
+
+/**
+ * used to inject a component with componentID & fallback on a component given if not available
+ * @param {function} getComponent the method used to resolve the component
+ * @param {string} componentId the id to fetch with getComponent method
+ * @param {object} Component The component to fallback to
+ */
+Inject.get = function injectGet(getComponent, componentId, Component) {
```
jsomsanith: Can you rename `Component` to `DefaultComponent` ?

packages/components/src/Inject/Inject.component.js
```diff
@@ -0,0 +1,108 @@
...
+		'aside': [{ component: 'MyConnectedSidePanel', toto: 'mdr' }],
+	}
+ */
+Inject.all = function injectAll(getComponent, components) {
+	if (!getComponent || !components) {
+		return nothing;
+	}
+	return key => {
+		if (Array.isArray(components[key])) {
+			return components[key].map(props => <Inject getComponent={getComponent} {...props} />);
```
jsomsanith: This is Inject.map()

output/components.eslint.txt
```diff
@@ -11,6 +11,9 @@ The react/require-extension rule is deprecated. Please use the import/extensions
 /home/travis/build/Talend/ui/packages/components/src/HeaderBar/HeaderBar.snapshot.test.js
   1:28  error  Unable to resolve path to module '@storybook/addon-storyshots'  import/no-unresolved
 
+/home/travis/build/Talend/ui/packages/components/src/Inject/Inject.component.js
+  55:1  error  Line 55 exceeds the maximum line length of 100  max-len
```
jsomsanith: Fix that please

packages/containers/examples/ExampleSidePanel.js
nicomaligne: Yep sorry about that

PR #959, opened 2018-01-05
docs: add and update existing doc

jmfrancois:

Lot's of people start to use CMF and ask questions.
There is few docs (thanks @jsomsanith for that)
The PR demo site on surge point to master documentation.

First we move to simple markdown in github.
So the start point is the README.md of CMF
Use relative link to each other sub documentation.
So each part of CMF may have it's own markdown.
We have ad one for cmfConnect.


packages/cmf/src/sagaRouter/index.md
```diff
@@ -0,0 +1,48 @@
...
+
+Those saga can now provide some side effect based on route context.
+
+## Problems example that are solved.
+
+### Basic routing problems
+
+Lets say that our application handle shortcut navigation, "Escape" button triggering change of route.
+
+We can now declare for each route, where the `ESCAPE` action type should direct the user.
```
acateland: where the `CANCEL_ACTION` action

packages/cmf/src/sagaRouter/index.md
```diff
@@ -0,0 +1,48 @@
...
+
+To provide a way to start or stop saga, depending on current route of the application.
+
+Those saga can now provide some side effect based on route context.
+
+## Problems example that are solved.
+
+### Basic routing problems
+
+Lets say that our application handle shortcut navigation, "Escape" button triggering change of route.
```
acateland: "Escape" button trigerring a `CANCEL_ACTION` to be dispatched wich should trigger a route change

packages/cmf/src/middlewares/http/index.md
```diff
@@ -0,0 +1,121 @@
...
+dispatch(action);
+```
+
+| Argument | Type | Description | Mandatory |
+|---|---|---|---|
+| type | string | One of the supported HTTP methods. | true |
+| body | object | The body to send. | false |
+| transform | function | The transformation method to call on response, to adapt it before dispatch. | false |
+| onSend | string | The action type to dispatch before fetch. This is optional but CMF will still dispatch an `@@HTTP/REQUEST` action on every request. By providing this option, CMF will dispatch an action for your specific case after the global one. | false |
+| onError | string &#124; function | The action or action creator to dispatch on fetch error. This is optional because all http error will trigger an `@@HTTP/ERRORS` action with error details. By providing this option, CMF dispatch an action for your specific case after the global one.| false |
```
romainseb: Is this still the case ? As we don't have `@@HTTP/ERRORS` when we give an onErrorHandler

packages/cmf/src/sagas/index.md
```diff
@@ -0,0 +1,158 @@
+# Sagas
+
+This modules contains a set of saga ready to use in CMF to write your business code
+
+## Matching pattern
```
acateland: all this to the bottom of the file belongs to sagaRouter/index.md

PR #966, opened 2018-01-09
feat(cmf/http): send dispatch errors when onError is a string

jare-talend:

If I put onError: 'MY_ERROR_TYPE' in an http action the global HTTP_ERROR action is not sent.
This has been done to being able to escape the global http error managment (needed in some cases)

This a side effect, it means you need to handle the the global onError your self.

In the case of onError: 'MY_ERROR_TYPE' the global error has to be handled in a saga or in a reducer. this is boring.

Let say if you need to do so just use a onError function: 

```javascript
import get from 'lodash/get';

return actions.http.get('/api/may-not-exists', {
	onError(error) {
		if (get(error, 'stack.status') !== 404) {
			return api.actions.http.httpError(error);
		}
		return {
			type: NOT_FOUND,
			cmf: {
				collectionId: `my-not-found`,
			},
			response: true,
		};
	},
	cmf: {
		collectionId: `my-found`,
	},
});
```

So just use a function to custom the handle of the error else we dipatch onError if exists and HTTP_Errors


packages/cmf/__tests__/middlewares/http.test.js
```diff
@@ -363,6 +363,41 @@ describe('CMF http middleware', () => {
 			done();
 		});
 	});
+
+	it('should httpMiddleware handle a specific TEXT onError', done => {
```
acateland: need rephrasing

packages/cmf/src/middlewares/http/middleware.js
```diff
@@ -162,7 +162,9 @@ export const httpMiddleware = ({ dispatch }) => next => action => {
 					} finally {
 						if (httpAction.onError) {
 							dispatch(onError(httpAction, errorObject));
-						} else {
+						}
+
+						if (typeof httpAction.onError !== 'function') {
```
acateland: so it is dispatched with two different way ? PR description does not reflect that

packages/cmf/src/actions/http.js
```diff
@@ -1,4 +1,71 @@
...
+
+export function httpRequest(url, config) {
+	return {
+		type: ACTION_TYPE_HTTP_REQUEST,
+		url,
+		config,
+	};
+}
+
+export function httpReducerError(error, action) {
```
acateland: i think the naming including `reducer` is wrong, since its an action

packages/cmf/src/middlewares/http/index.md
```diff
@@ -119,3 +119,35 @@ dispatch(fetchDataSets());
...
 | config | object | The rest of action configuration. | false |
+
+# HTTP Error
+
+Most of the time the error managment is centralized. You don't want to handle
+all http error code for each request.
+
+Sometimes you want to handle one particular case. For this you have onError param.
+
+In this case we want to handle 404. If not 404 stay on global HTTP error
```
acateland: Reading the code and the test case this statement is wrong shouldn't it be 

"In this case we want to handle the 404, yet the 404 is still dispatched and handled in the global error system.

packages/cmf/src/actions/http.js
jare-talend: This is an action triggered by the middleware when an error JS happens during the fetch. It is correct.

PR #968, opened 2018-01-09
fix(containers/Redirect): add cmf connect to redirect container

romainseb:
Currently, redirect container is not cmfconnected, so we can't use expressions with it
Add cmf connect to this container


packages/containers/src/Redirect/Redirect.test.js
```diff
@@ -1,18 +1,14 @@
 import React from 'react';
-import Redirect from './Redirect.component';
+import { RedirectContainer } from './Redirect.component';
```
sgendre: Maybe the file could be renamed to Redirect.container

PR #975, opened 2018-01-11
feat: add flag to show toggleAll checkbox

Innes333:
https://jira.talendforge.org/browse/TPSVC-4033
Toggle all checkbox should not be visible

toggleAllShown flag is added


packages/components/src/ListView/Items/Items.component.js
```diff
@@ -153,13 +153,15 @@ Items.propTypes = {
 	getItemHeight: PropTypes.func,
 	searchCriteria: PropTypes.string,
 	toggleAllChecked: PropTypes.bool,
+	toggleAllShown: PropTypes.bool,
```
jsomsanith: Can you rename it to match other props format ?
getItemHeight, searchCriteria, .. ---> `showToggleAll`

packages/components/src/ListView/Items/items.snapshot.test.js
```diff
@@ -32,4 +33,12 @@ describe('Items', () => {
 		// then
 		expect(toJson(wrapper)).toMatchSnapshot();
 	});
+
+	it('should render without toggleAll checkbox', () => {
+		// when
+		const wrapper = mount(<Items {...props} toggleAllShown={false} />);
+
+		// then
+		expect(toJson(wrapper)).toMatchSnapshot();
```
jsomsanith: The snapshot is pretty big, so it's hard to check if it's correct.
Can you replace it with a wrapper.find('.tc-listview-toggle-all') and you check if you find any ?

packages/components/src/ListView/Items/Items.component.js
```diff
@@ -36,7 +36,7 @@ class Items extends React.PureComponent {
 		this.renderToggleAllOrItem = this.renderToggleAllOrItem.bind(this);
 		this.renderToggleAll = this.renderToggleAll.bind(this);
 
-		this.hasToggleAll = this.props.items.length > 1;
+		this.hasToggleAll = this.props.items.length > 1 && this.props.toggleAllShown;
```
jsomsanith: Can you test the props before the length ? It is the most determinant condition.

packages/components/src/ListView/Items/items.snapshot.test.js
```diff
@@ -32,4 +33,12 @@ describe('Items', () => {
 		// then
 		expect(toJson(wrapper)).toMatchSnapshot();
 	});
+
+	it('should render without toggleAll checkbox', () => {
+		// when
+		const wrapper = mount(<Items {...props} toggleAllShown={false} />);
+
+		// then
+		expect(wrapper.find('.tc-listview-toggle-all').exists()).toBeFalsy();
```
jsomsanith: This test is not ok.
`toggleAllShown` --> `showToggleAll`

It is weird that it passes with the props name error, I looked into the snapshots and it's not a class but an id.
`.tc-listview-toggle-all` --> `#tc-listview-toggle-all`

Since it doesn't contains only snapshot tests, can you rename the file please ?

PR #987, opened 2018-01-13
chore(cmf): refactor actionCreator api

romainseb:
Currently, we use action api to register / get the actionCreator.
But actions is not the same concept as actioncreator
Create an action creator api

### DIFF
--- a/packages/cmf/src/actionCreator.js
+++ b/packages/cmf/src/actionCreator.js
@@ -1,0 +1,46 @@
+import registry from './registry';
+import CONST from './constant';
+
+/**
+ * return a function from the registry
+ * @param  {object} context
+ * @param  {string} id the id of the action creator
+ * @return {function}
+ */
+function get(context, id) {
+	const creator = context.registry[`${CONST.REGISTRY_ACTION_CREATOR_PREFIX}:${id}`];
+	if (!creator) {
+		throw new Error(`actionCreator not found in the registry: ${id}`);
+	}
+	return creator;
+}
+
+/**
+ * register your action creator. The action creator is a function with
+ * the following arguments:
+ * - event which trigger this action
+ * - data attached to the action (could contains anything)
+ * - context of the current react app (could contains registry, getState, ...)
+ * @param  {String} id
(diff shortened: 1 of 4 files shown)


packages/cmf/src/actionCreator.js
```diff
@@ -0,0 +1,46 @@
...
+function register(id, actionCreator, context) {
+	registry.addToRegistry(`${CONST.REGISTRY_ACTION_CREATOR_PREFIX}:${id}`, actionCreator, context);
+}
+
+/**
+ * This function allow to register an object with some action creators
+ * @param {object} actionCreators map of action creators
+ * @param {object} context optional context to get the registry
+ */
+function registerMany(actionCreators, context) {
```
sgendre: That's really look like component API, maybe we can make all that stuff one step further generic

packages/cmf/src/registry.js
```diff
@@ -84,6 +85,18 @@ function getFromRegistry(id, context) {
...
+ * @param {function} registerFn a function that register a item in the registry
+ */
+function getRegisterMany(registerFn) {
+	return (itemsToRegister, context) => {
+		Object.keys(itemsToRegister).forEach(key => {
+			registerFn(key, itemsToRegister[key], context);
+		});
+	};
+}
+
```
nicomaligne: Some TU will be nice :)

packages/cmf/src/registry.js
```diff
@@ -54,22 +54,23 @@ function addToRegistry(id, item, context) {
...
+				'Please check your CMF configuration, it should not move after the initial ' +
+				'configuration before bootstrap.',
 		);
 	}
 
 	const registry = getRegistry(context);
 	if (registry[id]) {
-		console.warn( // eslint-disable-line no-console
+		// eslint-disable-next-line no-console
+		console.warn(
```
nicomaligne: Not sure about that, but can we use invariant instead of console log ?

nicomaligne: And is it better ?

PR #994, opened 2018-01-15
fix(UIForms): datalist value update from props

jsomsanith:
*WARNING: this fix is only on UIForms, not the current forms*
Datalist widget holds an internal state to allow to reset the value while editing. But this state doesn't update when value changes from props.

This is visible for example with datalist within arrays.
when reordering array items, the values change, the key to reference the item is based on the index. So the widget stays and is updated. But the value in the datalist is not moved with the item reorder.

Update state within component lifecycle hook


jsomsanith:
Uni test added, thanks @ncomont

PR #996, opened 2018-01-16
feat(UIForms): add columns common title

jsomsanith:
*WARNING: this fix is only on UIForms, not the current forms*

Columns is only a box manager, but when we want a common title on top of a row of columns, the only solution is to set the title on the first columns.  
But doing that makes the columns inputs not aligned anymore.

Columns widget now accepts and displays a title, like any other "fieldset" widget.

### DIFF
--- a/packages/forms/src/UIForm/fieldsets/Columns/Columns.component.js
+++ b/packages/forms/src/UIForm/fieldsets/Columns/Columns.component.js
@@ -10,11 +10,11 @@ export default function Columns(props) {
 
 	return (
 		<div className={classNames('tf-columns', theme['tf-columns'])}>
-			{schema.title && (<legend>{schema.title}</legend>)}
+			{schema.title && <legend>{schema.title}</legend>}
 			<div className={classNames('tf-columns-items', theme.items)}>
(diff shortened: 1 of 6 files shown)


packages/forms/src/UIForm/fieldsets/Columns/Columns.component.js
```diff
@@ -10,9 +10,12 @@ export default function Columns(props) {
 
 	return (
 		<div className={classNames('tf-columns', theme['tf-columns'])}>
-			{schema.items.map(
-				(colSchema, index) => (<Widget {...restProps} key={index} schema={colSchema} />)
-			)}
+			{schema.title && (<legend>{schema.title}</legend>)}
```
acateland: i think we are missing a test here

jsomsanith: It's part of the test that is already in place, I just forgot to update the snapshot :/

jsomsanith: In fact I misread the test . Added it in the test

PR #1002, opened 2018-01-17
feat(TFD-2449): websocket deal with ping timeout optional parameters

sgendre:
The websocket doesn't dispatch any information about a possible timeout on websocket.
In particular, when the device looses the entire Internet connectivity.

Base on provided parameters and a new socketListener in your webapp (that uses cmf-cqrs/websocket), an action is dispatched when a timeout is reached.


packages/cmf-cqrs/src/middleware/smartWebsocket.js
```diff
@@ -54,15 +57,30 @@ export function startWebsocket(url, offlinebuffer, options) {
...
+		}
 	};
 	ws.onerror = function onerror(event) {
 		if (typeof onError === 'function') {
 			onError(event);
 		}
 	};
 	ws.ping = function ping() {
+		if (typeof onPing === 'function') {
+			if (!isNaN(options.pingTimeoutDelay)) {
```
talend-glorieux: Why check for `NaN`? What else could it be?

sgendre: As it is provided by app parameters on createWebsocketMiddleware and that there is no strongly type, UI can't ensure that this is a Number, which if not the case, will lead provided parameter to be replaced by 0 and will spam the dispatch of timeout.

I like rather testing it and be defensive, that is far less confusing.

packages/cmf-cqrs/src/middleware/smartWebsocket.js
```diff
@@ -1,3 +1,5 @@
+import { setTimeout } from 'timers';
```
jmfrancois: Why do you need this ?
I don't see any package.json modification, is it already availalble as a dependency ?

PR #1004, opened 2018-01-19
feat(component/E2E/List): add method to get all buttons

atriki:

The guideline mention add button must be 'info', not primary

We don't want to break current tests so we think add is always the first button.
We have added 'getButtons' to get all buttons


packages/components/e2e/component-objects/src/main/java/org/talend/component/List.java
```diff
@@ -15,7 +15,7 @@
 
     static final String LIST_SELECTOR = ".tc-list";
 
-    static final String ADD_BTN_SELECTOR = ".tc-actionbar-container .btn-primary";
+    static final String ADD_BTN_SELECTOR = ".tc-actionbar-container";
```
frassinier: What do you think about that:
```java
public class List extends Component {
    //...

    static final String ACTIONBAR_SELECTOR = ".tc-actionbar-container";
        
    static final String BTN_SELECTOR = ACTIONBAR_SELECTOR + " .btn";
    
    static final String ADD_BTN_SELECTOR = ACTIONBAR_SELECTOR + " .btn-primary";

    /**
     * Get add button of the list.
     *
     * @return WebElement button
     */
    public WebElement getAddButton() {
        return this.getElement().findElement(By.cssSelector(ADD_BTN_SELECTOR));
    }

    /**
     * Get all buttons of the list toolbar.
     *
     * @return WebElement button
     */
    public java.util.List<WebElement> getButtons() {
        return this.driver.findElements(By.cssSelector(BTN_SELECTOR));
    }

    //...
}
```

PR #1008, opened 2018-01-22
chore(List): update with Inject API

jmfrancois:

List container is quite old, it do not pass all props (inProgress for example)
It doesn't use last API nor inject (still on renderers)

dependency on need #1008
move from `renderers` to `Inject` in Action
add deprecated message on all cmf's api.action function


packages/components/src/List/List.component.js
```diff
@@ -95,26 +103,45 @@ ListToolbar.propTypes = {
...
+	rowHeight,
+	getComponent,
+	components = {},
+}) {
 	const classnames = classNames('tc-list', theme.list);
+	const injected = Inject.all(getComponent, omit(components, ['toolbar', 'list']));
+
 	return (
 		<div className={classnames}>
+			{injected('before-toolbar')}
```
frassinier: `{injected('before-list-wrap')}` ? Could be nice for something like `Breadcrumb`

packages/components/src/List/List.md
```diff
@@ -0,0 +1,48 @@
+# The List component
+
+This component display a list with two optional toolbars
```
frassinier: displays

packages/components/src/List/List.component.js
jmfrancois: yes it's for customization !! or tabs for example

packages/components/src/List/List.component.js
```diff
@@ -32,7 +33,14 @@ function ListToolbar({ id, toolbar, displayMode, list, renderers }) {
 		};
 	}
 
-	return <Toolbar {...toolbarProps} sort={!shouldHideSortOptions && toolbarProps.sort} />;
+	return (
+		<Toolbar
+			{...toolbarProps}
+			sort={!shouldHideSortOptions && toolbarProps.sort}
+			getComponent={getComponent}
```
jare-talend: You spread toolbarProps with getComponent and you pass again getComponent & components. You could remove them.

packages/containers/src/List/List.container.js
```diff
@@ -144,16 +150,13 @@ class List extends React.Component {
...
-			props.toolbar = {
-				display: {
-					...this.props.toolbar.display,
-					onChange: (e, p) => {
-						this.onSelectDisplayMode(e, p);
-					},
+		if (props.toolbar) {
+			props.toolbar.display = {
+				...props.toolbar.display,
+				onChange: (e, p) => {
```
jare-talend: Could you rename e & p by variable reader

packages/containers/src/List/List.container.js
```diff
@@ -178,36 +180,37 @@ class List extends React.Component {
...
-								{ startIndex, itemsPerPage },
-								this.context,
-							);
-						} else {
-							this.onChangePage(startIndex, itemsPerPage);
-						}
-					},
-				};
+			if (props.toolbar.pagination) {
+				const pagination = props.toolbar.pagination;
```
jare-talend: could replace this by:

pagination = {
... props.toolbar.pagination,
...pick(state, ['totalResults', 'itemsPerPage', 'startIndex']
}

more reader no ?

jmfrancois: yes but it break the reference so it will not update props.toolbar.pagination

PR #1009, opened 2018-01-22
feat(UIForms): conditional render

jsomsanith:
*WARNING : this is only on UIForms, the next version of forms, not the current one*
No conditional render.

Define a UISchema condition format to enable/disable fields depending on the form data.

conditional-render


packages/forms/src/UIForm/Widget/Widget.component.js
```diff
@@ -1,15 +1,27 @@
...
 import { sfPath } from 'talend-json-schema-form-core';
 
 import defaultWidgets from '../utils/widgets';
 import { getValue } from '../utils/properties';
 
+function shouldRender(conditions, properties) {
+	return (
+		!conditions ||
+		conditions.every(cond => {
+			const value = get(properties, cond.path);
```
sgendre: would it be useful to add a negate property on condition? it will avoid to list all values minus 1 and shorten form definition if such scenario may exists.

sgendre: - indeed operators function + filter + length can match that goal rather than sticking to equality condition only
- this implementation has no option about configuring something like "one of all conditions" to be true instead of "all conditions" to be true 

once again if such scenarios may exists

jsomsanith: Yes, baby steps :)
the default mode is the one implemented. When we need another one, we'll introduce operators

talend-glorieux: No need for that constant. It's only used once.

jsomsanith: Thanks, removed

PR #1012, opened 2018-01-23
feat(components/loader): customize size

ncomont:

Today, it's not possible to specify a size for the Loader component. The guideline specify that full screen loaders should be bigger than components loader.

Add a new `size` prop to the loader, then give it to the `CircularProgress` component. `SIZE` const as also be externalized and exported as `CIRCULAR_PROGRESS_SIZE`.


packages/components/stories/Loader.js
```diff
@@ -2,12 +2,31 @@ import React from 'react';
...
 import { action } from '@storybook/addon-actions';
 
-import { Loader } from '../src/';
+import { Loader, CIRCULAR_PROGRESS_SIZE as SIZE } from '../src/';
 
 
+const containerStyle = {
+	border: '1px solid black',
+	width: '500px',
+	height: '500px',
```
frassinier: What's about: 
```javascript
{
  width: '100vw',
  height: '100vh',
}
```

packages/components/stories/Loader.js
```diff
@@ -2,12 +2,31 @@ import React from 'react';
...
 storiesOf('Loader', module)
 	.addWithInfo('default', () => (
-		<section>
+		<div style={containerStyle}>
 			<Loader />
-		</section>
+		</div>
+	))
+	.addWithInfo('small', () => (
+		<div style={containerStyle}>
```
frassinier: DRY you should use decorator for wrapper

packages/components/src/Loader/Loader.snapshot.test.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
+
+import Loader from './Loader.component';
+import { CIRCULAR_PROGRESS_SIZE as SIZE } from '../constants';
+
+jest.mock('react-dom');
+
+
+describe('Loader', () => {
```
frassinier: Thanks for those tests :relaxed:

packages/components/src/constants.js
```diff
@@ -1,3 +1,8 @@
 const I18N_DOMAIN_COMPONENTS = 'tui-components';
+const CIRCULAR_PROGRESS_SIZE = {
+	small: 'small',
+	default: 'default',
+	large: 'large',
+};
 
-export default I18N_DOMAIN_COMPONENTS;
+export { I18N_DOMAIN_COMPONENTS, CIRCULAR_PROGRESS_SIZE };
```
frassinier: :scream: **Wow! Be careful!**
There is a lot of usage of `I18N_DOMAIN_COMPONENTS`.
That's why snapshots are updated...

ncomont: :D good catch, thanks

packages/components/stories/Loader.js
ncomont: 👍

packages/components/src/Loader/Loader.snapshot.test.js
```diff
@@ -0,0 +1,28 @@
...
+	it('should render a loader', () => {
+		const wrapper = renderer.create(<Loader />).toJSON();
+
+		expect(wrapper).toMatchSnapshot();
+	});
+
+	it('should render a small loader', () => {
```
jsomsanith: Your small and large tests produce the same snapshot as default :/
It's because of css modules, the classname is set to undefined.

We should modify the css module mock. Let's keep them so that we have tests when we change that.

PR #1016, opened 2018-01-24
chore: css-modules mock

ncomont:

Our css-modules mock doesn't render class names in snapshots.

Define a new Proxy, widely inspired by `identity-obj-proxy`.


packages/components/src/Loader/Loader.component.js
```diff
@@ -12,6 +12,7 @@ function Loader({ id, className, size = SIZE.default }) {
 
 	return (
 		<div id={id} aria-atomic="true" aria-busy="true" className={loaderClassNames}>
+			wehs : {size}
```
romainseb: Is this wanted ?

packages/components/src/ListView/Items/__snapshots__/items.test.js.snap
```diff
@@ -7793,7 +7793,7 @@ exports[`Items should render with provided id 1`] = `
                                 type="checkbox"
                               />
                               <span
-                                className="tc-listview-item-label"
+                                className="tc-listview-item-label tc-listview-item-label"
```
acateland: doesn't look like it always work fine

ncomont: Item's classnames is defined like that : `classNames(theme['tc-listview-item-label'], 'tc-listview-item-label');`

So yes, I think this is working as expected.

ncomont: As seen with @acateland, theme classes are now prefixed by `theme-`.

PR #1017, opened 2018-01-25
 fix(UIForms): triggers and buttons

jsomsanith:
* The buttons have no labels anymore because they get schema.label instead of schema.title (which is aligned with all other widget schema)
* triggers are not triggered anymore because it is omitted between container/component
* triggers format is the old format but we need other info for the new tcomp

* buttons : support title and label as label
* triggers are passed down again (with onReset, and other omitted props that should not)
* split into 2 apis depending on the `moz` flag (like onChange and onSubmit)


packages/forms/src/UIForm/UIForm.component.js
```diff
@@ -112,6 +112,7 @@ export default class UIForm extends React.Component {
 		}
 		this.props.setErrors(errors);
 
+		// run trigger
```
talend-glorieux: Please remove that useless comment.

packages/forms/src/UIForm/UIForm.component.js
```diff
@@ -125,9 +126,12 @@ export default class UIForm extends React.Component {
 					}
 				});
 				propertyName = schema.key[schema.key.length - 1];
+				const formId = this.props.id;
+				this.onTrigger(event, { formData, formId, propertyName, value });
```
talend-glorieux: Looks like that constant is only used once. Please remove it.

packages/forms/src/UIForm/UIForm.component.js
```diff
@@ -125,9 +126,12 @@ export default class UIForm extends React.Component {
...
+			} else {
+				const payload = { trigger: schema.triggers[0], schema, properties: formData };
+				this.onTrigger(event, payload);
```
talend-glorieux: Same here. No need to define a constant if it's to use it only once right after.

PR #1018, opened 2018-01-26
feat(cmf/sagaRouter): add isExact to saga router

yyanwang:
If you defined a router saga on /resources and /resources/:id, and then when you access /resources/:id, saga of /resources will be triggered too.

should let user know if it's an exact match.
add `isExact` param to route saga generator, also add configuration of `restartOnRouteChange` to make route saga restarted when change to another matching route.


packages/cmf/__tests__/sagaRouter/router.test.js
```diff
@@ -229,11 +268,53 @@ describe('sagaRouter route and route params', () => {
...
+						return {
+							pathname: '/resources',
+						};
+					}
+					return {
+						pathname: '/resources/action',
+					};
+				},
+			};
+		}
```
talend-glorieux: Looks like that function is duplicated here https://github.com/Talend/ui/pull/1018/files#diff-aa898b17c94b680d677cafaa355f9b62R182. Could you extract it to define it only once?

acateland:
find a missing pattern to be fixed, added a failling test.

packages/cmf/__tests__/sagaRouter/router.test.js
```diff
@@ -165,12 +170,46 @@ describe('sagaRouter RouteChange', () => {
...
 		expect(anotherGen.next({ type: '@@router/LOCATION_CHANGE' }).value).toEqual(
-			spawn(alternateRoutes['/toCancelFirst'], {})
+			spawn(alternateRoutes['/toCancelFirst'], {}, true)
 		);
 		expect(anotherGen.next(mockTask).value).toEqual(take('@@router/LOCATION_CHANGE'));
 		const anotherExpectedCancelYield = cancel(mockTask);
 		expect(anotherGen.next({ type: '@@router/LOCATION_CHANGE' }).value).toEqual(anotherExpectedCancelYield);
 	});
+
+	it('saga will NOT be restarted when change to another matching location', () => {
```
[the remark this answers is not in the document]
yyanwang: yes, you're right. will remove this one.

packages/cmf/src/sagaRouter/router.js
```diff
@@ -146,27 +154,33 @@ function parseSagaState(routeFragment, sagas, currentLocation) {
 export default function* sagaRouter(history, routes) {
 	const sagas = {};
 	const routeFragments = Object.keys(routes);
-	while (true) {  // eslint-disable-line no-constant-condition
+	while (true) {
+		// eslint-disable-line no-constant-condition
```
jmfrancois: this doesn't work, bad linter ?

output/cmf.eslint.txt
```diff
@@ -17,8 +17,9 @@ The react/require-extension rule is deprecated. Please use the import/extensions
   50:4  warning  Unexpected console statement  no-console
 
 /home/travis/build/Talend/ui/packages/cmf/src/sagaRouter/router.js
-  153:54  error  There should be no spaces inside this paren  space-in-parens
-  166:51  error  There should be no spaces inside this paren  space-in-parens
+  157:2   warning  Unexpected constant condition                no-constant-condition
```
jmfrancois: see below

packages/cmf/__tests__/sagaRouter/router.test.js
yyanwang: I've modify this test case, now they test on different cases.

packages/cmf/src/sagaRouter/router.js
yyanwang: yes, it's committed by CI's prettier :disappointed:

PR #1025, opened 2018-01-29
feat(UIForm): templates system including default Array template

jsomsanith:
*WARNING : this is on UIForm, the next version of forms, not the current one*

We need a system to customize some template.
Today it's possible to override a widget, but this implies that we need to redevelop all interactions.
Customizing a template will allow to redefining the rendering, but calling the existing actions.

Accept a `props.templates`.
For now, only `array` is available.
Array README updated.


packages/forms/src/UIForm/UIForm.component.js
```diff
@@ -43,7 +46,14 @@ export default class UIForm extends React.Component {
 			return;
 		}
 		if (Object.keys(jsonSchema).length) {
-			this.setState(merge(jsonSchema, uiSchema));
+			const merged = merge(jsonSchema, uiSchema);
```
acateland: do we really need to leverage state for this, when its value are allways derived from props and not from a side effect like user input etc ?

packages/forms/src/UIForm/UIForm.component.js
```diff
@@ -16,11 +16,14 @@ export default class UIForm extends React.Component {
 	constructor(props) {
 		super(props);
 		const { jsonSchema, uiSchema } = props;
+
+		const state = {};
 		if (Object.keys(jsonSchema).length) {
-			this.state = merge(jsonSchema, uiSchema);
-		} else {
-			this.state = {};
+			Object.assign(state, merge(jsonSchema, uiSchema));
```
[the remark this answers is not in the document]
jsomsanith: No in fac I want to spred the result of `merge()` and I add `widgets` just after, all those properties in the same object.

packages/forms/src/UIForm/UIForm.component.js
jsomsanith: Yes because merge can be heavy depending on the uiSpec. Executing that at each render can result in perf issues. 
Here we calculate this only if the uiSpec changes

packages/forms/src/UIForm/fieldsets/Array/DefaultArrayTemplate.component.js
```diff
@@ -0,0 +1,50 @@
...
+							isClosed={itemValue.isClosed}
+						>
+							{renderItem(index)}
+						</ArrayItem>
+					</li>
+				))}
+			</ol>
+			<div>
+				<button type="button" className="btn btn-info" onClick={onAdd}>
+					New Element
```
frassinier: It could not be translated :(

packages/forms/src/UIForm/fieldsets/Array/DefaultArrayTemplate.component.js
```diff
@@ -0,0 +1,50 @@
...
+							onReorder={canReorder && onReorder}
+							isClosed={itemValue.isClosed}
+						>
+							{renderItem(index)}
+						</ArrayItem>
+					</li>
+				))}
+			</ol>
+			<div>
+				<button type="button" className="btn btn-info" onClick={onAdd}>
```
frassinier: Can we use Action?

packages/forms/src/UIForm/fieldsets/Array/__snapshots__/DefaultArrayTemplate.component.test.js.snap
```diff
@@ -0,0 +1,77 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Default Array Template component should render default array template 1`] = `
+<div>
+  <legend>
```
frassinier: :warning: `legend` is designed to have 1:1 relation to its related `fieldset`.

PR #1026, opened 2018-01-30
fix(cmf-cqrs): Handle websocket timeout on creation

sgendre:
On app launch and websocket creation, if the backend service is unavailable, user get notified way too late, something like near 1 minute.
Align it with the timeout defined for ping-pong heartbeat.

Please note that the mock websocket lib does not provide enough precision to test this use case.


packages/cmf-cqrs/src/middleware/smartWebsocket.js
```diff
@@ -79,6 +80,15 @@ export function startWebsocket(url, offlinebuffer, options) {
...
 	};
+
+	function wsStillNotYetOpened() {
+		if (ws.readyState === WebSocket.CONNECTING) {
+			ws.onpingtimeout();
+		}
+		clearInterval(wsCreationIntervalId);
+	}
+
+	wsCreationIntervalId = setInterval(wsStillNotYetOpened, options.pingTimeoutDelay);
```
jmfrancois: options.pingTimeoutDelay is not required, so please check it exists

packages/cmf-cqrs/src/middleware/smartWebsocket.js
```diff
@@ -79,6 +80,18 @@ export function startWebsocket(url, offlinebuffer, options) {
 		}
 		ws.close();
 	};
+
+	function wsStillNotYetOpened() {
+		if (ws.readyState === WebSocket.CONNECTING) {
+			ws.onpingtimeout();
+		}
+		clearInterval(wsCreationIntervalId);
```
jsomsanith: 1. you init an interval to notify if the ws is still in connecting state
2. **on the first interval call**, you notify if needed, but in all cases, you clear the interval.
I think you forgot an `else`

jsomsanith:
`Please note that the mock websocket lib does not provide enough precision to test this use case.`
--> you init the smart ws, the mock is in CONNECTING state.
--> you use jest fake timer to flush tu timeouts, and check that onpingtimeout() is called

packages/cmf-cqrs/src/middleware/smartWebsocket.js
sgendre: If the ws is not in connecting state, the interval becomes useless.
If the ws is in connecting mode, the interval is useless as as ws.onpingtimeout will close teh ws and that another is reopen by design of smartwebsocket.

I will indeed change setInterval to setTimeout

sgendre:
@jsomsanith I tried that but the point is that I can't find a way to make the mock lib staying in CONNECTING state.
Using jest.advanceTimersByTime and praying that mock lib is still before OPEN state seems unreliable to me.

jsomsanith:
Looking quickly at the ws mock code, I only see a ` this.readyState = WebSocket.OPEN;` when you call :
```
send() => delay() => if(server){ ... this.readyState = WebSocket.OPEN; }
```

Did I missed something ?
Did you had a readyState change when you tried to test it with only 
* use `jest.useFakeTimers();`
* instanciate the socket
* call `jest.runAllTimers();`

sgendre:
from what I see `constructor => delay() => if(server){ ... this.readyState = WebSocket.OPEN; or this.readyState = WebSocket.CLOSED; }`

I provide failling tests cases directly depeneding of the ws mock implementation imho, any contribution is welcome.

jsomsanith:
The problem is not in the mock, the constructor doesn't call delay or send.
The problem is that SmartWebSocket send a ping message on open. This calls send() => delay() => set status to OPEN.

The delay of send is set to 4ms. So setting a pingTimeout to 1ms does the trick.

packages/cmf-cqrs/src/middleware/smartWebsocket.test.js
```diff
@@ -187,5 +187,33 @@ describe('smart websocket tests', () => {
...
+				checkInterval: 70,
+				// < 4ms, that is the delay from ws mock to send
+				// smartWebStocket send a ping on connect, so to test the connect timeout
+				// we need to trigger timeout before any send delay
+				pingTimeoutDelay: 1,
+			};
+
+			// when
+			const result = SmartWebsocket(url, options); // eslint-disable-line
+			result.send = () => {};
```
jmfrancois: why not jest.fn ? is it called ? should it be ?

packages/cmf-cqrs/src/middleware/smartWebsocket.test.js
```diff
@@ -187,5 +187,33 @@ describe('smart websocket tests', () => {
...
+			expect(result.getReadyState()).toEqual(WebSocket.CONNECTING);
+			jest.runTimersToTime(1);
+
+			// then
+			expect(result.getReadyState()).toEqual(WebSocket.CONNECTING);
+			expect(options.onPingTimeout).toBeCalled();
```
jmfrancois: please add this test previous the runTimer to be sure it has not been called

packages/cmf-cqrs/src/middleware/smartWebsocket.test.js
jsomsanith: Oops that was a test but not needed. I'll remove it

sgendre:
lgtm, and thanks for that 4ms test range !

PR #1031, opened 2018-01-31
chore(containers): remove Form from default export

romainseb:
With Form container in default export, we must have form as dependency on the project that import containers
Remove Form from default export and import it _à la_ lodash when you need it


packages/containers/src/index.js
```diff
@@ -48,7 +47,6 @@ export {
 	Drawer,
 	DeleteResource,
 	FilterBar,
-	Form,
```
jare-talend: just indicating on the PR there is a breaking change.

PR #1044, opened 2018-02-05
feat(containers/List): pagination

yyanwang:
List container doesn't handle pagination now.
Add client side pagination - only calculate paged items from current data in collection.
A new selector is added to compute items in a page base on `componentState`.


packages/containers/src/List/List.connect.js
```diff
@@ -22,6 +26,15 @@ function getItems(state, config) {
 	return [];
 }
 
+function getPagedItems(state, config) {
+	const items = configureGetPagedItems(config)(state);
+
+	if (items) {
+		return items.toJS();
```
jmfrancois: you should never do a toJS in mapStateToProps.
This should only done during a rendering.

packages/containers/src/List/List.connect.js
```diff
@@ -16,10 +20,13 @@ function componentId(ownProps) {
 function getItems(state, config) {
 	const items = configureGetFilteredItems(config)(state);
 
-	if (items) {
-		return items.toJS();
-	}
-	return [];
+	return items || [];
```
jmfrancois: So we are in Immutable or not. if return [] the `.size` will not work
So I guess here it's more new Immutable.List()

packages/containers/src/List/List.connect.js
```diff
@@ -16,10 +20,13 @@ function componentId(ownProps) {
...
+}
+
+function getPagedItems(state, config) {
+	const items = configureGetPagedItems(config)(state);
+
+	return items || [];
```
jmfrancois: same here

packages/containers/src/List/selector.js
```diff
@@ -102,8 +103,30 @@ export function configureGetFilteredItems(configure) {
...
+	const getPagedList = createSelector(
+		[getSortedList, getComponentState(configure.collectionId)],
+		(items, componentState) => {
+			let results = items;
+			if (componentState) {
+				const startIndex = componentState.get('startIndex');
+				const itemsPerPage = componentState.get('itemsPerPage');
+
+				if (itemsPerPage > 0 && startIndex > 0) {
+					results = results.slice(
```
jmfrancois: I would like to be sure this result is cached by the selector so we don't have re-render for nothing

PR #1046, opened 2018-02-06
feat(cmf/cmfConnect): spread viewProps in mapStateToProps

yyanwang:
I want to be able to do expressions evaluation in my mapStateToProps, since currently cmf only handle expression on first level of props.

For example I can do expression under props.list for the List component

Spread viewProps when calling mapStateToProps so I can handle more stuff here.
Also it's important to notice we don't want to split ownProps and viewProps here which are props from the outside which are applied to the component.


packages/cmf/src/cmfConnect.js
```diff
@@ -111,7 +111,7 @@ export function getStateToProps({
 
 	let userProps = {};
 	if (mapStateToProps) {
-		userProps = mapStateToProps(state, ownProps, cmfProps);
+		userProps = mapStateToProps(state, { ...ownProps, ...viewProps }, cmfProps);
```
jare-talend: instead { ...ownProps, ...viewProps }, use rather { ...viewProps, ...ownProps } to allow to override the values by ownProps.

packages/cmf/src/cmfConnect.js
```diff
@@ -111,7 +111,7 @@ export function getStateToProps({
 
 	let userProps = {};
 	if (mapStateToProps) {
-		userProps = mapStateToProps(state, ownProps, cmfProps);
+		userProps = mapStateToProps(state, { ...viewProps, ...ownProps }, cmfProps);
```
jare-talend: see with @jmfrancois, overrides the props by cmf settings. We have to override the props by the settings. Just add a doc on this.

PR #1048, opened 2018-02-06
TPSVC-4185: Fixed: [IE] Continue adding roles or groups doesn't work in user drawer in IE browser

andrii-holovko-dev:
Multi Select Tag Widget behaves wrong in IE.
1. Dropdown doesn't disappear on blur (styles doesn't work in IE). 
2. When you select item from dropdown blur it doen't trigger blur event.

1. Apply styles another way.
2. Force blur event inside onSelect callback.

### DIFF
--- a/packages/forms/src/widgets/MultiSelectTagWidget/MultiSelectTagWidget.scss
+++ b/packages/forms/src/widgets/MultiSelectTagWidget/MultiSelectTagWidget.scss
@@ -35,7 +35,7 @@ $tf-typeahead-section-border-color: #e0e0e0 !default;
 			width: 100%;
 		}
 
-		&[class*="container-open"] .items-container {
+		&[class*='container-open'] .items-container {
 			display: block;
 		}
 	}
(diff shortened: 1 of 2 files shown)


packages/forms/src/widgets/MultiSelectTagWidget/MultiSelectTagWidget.scss
```diff
@@ -34,9 +34,14 @@ $tf-typeahead-section-border-color: #e0e0e0 !default;
 		> div {
 			width: 100%;
 		}
+
+		&[class*="container-open"] .items-container {
```
acateland: doesn't `&.container-open .items-container` work here ?

andrii-holovko-dev: css modules generates unique ending for classname. your solution exists in typeahead styles and works everywhere except IE. that's why i added this workaround to not touch existing workable code

PR #1049, opened 2018-02-07
fix(Notifications): failSafe

sgendre:
If an early notification is dispatched, the component may not be ready.
eg : fetch of ressource on app loading that may encounter a 403 or 500

Provide an early exit and put the error in the console.


packages/containers/src/Notification/pushNotification.js
```diff
@@ -7,9 +7,19 @@ export default function pushNotification(state, notification) {
 	}
 	const path = ['Container(Notification)', 'Notification', 'notifications'];
 	let notifs = state.cmf.components.getIn(path);
-	notifs = notifs.push(Object.assign({
-		id: objectId(),
-	}, notification));
+	if (!notifs) {
+		// eslint-disable-next-line no-console
+		console.error('Notifications are not yet available.', notification);
```
nicomaligne: Don't we prefer the use of invariant instead of console ?

sgendre: I am not fan of invariant invariant(true, ...) and invariant(notifs, ...) will only evaluate a second time the content of the if statement

talend-glorieux: I agree with @nicomaligne here. Please use `invariant` instead of plain `console.log`. It has many advantages over it.

sgendre:
@ncomont  UT added

PR #1051, opened 2018-02-08
feat(forms/array): update look & feel

acateland:
Array of object and collapsible field are not really fitting in with other form component style and ui element style

https://xd.adobe.com/spec/9767ac9f-05da-4c86-a2f5-86c193dc3bd7
enhance look & feel so it is more consistent with other form and ui element


packages/forms/src/templates/ArrayFieldTemplate.scss
```diff
@@ -87,23 +108,25 @@ $tf-filterwidget-shadow-color: $shadow !default;
 			}
 
 			fieldset {
-				> div:first-child {
-					border-bottom: 1px solid $wild-sand;
+				> :global(.form-group:nth-child(2)) {
+					margin-top: 0px;
+				}
+				> div[role='button'] {
```
jmfrancois: may be add a class on it ?

packages/forms/src/templates/ArrayFieldTemplate.scss
```diff
@@ -87,23 +108,25 @@ $tf-filterwidget-shadow-color: $shadow !default;
 			}
 
 			fieldset {
-				> div:first-child {
-					border-bottom: 1px solid $wild-sand;
+				> :global(.form-group:nth-child(2)) {
```
jmfrancois: I'm not sure this work well on all supported browser

packages/forms/src/templates/ArrayFieldTemplate.scss
```diff
@@ -118,26 +141,32 @@ $tf-filterwidget-shadow-color: $shadow !default;
...
 			}
 		}
 	}
 }
+
+.ArrayFieldTemplate::before {
+	content: attr(data-content);
+	display: inline-block;
+	text-transform: uppercase;
+	margin: 10px 0;
```
jmfrancois: use theme var please

jmfrancois:
the demo Form -> Widgets doesn't work. could you please take a look ?
`TypeError: Cannot read property 'join' of undefined`

acateland:
@jmfrancois this does happen on master, maybe lets fix it in another PR ?

packages/forms/src/templates/ArrayFieldTemplate.scss
acateland: https://caniuse.com/#feat=css-sel3

packages/forms/src/UIForm/fieldsets/Array/Array.scss
```diff
@@ -1,5 +1,29 @@
...
+		text-transform: uppercase;
+		margin: 10px 0;
+	}
+
+	fieldset > legend {
+		position: absolute;
+		width: 1px;
+		height: 1px;
+		padding: 0;
+		margin: -1px;
```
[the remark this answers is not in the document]
acateland: going to explain :p

packages/forms/src/templates/ArrayFieldTemplate.js
```diff
@@ -8,46 +8,42 @@ import IconsProvider from '@talend/react-components/lib/IconsProvider';
...
-								<Icon name="talend-caret-down" />
-							</button>
-						</div>
-					)}
-				</div>
-			}
-			<div className={theme.element}>{element.children}</div>
+		<div className={theme.arrayElement}>
+			<div className={theme.control}>
+				<button
```
sgendre: like said on top, can you use <Action ...> ?

packages/forms/src/widgets/ColumnsWidget/ColumnsWidget.scss
```diff
@@ -4,14 +4,15 @@
 
 	.column {
 		flex: 1;
+		// added padding so column are nicely separated
+		margin: $padding-normal 4px 0 4px;
```
sgendre: can you use one of the already defined $padding variables?

packages/forms/src/templates/ArrayFieldTemplate.js
acateland: like i said to @jmfrancois its the legacy Form API and i don't really wish to spend more time than necessary since we are going to move onto the new Form API

PR #1055, opened 2018-02-08
feat(cmf): add redux-saga integration at component level

jmfrancois:

Today we have the saga router, so you can start / stop saga on each route.
You can't do that at component level

If your component is cmfConnected you can pass a props.saga (string) which will be start & stop by the saga `api.sagas.component.handle` that you should register in your saga middleware.

to support the saga as a string cmf pattern you have the following api to use in your configure.js

```javascript
import bootstrap from './sagas/bootstrap';

api.sagas.register('bootstrap', bootstrap);
```


packages/cmf/src/sagas/collection.js
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * this saga ends when the collection is available
+ * @param {string} id of the collection to wait for
+ * @param {number} interval in ms
+ */
+export function* waitFor(id, interval = 10) {
+	// eslint-disable-next-line no-constant-condition
+	while (true) {
+		const collection = yield select(state => state.cmf.collections.get(id));
```
nicomaligne: Make a dediacted function instead of a anonymous one would be better no ?
By the way i think that you didn't do the TU here.

packages/cmf/src/sagas/component.js
```diff
@@ -0,0 +1,20 @@
...
+import { fork, cancel, take, takeEvery } from 'redux-saga/effects';
+import CONST from '../constant';
+import { get } from '../saga';
+
+export function* onSagaStart(action) {
+	const saga = get(action.saga);
+	if (!saga) {
+		invariant(process.env.NODE_ENV !== 'production', `The saga ${action.saga} is not registred`);
+	} else {
+		const task = yield fork(saga);
```
nicomaligne: It might be a dumb question, but do we really always want to do a fork of the saga ?

output/cmf.eslint.txt
```diff
@@ -5,7 +5,7 @@
 The react/require-extension rule is deprecated. Please use the import/extensions rule from eslint-plugin-import instead.
 
 /home/travis/build/Talend/ui/packages/cmf/src/cmfConnect.js
-  151:4  warning  Unexpected console statement  no-console
+  152:4  warning  Unexpected console statement  no-console
```
nicomaligne: Maybe use invariant instead

packages/cmf/src/api.js
```diff
@@ -38,5 +44,7 @@ export default {
 	expression,
 	route,
 	registry,
+	registerInternals,
+	sagas,
```
romainseb: sagas & saga feel me strange, could be ambiguous
could one of those ( sagas ) be renamed or moved maybe ?

packages/cmf/src/sagas/collection.js
```diff
@@ -0,0 +1,18 @@
+import { call, select } from 'redux-saga/effects';
+import { delay } from 'redux-saga';
+
+/**
+ * this saga ends when the collection is available
+ * @param {string} id of the collection to wait for
+ * @param {number} interval in ms
+ */
+export function* waitFor(id, interval = 10) {
```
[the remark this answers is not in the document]
jmfrancois: it's already in a 'collection' module :/

output/cmf.eslint.txt
jmfrancois: this one is for a deprecation warning, invariant throw an error, so it block the app in dev.

packages/cmf/src/api.js
jmfrancois: merged into one

PR #1061, opened 2018-02-13
fix(components): Subheader form title is required

jmfrancois:

Subheader doesn't handle the empty form, (you can submit empty title).
We believe we should not able to do that.

Wrap the input into a bootstrap form-group and add the has-error class if empty.
Disable the validate button if the form is empty


sgendre:
the fix of the CI is important enough to have a dedicated PR imho

jmfrancois:
title is limited to 70 characters, it's quite hard

talend-glorieux:
@jmfrancois You can add it to the description. We'll keep it when merging.

PR #1064, opened 2018-02-13
fix(TDS-3104/List): toolbar dropdown value should be in capitalize

jlame:
The sort by are displayed in uppercase
Display them in capitalize


packages/components/src/List/Toolbar/Toolbar.scss
```diff
@@ -26,6 +26,9 @@ $tc-list-toolbar-background-color: $concrete !default;
 	:global(.navbar-nav) {
 		li a {
 			color: $black;
+			&:global(.dropdown-toggle) {
+				text-transform: capitalize;
+			}
```
talend-glorieux: Why did you need to add a global `text-transform`?

jlame: if i'm correct its because the top class is scoped locally, could have wrote it that way though : 

:global {
		.navbar-nav li a {
			color: $black;
			&.dropdown-toggle {
				text-transform: capitalize;
			}
		}
	}

which might be nicer, what do you think ?

jlame:
@jare-talend its already part of the guideline at least in the screenshots

PR #1069, opened 2018-02-15
feat(cmf): add common expressions

jmfrancois:

We have no common expressions related to cmf so we must write code

provide default expressions:

* cmf.collections.get
* cmf.components.get

TODO: Wait for #1055 to register in api.registerInternals();


packages/cmf/src/expressions/immutable.js
```diff
@@ -0,0 +1,12 @@
...
+import curry from 'lodash/curry';
+
+function getInState(statePath, { context }, immutablePath, defaultValue) {
+	return _get(context.store.getState(), statePath, new Immutable.Map()).getIn(
+		immutablePath.split('.'),
+		defaultValue,
+	);
+}
+
+export default curry(getInState);
```
nicomaligne: This super selectors seems odd and a little bit complex.
I have imagined selectors  to  be more precise (collectionSelectors, componentSelectors etc).
Not a big fan of this.

jmfrancois: The first version was written this way and we add exactly the same code with just 'collections' and 'components' difference.

nicomaligne: Don't you think immutable.js is and odd name for a file ? 
It can be misleading and confusing with the library immutable.

PR #1081, opened 2018-02-16
feat(forms/templates) : add translation in ArrayFieldTemplate

smouillour:
The template ArrayFieldTemplate is not translated
and fix a bug with i18n-next to extract keys for tui-forms

Add translation on this template
Fix the bug with i18n-next


packages/forms/src/UIForm/fieldsets/Array/DefaultArrayTemplate.component.test.js
```diff
@@ -97,7 +97,9 @@ describe('Default Array Template component', () => {
 				onReorder={jest.fn()}
 				renderItem={index => <div>Render item {index}</div>}
 				schema={schema}
-				t={(key, options) => options.defaultValue}
+				t={function translate(key, options) {
```
jare-talend: I can see the edit of the package.json to remove the errors with i18next-parser with an options to ignore the variables.
I think that it is here your issue with [error] in your console. I prefer use fileFilter option i18next-parser with this file to know what is the file concerned instead of to use.

smouillour: the problem is here when the file is in node-modules (folder where i18n-next extract the keys). the arrow function is transformed by a function named t(key, option). so to fix the problem i transformed the arrow function to a standard function. I should not need to use the option "ignore-variables".

packages/forms/src/templates/ArrayFieldTemplate.js
```diff
@@ -17,7 +21,7 @@ function FieldTemplate({ element, cantDelete }) {
 					name={`btn-move-element-up-${element.index}`}
 					disabled={!element.hasMoveUp}
 					onClick={element.onReorderClick(element.index, element.index - 1)}
-					title="Move Up"
+					title={t('FIELD_TEMPLATE_ACTION_MOVE_UP', { defaultValue: 'Move Up' })}
```
jsomsanith: On those 3 translated keys, you should prefix it with `ARRAY_FIELD_TEMPLATE` instead of `FIELD_TEMPLATE`.
This FieldTemplate component exists only for readability, but all the file is part of the array template.

PR #1083, opened 2018-02-18
fix(cmf): add console.error on http JSError

jmfrancois:

When you do a http request and for any reason something goes bad on the JS side
CMF dispatch an action. But the first reflex as a user is to check the console.

some proptypes warning happens sometimes complains about the following props to be not supported by a tag (let say div for example).

add a console.error in that case.

exclude all props found in the warning.


output/cmf.eslint.txt
```diff
@@ -4,8 +4,11 @@
 
 The react/require-extension rule is deprecated. Please use the import/extensions rule from eslint-plugin-import instead.
 
+/home/travis/build/Talend/ui/packages/cmf/src/actions/http.js
+  30:2  warning  Unexpected console statement  no-console
```
jsomsanith: Should we suppress the warning ?

nicomaligne: Same as Jimmy

PR #1086, opened 2018-02-19
fix(cmf/sagas/collection/waitFor): add id to selector

jmfrancois:

waitFor doesn't do it's job

add missing id to the selector + add tests


packages/datagrid/lib/components/boolean-cell-renderer/default-boolean-cell-renderer.component.js
```diff
@@ -0,0 +1,32 @@
+'use strict';
```
ncomont: I think these files should not be here (?)

PR #1087, opened 2018-02-19
fix(confirm-dialog): putting button computing in mapStateToProps

jlame:
The confirm dialog container was not updating when using expressions even if it was cmfconnected. The render function wasn't called when the props were unchanged which resulted in not evaluating expressions even when there was a change.

Computing validateAction and cancelAction in mapStateToProps in order to force a render when cmf stateToProps is called.

### DIFF
--- a/packages/containers/src/ConfirmDialog/ConfirmDialog.connect.js
+++ b/packages/containers/src/ConfirmDialog/ConfirmDialog.connect.js
@@ -9,10 +9,9 @@ export function mapStateToProps(state, props, cmfProps) {
 			getState: () => state,
 		},
 	};
-	const validateAction = cmfProps.state ? cmfProps.state.getIn(['validateAction']) : undefined;
-	const cancelAction = cmfProps.state ? cmfProps.state.getIn(['cancelAction']) : undefined;
-	const model = cmfProps.state ? cmfProps.state.getIn(['model']) : undefined;
-
+	const validateAction = cmfProps.state ? cmfProps.state.get('validateAction') : undefined;
+	const cancelAction = cmfProps.state ? cmfProps.state.get('cancelAction') : undefined;
+	const model = cmfProps.state ? cmfProps.state.get('model') : cmfProps.model;
 	return {
 		validateAction: getActionsProps(context, validateAction, model),
 		cancelAction: getActionsProps(context, cancelAction, model),
--- a/packages/containers/src/ConfirmDialog/ConfirmDialog.connect.js
+++ b/packages/containers/src/ConfirmDialog/ConfirmDialog.connect.js
@@ -1,7 +1,25 @@
-import { cmfConnect } from '@talend/react-cmf';
+import { cmfConnect, api } from '@talend/react-cmf';
+import { getActionsProps } from '../actionAPI';
 import Container, { DEFAULT_STATE } from './ConfirmDialog.container';
+
+export function mapStateToProps(state, props, cmfProps) {
+	const context = {
+		registry: api.registry.getRegistry(),
+		store: {
+			getState: () => state,
+		},
+	};
+	const validateAction = cmfProps.state ? cmfProps.state.getIn(['validateAction']) : undefined;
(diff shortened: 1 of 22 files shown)


packages/containers/src/ConfirmDialog/ConfirmDialog.test.js
```diff
@@ -51,6 +71,13 @@ describe('Connected ConfirmDialog', () => {
 	});
 });
 
+describe('Connected ConfirmDialog', () => {
```
jsomsanith: Remove that, you copied/pasted the block

packages/containers/src/ConfirmDialog/ConfirmDialog.test.js
```diff
@@ -1,18 +1,36 @@
...
 
-jest.mock(
-	'@talend/react-components',
-	() => ({ ConfirmDialog: props => (<div className="tc-confirm-dialog" {...props} />) })
-);
+jest.mock('@talend/react-components', () => ({
+	ConfirmDialog: props => <div className="tc-confirm-dialog" {...props} />,
+}));
+
+describe('Action.mapStateToProps', () => {
```
jsomsanith: `Connected ConfirmDialog mapStateToProps `
Or better, move that into `describe('Connected ConfirmDialog')`

packages/containers/src/ConfirmDialog/ConfirmDialog.connect.js
```diff
@@ -1,7 +1,26 @@
...
 
+export function mapStateToProps(state, props, cmfProps) {
+	const context = {
+		registry: api.registry.getRegistry(),
+		store: {
+			getState: () => state,
+		},
+	};
+	const validateAction = cmfProps.state ? cmfProps.state.getIn(['validateAction']) : undefined;
+	const cancelAction = cmfProps.state ? cmfProps.state.getIn(['cancelAction']) : undefined;
```
jmfrancois:
```javascript
const validateAction = cmfProps.state ? cmfProps.state.getIn(['validateAction']) : cmfProps.validateAction;
const cancelAction = cmfProps.state ? cmfProps.state.getIn(['cancelAction']) : cmfProps.cancelAction;
```

packages/containers/src/ConfirmDialog/ConfirmDialog.connect.js
```diff
@@ -1,7 +1,26 @@
...
+	const model = cmfProps.state ? cmfProps.state.getIn(['model']) : undefined;
```
jmfrancois: same here

packages/containers/src/ConfirmDialog/ConfirmDialog.container.js
```diff
@@ -30,12 +31,9 @@ class ConfirmDialog extends React.Component {
 		// and adding onClick event using the retrieved  action creator
 		state.validateAction = getActionsProps(this.context, state.validateAction, state.model);
 		state.cancelAction = getActionsProps(this.context, state.cancelAction, state.model);
+		const props = { ...omit(this.props, cmfConnect.INJECTED_PROPS), ...state };
```
jmfrancois: remove line 32 and 33
```javascript
const props = { ...omit(this.props, cmfConnect.INJECTED_PROPS), ...omit(state, ['validateAction', 'cancelAction']) };
```

packages/containers/src/ConfirmDialog/ConfirmDialog.connect.js
jmfrancois: replace with get and cmfProps.model

packages/containers/src/ConfirmDialog/ConfirmDialog.container.js
jlame: as discussed, we're keeping them for now

jmfrancois:
I approve this PR but a big re-work is needed:
Do not use getActionProps at all.
We should use the Inject mecanism with just cmfConnect is enough. So the action Id are resolved by the Action container.

PR #1092, opened 2018-02-21
chore(icons): update AWS-kinesis.svg

mloussot:
The icon kinesis has now Streams colors, it will be used in more apps, so should get a neutral color

Colors were changed


talend-glorieux:
@jmfrancois What? How is this a breaking change? I'm not OK to introduce upper/lower case constrains on icon names.

jsomsanith:
We should introduce constrains in svg file names, in order to align them.
+1 with @glorieux it should be in lower case

PR #1093, opened 2018-02-22
fix(Action): move from actionAPI to props

jmfrancois:

Multiple problems:

# 1. Action is still on renderers API not the new Inject API
# 2. Action name props resolve as the actionId, so two props to do the same things.
```javascript
<Action name="my:action" />
```

is resolved but we have the following problem:

if we put actionId="my:action" and this resolve a props.name = "anAction" it will break with the error: action "anAction" not found.

`name` is an existing html props, so it should not be used to do something else.

# 3. cmf "actions" settings and API must be replace with "props" settings.

actions api has multiple problems:

* dev use them in render (it's a map state to props)
* they are less flexible than cmf props
* this api is redondant and is just used to map the settings to props.

need #1097

The CMF action api is deprecated. You should use your own component.
This api is redondant and is just used to map the settings to props.
So it will still working but this PR try to support the following migration in settings
A PR with a migration script will be done later but the idea is to go from

```javascript
{
	"actions": {
		"my-action": { ... props }
	}
}
```
to

```javascript
{
	"props": {
		"MyActionComponent#my-action": {
			...props
		}
	}
}
```

This PR ensure all Action component use the inject API


packages/containers/src/actionAPI.js
```diff
@@ -1,4 +1,9 @@
 import { api } from '@talend/react-cmf';
+import deprecated from '@talend/react-cmf/lib/deprecated';
+
+const msg = `@talend/react-containers/lib/actionAPI should not be used any more
```
acateland: a link to the doc would be awesome

packages/containers/src/Action/Action.connect.js
```diff
@@ -1,41 +1,31 @@
...
-		},
-		ownProps.actionId || ownProps.name,
-	);
+	return props;
+}
 
-	props.actionId = ownProps.actionId || ownProps.name;
-	props.displayMode = info.displayMode;
+export function mergeProps(stateProps, dispatchProps, ownProps) {
+	const props = Object.assign({}, ownProps, stateProps, dispatchProps);
```
acateland: beware Object.assign does a shallow merge, use https://lodash.com/docs/4.17.5#merge if it is intended to have a recursive merge

output/components.eslint.txt
```diff
@@ -12,6 +12,13 @@ The react/require-extension rule is deprecated. Please use the import/extensions
   1:1  error  '@storybook/react' should be listed in the project's dependencies, not devDependencies  import/no-extraneous-dependencies
   4:2  error  Unexpected require()                                                                    global-require
 
+/home/travis/build/Talend/ui/packages/components/src/HeaderBar/HeaderBar.component.js
```
acateland: hur

jmfrancois: thanks to prettier :/

PR #1097, opened 2018-02-23
feat(cmf): add dispatch to props

jmfrancois:

We can read anything in the state using expressions but we can't dispatch action and create handler using the settings.

Any props that starts with 'on' and finish with either 'Dispatch' or 'ActionCreator' will be processed by the cmfConnect to create the handler.
Working example: 'onClickDispatch', 'onClickActionCreator', ...
Possible values:

```javascript
{
	"props": {
		"SimpleButton#default": {
			"label": "What you want even an expression",
			// just to dispatch static action onClick
			"onClickDispatch": {
				"type": "SIMPLE_BUTTON_CLICKED",
				"what": "you want more"
			}
			// to dispatch action creator onClick with event, data (props) and context
			"onClickActionCreator": "mycustomactioncreator"
			// to dispatch action creator onClick with controlled arguments
			"onClickActionCreator": {
				"id": "cmf.http",
				"args": [{
					"method": "GET",
					"url": "/api/v1/foo",
					"cmf": {
						"collectionId": "foo"
					}
				}]
			}
		}
	}
}
```


packages/cmf/src/cmfConnect.js
```diff
@@ -272,14 +311,63 @@ export default function cmfConnect({
...
+						const newKey = key.replace('ActionCreator', '');
+						let actionCreator = this.props[key];
+						let args;
+						if (typeof this.props[key] === 'object') {
+							actionCreator = this.props[key].id;
+							args = this.props[key].args;
+						}
+						// eslint-disable-next-line no-param-reassign
+						props[newKey] = (event, data) => {
+							if (!args) {
```
jsomsanith: I'm not sure this ensure that the right arguments are provided to dispatchActionCreator

`dispatchActionCreator(actionCreatorId, event, data, context)`

```
let actionCreator = this.props[key];
let args;
if (typeof this.props[key] === 'object') {
	actionCreator = this.props[key].id;
	args = this.props[key].args; // QUESTION : this is data param, right ? It's from settings so it cannot have event or context, right ? But in your unit test, it's an array, not sure why.
}
if (!args) {
	args = [{ props: this.props, data }];
}
this.dispatchActionCreator(actionCreator, event, args);
```

packages/cmf/src/cmfConnect.js
```diff
@@ -272,14 +311,63 @@ export default function cmfConnect({
...
+				return props;
+			}
+
 			dispatchActionCreator(actionCreatorId, event, data, context) {
 				const extendedContext = Object.assign({}, this.context, context);
 				this.props.dispatchActionCreator(actionCreatorId, event, data, extendedContext);
 			}
 
 			render() {
+				const { toOmit, ...handlers } = this.onEvent();
```
jsomsanith: Why is that in render instead of `getDispatchToProps` ?
Other props manipulation are in `getDispatchToProps` or `getStateToProps`

packages/cmf/src/cmfConnect.md
```diff
@@ -157,6 +159,61 @@ Here instead of having click handler dispatcher hard coded into that component,
...
+					"cmf": {
+						"collectionId": "foo"
+					}
+				}]
+			}
+		}
+	}
+}
+
+// the first and thirst option will dispatch an action in redux with the event serialized.
```
frassinier: third

frassinier: with the serialized event

packages/cmf/src/cmfConnect.md
```diff
@@ -157,6 +159,61 @@ Here instead of having click handler dispatcher hard coded into that component,
...
+
+function SimpleButton ({label, onClick}) {
+    return (
+      <button onClick={props.onClick}>{label}</button>
+    );
+  }
+}
+export default cmfConnect({})(SimpleButton);
+
+// using the follwing configuration
```
frassinier: following

packages/cmf/src/cmfConnect.js
```diff
@@ -42,6 +44,40 @@ import { mapStateToViewProps } from './settings';
...
+	'screenY',
+];
+
+function serializeEvent(event) {
+	if (event.nativeEvent) {
+		return Object.assign(
+			{
+				target: pick(event.target || {}, ['type', 'value', 'checked']),
+			},
+			pick(event, KEEP_EVENT_ATTRIBUTES),
```
frassinier: Is this really needed?

packages/cmf/__tests__/cmfConnect.test.js
```diff
@@ -522,5 +522,106 @@ describe('cmfConnect', () => {
...
+			expect(context.store.dispatch).toHaveBeenCalledWith({
+				type: 'MY_BUTTON_CLICKED',
+				event: {
+					type: 'click',
+				},
+			});
+		});
+		it('should transform onEventActionCreator props to onEvent handler', () => {
+			const Button = ({ onClick, label }) => (<button onClick={onClick} >{label}</button>);
+			const CMFConnectedButton = cmfConnect({})(Button);
```
frassinier: This could be defined once by using `before` statement.

packages/cmf/__tests__/cmfConnect.test.js
```diff
@@ -522,5 +522,106 @@ describe('cmfConnect', () => {
...
+			const onClickActionCreator = 'myactionCreator';
+			const context = mock.context();
+			context.store.dispatch = jest.fn();
+			context.registry = {
+				'actionCreator:myactionCreator': () => ({ type: 'FETCH_STUFF' }),
```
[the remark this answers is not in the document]
jmfrancois: Good catch the idea was to do the same as payload in action (mean static things) but lets add it

packages/cmf/src/cmfConnect.js
jmfrancois: this built in render because you need the router in the context.
This is an option I would like to do as optimisation in an other PR:
synchronise all router info into the store so we can remove the container of cmfConnect. This will increase performance I'm sure.
the issue:
mapStateToProps -> state
mapDispatch -> dispatch
merge -> only props.
So I can't have both.
The router is not accessible at this level at the moment so we can't do better without lots of change, but this is should be  added to the roadmap.

packages/cmf/src/cmfConnect.js
jmfrancois: look at event pooling: https://reactjs.org/docs/events.html#event-pooling

frassinier: Ok 👌

packages/cmf/src/cmfConnect.js
```diff
@@ -42,6 +44,40 @@ import { mapStateToViewProps } from './settings';
...
+	'buttons',
+	'clientX',
+	'clientY',
+	'pageX',
+	'pageY',
+	'screenX',
+	'screenY',
+];
+
+function serializeEvent(event) {
```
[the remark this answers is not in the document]
frassinier: But why a white list and not a hard copy?

packages/cmf/src/cmfConnect.js
```diff
@@ -272,14 +311,63 @@ export default function cmfConnect({
 				}
 			}
 
+			onEvent() {
+				return Object.keys(this.props).reduce(this.onEventDispatch, { toOmit: [] });
+			}
+
+			onEventDispatch(props, key) {
+				if (key.startsWith('on')) {
```
jare-talend: you can use a regular expression to check that and use a constant

packages/cmf/src/cmfConnect.js
```diff
@@ -272,14 +311,63 @@ export default function cmfConnect({
...
 				const extendedContext = Object.assign({}, this.context, context);
 				this.props.dispatchActionCreator(actionCreatorId, event, data, extendedContext);
 			}
 
 			render() {
+				const { toOmit, ...handlers } = this.onEvent();
 				const props = {
-					...this.props,
+					...omit(this.props, toOmit),
+					...handlers,
```
jmfrancois: issue here, thanks @jare, we drop onEvent if it was on the props.
We must take care of composition in the build of the function onEvent

packages/cmf/src/cmfConnect.md
```diff
@@ -115,6 +115,8 @@ function MyArticle(props) {
 
 The titleExpression will be evaluated and injected as title props.
 
+So expression are state reader, they are a mapStateToProps available in JSON.
```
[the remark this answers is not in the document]
jmfrancois: evaluate = eval ... 
`Expressions are a way to read the state`

packages/cmf/src/cmfConnect.js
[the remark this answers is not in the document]
jsomsanith: Your link indicates that if you want to use it in async code, you should call `event.persist()`. Isn't that easier to maintain that your adaptation function that select a subset ?

```
If you want to access the event properties in an asynchronous way, you should call event.persist() on the event, which will remove the synthetic event from the pool and allow references to the event to be retained by user code.
```

packages/cmf/src/cmfConnect.js
jmfrancois: The main idea here is to provide static argument.
For example let say the following action creator has static arguments:
* actions.collections.addOrReplace
* actions.http
* ...
in this use case (static argumetn, it's ok)

Note there is a diff between this.dispatchActionCreator and this.props.dispatchActionCreator.
The first one insert the context for you, so here we don't need to care about.

Then `args` is an array yes, it's like expressions, so it's why it's spread.
If we remove the spread, It forces event, data, context every where. just don't know.

packages/cmf/src/cmfConnect.js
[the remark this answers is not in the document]
jmfrancois: it looks like there is still issue with event persist, but let's try: https://github.com/facebook/react/issues/2857

packages/cmf/src/cmfConnect.js
```diff
@@ -34,14 +34,24 @@ import React, { createElement } from 'react';
...
+import CONSTANT from './constant';
 
 import { statePropTypes, initState, getStateAccessors, getStateProps } from './componentState';
 import { mapStateToViewProps } from './settings';
 
 let newState;
 
+function serializeEvent(event) {
+	if (event.persist) {
+		debugger;
```
jare-talend: why do you add a debugger here. Could you replace by console.error/warn is really necessary ?

jmfrancois: my bad :(

PR #1098, opened 2018-02-25
chore: @talend/scripts module

jsomsanith:
Each app has it's build/test/lint configuration, even if they are the same or almost the same.

Create a module `@talend/scripts` that exposes bins to run common scripts with common configuration. Those configurations can be overridden or enhanced by app custom configuration.

This PR is only on build (webpack). Please refer to the README to understand how it works.
It is still on webpack 3 because there are plugins and loaders that are not ready yet :
 - resolve-url-loader : https://github.com/bholloway/resolve-url-loader/issues/79
 - extract-text-webpack-plugin : https://github.com/webpack-contrib/extract-text-webpack-plugin/issues/701
 - html-webpack-plugin : for now html-webpack-plugin@webpack-contrib/html-webpack-plugin


packages/scripts/scripts/utils.js
```diff
@@ -0,0 +1,175 @@
...
+		'          | |   \'.___`-.   | || |  | |         | || |   |  __ /    | || |      | |     | || |    |  ___/   | || |     | |      | || |   \'.___`-.   | |\n' +
+		'          | |  |`\\____) |  | || |  \\ `.___.\'\\  | || |  _| |  \\ \\_  | || |     _| |_    | || |   _| |_      | || |    _| |_     | || |  |`\\____) |  | |\n' +
+		'          | |  |_______.\'  | || |   `._____.\'  | || | |____| |___| | || |    |_____|   | || |  |_____|     | || |   |_____|    | || |  |_______.\'  | |\n' +
+		'          | |              | || |              | || |              | || |              | || |              | || |              | || |              | |\n' +
+		'          | \'--------------\' || \'--------------\' || \'--------------\' || \'--------------\' || \'--------------\' || \'--------------\' || \'--------------\' |\n' +
+		'           \'----------------\'  \'----------------\'  \'----------------\'  \'----------------\'  \'----------------\'  \'----------------\'  \'----------------\' \n' +
+		'###################################################################################################################################################################');
+}
```
talend-glorieux: Remove that useless ascii art.

packages/scripts/scripts/utils.js
```diff
@@ -0,0 +1,175 @@
...
+ * @param executable The executable name (in case the executable has a different name)
+ * @param cwd The execution path
+ * @returns {*} The executable path
+ */
+function resolveBin(modName, { executable = modName, cwd = process.cwd() } = {}) {
+	let pathFromWhich;
+	try {
+		pathFromWhich = fs.realpathSync(which.sync(executable));
+	} catch (_error) {
+		// ignore _error
```
talend-glorieux: Remove that try/catch if you don't intend of using the error.

packages/scripts/scripts/utils.js
```diff
@@ -0,0 +1,175 @@
...
+/**
+ * Resolve the bin module executable path.
+ * This is from kcd-scripts (https://github.com/kentcdodds/kcd-scripts/blob/master/src/utils.js#L21)
+ * @param modName The bin module name
+ * @param executable The executable name (in case the executable has a different name)
+ * @param cwd The execution path
+ * @returns {*} The executable path
+ */
+function resolveBin(modName, { executable = modName, cwd = process.cwd() } = {}) {
+	let pathFromWhich;
```
talend-glorieux: Rename `pathFromWhich` to something meaningful.

packages/scripts/scripts/utils.js
```diff
@@ -0,0 +1,175 @@
+/* eslint-disable global-require */
```
talend-glorieux: Please properly rename that script into something that describes its content.

packages/scripts/scripts/utils.js
```diff
@@ -0,0 +1,175 @@
...
+
+/**
+ * Resolve relative path from cwd
+ * @param dirname The folder the path starts from
+ * @param p The path
+ * @returns {string} The relative path from cwd
+ */
+function hereRelative(dirname, p) {
+	return path
+		.join(dirname, p)
```
talend-glorieux: Rename `p` to path.

packages/scripts/scripts/utils.js
```diff
@@ -0,0 +1,175 @@
...
+		return JSON.parse(process.env.TALEND_SCRIPTS_CONFIG);
+	}
+	return process.env.TALEND_SCRIPTS_CONFIG;
+}
+
+/**
+ * Create a user configuration getter
+ * @returns {getUserConfig} The user configuration getter
+ */
+function createUserConfigGetter() {
```
talend-glorieux: I don't understand why you need a getter creator. Can't you just do the call from inside the getter?

packages/scripts/scripts/utils.js
jsomsanith: path is already used by nodeJs' path

jsomsanith: I'll find something

packages/scripts/scripts/utils.js
jsomsanith: It's to memoize the talendScriptsConfig.
It is transported in environment var, so it is serialized (JSON.stringify). To get it, I parse it (JSON.parse). This allows to parse it 1 time and then each call to get just use it.

packages/scripts/scripts/utils.js
jsomsanith: No it's to not throw the error from `which` module call.

PR #1101, opened 2018-02-26
fix(components/typeahead): use a flag for the display mode

ncomont:

The only way to display the typeahead input when using the toggle functionnality is to delete the `onToggle` property, which is obviously a bad thing.

Add a `docked` property.

### DIFF
--- a/packages/components/src/Typeahead/Typeahead.component.js
+++ b/packages/components/src/Typeahead/Typeahead.component.js
@@ -19,8 +19,8 @@ import { Action } from '../Actions';
  *
  * <Typeahead {...props} />
  */
-function Typeahead({ onToggle, icon, position, ...rest }) {
-	if (onToggle) {
+function Typeahead({ onToggle, icon, position, docked, ...rest }) {
+	if (docked === true && onToggle) {
 		return (
 			<Action
 				onClick={onToggle}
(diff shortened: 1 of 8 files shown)


packages/components/src/Typeahead/Typeahead.component.js
```diff
@@ -19,8 +19,8 @@ import { Action } from '../Actions';
  *
  * <Typeahead {...props} />
  */
-function Typeahead({ onToggle, icon, position, ...rest }) {
-	if (onToggle) {
+function Typeahead({ onToggle, icon, position, docked, ...rest }) {
+	if (docked === true && onToggle) {
```
frassinier: `if (!!docked && onToggle)`

packages/components/src/Typeahead/Typeahead.snapshot.test.js
```diff
@@ -54,9 +58,7 @@ describe('Typeahead', () => {
 			};
 
 			// when
-			const wrapper = renderer
-				.create(<Typeahead {...props} />)
-				.toJSON();
+			const wrapper = renderer.create(<Typeahead {...props} />).toJSON();
```
frassinier: where are the snap?

ncomont: Why would you want to see it modified by this test ?

frassinier: Dunno, right.

PR #1057, opened 2018-02-09
feat(VirtualizedList): add a CellDatetimeRenderer

sgendre:
**What is the problem this PR is trying to solve?**
DateTime are transformed in text too early, meaning that sorting will deal with computed lexicographic value which are wrong for comparaison.

**What is the chosen solution to this problem?**
Keep initial timestamp and provide a specific renderer that can be parameterized from within settings.

**Please check if the PR fulfills these requirements**
- [x] The PR commit message follows our [guidelines](https://github.com/talend/tools/blob/master/tools-root-github/CONTRIBUTING.md)
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)
- [ ] Related design / discussions / pages (not in jira), if any, are all linked or available in the PR

**[ ] This PR introduces a breaking change**

PR #1102, opened 2018-02-26
chore: set date-fns as dep, not peer dep

jsomsanith:
#1057 introduced a kind of breaking change, as it added `date-fns` as peer dependency but use it as dependency.
Apps have to add manually the dependency to avoid a `Module not found: Error: Can't resolve 'date-fns'`

Set it as dependency instead of peer dependency


packages/components/package.json
```diff
@@ -44,6 +44,7 @@
     "url": "https://github.com/Talend/ui.git"
   },
   "dependencies": {
+    "date-fns": "1.27.2",
```
jmfrancois: if we start to do that, we should use `^` inside deps and no `^`
So the project is still able to control versions of the dependencies.

frassinier: I don't understand your point?

jsomsanith: I don't understand your first sentence. But I think I get the idea. You said that we should add a `^` wildcard to avoid apps to get 2 versions  if they rely on a version that can fit `^1.27.2` right ?

jmfrancois: yes. this need to re-work the version.js

jsomsanith: I'll change it and merge this PR

PR #1106, opened 2018-02-27
fix(forms): add static class and dynamic class name to arrayField elements

acateland:
no element was present for QA to target a specific items in an arrayField

Added a class generated on index, keyed on render, added the static class to.


packages/forms/src/UIForm/fieldsets/Array/DefaultArrayTemplate.component.js
```diff
@@ -25,7 +25,11 @@ export function DefaultArrayTemplate(props) {
 			/>
 			<ol id={id} className={classNames(theme['tf-array'], 'tf-array')}>
 				{value.map((itemValue, index) => (
-					<li className={theme.item} key={index}>
+					<li
+						className={classNames(theme.item, 'item', `item-${index}`)}
+						key={index}
+						id={`tf-array-item-${index}`}
```
jsomsanith: If you have 2 arrays you'll have id collision.
Can you include the props `id` which is specific to the property please ?

acateland: thx for the catch

PR #1107, opened 2018-02-27
chore(containers): register all components and containers

jmfrancois:

Each container are registred one by one; All components in @talend/react-components are not.
Bootstrap component like NavItem can't be used by inject.

Just register all the components and containers available in the default index.
bootstrap component are not design to be connected in cmf. They don't have Inject.
Add a wrap function to connect them and adding only one slot: 'children'


packages/containers/src/register.js
```diff
@@ -1,66 +1,26 @@
...
-	api.component.register('ObjectViewer', ObjectViewer);
-	api.component.register('Redirect', Redirect);
-	api.component.register('ShortcutManager', ShortcutManager);
-	api.component.register('SelectObject', SelectObject);
-	api.component.register('SidePanel', SidePanel);
-	api.component.register('SubHeaderBar', SubHeaderBar);
-	api.component.register('TooltipTrigger', TooltipTrigger);
-	api.component.register('TreeView', TreeView);
-	api.component.register('Typeahead', Typeahead);
+	const mycontainers = omit(containers, ['actionAPI']);
```
talend-glorieux: Replace `my` with something meaningful.

packages/containers/src/register.js
```diff
@@ -1,66 +1,26 @@
...
+	const toOmit = Object.keys(mycontainers);
+	const connected = {};
+	// connect our component which provide a great api
+	Object.keys(omit(components, toOmit)).forEach(key => {
+		if (components[key]) {
+			connected[key] = cmfConnect({})(components[key]);
+		}
+	});
```
talend-glorieux: Use a `map` instead of `forEach`. It avoids creating a contant which isn't one.

jmfrancois: `map` is not appropriate to build an object. You mean `reduce` ?

output/containers.eslint.txt
```diff
@@ -11,13 +11,18 @@ The react/require-extension rule is deprecated. Please use the import/extensions
...
 /home/travis/build/Talend/ui/packages/containers/src/renderers.js
   5:33  error  Unexpected parentheses around single function argument  arrow-parens
 
 /home/travis/build/Talend/ui/packages/containers/src/SidePanel/SidePanel.connect.js
   3:1  error  'react-router' should be listed in the project's dependencies, not devDependencies  import/no-extraneous-dependencies
 
-✖ 5 problems (5 errors, 0 warnings)
+/home/travis/build/Talend/ui/packages/containers/src/wrap.test.js
+  7:20  error  'onClick' is missing in props validation   react/prop-types
+  7:29  error  'children' is missing in props validation  react/prop-types
```
talend-glorieux: To fix.

PR #1114, opened 2018-02-28
feat(Scripts): lint:es command

jsomsanith:
Continue with @talend/scripts module.

Implement `lint:es` command.

Refactor existing scripts : 
* 1 entry point `talend-scripts` that accept a command argument (ex: talend-scripts start)
* build child environment and presetApi once, and pass it to each command


packages/scripts/README.md
```diff
@@ -22,10 +22,11 @@ yarn add --dev @talend/scripts
 3. Define the npm scripts
 ```json
 {
-  "start": "talend-scripts-start",
-  "prepublish": "talend-scripts-build",
-  "test": "talend-scripts-test",
-  "lint": "talend-scripts-lint"
+  "start": "talend-scripts start",
+  "prepublish": "talend-scripts build",
+  "test": "talend-scripts test", // TODO
```
frassinier: Deal with todo out of this JSON snippet 🙏

packages/scripts/README.md
```diff
@@ -65,22 +66,31 @@ A preset is a node module `talend-scripts-preset-${presetName}`. It exports 3 fu
...
-	getJestConfiguration(presetApi) {}, // TODO
-	getEslintConfiguration(presetApi) {} // TODO
+	getWebpackConfiguration(presetApi) { },
+	getEslintConfiguration(presetApi) { },
 }
 ```
+| Preset function | Description |
+|---|---|
+| getWebpackConfiguration | Returns the webpack configuration object. |
+| getEslintConfiguration | Returns the path to .eslintrc file. It will be passed to eslint --config option. This path should be absolute or relative from cwd. |
```
frassinier: here it's a path and it's an object for Webpack? 🤔 What do you think of `getEslintConfigurationPath` instead?

packages/scripts/README.md
```diff
@@ -65,22 +66,31 @@ A preset is a node module `talend-scripts-preset-${presetName}`. It exports 3 fu
...
+	getUserConfig,
+	hereRelative,
+} = presetApi;
 ```
 
 | Preset api | Description |
 |---|---|
 | mode | The run mode : `development` or `production` |
 | getUserConfig(path: array or string, defaultValue: any) | User configuration getter. They are defined in the `talend-scrips.json` file. It uses lodash.get under the hood, the object path can be an array of keys of a dot-separated string. |
+| hereRelative(dirname: string, relativePathFromDirname: string) | Resolve a relative path from cwd, based on the current folder and the file relative path |
```
frassinier: here relative? Do you want a `getRelativePath`as helper?

packages/scripts/index.js
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env node
+/* eslint-disable global-require,no-console */
+
+const { getEnv } = require('./utils/env');
```
frassinier: utils folder! shame! ping @glorieux

packages/scripts/scripts/build.js
```diff
@@ -1,39 +1,24 @@
...
 
-printSeparator('CONFIGURATION');
+module.exports = function build(env) {
+	// remove old builds folder
+	const rmResult = spawn.sync(
+		rimraf,
+		['./build', './dist'],
+		{ stdio: 'inherit', env }
+	);
+	if (rmResult.status === 0) {
```
frassinier: if it fails, it will stop I guess? Is this check really useful?

packages/scripts/utils/env.js
```diff
@@ -4,6 +4,17 @@ const fs = require('fs');
 const get = require('lodash.get');
 const path = require('path');
 
+/**
+ * Get talend-scripts.json
+ */
+function getScriptsConfig() {
+	const userConfigFilePath = path.join(process.cwd(), 'talend-scripts.json');
+	if (fs.existsSync(userConfigFilePath)) {
+		return require(userConfigFilePath);
```
frassinier: Be careful with several `require` executions, require has a cache system.
https://stackoverflow.com/questions/9210542/node-js-require-cache-possible-to-invalidate

packages/scripts/index.js
jsomsanith: If I rename it to `common` will you be happy ?

packages/scripts/scripts/build.js
jsomsanith: We spawn a child process, so its failure doesn't impact the current process.
That's why in `index.js` we call process.exit with the child process status

packages/scripts/utils/env.js
jsomsanith: The goal is not to support any change on the fly. Does the cache persist between calls ?
 What I understand is that you run your process, and everytime to require, it returns the one cached. It is cached the first time it is called --> within the process.

packages/scripts/README.md
jsomsanith: It's to have path from cwd. But in fact you're right, it's not usefull, we can request an absolute path from `getEslintConfigurationPath` to pass to eslint cli.
Let's remove that

packages/scripts/utils/env.js
frassinier: Right

packages/scripts/index.js
frassinier: No I don't care, just kidding :)

PR #1120, opened 2018-03-02
feat(cmf): support simple name in settings props

jmfrancois:

Most of the time it's hard to know what name we can use in the `props` section.
Anwser it's the displayName of the component you cmfConnect.
So you can have `Container(ActionButton)#default` or `Translate(HeaderBar)#default`
This also means if we add an HOC we need to update props so it's a breaking change .

Let's remove HOC from the display name keeping compatibility.
So `ActionButton#default` and `HeaderBar#default` will match

### DIFF
--- a/packages/cmf/src/settings.js
+++ b/packages/cmf/src/settings.js
@@ -27,7 +27,11 @@ export function generateDefaultViewId(viewId, componentName, componentId) {
  * @return {String} MyComponent#default
  */
 function withoutHOC(componentName) {
-	return componentName.split('(').slice(-1).pop().split(')')[0];
+	return componentName
+		.split('(')
+		.slice(-1)
+		.pop()
+		.split(')')[0];
 }
 
 /**
--- a/packages/cmf/src/settings.js
+++ b/packages/cmf/src/settings.js
@@ -22,7 +22,7 @@ export function generateDefaultViewId(viewId, componentName, componentId) {
 }
 
 /**
- * this function remove all HOC from the componentName
+ * Removes all HOC from the componentName
  * @param {String} viewId Connect(CMF(Container(MyComponent)))#default
  * @return {String} MyComponent#default
  */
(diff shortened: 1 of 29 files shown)


packages/cmf/src/settings.js
```diff
@@ -21,6 +21,19 @@ export function generateDefaultViewId(viewId, componentName, componentId) {
 	return viewId;
 }
 
+/**
+ * this function remove all HOC from the componentName
+ * @param {String} viewId Connect(CMF(Container(MyComponent)))#default
+ * @return {String} MyComponent#default
+ */
+function withoutHOC(componentName) {
+	return componentName
```
jare-talend: It seems that you don't like the regular expression :)

packages/cmf/src/settings.js
```diff
@@ -36,6 +49,11 @@ export function nonMemoized(state, ownProps, componentName, componentId) {
 
 	if (viewId && state.cmf.settings.props[viewId]) {
 		viewProps = state.cmf.settings.props[viewId] || {};
+	} else if (componentName && componentName.indexOf('(') !== -1) {
```
frassinier: I found that more readable:

`if (componentName && componentName.includes('(')) {`

packages/cmf/src/settings.js
```diff
@@ -21,6 +21,19 @@ export function generateDefaultViewId(viewId, componentName, componentId) {
 	return viewId;
 }
 
+/**
+ * this function remove all HOC from the componentName
```
frassinier: removes

packages/cmf/src/settings.js
talend-glorieux: No need for Regexp if they are not making more sense than simple string manipulation or are significantly faster.

jare-talend: split is lower than regexp. https://jsperf.com/split-regexp42
IMO, to get something in a string, I prefer use a regexp.

jmfrancois: done

PR #1124, opened 2018-03-05
chore(components): remove skeleton constants from index

jare-talend:
removing the skeleton constants from index to get only components to import by the register

set the constants on the component by  converting the component to a class component


romainseb:
We gona go on a stateless way here ->
https://github.com/Talend/ui/pull/1125

PR #1129, opened 2018-03-06
fix(HeaderBar): remove useless tooltip, that hides menu

jsomsanith:
In header bar, the username menu has a tooltip:
* It's useless because the name is never trimed
* it displays over the first menu item, hiding it

Remove it


packages/components/src/HeaderBar/HeaderBar.component.js
```diff
@@ -164,7 +164,6 @@ function User({ name, firstName, lastName, getComponent, ...rest }) {
 				icon="talend-user-circle"
 				pullRight
 				tooltipPlacement="bottom"
```
frassinier: Should be removed too

PR #1130, opened 2018-03-06
chore(ui): update process to hightlight a breaking change

jare-talend:
process to highlight the break changes.

Modify the template with a comment to replace if there is a breaking change into the pull request.


jmfrancois:
Do we remove the BREAKING_CHANGES.md file ?
We have some links that point to it, so I would say no in a first time. 
We need to udpate all underline README first
And update some of the last release tags

PR #1131, opened 2018-03-06
feat(forms): add validation check

jmfrancois:

There are many validator which exists.
Our current form lib use `jsonschema` library to validate the things.
The new uiform lib use `tv4` which is known as having issues in validation and is not performant.
The most used and fast library for jsonschema is `ajv`

let's provide external validation function you can import in your project to write cmf expressions.

How to use:
```javascript
import validate from '@talend/react-forms/lib/validate';

function isvalide(options) {
	return validate(options.jsonSchema, options.data);
}
```


packages/forms/README.md
```diff
@@ -138,6 +138,18 @@ You can use them by importing the DataPropTypes and ActionsPropTypes functions.
...
 ```
 
+### Validation
+
+You can use validation from outside (let say button outside the form) this way:
+
+```javascript
+import validate from '@talend/react-forms/lib/validate';
+
+function isvalide(options) {
```
acateland: isValid

how do i use isValid with a specific form ?
Or is this not intented to allow to supersed the basic form validation mechanism ?

packages/forms/src/validate.test.js
```diff
@@ -0,0 +1,37 @@
...
+};
+
+const typeNotValidPayload = {
+	label: 3,
+};
+
+const emptyPayload = {};
+
+describe('Form.validate', () => {
+	it('should return true if valid data', () => {
```
acateland: i think such core feature would require a bit more test with some convoluted case.

packages/forms/README.md
jmfrancois: your specific form is using jsonSchema and has data. But those can be anywhere (in a redux state or not).
In some way the json schema can be built on the fly by your component.
So this is why it's considered as external validation.
You may want to disable a button outside the form when the form is not valid.

So our form lib will use it's own validation mecanism internally to display error results.
Here we already have two libs. I have played with thoses two libs and have issues on validations.
So here I propose to use ajv which is far better, faster and looks good at the errors.
May be one day our form lib will use ajv ?

packages/forms/src/validate.js
```diff
@@ -0,0 +1,10 @@
+import Ajv from 'ajv';
+
+export default function validate(jsonSchema, data) {
+	if (!data || !jsonSchema) {
+		return false;
+	}
+	const validator = new Ajv().compile(jsonSchema);
+	const valid = validator(data);
```
sgendre: could you be more defensive about validator that might be `null` according to the lib implementation ?

jmfrancois: I don't understand null is falsy so it should not be a problem here.
you want a `return !!valid` ?

sgendre: Validator may be valued to undefined by ajv library

ncomont: So `return validator && validator(data);` could do a better job, I agree.

PR #1137, opened 2018-03-06
feat(cmf): console.log service

frassinier:
There is no service to wrap browser console

Provide a new service


packages/cmf/src/console.js
```diff
@@ -0,0 +1,59 @@
...
+	logger[method] = (...args) => {
+		if (process.env.NODE_ENV !== 'production' || LOGGER_METHODS_FOR_PRODUCTION.includes(method)) {
+			const windowConsoleMethod = windowConsole[method] || windowConsole.log;
+			if (windowConsoleMethod) {
+				const consoleArgs = Array.prototype.slice.call(args);
+				consoleArgs.unshift(logger.getPrefix());
+				windowConsoleMethod.apply(windowConsole, consoleArgs);
+			}
+		}
+		return Function.prototype;
```
jare-talend: why do you return Function.prototype ?

frassinier: NOOP, an equivalent for `() => {}`

packages/cmf/src/console.js
```diff
@@ -0,0 +1,59 @@
...
+	},
+	getPrefix() {
+		return this.prefix;
+	},
+};
+
+/**
+ * First, be sure that window.console exists
+ */
+const windowConsole = typeof window !== 'undefined' ? window.console : [];
```
jmfrancois: Do you have tested this one on IE ?

jmfrancois:
please merge it only if you have tested it on IE in your project ;)

packages/cmf/__tests__/console.test.js
```diff
@@ -0,0 +1,209 @@
...
+		it('should produce output with error method', () => {
+			// given
+			const text = 'error';
+
+			// when
+			cmfConsole.error(text);
+
+			// then
+			expect(console.error).toBeCalledWith(LOGGER_PREFIX, text);
+		});
```
ncomont: All these tests could be replaced by something like that : 

```javascript
it('should produce output with every methods', () => {
	LOGGER_METHODS.forEach(level => {
		cmfConsole[level](level);
		expect(console[level]).toBeCalledWith(LOGGER_PREFIX, level);
	});
});
```

packages/cmf/__tests__/console.test.js
```diff
@@ -0,0 +1,209 @@
...
+
+		it('should produce output with error method', () => {
+			// given
+			const text = 'error';
+
+			// when
+			cmfConsole.error(text);
+
+			// then
+			expect(console.error).toBeCalledWith(LOGGER_PREFIX, text);
```
ncomont: All these tests could be replaced by something like that : 

```javascript
it('should produce output with every methods', () => {
	LOGGER_METHODS_FOR_PRODUCTION.forEach(level => {
		cmfConsole[level](level);
		expect(console[level]).toBeCalledWith(LOGGER_PREFIX, level);
	});
});
```

packages/cmf/__tests__/console.test.js
```diff
@@ -0,0 +1,209 @@
...
+			expect(console.log).not.toBeCalled();
+		});
+
+		it('should not produce output with info method', () => {
+			// when
+			cmfConsole.info('info');
+
+			// then
+			expect(console.info).not.toBeCalled();
+		});
```
ncomont: All these tests could be replaced by something like that : 

```javascript
it('should produce output with every methods', () => {
	difference(LOGGER_METHODS, LOGGER_METHODS_FOR_PRODUCTION)
		.forEach(level => {
			cmfConsole[level](level);
			expect(console[level]).not.toBeCalled();
		});
});
```

ncomont: ( `import difference from 'lodash/difference';` )

packages/cmf/src/console.js
ncomont: I think that `() => {}` is clearer

packages/cmf/src/reducers/componentsReducers.js
```diff
@@ -6,6 +6,7 @@ import get from 'lodash/get';
 import { Map, fromJS } from 'immutable';
 import invariant from 'invariant';
 import ACTIONS from '../actions';
+import console from './../console';
```
ncomont: `../`

packages/cmf/src/reducers/settingsReducers.js
```diff
@@ -5,6 +5,7 @@
 import get from 'lodash/get';
 import invariant from 'invariant';
 import * as ACTIONS from '../actions/settingsActions';
+import console from './../console';
```
ncomont: `../`

packages/cmf/src/actions/http.js
```diff
@@ -27,7 +28,7 @@ function onRequest(url, config) {
 }
 
 function onJSError(error, action) {
-	console.error(error); // eslint-disable-line no-console
+	console.error(error);
```
romainseb: The same way we may want to use Immutable.Map instead of Map to avoid confusion, should we rename this import to something like cmfConsole to prevent the confusion with the actual console ? 
Was just a thought

PR #1147, opened 2018-03-08
chore(e2e): headerbar + notifications + large display

jlame:
HeaderBar: get/click menuItem on label
List:  no Large display selector
Notification: no notification selector

HeaderBar: get/click menuItem by id
List: create Large display selector that use existing items/cells
Notification: create notification selector


packages/components/e2e/component-objects/src/main/java/org/talend/component/HeaderBar.java
```diff
@@ -0,0 +1,55 @@
...
+
+public class HeaderBar extends Component {
+
+    private static final Logger LOGGER = LogManager.getLogger(AppHeaderBar.class);
+
+    static final String HEADER_NAME = "AppHeaderBar";
+
+    static final String HEADER_SELECTOR = ".tc-header-bar";
+
+    static final String HELP_SELECTOR = "header-help";
```
tlnd-jbrun: It is not really a selector...Maybe it would be better to rename this

packages/components/e2e/component-objects/src/main/java/org/talend/component/HeaderBar.java
```diff
@@ -0,0 +1,55 @@
...
+
+    static final String USER_SELECTOR = "header-user";
```
tlnd-jbrun: the same : not really a selector

packages/components/e2e/component-objects/src/main/java/org/talend/component/HeaderBar.java
```diff
@@ -0,0 +1,55 @@
...
+
+    static final String USER_DROP_DOWN_SELECTOR = ".dropdown-menu.dropdown-menu-right";
+
+    static final String ABOUT_SELECTOR = "//li/a[@label='About']";
```
tlnd-jbrun: is it possible to avoid xpath request ?

packages/components/e2e/component-objects/src/main/java/org/talend/component/List.java
```diff
@@ -56,20 +61,47 @@ public WebElement getAddButton() {
...
+     */
+    public WebElement getToolBarItemById(String id) {
+        return this.getElement().findElement(By.cssSelector(String.format(TOOL_BAR_ITEM_SELECTOR, id)));
+    }
+
     /**
      * Get the table display element manager.
      *
-     * @return Table the table element manager
+     * @return Large the table element manager
```
tlnd-jbrun: why did you change the javadoc ?

packages/components/e2e/component-objects/src/main/java/org/talend/component/List.java
```diff
@@ -56,20 +61,47 @@ public WebElement getAddButton() {
...
+    public Large getLarge() {
+        return new Large(driver);
+    }
+
     /**
      * Get the table display element manager identified by id.
      *
      * @param id The table id
-     * @return Table the table element manager
+     * @return Large the table element manager
```
tlnd-jbrun: same thing

packages/components/e2e/component-objects/src/main/java/org/talend/component/Notification.java
```diff
@@ -0,0 +1,32 @@
...
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+
+public class Notification extends Component {
+
+    private static final String NOTIFICATION_NAME = "Notification";
+
+    private static final String NOTIFICATION_SELECTOR = ".tc-notification-container";
+
+    private static final String NOTIFICATION_MESSAGE = ".tc-notification-error .tc-notification-message";
```
tlnd-jbrun: maybe it would be a good thingto differentiate these selectors ?

packages/components/e2e/component-objects/src/main/java/org/talend/component/list/large/Large.java
```diff
@@ -0,0 +1,131 @@
...
+                .findElements(By.cssSelector(TABLE_ITEM_SELECTOR)) //
+                .stream() //
+                .map(webElement -> new ItemLarge(driver, webElement)) //
+                .collect(toList());
+    }
+
+    /**
+     * Scroll to top
+     */
+    public void scrollToTop() {
```
tlnd-jbrun: it might be a good thing to mutualize the different functions (use an argument for exemple)
scrollToTop
canScrollDown
getItem
hasItem

because the differences are really minimal with the classe Table

packages/components/e2e/component-objects/src/main/java/org/talend/component/HeaderBar.java
tlnd-jbrun: li > a[label=About]

jlame: in fact it should be replace by a getBtnById something like that

packages/components/e2e/component-objects/src/test/java/org/talend/component/ListTest.java
```diff
@@ -1,7 +1,14 @@
 package org.talend.component;
 
+import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
+import org.openqa.selenium.By;
```
jsomsanith: Clean unused imports

packages/components/e2e/component-objects/src/main/java/org/talend/component/List.java
```diff
@@ -74,4 +108,19 @@ public Table getTable() {
...
+
+    /**
+     * Get the table display element manager.
+     *
+     * @return table the table element manager
+     */
+    public Display getDisplay() {
+        Large large = new Large(driver);
+        try {
+            large.getElementToScroll();
```
jsomsanith: Why do you get the scroll element ?
Use `getElement()` instead please. Or use display selectors, I think there is a `tc-list-large` or classes like that

packages/components/e2e/component-objects/src/test/java/org/talend/component/ListTest.java
```diff
@@ -29,4 +36,15 @@ public void should_get_table_with_id() {
         // then
         // should not throw NotFoundException
     }
+
+    @Test
+    public void should_get_table_using_display() {
+        // when
+        // should not throw NotFoundException
+        WebElement element = listObject.getDisplay().getElement();
```
jsomsanith: This var is not used

packages/components/e2e/component-objects/src/test/java/org/talend/component/display/LargeTest.java
```diff
@@ -0,0 +1,110 @@
+package org.talend.component.display;
+
+import org.junit.Before;
+import org.junit.Test;
+import org.openqa.selenium.WebElement;
+import org.talend.component.StorybookTest;
+import org.talend.component.list.Item;
+import org.talend.component.list.display.Cell;
```
jsomsanith: Clean unused import

packages/components/e2e/component-objects/src/main/java/org/talend/component/AppHeaderBar.java
```diff
@@ -152,4 +152,13 @@ public WebElement getHelp() {
     public void clickOnHelp() {
         this.getHelp().click();
     }
+
+    /**
+     * Retrieve an item by its id
+     * @param id
+     * @return the header bar element with the given id
+     */
+    public WebElement getItemById(String id) {
```
jsomsanith: Hmm not sure that this is the kind of thing we want. We create and maintain generic selectors so they follow any structure change of the component.
Ids should be used to get the component (ex: if you have 2 lists in the page, use the id to identify the list) but to work within the component, you should use the utility function we provide, or else you don't need the component-objects

jlame: well I did it because of the translation ... So i didn't want to depend on labels as they can change, didn't find better than that.

packages/components/e2e/component-objects/src/main/java/org/talend/component/Notification.java
```diff
@@ -0,0 +1,43 @@
+package org.talend.component;
+
+import org.openqa.selenium.By;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+
+import java.util.List;
+import java.util.Optional;
```
jsomsanith: Unused import

PR #1159, opened 2018-03-09
fix(components/Dialog): spread props [BR]

jmfrancois:

You can't add id, className, and other props to the modal you have to use bsDialogProps.

Spread all props to the Modal component
And remove bsDialogProps (BREAKING CHANGE)

- [x] **This PR introduces a breaking change**

Before
```javascript
	const bsDialogProps = {
		dialogClassName: 'tmc-conductor-dialog',
		keyboard: true,
		onHide: dispatchProps.cancelItemsRemove,
	};
        return <Dialog bsDialogProps={bsDialogProps}/>
```

after

```javascript
	const bsDialogProps = {
		dialogClassName: 'tmc-conductor-dialog',
		keyboard: true,
		onHide: dispatchProps.cancelItemsRemove,
	};
        return <Dialog {...bsDialogProps} />
```

tips: if you use the connected version you can do `onHideDispatch`


packages/components/src/Dialog/Dialog.component.js
```diff
@@ -50,16 +71,16 @@ Dialog.defaultProps = {
 
 Dialog.propTypes = {
 	header: PropTypes.string,
-	size: PropTypes.oneOf(['small', 'large']),
+	size: PropTypes.oneOf(['sm', 'small', 'lg', 'large']),
```
smouillour: why have 'sm' and 'small' (same for 'lg' and 'large') ?

smouillour: one alone is not enough?

jmfrancois: yes they are the same

PR #1163, opened 2018-03-12
fix(component/wrap): remove some attributes and explain why

jmfrancois:

Current modal throw the following error in log:
`Warning: Modal(...): childContextTypes cannot be defined on a functional component.`

No more log


packages/generator/package.json
```diff
@@ -5,6 +5,7 @@
   "license": "Apache-2.0",
   "scripts": {
     "start": "echo nothing to start",
+    "prepublish": "echo nothing to prepublish",
```
jmfrancois: I have added that so you can do 
`lerna exec yarn run prepublish`
this is the only package which have no prepublish script

packages/components/src/wrap.js
```diff
@@ -35,22 +44,23 @@ export default function wrap(Component, key) {
 			</Component>
 		);
 	};
-	Object.keys(Component).forEach(attr => {
-		Wrapper[attr] = Component[attr];
-	});
+	Object.keys(Component)
+		.filter(filterAttr)
```
frassinier: `isNotBlackListedAttr` instead of `filterAttr`?

packages/generator/package.json
frassinier: ok

packages/components/src/wrap.js
```diff
@@ -35,22 +44,23 @@ export default function wrap(Component, key) {
 			</Component>
 		);
 	};
-	Object.keys(Component).forEach(attr => {
-		Wrapper[attr] = Component[attr];
-	});
+	Object.keys(Component)
```
acateland: would be nice to use reduce to limit the number of iteration
would be nice to add some form of memoize to limit the number of execution

non blocking comment could be nice to add as a comment to the code because if im not mistaken it is executed A LOT

jmfrancois: This is executed once at bootstrap to expose bootstrap components and the result is exported.

PR #1165, opened 2018-03-13
fix(CMF): Register internal expressions

yyanwang:
internal expressions (eg: cmf.collections.get) are not registered when calling `api.registerInternals()`
register internal expressions so CMF users don't have do it in their apps.


packages/cmf/src/api.js
```diff
@@ -35,6 +35,7 @@ import component from './component';
 function registerInternals(context) {
 	actionCreator.register('cmf.saga.start', actions.saga.start, context);
 	actionCreator.register('cmf.saga.stop', actions.saga.stop, context);
+	Object.keys(expressions).forEach(id => expression.register(id, expressions[id], context));
```
jmfrancois: please use expression.registerMany

PR #1175, opened 2018-03-13
fix(datagrid): fix error when the schema change

romainseb:
When the schema change to remove some columns, the grid was broken and nothing could be updated anymore. This is cause by an error when some cells are rendering without values.
render the cell only if there is a "value"


packages/datagrid/src/components/DefaultCellRenderer/DefaultCellRenderer.component.js
```diff
@@ -11,6 +11,9 @@ import theme from './DefaultCell.scss';
 export const CELL_RENDERER_COMPONENT = 'cellRenderer';
 
 export default function DefaultCellRenderer({ avroRenderer, colDef, value, getComponent }) {
+	if (value === undefined) {
+		return null;
```
jmfrancois: should not we render empty div ?
or even better support a props with noValueMessage ?
What if the value === null ? or is a string ?
seems lots of case are not covered here

PR #1176, opened 2018-03-14
chore(components/e2e) : minor components modification

tlnd-jbrun:
Minor refactos : naming, mutualize WebDriverWait in the parent component,...


packages/components/e2e/component-objects/src/main/java/org/talend/component/Component.java
```diff
@@ -37,20 +43,22 @@ public Component(WebDriver driver, String name, String selector) {
...
-     * @param root Component root element
+     * @param root   Component root element
      */
     public Component(WebDriver driver, String name, WebElement root) {
         LOGGER.info("Component " + name + " from provided root");
         this.driver = driver;
         this.jsExec = (JavascriptExecutor) driver;
         this.name = name;
         this.root = root;
+        this.wait = new WebDriverWait(this.driver, 7);
```
tlnd-jbrun: should be passed as parameter

packages/components/e2e/component-objects/src/main/java/org/talend/component/Application.java
```diff
@@ -29,11 +30,36 @@
      * @param driver Selenium WebDriver
      */
     public Application(WebDriver driver) {
+        LOGGER.debug("Construct the Component Application");
```
ncomont: Could you remove this log, please ?

packages/components/e2e/component-objects/src/main/java/org/talend/component/Application.java
```diff
@@ -13,6 +13,7 @@
 
     public WebDriver driver;
 
+    // TODO (breaking change) : Rename theses variable in order to respect the Java Convention, and set them private in order to use getter
     public SidePanel SidePanel;
```
ncomont: Why didn't you made this change ?

packages/components/e2e/component-objects/src/main/java/org/talend/component/Component.java
ncomont: Why 7 ?

packages/components/e2e/component-objects/src/main/java/org/talend/component/Component.java
```diff
@@ -67,17 +75,16 @@ public Component(WebDriver driver, String name, WebElement root) {
      * Get component from its selector
      *
      * @return WebElement found with selector
-     * @throws NotFoundException if no elements are found or if more than one element are found
      */
-    public WebElement getElement() throws NotFoundException {
+    public WebElement getElement() {
```
ncomont: Why did you remove that ?

packages/components/e2e/component-objects/src/main/java/org/talend/component/AppHeaderBar.java
```diff
@@ -32,20 +32,22 @@
 
     /**
      * AppHeaderBar constructor
+     * public constructor in order to be extended in a more specific way for each project
      *
      * @param driver Selenium WebDriver
      */
-    AppHeaderBar(WebDriver driver) {
-        super(driver, NAME, SELECTOR);
+    public AppHeaderBar(WebDriver driver) {
```
ncomont: Should be private

packages/components/e2e/component-objects/src/main/java/org/talend/component/AppHeaderBar.java
```diff
@@ -12,9 +12,9 @@
  */
 public class AppHeaderBar extends Component {
 
-    private static final String NAME = "HeaderBar";
+    private static final String HEADER_BAR_NAME = "HeaderBar";
 
-    private static final String SELECTOR = ".tc-header-bar";
+    private static final String HEADER_BAR_SELECTOR = ".tc-header-bar";
```
talend-glorieux: No need to repeat `HEADER_BAR` since you are already inside the `AppHeaderBar` class.

packages/components/e2e/component-objects/src/main/java/org/talend/component/AppHeaderBar.java
```diff
@@ -12,9 +12,9 @@
  */
 public class AppHeaderBar extends Component {
 
-    private static final String NAME = "HeaderBar";
+    private static final String HEADER_BAR_NAME = "HeaderBar";
```
talend-glorieux: No need to repeat `HEADER_BAR` since you are already inside the `AppHeaderBar` class.

packages/components/e2e/component-objects/src/main/java/org/talend/component/Component.java
```diff
@@ -37,20 +43,22 @@ public Component(WebDriver driver, String name, String selector) {
         this.jsExec = (JavascriptExecutor) driver;
         this.name = name;
         this.selector = selector;
+        this.wait = new WebDriverWait(this.driver, 1);
```
talend-glorieux: No wait method should be used. 
It prevents reproducible builds and introduces latency.

packages/components/e2e/component-objects/src/main/java/org/talend/component/Component.java
talend-glorieux: No wait method should be used. 
It prevents reproducible builds and introduces latency.

packages/components/e2e/component-objects/src/main/java/org/talend/component/list/table/Item.java
```diff
@@ -17,7 +16,7 @@
  */
 public class Item extends Component {
 
-    private static final String NAME = "Item";
+    private static final String ITEM_NAME = "Item";
```
talend-glorieux: No need to specify `ITEM`. It's already a private constant of the `Item` class.

packages/components/e2e/component-objects/src/main/java/org/talend/component/list/table/Table.java
```diff
@@ -14,7 +17,7 @@
  */
 public class Table extends Component {
 
-    private static final String NAME = "Table";
+    private static final String TABLE_NAME = "Table";
```
talend-glorieux: No need to specify `TABLE`. It's already a private constant of the `Table` class.

packages/components/e2e/component-objects/src/main/java/org/talend/component/AppHeaderBar.java
tlnd-jbrun: i will discard my PR beacause this doen't have any interest but i am not sure that variable shadowing in this case it's a good practice

packages/components/e2e/component-objects/src/main/java/org/talend/component/Component.java
tlnd-jbrun: This introduce latency only if you use it, the goal is precisely to the latency. But this was already discussed with Jimmy who thinks that this is not the place to introduce it, but it is better to introduce the wait only when a component needs it. So this PR is useless

packages/components/e2e/component-objects/src/main/java/org/talend/component/Component.java
tlnd-jbrun: Because this exception is a RuntimeException so this code is useless

packages/components/e2e/component-objects/src/main/java/org/talend/component/Application.java
tlnd-jbrun: Because the variables names doesn't respect the classic Java checkstyle, and i think it is a bad practice to make them public. We should use getter. But this is a breaking change, and i'am not sure that these reasons are enough significant to deserve the breaking change. So i wanted to open the discussion.

PR #1180, opened 2018-03-16
feat(List): keep filter icon displayed when expanded

sgendre:
By default, the filter input display search-icon only when list FilterBar is collapsed.

Add a props to keep Icon visible even when list FilterBar is expanded / valued.


romainseb:
As i understand it, we would always keep the icon visible, currently on the pr's surge, this is what i get 
kapture 2018-03-19 at 8 01 26

PR #1183, opened 2018-03-19
fix: add checking on instanceOf Array, and fix some prop-types

VolodymyrKovalM:
When using Enumeration, sometimes we have item[key] that is a string.
In Item component we expect it to be an array, and because of this we do unnecessary mapping in order to convert each name from string to array.
This also causes a bug.

Add checking to Item component


packages/components/src/Enumeration/Items/Item/Item.component.js
```diff
@@ -37,6 +37,7 @@ function Item({ id, item, searchCriteria, showCheckboxes }) {
 		actions,
 		onSelectItem,
 	} = item.itemProps;
+	const actualLabel = item[key] instanceof Array ? item[key].join(',') : item[key];
```
jare-talend: could you add some test please

packages/components/src/Enumeration/Items/Item/item.snapshot.test.js
```diff
@@ -62,11 +62,22 @@ describe('Item', () => {
...
+
+	it('should display value if it is a string', () => {
+		const props = {
+			item: {
+				...item,
+				values: 'toto',
+			},
+		};
+
+		const wrapper = renderer.create(<Item {...props} />).toJSON();
```
jmfrancois: for this kind of test I prefer the use of shallow rendering + find (no snapshot please)

packages/components/src/Enumeration/Items/Item/Item.component.js
```diff
@@ -32,11 +32,8 @@ function itemDefaultActionsClasses() {
 }
 
 function Item({ id, item, searchCriteria, showCheckboxes }) {
-	const {
-		key,
-		actions,
-		onSelectItem,
-	} = item.itemProps;
+	const { key, actions, onSelectItem } = item.itemProps;
+	const actualLabel = item[key] instanceof Array ? item[key].join(',') : item[key];
```
talend-glorieux: What do you mean by `actualLabel`?

VolodymyrKovalM: We use this variable two times in Item component, in both cases it is text inside button, aka button label, and it also depends on what will be in "item[key]", whether array or a string
So, because of this dependance I decided to name it like this (I also was thinking about joinedLabel)
How would You propose to name it?

packages/components/src/Enumeration/Items/Item/Item.test.js
```diff
@@ -61,4 +63,16 @@ describe('Item', () => {
...
+	it('should display a label if "item[key]" is a string', () => {
+		const props = {
+			item: {
+				...item,
+				values: 'toto',
+			},
+		};
+
+		const wrapper = mount(<Item {...props} />);
+		expect(wrapper.find('span').text()).toEqual('toto');
```
jmfrancois: 👍

PR #1186, opened 2018-03-20
feat(cmf): add onEventSetState

jmfrancois:

We want to control the `inProgress` and some other props supported by some componetn like Action, List and so on, using simple description

At the moment some solution using saga + collection like penders exist but look complex and need saga to achieve that.

`onEventSetState`

State is immutable structure and we want to change one or more value in the state when an event is sent (most of the time will be onClick).

So let's add support of onEventSetState props into cmfConnect.

`spreadCMFState`

To make the previous one usefull we need to ask cmf to spread that state while keeping current behavior for the sake of compatibility.

Let's take the state and spread all the values as props to the pure component.

## Examples

to make action inProgress onClick (using static value)

```javascript
{
	"actions": {
		"myaction": {
			"id": "myaction",
			"icon": "talend-plus-circle",
			"label": "Add data",
			"onClickSetState": {
				"inProgress": true
			},
			"onClickActionCreator": "my:action",
			"initialState": {
				"inProgress": false
			},
			"spreadCMFState": true
		}
	}
}
```

to make it save dynamic data with onFilter(event, value) function

```javascript
{
	"props": {
		"List#mylist": {
			"spreadCMFState": true,
			"onFilterSetState": {
				"query": [1],
				"query": [0, 'target.value'] // ===
			}
		}
	}
}
```

the toggle case (SidePanel)

```javascript
{
	"props": {
		"SidePanel#default": {
			"spreadCMFState": true,
			"onToggleDockSetState": {
				"docked": "toggle"
			}
		}
	}
}
```


talend-glorieux:
@jmfrancois Wouldn't it be possible to make any `onBlaBlaSetState` method automaticaly spread to CMF state?

sgendre:
"onFilterSetState": {
	"query": [1],
	"query": [0, 'target.value'] // ===
}

is not valid json

jmfrancois:
@sam it's to show they are the same !
handler is doing that call onFilter(event, event.target.value)

packages/cmf/src/cmfConnect.js
```diff
@@ -290,40 +285,27 @@ export default function cmfConnect({
 
 			onEventDispatch(props, key) {
 				if (CONSTANT.IS_HANDLER_DISPATCH_REGEX.test(key)) {
-					const handlerKey = key.replace(CONSTANT.IS_HANDLER_DISPATCH, '');
 					props.toOmit.push(key);
```
sgendre: those 4 lines looks very similar on the 3 cases, a little function used like setHandler(props, expressionSuffix, getHandler) can shorten the code a little bit

packages/cmf/src/onEvent.js
```diff
@@ -0,0 +1,79 @@
+import get from 'lodash/get';
+
+function serializeEvent(event) {
+	if (event.persist) {
+		event.persist();
+	}
+	return event;
+}
+
+function getOnEventActionCreatorHandler(instance, config, original) {
```
sgendre: why don't you name the third argument like in the 2 other functions : currentHandler? 
It has semantic which is better for comprehension than old/original/previous.

packages/cmf/src/cmfConnect.js
```diff
@@ -290,40 +285,27 @@ export default function cmfConnect({
...
-						if (this.props[handlerKey]) {
-							this.props[handlerKey](event, data);
-						}
-					};
+					props[handlerKey] = onEvent.getOnEventActionCreatorHandler(
+						this,
+						this.props[key],
+						original,
+					);
+				} else if (CONSTANT.IS_HANDLER_SETSTATE_REGEX.test(key)) {
```
jsomsanith: We have here exactly 3 same patterns, with 3 variables : 
* the regex to test if we have an action creator, a dispatch, or a setState
* the constant string to remove from name, a suffix
* the getHandler() function

Perhaps we can factorize those duplications
```
onEventDispatch(props, key) {
	const eventConfiguration = getEventActionTypeConfiguration(key); // TODO
	if (eventConfiguration) {
		props.toOmit.push(key);
		const handlerKey = key.replace(eventConfiguration.SUFFIX, '');
		const originalHandler = props[handlerKey];
		// eslint-disable-next-line no-param-reassign
		props[handlerKey] = eventConfiguration.getHandler(this, this.props[key], originalHandler);
	}
}
```

packages/cmf/src/onEvent.md
```diff
@@ -0,0 +1,84 @@
+# Introduction
+
+The concept here is to create handler (`onClick` for example).
+
+A handler is function that recieve most of the time an event as first param and maybe some other params.
```
jsomsanith: typo `received`

packages/cmf/src/onEvent.js
jsomsanith: +1 please rename this one

packages/cmf/src/onEvent.md
```diff
@@ -0,0 +1,84 @@
...
+        >
+            Click me once
+        </MyButton>;
+}
+```
+
+The initialState props will tell cmfConnect to initialize this MyButton component state in redux.
+
+The `spreadCMFState` props will request cmfConnect to
+transform the content of the state into props
```
jsomsanith: Nice doc !
Can you add the `toggle` special case please ?

packages/cmf/src/onEvent.js
```diff
@@ -0,0 +1,79 @@
...
+					acc[key] = args[value[0]];
+				} else if (value.length === 2) {
+					// eslint-disable-next-line no-param-reassign
+					acc[key] = get(args[value[0]], value[1]);
+				} else {
+					throw new Error('onEventSetState array must have 1 or 2 element for ', key, args);
+				}
+			} else if (value === 'toggle') {
+				// because toggle need to read the state we dispatch it with a function
+				instance.props.setState(props => props.setState({ [key]: !props.state.get(key) }));
```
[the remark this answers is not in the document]
jmfrancois: it's the cmf setState

PR #1187, opened 2018-03-20
fix(components/e2e): tabbar constructors should be public

ncomont:

TabBar constructors are not public, so QA could not instanciate it.

Make TabBar constructors public.


jsomsanith:
Hmm, none is public because the idea was to not instantiate things except the `Application` object.
Then you can write things like `application.AppHeaderBar.clickOnMenuItem(id)`.

Here perhaps we should have an `application.getTabBar(id)` instead.
We already have this with `application.List.getTable(id)`

PR #1189, opened 2018-03-20
feat(components/Actions): pass through all aria and data attributes

frassinier:
It's hard to track buttons clicks for analytics purpose (by using inspector).

Let's add `data-feature` attribute on `<*Action/>` components in order to simplify features tracking.


packages/components/src/utils/getPropsFrom.js
```diff
@@ -15,10 +15,12 @@ function extractComponentProps(Component, props) {
 
 function extractNativeProps(props) {
 	const extractedProps = {};
-	NATIVE_PROPS.forEach(propName => {
-		if (props[propName] !== undefined) {
-			extractedProps[propName] = props[propName];
-		}
+	NATIVE_PROPS.forEach(nativeProp => {
+		Object.keys(props).forEach(propName => {
+			if (propName.includes(nativeProp)) {
```
jmfrancois: startswith looks safer to me.
or fill the NATIVE_PROPS with regex so you just do nativeProp.test(propName) ?

frassinier: I switched for `startsWith` :+1:

packages/components/src/utils/getPropsFrom.js
```diff
@@ -15,10 +15,12 @@ function extractComponentProps(Component, props) {
 
 function extractNativeProps(props) {
 	const extractedProps = {};
-	NATIVE_PROPS.forEach(propName => {
-		if (props[propName] !== undefined) {
-			extractedProps[propName] = props[propName];
-		}
+	NATIVE_PROPS.forEach(nativeProp => {
+		Object.keys(props).forEach(propName => {
```
jsomsanith: `m` the length of `NATIVE_PROPS`
`n` the length of `props` properties

Here with this implementation you always have a complexity of 0(m*n) because event if there is a property that is taken, you test it over the other `NATIVE_PROPS ` elements.

If you take the reverse : 
* go over props elements
* for each prop, you `find()` the first element that match (`startsWith`). This will stop on the first match.
* if there is a result, you take it

On the best scenario you'll have a complexity of O(m), and the worst scenario O(m*n)

PR #1190, opened 2018-03-21
fix(SidePanel): colors managed in the component

jsomsanith:
Regression with #1177
The SidePanel menu buttons style is managed by ActionList which is basically a nav-pill, so styled by the theme.
But the rest of the the SidePanel has transparent background on default style, and have a color set on reverse mode.

To be consistent, in default mode, we now set the brand primary color as background, which is consistent with the nav-pills style.

The two column layout now forces its sidepanel content to take its full height, so no color to synchronise between layout and sidepanel.

I removed all the custom background colors from Side Panel stories.
The container SidePanel with injected content story has been styled within storybook to have a better demo.


packages/containers/.storybook/preview-head.html
```diff
@@ -20,4 +20,22 @@
...
+    }
+    .tc-side-panel  > form > .form-group > input {
+        margin: 0;
+        color: white;
+    }
+    .tc-side-panel > button svg.tc-svg-icon,
+    .tc-side-panel > form svg.tc-svg-icon {
+        margin: 0 10px;
+        color: white;
+    }
```
jmfrancois: Custom CSS !
This should go into filterbar.scss with a :global(.tc-side-panel) selector

jsomsanith: This is only for demo purpose. We don't need it, so adding this style in the component won't make sense. It will be integrated in the bundles and shipped with apps.

PR #1191, opened 2018-03-21
fix(Inject): pass props from inject.all

jare-talend:

I inject a component with a nested route. The props haven't spread when i inject a component to show the children

```javascript
const components = {
	content: { component: 'SubHeader', },
};
const inject = Inject.all(getComponent, components);
inject('content', { children }));
```

spread props


packages/components/src/Inject/Inject.component.js
```diff
@@ -67,11 +67,11 @@ Inject.all = function injectAll(getComponent, components, CustomInject = Inject)
 	if (!getComponent || !components) {
 		return nothing;
 	}
-	return key => {
+	return (key, props) => {
```
jmfrancois: add props = {} as add default value

jsomsanith: No need too, spread of undefined works as expected

PR #1192, opened 2018-03-21
fix(i18n-form) : add mapping for the prop language in form container

smouillour:
When you use uiform via form with using the props uiform and language (for translation), the prop language is not mapped in the form-container.
So the translation don't work correctly in this case.

I added the mapping for language in the form-container.


packages/containers/src/Form/Form.test.js
```diff
@@ -34,6 +35,79 @@ describe('Container(Form)', () => {
...
+		);
+		expect(
+			wrapper
+				.dive()
+				.dive()
+				.getElement().type.name,
+		).toEqual(UIForm.name);
+	});
+
+	it('should render UIForm with not language prop set', () => {
```
jsomsanith: Typo ?
--> `should render UIForm with language prop`

packages/containers/src/Form/Form.test.js
```diff
@@ -34,6 +35,79 @@ describe('Container(Form)', () => {
...
+		);
+		expect(
+			wrapper
+				.dive()
+				.dive()
+				.getElement().props.language.OBJECT_REQUIRED,
+		).toEqual('Field translated');
+	});
+
+	it('should render UIForm with language prop not set', () => {
```
jsomsanith: This one is not necessary, it tests the default UIForms messages which is not useful in your component. You can remove that

packages/containers/src/Form/Form.test.js
```diff
@@ -34,6 +35,79 @@ describe('Container(Form)', () => {
...
+				actions={[]}
+				formProps={{ other: true }} // extra props
+				uiform
+				language={{ OBJECT_REQUIRED: 'Field translated' }}
+			/>,
+		);
+		expect(
+			wrapper
+				.dive()
+				.dive()
```
jsomsanith: Instead of `.dive()` that is not very clear, can you replace it by `.find(UIForm)` ?

PR #1194, opened 2018-03-22
feat(containers): add AppLoader container

romainseb:
To bootstrap our application, we have some specific code in each app.
Share an AppLoader container that make http call & wait some collections to be filled in the store to render the app


packages/containers/src/AppLoader/AppLoader.md
```diff
@@ -0,0 +1,45 @@
+# App Loader
```
acateland: :+1:

packages/containers/src/AppLoader/AppLoader.md
```diff
@@ -0,0 +1,45 @@
...
+## How to configure it ?
+
+This is a sample of configuration :
+
+```json
+"AppLoader#default":{
+    "saga": "appLoaderSaga",
+    "steps":[
+        { "actionCreators":["user:fetchIdentity"] },
+        { "waitFor":["identity"] },
```
acateland: i have the feeling that the waitFor declaration is not really usefull ? can each block of actionCreators wait for the positive resolution of their predecessors ?

romainseb: Don't know if i can make it this way. Here if for `user:fetchIdentity` we have this piece of code :
```
function fetchIdentity() {
	return actions.http.get(API.identity, {
		cmf: {
			collectionId: 'identity',
		},
	});
}
```
When the actionCreator would be resolved, the http call would be launched but we are far away from having the identity's collection provisioned in the store.

romainseb:
i agree well on the handle failure case. Currently, i'm not sure about what we want to do if there is some error on the init.

output/containers.eslint.txt
```diff
@@ -1,12 +1,15 @@
...
+> @talend/react-containers@0.175.0 lint:es /home/travis/build/Talend/ui/packages/containers
 > eslint --config ../../.eslintrc src
 
 The react/require-extension rule is deprecated. Please use the import/extensions rule from eslint-plugin-import instead.
 
 /home/travis/build/Talend/ui/packages/containers/src/ActionSplitDropdown/ActionSplitDropdown.test.js
   4:10  error  'ActionSplitDropdown' is defined but never used  no-unused-vars
 
+/home/travis/build/Talend/ui/packages/containers/src/AppLoader/AppLoader.saga.js
+  3:1  error  'redux-saga' should be listed in the project's dependencies, not devDependencies  import/no-extraneous-dependencies
```
smouillour: could you fix this error?

packages/containers/src/AppLoader/AppLoader.connect.js
```diff
@@ -0,0 +1,51 @@
...
+};
+
+/**
+ * calculate the loading attribute with the given props
+ * @param {object} state the redux state
+ * @param {object} ownProps the component props
+ */
+export function mapStateToProps(state, ownProps) {
+	const newProps = { loading: false };
+	get(ownProps, 'hasCollections', []).forEach(collectionName => {
```
jare-talend: could you use newsProps.loading = get(ownProps, 'hasCollections', []).every(collectionName => state.cmf.collections.has(collectionName))

packages/containers/examples/ExampleAppLoader.js
```diff
@@ -0,0 +1,42 @@
...
+function ac2() {
+	return actions.http.get('https://jsonplaceholder.typicode.com/photos/', {
+		cmf: {
+			collectionId: 'col2',
+		},
+	});
+}
+
+api.actionCreator.register('ac1', ac1);
+api.actionCreator.register('ac2', ac2);
```
jmfrancois: they are registred globaly when you do that so other story can rely on it.
So move them to story/config with better name please

packages/containers/src/AppLoader/AppLoader.connect.test.js
```diff
@@ -0,0 +1,62 @@
...
+
+describe('AppLoader container', () => {
+	describe('rendering', () => {
+		it('should render child if not loading', () => {
+			const wrapper = shallow(
+				<AppLoaderContainer>
+					<div className="child" />
+				</AppLoaderContainer>,
+			);
+			expect(wrapper.getElement()).toMatchSnapshot();
```
jmfrancois: please try to find '.child' instead of snapthost

packages/containers/src/AppLoader/AppLoader.connect.test.js
```diff
@@ -0,0 +1,62 @@
...
+		});
+
+		it('should render the app loader if loading', () => {
+			const wrapper = shallow(
+				<AppLoaderContainer loading>
+					<div className="child" />
+				</AppLoaderContainer>,
+			);
+			expect(wrapper.getElement()).toMatchSnapshot();
```
jmfrancois: please check '.child' is not found
and you find 'AppLoader'

packages/containers/src/AppLoader/AppLoader.md
```diff
@@ -0,0 +1,45 @@
...
+
+This component is here to make http calls before the first render of the application. During the loading, it show the application's loader.
+
+## Where i put it ?
+
+On your top level component, you can place it this way :
+
+```html
+<AppLoader>
+    <div>
```
jmfrancois: is that div needed ?

packages/containers/src/AppLoader/AppLoader.md
jmfrancois: No they are redux actions so non blocking.
In fact we have the two cases, the idea here to be able to make some wait in steps.
But not required

packages/containers/src/AppLoader/AppLoader.md
```diff
@@ -0,0 +1,45 @@
...
+        { "waitFor":["identity"] },
+        { "actionCreators":["datastore:fetch.types", "datastore:fetch:cloudAgents"]}
+        { "waitFor":["datastore-types", "cloud-agents"] },
+        { "actionCreators":["datastore:fetchAll", "dataset:fetchAll"] }
+    ],
+    "hasCollections": ["identity", "datastore-types", "datastores", "datasets"]
+},
+```
+
+### Attributes
```
jmfrancois: props ?

packages/containers/src/AppLoader/AppLoader.connect.js
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * This container show the application's loader & bootstrap the app
+ * @param {object} props the component props
+ * @param {boolean} props.loading tell if the app loader should show the loader or the content
+ * @param {object} props.children react element to show
+ */
+export function AppLoaderContainer({ loading, children }) {
+	if (loading) {
+		return <AppLoader />;
```
jmfrancois: please spread the rest of props to AppLoader

packages/containers/src/AppLoader/AppLoader.md
romainseb: to make it work with react 15, i think this is required, no ?

jmfrancois: So I think the div must be in the AppLoader container.
So we can inject before and after children and that will let us remove our custom App components in every projects

packages/containers/src/AppLoader/AppLoader.connect.js
```diff
@@ -0,0 +1,49 @@
...
+ * @param {object} props the component props
+ * @param {boolean} props.loading tell if the app loader should show the loader or the content
+ * @param {object} props.children react element to show
+ */
+export function AppLoaderContainer({ loading, children, ...rest }) {
+	if (loading) {
+		return <AppLoader {...rest} />;
+	}
+
+	return children || null;
```
jmfrancois: what do you think about:
```
		const injected = Inject.all(this.props.getComponent, this.props.components, CustomInject);
		return (
			<div>
				{injected('before-children')}
				{this.props.children || null}
				{injected('after-children')}
			</div>
		);
```

packages/containers/src/AppLoader/AppLoader.saga.js
```diff
@@ -0,0 +1,52 @@
...
+		yield call(delay, interval);
+	}
+}
+
+/**
+ * This function handle a specific step
+ * @param {object} step a bootstrap step that could contain a actionCreator list or a waitList
+ */
+export function* handleStep(step) {
+	if (step[ACTION_CREATORS]) {
```
jare-talend: On TDC, we have some additional fetaures like TAKE_ACTION. so we can update it to reuse this.

packages/containers/.storybook/config.js
```diff
@@ -81,22 +83,51 @@ function selectTab(event, data) {
...
-registerActionCreator('subheaderbar:goback', actionsCreatorsSubHeader.goBackSubHeaderBar);
 
-registerActionCreator('tabbar:select', selectTab);
+api.actionCreator.register('http:get:photos1', httpPhotosGet1);
+api.actionCreator.register('http:get:photos2', httpPhotosGet2);
+api.actionCreator.register('object:view', objectView);
+api.actionCreator.register('cancel:hide:dialog', hideDialog);
+api.actionCreator.register('confirm:dialog', confirmDialog);
+api.actionCreator.register('item1:action', chooseItem1);
+api.actionCreator.register('item2:action', chooseItem2);
```
jsomsanith: You register the 5 previous action creators, but they are already registered above.
Plus, you `aoi.actionCreator.register` but it is extarcted into `registerActionCreator` var. Please align all the register calls.

packages/containers/src/AppLoader/AppLoader.md
```diff
@@ -0,0 +1,47 @@
+# App Loader
+
+This component is here to make http calls before the first render of the application. During the loading, it show the application's loader.
+
+## Where i put it ?
```
jsomsanith: Where `do` I put it ?

packages/containers/src/AppLoader/AppLoader.md
jsomsanith: Hmm I'm not sure to understand the `to make it work with react 15`.  Without the div, that makes all sub components the children. So in your AppLoader container, you use `{children}` in your JSX to render them, that's fine (and already the case).

(Don't forget to remove this div in the doc)

packages/containers/src/AppLoader/AppLoader.md
```diff
@@ -0,0 +1,47 @@
...
+        <ShortcutManager view="shortcuts" />
+        <IconsProvider icons={mockIconDefinition} />
+        <ACKDispatcher />
+        <Notification />
+        {props.children}
+    </div>
+</AppLoader>
+```
+
+Another way to do it could be to use the inject placeholders to inject all the stuff
```
jsomsanith: Can you add some code here for example, so we don't have to go in the code to get the placeholder names or go over inject's code to know how to use it ?

packages/containers/src/AppLoader/AppLoader.md
```diff
@@ -0,0 +1,47 @@
...
+    ],
+    "hasCollections": ["identity", "datastore-types", "datastores", "datasets"]
+},
+```
+
+### Props
+
+* saga : required cause this is how the action creator are dispatched
+* steps, you can pass objects, this objects can have an attribute ( exclusive ) :
+  * actionCreators : an array of action creator to dispatch in parallel.
```
jsomsanith: Typo: action creator`s`

packages/containers/src/AppLoader/AppLoader.md
```diff
@@ -0,0 +1,47 @@
...
+  * waitFor : an array of collection we have in the cmf store to go on next step
+* hasCollections : an array of collections we have in the cmf store to render the children
```
jsomsanith: `an array of collection ids to have in cmf store before it can render the children`

packages/containers/src/AppLoader/AppLoader.md
```diff
@@ -0,0 +1,47 @@
...
+    "hasCollections": ["identity", "datastore-types", "datastores", "datasets"]
+},
+```
+
+### Props
+
+* saga : required cause this is how the action creator are dispatched
+* steps, you can pass objects, this objects can have an attribute ( exclusive ) :
+  * actionCreators : an array of action creator to dispatch in parallel.
+  * waitFor : an array of collection we have in the cmf store to go on next step
```
jsomsanith: `an array of collection ids to have in cmf store before we can trigger the next step`

packages/containers/src/AppLoader/AppLoader.connect.js
```diff
@@ -0,0 +1,57 @@
...
+};
+
+/**
+ * calculate the loading attribute with the given props
+ * @param {object} state the redux state
+ * @param {object} ownProps the component props
+ */
+export function mapStateToProps(state, ownProps) {
+	return {
+		loading: !get(ownProps, 'hasCollections', []).every(collectionName =>
```
jsomsanith: loading = not(every collection id is in store)
`loading: !get(ownProps, 'hasCollections', []).every(collectionId => state.cmf.collections.has(collectionName))`

vs

loading = at least one collection id is not is store
`loading: get(ownProps, 'hasCollections', []).some(collectionId => !state.cmf.collections.has(collectionName))`

That doesn't make a lot of difference, I let you choose, :D

packages/containers/src/AppLoader/AppLoader.connect.test.js
```diff
@@ -0,0 +1,64 @@
...
+			// given
+			const state = { cmf: { collections: Immutable.Map({ test2: Immutable.Map() }) } };
+			const ownProps = { hasCollections: ['test', 'test2'] };
+			// when
+			const result = mapStateToProps(state, ownProps);
+			// then
+			expect(result).toEqual({ loading: true });
+		});
+
+		it('should return loading to false if everithing to wait is present in collections', () => {
```
jsomsanith: Typo: ever`y`thing

PR #1198, opened 2018-03-23
fix(cmf): invariant condition for 'register undefined as component' error

yyanwang:
'register undefined as component' error is only threw on production environment. should throw it only on dev environment.
negative this invariant condition

### DIFF
--- a/packages/cmf/src/sagas/component.js
+++ b/packages/cmf/src/sagas/component.js
@@ -27,7 +27,7 @@ export const registerMany = registry.getRegisterMany(register);
 export function* onSagaStart(action) {
 	const saga = get(action.saga);
 	if (!saga) {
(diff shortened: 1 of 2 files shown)


packages/cmf/src/sagas/component.js
```diff
@@ -27,7 +27,7 @@ export const registerMany = registry.getRegisterMany(register);
 export function* onSagaStart(action) {
 	const saga = get(action.saga);
 	if (!saga) {
-		invariant(process.env.NODE_ENV !== 'production', `The saga ${action.saga} is not registred`);
+		invariant(process.env.NODE_ENV === 'production', `The saga ${action.saga} is not registred`);
```
jmfrancois: 'You cannot register undefined as a saga for id "%s"

PR #1208, opened 2018-03-27
fix(theme): fix filter bar styling in forms

ybaskaran:
Filter bar styling not applied after the recent changes made from https://jira.talendforge.org/browse/TUI-87

Applying custom styles to search fields under form-group class

### DIFF
--- a/packages/theme/src/theme/_forms.scss
+++ b/packages/theme/src/theme/_forms.scss
@@ -30,10 +30,12 @@ form {
 		input[type='password'],
 		input[type='email'],
 		input[type='number'],
+		input[type='search'],
 		[type='text'].form-control,
 		[type='password'].form-control,
 		[type='email'].form-control,
 		[type='tel'].form-control,
+		[type='search'].form-control,
 		[contenteditable].form-control {
 			&:focus:not([readonly]) + label {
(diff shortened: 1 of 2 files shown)


packages/theme/src/theme/_forms.scss
```diff
@@ -30,10 +30,12 @@ form {
 		input[type='password'],
 		input[type='email'],
 		input[type='number'],
+		input[type='search'],
 		[type='text'].form-control,
 		[type='password'].form-control,
 		[type='email'].form-control,
 		[type='tel'].form-control,
+		[type='search'].form-control,
```
sgendre: even as it is fixing the styling issue on FilterBar, there  are plenty of type available on <input> https://developer.mozilla.org/fr/docs/Web/HTML/Element/input

why do you limit to those ?

ybaskaran: its fixed now, initially it was removed to fix sass-lint errors but it makes sense to have it for all inputs under form-group class so adding it back

sgendre: I mean this style can also be useful on input type='date' or 'number' and maybe all other text related input.

PR #1210, opened 2018-03-27
fix(cmf/httpMiddleware): override default headers

jlame:
Content-type is always sent which causes some issue with file upload

Removing the content type from header when a formdata is sent (like in sagas/http)


packages/cmf/src/middlewares/http/middleware.js
```diff
@@ -101,10 +101,16 @@ export function getMethod(action) {
...
 export function mergeConfiguredHeader(config) {
 	// still need to keep the previous header added by action
-	return options => ({
-		...options,
-		headers: { ...options.headers, ...DEFAULT_HTTP_HEADERS, ...config.headers },
-	});
+	return options => {
+		const headerMergedConfig = {
+			...options,
+			headers: { ...DEFAULT_HTTP_HEADERS, ...options.headers, ...config.headers },
```
romainseb: Shouldn't be more like 
`{ ...DEFAULT_HTTP_HEADERS, ....config.headers, ...options.headers }`
to allow specific options to override global config ?

jlame: you're right about that options are the one coming from the action so it should be the last one overriding

packages/cmf/src/middlewares/http/middleware.js
```diff
@@ -101,10 +101,16 @@ export function getMethod(action) {
...
-	return options => ({
-		...options,
-		headers: { ...options.headers, ...DEFAULT_HTTP_HEADERS, ...config.headers },
-	});
+	return options => {
+		const headerMergedConfig = {
+			...options,
+			headers: { ...DEFAULT_HTTP_HEADERS, ...config.headers, ...options.headers },
+		};
+		if (headerMergedConfig.body && headerMergedConfig.body instanceof FormData) {
```
jmfrancois: why the first condition is required ? If no .body it will not be an instance of FormData right.

jlame: right changing

PR #1213, opened 2018-03-28
fix(HeaderBar): remove username text transformation

jsomsanith:
HeaderBar component transforms the user lastname to uppercase.
But the strategy of user inputs is not to transform them, display them as user typed it.

In [guideline](http://guidelines.talend.com/document/92132#/design-principles/font-styles)
```
All other UI elements, if not in this list, are displayed in lower case, or with the original case coming out from their source (database, etc.)
```
(The username is not in the exception list)

Remove the extra elements and style to perform the text transformation

### DIFF
--- a/packages/components/src/HeaderBar/HeaderBar.component.js
+++ b/packages/components/src/HeaderBar/HeaderBar.component.js
@@ -151,11 +151,7 @@ function User({ name, firstName, lastName, getComponent, ...rest }) {
 
 	function getDisplayName(params) {
 		if (params.firstName && params.lastName) {
-			return (
-				<span className={classNames(theme['user-name'], 'user-name')}>
-					{params.firstName} {params.lastName}
-				</span>
-			);
+			return `${params.firstName} ${params.lastName}`;
 		}
 		return params.name;
 	}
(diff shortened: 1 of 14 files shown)


packages/components/src/HeaderBar/HeaderBar.component.js
```diff
@@ -153,12 +153,7 @@ function User({ name, firstName, lastName, getComponent, ...rest }) {
 		if (params.firstName && params.lastName) {
 			return (
 				<span className={classNames(theme['user-name'], 'user-name')}>
-					<span className={classNames(theme['user-firstname'], 'user-firstname')}>
-						{params.firstName}
-					</span>
-					<span className={classNames(theme['user-lastname'], 'user-lastname')}>
-						{params.lastName}
-					</span>
+					{params.firstName} {params.lastName}
```
frassinier: if `params.firsName` is empty then you will get an extra space

WDYT ? `{params.firstName ? params.firstName + ' ' : '' }{params.lastName}`

jsomsanith: Hmm I don't think so, there will be an extra space in the html, but not interpreted. Here it is interpreted because it is between 2 words.

jsomsanith: I just saw that my fix is not totally write. I'll remove the extra span. But in fact there is a test, if we have firstname *and* lastname it will display them. Else it will display name props

PR #1214, opened 2018-03-28
feat(cmf): add renderIf feature

VolodymyrKovalM:
Currently we can hide the Action button if it is not available,
but we can not hide particular item from dropdown if it is not available
(We have this case on engines tab in TMC, where adding elastic engine is not always available)

Add checking availability to cmfConnected


packages/cmf/src/cmfConnect.js
```diff
@@ -295,6 +296,9 @@ export default function cmfConnect({
 			}
 
 			render() {
+				if (this.props.renderIf === false) {
+					return null;
```
jsomsanith: Can you add a test please ?

VolodymyrKovalM: Yes, sure, I will add

PR #1218, opened 2018-03-29
fix(component/datalist): checking if value exist on update

jlame:
When typing search if the input was blurred or enter key pressed, the search value was updated as value whereas it does not exist in the suggestions.

If the value doesn't exist, it is updated with the previous selected value or by empty string.


output/components.eslint.txt
```diff
@@ -8,6 +8,9 @@ The react/require-extension rule is deprecated. Please use the import/extensions
   1:1  error  '@storybook/react' should be listed in the project's dependencies, not devDependencies  import/no-extraneous-dependencies
   4:2  error  Unexpected require()                                                                    global-require
 
+/home/travis/build/Talend/ui/packages/components/src/Datalist/Datalist.component.test.js
+  148:9  error  'payload' is assigned a value but never used  no-unused-vars
```
jsomsanith: Please fix linter error

packages/components/src/Datalist/Datalist.component.js
```diff
@@ -212,8 +211,22 @@ class Datalist extends Component {
 					}
 				}
 			}
-			const payload = { value: get(enumValue, 'value', value) };
-			this.props.onChange(event, payload);
+			const payload = { value: get(enumValue, 'value') };
+			if (payload.value || !this.props.restricted) {
+				this.props.onChange(
+					event,
+					payload.value ? payload : { value: get(enumValue, 'value', value) },
```
jsomsanith: This block is quite hard to read and you fetch the enumValue twice. What about
```
const selectedEnumValue = get(enumValue, 'value');
if (selectedEnumValue || !this.props.restricted) {
	this.props.onChange(
		event,
		payload: { value:  selectedSuggestionValue || value }
	);
	this.setState({
		previousValue: typeof previousValue === 'object' ? previousValue.title : previousValue,
	});
}
```

packages/components/src/Datalist/Datalist.component.js
```diff
@@ -212,8 +211,22 @@ class Datalist extends Component {
...
+				);
+				// setting the previous value to current only if the current value exists
+				this.setState({
+					previousValue: typeof previousValue === 'object' ? previousValue.title : previousValue,
+				});
+			} else {
+				this.setState({
```
jsomsanith: You should call `resetValue()`

packages/components/src/Datalist/Datalist.component.test.js
```diff
@@ -111,15 +111,69 @@ describe('Datalist component', () => {
...
+		// when
+		input.simulate('blur');
+
+		// then
+		const payload = { value: 'foobar' };
+		expect(onChange).toBeCalledWith(expect.anything(), payload);
+		expect(wrapper.find(Typeahead).props().items).toBe(null);
+	});
+
+	it('should reset suggestions and change value on blur when not in restricted mode', () => {
```
jsomsanith: This one is the exact same as the previous one, you can remove it.

jlame: not exactly, this one is to check that the value is selected even if it does not exist in the suggestions

PR #1220, opened 2018-03-29
fix(components): Action instead of Button from react-bootstrap

frassinier:
We're using Button from react-bootstrap when using our `Action` is enough

Switch to `Action`


packages/components/src/ListView/Items/__snapshots__/items.test.js.snap
```diff
@@ -7788,36 +7791,3633 @@ exports[`Items should render with nested items 1`] = `
...
-                                  type="button"
                                 >
-                                  <Icon
-                                    name="talend-caret-down"
+                                  <ActionButton
+                                    available={true}
+                                    bsStyle="link"
+                                    disabled={false}
+                                    i18n={
+                                      I18n {
```
smouillour: Could you make something to escape I18N? please

packages/components/src/ListView/Items/__snapshots__/items.test.js.snap
```diff
@@ -7788,36 +7791,3633 @@ exports[`Items should render with nested items 1`] = `
...
                                   onClick={[Function]}
-                                  type="button"
                                 >
-                                  <Icon
-                                    name="talend-caret-down"
+                                  <ActionButton
+                                    available={true}
+                                    bsStyle="link"
+                                    disabled={false}
+                                    i18n={
```
jmfrancois: Why do we have this big object here :(

PR #1241, opened 2018-04-03
tool(non-reg): add possibility to retry when it get an error 404 with surge

smouillour:
the script don't manage the error 404 when the page in surge is not ready
Add the possibility for the script to retry to access at the surge page after a timeout as 30s (by default)

it is possible to change the timeout by using the option -t or --timeout

### DIFF
--- a/screenshots.js
+++ b/screenshots.js
@@ -9,9 +9,11 @@ program
 	.option('-p, --pullrequest [pr]', 'Pull request')
 	.option('-c, --config [config]', 'JSON config file')
 	.option('-v, --verbose', 'Verbose')
+	.option('-t, --timeout [time]', 'timeout for waiting surge upload')
 	.parse(process.argv);
 
 const PR = program.pullrequest;
+const surgeTimeout = program.timeout || 30000;
 
 if (!PR) {
 	console.error('you must precise a PR number using -p or --pullrequest');
@@ -59,6 +61,33 @@ const TMP_CONFIG = { postfix: '.png' };
 		return screenshot;
 	}
 
+	async function timeout(ms) {
+		return new Promise(resolve => setTimeout(resolve, ms));
+	}
+
+	async function goToPage(page, url, isRestart) {
+		if (isRestart) {
+			console.error('retry to access to :');
+			console.error(url);
+		}
+		await page.goto(url);
+		await isNotFound(page, url);
+	}
+
+	async function sleepBeforeRetry(page, url) {
+		await timeout(surgeTimeout);
+		return goToPage(page, url, true);
+	}
+
+	async function isNotFound(page, url) {
+		const element = await page.evaluate(() => document.querySelector('h1').textContent);
+		if (element === 'project not found') {
+			console.error(`this page is not ready yet : ${url}`);
+			console.error(`Waiting surge upload (${surgeTimeout})`);
+			await sleepBeforeRetry(page, url);
+		}
+	}
+
 	async function process(masterPage, branchPage, pageConfig) {
 		log(`Process ${pageConfig.name} ${pageConfig.selector}`);
 		const masterScreenShot = await getScreenShot(masterPage, pageConfig);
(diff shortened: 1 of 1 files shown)


screenshots.js
```diff
@@ -59,6 +61,33 @@ const TMP_CONFIG = { postfix: '.png' };
 		return screenshot;
 	}
 
+	async function timeout(ms) {
+		return new Promise(resolve => setTimeout(resolve, ms));
```
jmfrancois: it's used only once, please move it inside

screenshots.js
```diff
@@ -59,6 +61,33 @@ const TMP_CONFIG = { postfix: '.png' };
...
+		await isNotFound(page, url);
+	}
+
+	async function sleepBeforeRetry(page, url) {
+		await timeout(surgeTimeout);
+		return goToPage(page, url, true);
+	}
+
+	async function isNotFound(page, url) {
+		const element = await page.evaluate(() => document.querySelector('h1').textContent);
```
jmfrancois: you can either use page.$ or rename element to textContent.

screenshots.js
```diff
@@ -95,14 +109,23 @@ const TMP_CONFIG = { postfix: '.png' };
 			log('-- no diff');
 		}
 	}
+	function onResponse(page, isMaster) {
+		const pageUrls = Object.keys(config).map(path => `http://${PR}.talend.surge.sh/${path}`);
+		page.on('response', async res => {
+			if (pageUrls.includes(res.url()) && res.status() === 404) {
```
jsomsanith: Why do you check if the requested url is one of the PR's surge path ?

screenshots.js
```diff
@@ -95,14 +109,23 @@ const TMP_CONFIG = { postfix: '.png' };
...
+				console.error(`Surge is not ready yet ${isMaster ? 'for Master' : `for the PR ${PR}`}: ${res.url()}`);
```
jsomsanith: One of the condition to execute this is that the requested url is one of the PR's one, so it conflicts with the ternary checking `isMaster` here.
With my previous comment, I still don't understand what it is supposed to do exactly. It should be reworked or explained.

screenshots.js
annilsson: I remember this is because the response is not only the main page response, but also from assets etc. And it's only interesting to know if the main request is 404.

screenshots.js
```diff
@@ -96,33 +114,72 @@ const TMP_CONFIG = { postfix: '.png' };
 		}
 	}
 
+	function onResponse(page, isMaster) {
+		const pageUrls = Object.keys(config).map(path => `${getUrl(isMaster)}${path}`);
+		page.on('response', async res => {
+			if (pageUrls.includes(res.url()) && res.status() === 404) {
+				console.error(`Surge is not ready yet ${isMaster ? 'for Master' : `for the PR ${PR}`}: ${res.url()}`);
+				errorWithSurgeMaster = isMaster;
+				errorWithSurgePR = !isMaster;
```
jsomsanith: You modify the 2 flags, even if it's not the current surge's.
```
if (isMaster) {
	errorWithSurgeMaster = true;
} else {
	errorWithSurgePR = true;
}
```

screenshots.js
```diff
@@ -96,33 +114,72 @@ const TMP_CONFIG = { postfix: '.png' };
...
 	log('Open browser for PR pages');
 	const branchPage = await browser.newPage();
+	onResponse(branchPage, false);
 
 	await asyncForEach(Object.keys(config), async path => {
 		try {
 			await compare(masterPage, branchPage, path);
 		} catch (error) {
-			console.error(path, error);
+			console.log(error);
```
jsomsanith: Remove this `console.log`

screenshots.js
```diff
@@ -96,33 +114,72 @@ const TMP_CONFIG = { postfix: '.png' };
...
 	await browser.close();
-})();
+});
+
+async function runScreenshot(nbCurrentRetry) {
+	try {
+		await screenshotTest();
+	} catch (e) {
+		if (nbCurrentRetry < SURGE_MAX_RETRY) {
+			const newNbCurrentRetry = nbCurrentRetry + 1;
```
jsomsanith: This const is not needed, inline it in the retry call

screenshots.js
```diff
@@ -96,33 +114,72 @@ const TMP_CONFIG = { postfix: '.png' };
 		}
 	}
 
+	function onResponse(page, isMaster) {
+		const pageUrls = Object.keys(config).map(path => `${getUrl(isMaster)}${path}`);
```
jsomsanith: Please build that 1 time for master and PR instead of at each fetch

PR #1243, opened 2018-04-03
feat(DeleteResource): Add wording & i18n

romainseb:
Yep, it's back !

There is just the label of the entity showned
Add a sentense ( and translate it )


jare-talend:
Could you add this command on the package.json as the other i18n packages.
https://github.com/Talend/ui/blob/master/packages/components/package.json#L21

We have a issue with the female context where the key is a variable so the parser can't find it.

packages/containers/src/DeleteResource/DeleteResource.container.js
```diff
@@ -2,15 +2,18 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { componentState } from '@talend/react-cmf';
 import { ConfirmDialog } from '@talend/react-components';
+import { translate, Trans } from 'react-i18next'; // eslint-disable-line import/no-extraneous-dependencies
```
jsomsanith: You should not disable this rule.
It highlight an issue in your dependencies. You have (react-)i18next as dev deps, but that's wrong. You should either add it as peer dependencies or move to to dependencies.

packages/containers/src/translate.js
```diff
@@ -0,0 +1,5 @@
+import { createInstance } from 'i18next'; // eslint-disable-line import/no-extraneous-dependencies
```
jsomsanith: When you fix the dependencies, you can remove that

packages/containers/examples/config/i18n.js
```diff
@@ -1,8 +1,19 @@
-import i18n from 'i18next';
+import React from 'react';
+import i18n from 'i18next'; // eslint-disable-line import/no-extraneous-dependencies
```
jsomsanith: When you fix the dependencies, you can remove that

packages/containers/src/DeleteResource/DeleteResource.container.js
```diff
@@ -70,15 +81,25 @@ export default class DeleteResource extends React.Component {
...
 		return (
 			<ConfirmDialog
 				show
 				header={this.props.header}
 				cancelAction={cancelAction}
 				validateAction={validateAction}
 			>
-				<div>{resourceInfo.label}</div>
+				<div>
+					<Trans i18nKey={i18nKey}>
```
jsomsanith: 2 things on i18n : 
* add npm scripts to extract translation keys in root package.js
* this trans is not compatible with i18next-parser, I won't block this PR for that since it is really needed and the solution is the good one I think, but we need a solution

romainseb: 1) The npm script is on the container package.json
2) Currently yes, there is some work here on this side -> https://github.com/i18next/i18next-parser/pull/85

For now, i think this is still the best way to get the translations with html markup.
There is another issue currently, as the Trans component manage to make it work the count attribute but the "context" attribute to declare male or female  translation is not working as it is. I try to make a PR on this point in react-i18next ;)

jsomsanith: 1. it should be added to the root `package.json` like the others, so that `yarn extract-i18n` extract all packages translations in a single folder. Take a look at root npm scripts

PR #1247, opened 2018-04-05
fix(forms): ListViewWidget selection

jmfrancois:

The PR #1245 introduce a bug. you can't unselect item
Every mutation trigger a re-render, because the === doesn't work, options are always rebuild.

rework detection of changes using deepequal based on values so we don't change state for nothing.
Use current options for the value.


packages/forms/src/widgets/ListViewWidget/ListViewWidget.js
```diff
@@ -78,11 +90,16 @@ class ListViewWidget extends React.Component {
...
-			this.setState({ items, displayedItems: items });
+		const options = this.props.options;
+		const nextOptions = nextProps.options;
+		if (!areOptionsEqual(options, nextOptions)) {
+			const items = getItems(nextOptions, this.props.value, this);
+			this.setState({
+				items,
+				displayedItems: items,
+				searchCriteria: null,
+				toggleAllChecked: items.length === items.filter(i => i.checked).length,
```
jsomsanith: * `items.every(i => i.checked)` so it stops on first element that doesn't match
* can you mutualise the `{ item, displayedItems, toggleAllChecked }` from here + constructor ? Perhaps a `this.getDerivedStateFromOptions(newOptions )`

```
	/**
	* This computes new derived state when options change
	* Compute new items
	* Reset search and displayedItems
	* Update toggleAll status
	*/
	getDerivedStateFromNewOptions(options) {
		const items = getItems(options, this.props.value, onItemChange.bind(this));
		return {
			items,
			searchCriteria: null,
			displayedItems: items,
			toggleAllChecked: items.every(i => i.checked),
		};
	}
```

What do you think ?

jsomsanith:
@ncomont it is more complicated, and cannot be done with componentShouldUpdate (cSU)
This component maintains a state containing each item and its state. When the options change, we need update this state.
cWRP is supposed to allow that, whereas cSU is meant to test props/state to update the render or not.

packages/forms/src/widgets/ListViewWidget/ListViewWidget.js
jmfrancois: 👍

jmfrancois:
@ncomont I would really like to, but this is a big refactoring, and I only want to fix.
I would really like to know why we have that bunch of external handlers.
I have remove some of the stuff from the state.

packages/forms/src/widgets/ListViewWidget/ListViewWidget.js
```diff
@@ -29,6 +41,10 @@ function getItems(options, value, instance) {
 	}));
 }
 
+function getItemHeight() {
+	return DEFAULT_ITEM_HEIGHT;
```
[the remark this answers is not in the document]
jmfrancois: because the component behind wait for a function.
Here I have just remove the arrow function from constructor.

PR #1249, opened 2018-04-06
feat(components/icon): add name attr and spread props

jmfrancois:

We can't apply specific style to an icon using a simple css selector

let's add name attr (which is used to identify the icon).
And so let's spread all the props (aria- and so on) (Most of the file changes are snapshots)


packages/components/src/Icon/Icon.test.js
```diff
@@ -1,35 +1,40 @@
...
 
 	it('should render with provided className', () => {
-		const wrapper = renderer.create(<Icon name="svg-dd" className="custom-class" />).toJSON();
-		expect(wrapper).toMatchSnapshot();
+		const wrapper = shallow(<Icon name="svg-dd" className="custom-class" />);
+		expect(wrapper.getElement()).toMatchSnapshot();
+	});
+
+	it('should support extra props', () => {
+		const wrapper = shallow(<Icon name="svg-dd" className="custom-class" />);
```
jsomsanith: This is exactly the same test case as previous one. Let's add a new attribute to test that it passes extra props.

PR #1254, opened 2018-04-09
fix(cmf/onEventSetState): toggle call setState

jmfrancois:

If you use onClickSetState with 'toggle' feature you will have the following message:
`call function setState is undefined`

This is due to the fact the given props only contains the state not setState.

let's use instance.setState

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -3,7 +3,6 @@
   "parser": "babel-eslint",
   "rules": {
     "arrow-parens": [2, "as-needed"],
-    "import/no-extraneous-dependencies": ["error", {"devDependencies": ["**/*.test.js", "**/*.spec.js"]}],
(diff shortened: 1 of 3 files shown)


.eslintrc
```diff
@@ -3,7 +3,6 @@
   "parser": "babel-eslint",
   "rules": {
     "arrow-parens": [2, "as-needed"],
-    "import/no-extraneous-dependencies": ["error", {"devDependencies": ["**/*.test.js", "**/*.spec.js"]}],
```
frassinier: is this needed here? Because we maintain this file in tools too!

jmfrancois: Oh on my vscode it completly break eslint :/ I can re-add it but it just fails to start eslint server due to that

frassinier: May I ask you to do it in another PR? (tools first then go back here to fix it)

PR #1259, opened 2018-04-09
fix(containers/FilterBar): use props when passed

jlame:
The filterbar container was only using its state but in some case (when used inside of List) it needs to take the props.
https://jira.talendforge.org/browse/TDS-3254, this bug is also present on other apps.

Using props attribute when they are passed.

### DIFF
--- a/packages/containers/src/FilterBar/FilterBar.container.js
+++ b/packages/containers/src/FilterBar/FilterBar.container.js
@@ -13,6 +13,8 @@ export const DEFAULT_STATE = new Immutable.Map({
 
 export const DISPLAY_NAME = 'Container(FilterBar)';
 
+const DOCKED_ATTR = 'docked';
+
 class FilterBar extends React.Component {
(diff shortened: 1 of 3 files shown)


packages/containers/src/FilterBar/FilterBar.container.js
```diff
@@ -67,8 +67,8 @@ class FilterBar extends React.Component {
 	render() {
 		const state = this.props.state || DEFAULT_STATE;
 		const props = Object.assign({}, omit(this.props, cmfConnect.INJECTED_PROPS), {
-			docked: state.get('docked'),
-			value: state.get(QUERY_ATTR, ''),
+			docked: this.props.docked !== undefined ? this.props.docked : state.get('docked'),
```
ncomont: You should use `this.props.docked != null` to handle `null` and `undefined` (and add `const DOCKED_ATTR = 'docked';` at the top of this file :p )

PR #1276, opened 2018-04-12
feat(cmf): http cmf convention on http saga

sgendre:
Http request handled by sagas doesn't handle configuration from action with 
onResponse, onSend, onError, tranform, collectionId

Make those 2 work hand by hand.
+ add a "responseSelector" that allows to select within the response object what data actually matters

### DIFF
--- a/packages/cmf/src/sagas/http.js
+++ b/packages/cmf/src/sagas/http.js
@@ -2,6 +2,7 @@ import { call, put } from 'redux-saga/effects';
 import merge from 'lodash/merge';
 import curry from 'lodash/curry';
 import get from 'lodash/get';
+import { api } from '../api';
 
 import { mergeCSRFToken } from '../middlewares/http/csrfHandling';
 import {
@@ -10,6 +11,14 @@ import {
 	HTTP_STATUS,
 	testHTTPCode,
 } from '../middlewares/http/constants';
+import {
+	onRequest,
+	onResponse as onHttpResponse,
+	onError as onHttpError,
+	onActionResponse,
+	onActionError,
+	onJSError,
+} from '../actions/http';
 
 export class HTTPError extends Error {
 	constructor({ data, response }) {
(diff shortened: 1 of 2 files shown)


packages/cmf/src/cmfConnect.js
```diff
@@ -36,6 +36,8 @@ import ImmutablePropTypes from 'react-immutable-proptypes';
 import { connect } from 'react-redux';
 import omit from 'lodash/omit';
 import api from './api';
```
jmfrancois: I think we should also remove the api import

packages/cmf/src/sagas/http.js
```diff
@@ -45,7 +50,8 @@ export function handleBody(response) {
  * @return {Promise}           A promise that reject with the result of parsing the body
  */
 export function handleError(response) {
-	return handleBody(response).then(body => new HTTPError(body));
+	// BREAKING CHANGE to allow Go-like error handling
```
jmfrancois: I think this BR should exists in the PR description to think to move into the wiki on release

packages/cmf/src/sagas/http.js
```diff
@@ -111,6 +117,74 @@ export function httpFetch(url, config, method, payload) {
...
+			} catch (exception) {
+				errorToEnrich.stack.messageObject = 'Parsing response error';
+				return Promise.reject(errorToEnrich);
+			}
+		});
+}
+export function* errorProcessing(error, httpAction) {
+	const errorObject = {
+		name: error.name,
+		message: error.description || error.message,
```
jmfrancois: error.message is the official spec. what is error.description ?

packages/cmf/src/sagas/http.js
```diff
@@ -111,6 +117,74 @@ export function httpFetch(url, config, method, payload) {
...
+		number: error.number,
+		stack: error.stack,
+	};
+	const clone = get(error, 'stack.response.clone');
+	if (!clone) {
+		yield put(onJSError(errorObject, httpAction));
```
jmfrancois: I would like to add a console.error here. What do you think about this one ?
Most of the time it's an error in a reducer we just catch.
So we are sure the dev see this in the console which is not the case at the moment

packages/cmf/src/sagas/http.js
```diff
@@ -111,6 +117,74 @@ export function httpFetch(url, config, method, payload) {
...
+) {
+	const configEnhancer = handleDefaultConfiguration(defaultConfig);
+	if (isObject(urlOrBag)) {
+		return {
+			...urlOrBag,
+			config: configEnhancer(urlOrBag.config),
+		};
+	}
+
+	return {
```
jmfrancois: We should ensure url is a string, either throw an new Error

packages/cmf/src/sagas/http.js
```diff
@@ -163,14 +299,14 @@ export function* httpPost(url, payload, config, options) {
...
-	return yield* wrapFetch(url, config, HTTP_METHODS.PATCH, payload, options);
+export function* httpPatch(urlOrBag, payload, config, options, defaultConfig) {
+	return yield* wrapFetch(urlOrBag, config, HTTP_METHODS.PATCH, payload, options, defaultConfig);
 }
 
 /**
  * function - fetch a url with PUT method
  *
- * @param  {string} url     url to request
+ * @param  {string} urlOrBag    url to request or bag of config(http+cmf)
```
jmfrancois: {string|Object}

packages/cmf/src/sagas/http.js
```diff
@@ -179,68 +315,90 @@ export function* httpPatch(url, payload, config, options) {
...
+ * @param  {object} defaultConfig  	defaultConfig of the http client
+ * @example
+ * import { sagas } from '@talend/react-cmf';
+ * import { call } from 'redux-saga/effects'
+ * yield call(sagas.http.post, '/foo', {foo: 42});
+ */
+export function* http(urlOrBag, defaultConfig) {
+	return yield* wrapFetch(
+		urlOrBag,
+		urlOrBag.config,
```
jmfrancois: so this can't be url

packages/cmf/src/sagas/http.js
sgendre: I prefer to let developer use the signature they need, and avoid error but a console.warn about a migration would be fine to me.

packages/cmf/src/sagas/http.js
sgendre: :+1: I agree on providing an explicit message to developer

packages/cmf/src/sagas/http.js
sgendre: https://github.com/Talend/ui/blob/8386cc93a9c42e2de38aa5cd77a27882794a6b13/packages/cmf/src/middlewares/http/middleware.js#L181

PR #1281, opened 2018-04-17
chore(Datalist): adding noResultText prop

jlame:
noResultText cannot be passed from datalist to the typeahead underneath

add the prop and pass it


packages/components/src/Datalist/Datalist.component.js
```diff
@@ -315,6 +315,7 @@ class Datalist extends Component {
 					readOnly={this.props.readOnly || false}
 					theme={this.theme}
 					value={this.state.value}
+					noResultText={this.props.noResultText}
```
jmfrancois: please spread props here and remove autoFocus, disabled, ...

jlame: As discussed spreading the props would pass unwanted/unecessary one to the typeahead underneath so having this props whitelist is more suited. In addition we don't want to pass all props that the typeahead can manage cause the datalist is a specific usage of the typeahead.

PR #1282, opened 2018-04-17
fix(forms): switch to UIform if uischema is an array

jmfrancois:

1. If I want to use directly the UIform I need to pass uiform=true props which force the moz props.
2. With UIFrom, currently i can't used widgets
3. the functions onChange and onSubmit have different parameters and currently we can't use the Form.container for UIForm because of that
4. The implementations of onTrigger are Form and UIForm are different

1. I would like to just use @talend/react-forms passing a UISchema v2 (which is an array) and see it as it is.
2. Add props widget in Form.container
3. **[BREAKING CHANGE]** Adapt the component Form.js to call onChange and onSubmit with the same number of parameter as UIForm component. First is the event or null (we don't have the event with RJSF).

| Function | Before  | After |
| ------------- | ------------- | ------------- |
| onChange | `props.onChange(...args)` | `props.onChange(null, ...args)`  |
| onSubmit | `props.onSubmit(changes)` | `props.onSubmit(null, changes)`  |

4. delete the function onTrigger in Form.container and get it from props

This last change can impact projects using directly the component Form.js

**To summarize** : Now we can have three cases with usage of Form container
1. with a uiSpec v0 without the props 'uiform'
- that will use the RJSF as before
2. with a uiSpec v0 with the props 'uiform'
- that will use the UIForm with the merge.js function to adapt uiSpec
3. with a uiSpec v1 (the props uiform is not needed)
- that will use the UIForm without transformation

### DIFF
--- a/packages/forms/src/Form.js
+++ b/packages/forms/src/Form.js
@@ -120,7 +120,9 @@ class Form extends React.Component {
 	}
 
 	render() {
-		if (this.props.uiform) {
+		if (Array.isArray(this.props.data.uiSchema)) {
+			return <UIForm {...this.props} />;
+		} else if (this.props.uiform) {
 			const props = Object.assign({}, this.props);
 			props.moz = true;
 			if (props.widgets) {
@@ -131,9 +133,6 @@ class Form extends React.Component {
 					});
 			}
 			return <UIForm {...props} />;
-		}
-		if (Array.isArray(this.props.data.uiSchema)) {
-			return <UIForm {...this.props} />;
 		}
 		const schema = this.props.data && this.props.data.jsonSchema;
 		if (!schema) {
--- a/packages/forms/src/Form.js
+++ b/packages/forms/src/Form.js
@@ -132,6 +132,9 @@ class Form extends React.Component {
 			}
 			return <UIForm {...props} />;
 		}
+		if (Array.isArray(this.props.data.uiSchema)) {
+			return <UIForm {...this.props} />;
+		}
 		const schema = this.props.data && this.props.data.jsonSchema;
 		if (!schema) {
 			throw Error('You must provide data with valid JSON Schema');
--- a/packages/forms/src/Form.js
+++ b/packages/forms/src/Form.js
@@ -81,7 +81,7 @@ class Form extends React.Component {
 
 	handleSchemaSubmit(changes) {
 		if (this.props.onSubmit) {
-			this.props.onSubmit(changes);
+			this.props.onSubmit(null, changes);
 		}
 	}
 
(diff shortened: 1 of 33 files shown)


packages/forms/src/UIForm/UIForm.container.js
```diff
@@ -44,7 +44,7 @@ export default class UIForm extends React.Component {
 		if (!this.props.onChange) {
 			return;
 		} else if (this.props.moz) {
-			this.props.onChange(payload);
+			this.props.onChange(null, payload);
```
jsomsanith: Here you have the event (which can be null), just pass it. 
With that you can even remove this if/elseif/else by

```
if (this.props.onChange) {
    this.props.onChange(event, payload);
}
```

packages/forms/src/Form.js
```diff
@@ -132,6 +132,9 @@ class Form extends React.Component {
 			}
 			return <UIForm {...props} />;
 		}
+		if (Array.isArray(this.props.data.uiSchema)) {
```
jsomsanith: Not a big deal, but isn't that the most determinant condition ? If so it should be the first test in the `render` method.
1. if it's uiSpec v1 --> render directly the UIForm with all the props
2. if we have the uiForm flag, it means that we have uiSpec v0 but we want to use UIForm --> widget adaptation and render UIForm
3. else render RJSFForm

packages/forms/src/Form.test.js
```diff
@@ -373,3 +375,17 @@ describe('<Form/>', () => {
 		});
 	});
 });
+
+describe('<UIForm/>', () => {
```
jsomsanith: No need for another `describe` because you're still testing `<Form>`. Please move the test in the previous `describe`.

packages/forms/src/Form.test.js
```diff
@@ -373,3 +375,17 @@ describe('<Form/>', () => {
...
+	let props;
+	beforeEach(() => {
+		props = initProps();
+	});
+
+	it('should render uiform', () => {
```
jsomsanith: `should render UIForm with uiSchema array format` ?

packages/forms/src/Form.test.js
```diff
@@ -142,6 +144,13 @@ describe('<Form/>', () => {
 		expect(Form.displayName).toBe('TalendForm');
 	});
 
+	it('should render uiform', () => {
+		const props = initProps();
+		// when
+		wrapper = shallow(<Form data={dataUIForm} {...props} />);
+		expect(wrapper.getElement()).toMatchSnapshot();
```
jsomsanith: The shallow snapshot doesn't check that we have a UIForm.
1. Don't forget to check the snapshot when you write/modify one
2. Here you can check that without snapshot (which will log a lot more info than what you want to test)

jsomsanith: My bad I will learn to read

PR #1283, opened 2018-04-18
feat(cmf): add toJS memoized for expression / mapStateToProps

jmfrancois:

with CMF we want all data structure to be Immutable into the store for performance reason.
But many components want POJO, to the solution is to use toJS.
But it's known as a bad pratice because:
* it's slow
* it creates a new object on every mutation of the store so react-redux re-render the component

The solution comes with the following practice:
* you should write a selector to get your data from the state (without any other params than the state).
* then call the api.selectors.toJS(selector) to get your memoized selector

this memoized selector will return you the result of the toJS but it will be called only once
This garantie to do not have memory leak (only one instance is kept in cache)
And you keep good performances !

So if you had
```javascript
function mapStateToProps(state) {
  return {
    mystuff: state.cmf.collections.get('mycollection').toJS(),
  };
}
```

Which is very bad, you can now write the following

```javascript
function mapStateToProps(state) {
  return {
    mystuff: api.selectors.collections.toJS(state, 'mycollection'),
  };
}
```

### DIFF
--- a/packages/cmf/__tests__/selectors/toJS.test.js
+++ b/packages/cmf/__tests__/selectors/toJS.test.js
@@ -26,12 +26,27 @@ describe('toJS', () => {
 		const result2 = myselector(state);
 		expect(result1).toBe(result2);
 	});
+	it('the returned function should return a different result if store is has been modified', () => {
+		const myselector = toJS(selector);
+		const state = {
+			foo: new Immutable.Map({ bar: 'bar' }),
+		};
+		const result1 = myselector(state);
+		state.foo = state.foo.set('bar', 'baz');
+		const result2 = myselector(state);
+		expect(result1).not.toBe(result2);
+		expect(result2.bar).toBe('baz');
+	});
 	it('the returned function should throw an error if the selector return a not immutable data', () => {
 		const myselector = toJS(selector);
 		const state = {
 			foo: { bar: 'bar' },
 		};
 		const toThrow = () => myselector(state);
+		expect(toThrow).toThrow();
+	});
+	it('should throw if selector is not a function', () => {
+		const toThrow = () => toJS({});
 		expect(toThrow).toThrow();
 	});
 	it('the returned function should return undefined if the selector doesn t return data', () => {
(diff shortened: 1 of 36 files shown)


packages/cmf/src/selectors/toJS.js
```diff
@@ -0,0 +1,26 @@
+function toJS(data) {
+	if (data) {
+		if (data.toJS) {
```
acateland: i would also test with typeof to check if it is indeed a function

packages/cmf/src/selectors/toJS.js
```diff
@@ -0,0 +1,26 @@
...
+			return data.toJS();
+		}
+		throw new Error('the selector return a data which is not an immutable');
```
acateland: Enhancement of the error message
'The selector ${function.name} return a data which is not an immutable, and thus should bnot be used with `toJS` selector decorator'

for this you would propably need throw a more generic error to be catched into getToJSMemoized so it can fill the function name

packages/cmf/src/selectors/toJS.js
```diff
@@ -0,0 +1,26 @@
...
+	}
+	return undefined;
+}
+
+export default function getToJSMemoized(selector) {
```
acateland: js doc since its a public api

packages/cmf/src/selectors/toJS.js
```diff
@@ -0,0 +1,26 @@
...
+			if (cache.key === args[0]) {
+				return cache.value;
+			}
+			cache.key = args[0];
+			cache.value = func(...args);
+			return cache.value;
+		};
+	}
+	const toJSMemoized = memoize(toJS);
+	return state => toJSMemoized(selector(state));
```
acateland: check that selector is a function, to throw a proper error in dev mode

packages/cmf/__tests__/selectors/toJS.test.js
```diff
@@ -0,0 +1,42 @@
+import Immutable from 'immutable';
+import toJS from '../../src/selectors/toJS';
+
+describe('toJS', () => {
```
acateland: miss a test where the state get mutated between two calls, and the memoized function should invalidate its inner cache, to return the new proper result.

PR #1294, opened 2018-04-22
feat(cmf): allow to pass componentId in routes

romainseb:
Currently, we are stuck to the view to pass a componentId, we could use the "props" entry of the settings if the component id is handled

Before we do
```json
{
    "path": "datasets",
    "component": "HomeListView",
    "view": "HomeListView#datasets",
}
```

Let componentId be passed & set in the component if applicable, so in a route we could have :
```json
{
    "path": "datasets",
    "component": "HomeListView",
    "componentId": "datasets",
}
```


packages/cmf/__tests__/route.test.js
```diff
@@ -44,6 +44,22 @@ describe('loadComponent behavior', () => {
 		expect(mockItem.component.displayName).toBe('WithView');
 	});
 
+	it('should use the componentId', () => {
```
acateland: `should use the componentId` is a to vague statement to describe what the CMF route do here.

packages/cmf/src/route.js
```diff
@@ -76,12 +76,15 @@ function loadComponents(context, item, dispatch) {
 		item.component = component.get(item.component, context);
 		if (item.view && !item.component.CMFContainer) {
 			item.component = connectView(context, item.component, item.view);
-		} else if (item.view && item.component.CMFContainer) {
+		} else if ((item.view || item.componentId) && item.component.CMFContainer) {
 			const WithView = item.component;
-			item.component = props => <WithView view={item.view} {...props} />;
+			item.component = props => (
```
jmfrancois: really nice, could you please add a console.warn if item.view ?
`console.warn('DEPRECATION: please replace view by componentId and update your settings')`

PR #1295, opened 2018-04-23
feat(component): allow to pass classes to input file

romainseb:
We can't currently pass classes to input file
Use className attribute


packages/components/src/Actions/ActionFile/ActionFile.component.js
```diff
@@ -133,7 +135,7 @@ class ActionFile extends React.Component {
 					name={name}
 					id={localId}
 					disabled={inProgress || disabled}
-					className={classNames(theme['action-file-input'], 'sr-only')}
+					className={classNames(theme['action-file-input'], 'sr-only', className)}
```
acateland: should stay `classNames` since you can pass many class name to this props, you can however alias the `classNames` function on import.

packages/components/src/Actions/ActionFile/ActionFile.component.js
```diff
@@ -1,7 +1,7 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 import { OverlayTrigger } from 'react-bootstrap';
-import classNames from 'classnames';
+import classnames from 'classnames';
```
acateland: just a change on character case could lead to a lot on misunderstanding maybe something more relevant as name ?

PR #1305, opened 2018-04-24
feat(components/PieChartButton): add refs

nicomaligne:
There is no way to pass reference to the PieChartButton. 
But we need it to resolve some special case like with the focus on the hierarchical view.
Add ref props on the overlay and the button.


packages/components/src/PieChartButton/PieChartButton.component.js
```diff
@@ -190,18 +190,24 @@ export function setMinimumPercentage(model, minimumPercentage) {
  * @param {Element} overlayComponent the overlay component
  * @param {string} overlayId the id to be set for the overlay
  */
-export function decorateWithOverlay(btn, overlayPlacement, overlayComponent, overlayId) {
+export function decorateWithOverlay(
```
acateland: please update the above jsdoc

packages/components/src/PieChartButton/PieChartButton.component.js
```diff
@@ -190,18 +190,24 @@ export function setMinimumPercentage(model, minimumPercentage) {
...
 	}
-
 	return (
 		<span>
 			<OverlayTrigger
 				trigger="click"
 				rootClose
 				placement={overlayPlacement}
 				overlay={<Popover id={overlayId}>{overlayComponent}</Popover>}
+				ref={overlayRef}
```
acateland: what happen if overlayref is undefined ? does it trigger an error ?

PR #1309, opened 2018-04-24
chores(cmf): cleanup deprecation and tests

acateland:
Some deprecated function and method are still used internaly, and refrenced into documentation.
Removing or updating code and documentation to use non deprecated elements.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,6 +1,6 @@
 {
   "lerna": "2.5.1",
-  "version": "0.180.0",
+  "version": "0.181.0",
   "npmClient": "yarn",
   "useWorkspaces": true,
   "packages": [
--- a/packages/cmf-cqrs/package.json
+++ b/packages/cmf-cqrs/package.json
@@ -35,7 +35,7 @@
     "mock-socket": "^7.0.0"
   },
   "devDependencies": {
-    "@talend/react-cmf": "^0.180.0",
+    "@talend/react-cmf": "^0.181.0",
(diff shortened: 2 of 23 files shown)


acateland:
@jmfrancois 
https://github.com/Talend/ui/blob/master/packages/cmf/__tests__/cmfConnect.test.js#L211

im not sure to see the point of this test and the underlying feature
my understanding is that cmfConnect should not crash if a context is not provided, but it doesn't work as intended
first it create an error since `dispatchActionCreator` and maybe other cmf specific props are not handled properly and propagated to the children wich in case of this test is a div (could happen on live code to) and trigger a `Warning: Unknown prop `dispatchActionCreator` on <div> tag. Remove this prop from the element. For details, see https://fb.me/react-unknown-prop`

second i would prefer a big phat message telliming that something happened and that the context was not given to the component.

acateland:
@jmfrancois @jsomsanith shouldn't we remove those deprecated element soon they are 6 month olds

acateland:
@jmfrancois https://github.com/Talend/ui/blob/master/packages/cmf/src/route.js#L78
could use help to fix this one
`connectView` is deprecated
and cause the following errors
    console.warn src/deprecated.js:33
      DEPRECATED: The component _RouteComponentComponent must be connected using cmfConnect
    console.error ../../node_modules/react-redux/lib/utils/warning.js:14
      mapStateToProps() in Connect(_RouteComponentComponent) must return a plain object. Instead received undefined.

PR #1312, opened 2018-04-25
chore: update browserlist to >0.25% + IE11

jsomsanith:
(See https://jamie.build/last-2-versions)
`last 2 versions` as browserlist brings support to a lot more browsers than expected. More code is transpiled, and the bundle is heavier.
You can see the compatible browsers [here](http://browserl.ist/?q=last+2+versions)

Change it to `>0.25%`, `IE 11`, `not op_mini all`
You can see the compatible browsers [here](http://browserl.ist/?q=%3E0.25%25%2C+IE+11%2C+not+op_mini+all)

This is done on
* Talend/ui's babel and autoprefixer conf
* Talend/scripts, so when app migrate their conf on it, they will have the new conf


.babelrc
```diff
@@ -4,9 +4,10 @@
       "env",
       {
         "targets": {
-          "ie": 11,
           "browsers": [
-            "last 2 versions"
+            ">0.25%",
+            "not op_mini all",
+            "IE 11"
```
jmfrancois: http://browserl.ist/?q=%3E0.25%25 
ie 11 is already in

jsomsanith: Yes, but I added it the be future proof, when it falls under we have to ensure that we can remove the support before we really remove it

jsomsanith:
@acateland no clue, just that we drop some browsers, but I won't go over the list of supported ES6+ features on those browsers to dress a list of things that won't be transpiled :(

acateland:
@jsomsanith i think a before after breakdown of file size would be great no need to go in great details.

jsomsanith:
@acateland tested on 2 projects, no changes in build size. So no harm in pratice, and potential gain theoritically

jmfrancois:
Note: please add the same in tools: https://github.com/Talend/tools/blob/master/tools-javascript/compilation/react/.babelrc

PR #1320, opened 2018-04-26
feat(i18n/l10n): add crowdin script + integration into CI

jmfrancois:

We don't want to store translations into UI repository
We don't have an API to store them internally
We have some hard coded translation into components/stories

Let's try crowdin service.
Create a script with the following features:

* status: give a status of translations
* download: download all existing translations
* upload: upload the result of i18next extraction

Plug the upload to the CI when running on master (after each PR merged)

The main idea is to trigger the download with that script during the release process so the locales are available under `@talend/react-components` **/locales** and `@talend/react-forms`**/locales**

You can try with @talend/react-components and @talend/react-forms using the following version: **0.83.0-beta-2**

We have moved the parse-i18n into *prepublish* of each packages.


crowdin.js
```diff
@@ -0,0 +1,156 @@
...
+function getDestination(fileEntry) {
+	debug(fileEntry.toString());
+	const path = fileEntry.entryName;
+	if (fileEntry.name === 'tui-components.json') {
+		return `packages/components/locales/${path}`; // "ja/tui-components.json"
+	} else if (fileEntry.name === 'tui-forms.json') {
+		return `packages/forms/locales/${path}`; // "ja/tui-forms.json"
+	} else if (fileEntry.isDirectory) {
+		debug(`directory found supposed to be a language folder: ${path}`);
+		if (!fs.existsSync(`packages/components/locales/${path}`)) {
```
jsomsanith: `getDestination()` should not create anything, but just return a destination.
Perhaps you can move that into `extractAndSave()`. At the beginning, test if it's a folder, if so create the folders, else write.

crowdin.js
```diff
@@ -0,0 +1,156 @@
...
+
+function debug(...args) {
+	if (program.verbose) {
+		console.log(...args);
+	}
+}
+
+// TODO: prompt for the PROJECT_KEY
+
+const PROJECT_KEY = program.key;
```
jsomsanith: Can you rename it to `API_KEY` ?

crowdin.js
```diff
@@ -0,0 +1,156 @@
...
+}
+
+function handleDownloadRequest(error, response, body) {
+	onError(error);
+	const zip = new AdmZip(body);
+	zip.getEntries().forEach(extractAndSave);
+}
+
+function cleanUpBeforeDownload() {
+	if (!fs.existsSync('packages/components/locales')) {
```
jsomsanith: Shouldn't we remove all the folder content ?

crowdin.js
```diff
@@ -0,0 +1,156 @@
...
+		},
+		encoding: null,
+	};
+	debug(`download url OPTIONS ${options}`);
+	request(options, handleDownloadRequest);
+}
+
+function upload() {
+	console.log('please use curl');
+	const files = FILES.map(file => `\n  -F "files[/${file.name}]=@${file.path}"`);
```
jsomsanith: Could we run `yarn extract-i18n` so we don't have to do it before running the script ?

crowdin.js
```diff
@@ -0,0 +1,156 @@
...
+}
+
+function getDestination(fileEntry) {
+	debug(fileEntry.toString());
+	const path = fileEntry.entryName;
+	if (fileEntry.name === 'tui-components.json') {
+		return `packages/components/locales/${path}`; // "ja/tui-components.json"
+	} else if (fileEntry.name === 'tui-forms.json') {
+		return `packages/forms/locales/${path}`; // "ja/tui-forms.json"
+	} else if (fileEntry.isDirectory) {
```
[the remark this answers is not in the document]
jmfrancois: AFAIK containers has no translations. I see i18next call but no locales extracted

crowdin.js
jmfrancois: I don't think this is needed.
I have renamed the function ensureLocalesExists

crowdin.js
jmfrancois: Good idea, just switch it. I have added the call to crowdin download in the extract

crowdin.js
```diff
@@ -0,0 +1,178 @@
...
+const URL = `https://api.crowdin.com/api/project/${PROJECT_ID}`;
+
+if (!API_KEY) {
+	throw new Error('you must set -k or put a key in .crowdin config file');
+}
+
+const FILES = [
+	{
+		name: 'tui-components.json',
+		path: `${__dirname}/i18n/components/en/tui-components.json`,
```
jare-talend: we can store the template on each packages in packages/{{package}}/locales/template/{{namespace}}.

So, we can add some variants for a key which are not parsed by i18next-parser (plural, context, ...)

packages/components/stories/config/i18n.js
```diff
@@ -1,67 +1,31 @@
...
-			},
-		},
-	},
-	debug: false,
-	wait: true, // globally set to wait for loaded translations in translate hoc
-});
+let i18nInitialized = false;
+const languages = ['en', 'fr', 'ja'];
+
+if (process.env.NODE_ENV === 'production') {
```
jare-talend: you can remove this. On storybook, you doesn't need to check the env.

packages/components/.gitignore
```diff
@@ -41,5 +41,7 @@ hs_err_pid*
 # Talend web apps build directory
 build/
 lib/
-
+locales/fr
+locales/en
+locales/ja
```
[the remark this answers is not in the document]
jmfrancois: Yes I'm not proud of this one but I want to save the template; but not the locales

jmfrancois:
Crowdin doesn't cover the need

jare-talend:
can you explain why ?

PR #1322, opened 2018-04-27
fix(cmf/scripts/settings): manage case of undefined namespace or not define in the cmf.json file

smouillour:
currently the script don't manage the case where the namespace is not defined in the cmf.json file or if the namespace is not define in the settings files

manage those cases

doc : https://github.com/Talend/ui/blob/bb6c2976931f3f2241dbac24416681c76f38fc91/packages/cmf/scripts/index.md#exemple-of-settings-with-translation

in this case, it's not necessary to add namespace in key.


acateland:
Lets discuss before merging this.

I am not sure that creating an exception to something so simple as a real bonus value. I think it will be more complicated for developer to switch between namespaced project to a non namespaced project.


it feel trivial to me to not forget to add namespace.

smouillour:
@acateland see with you : we will adapt the tmc project to the script and not the other way around

packages/cmf/scripts/cmf-settings.merge.js
```diff
@@ -95,6 +95,7 @@ function merge(options, errorCallback) {
 			namespaces,
 			cmfconfig.settings.i18n.languages,
 			cmfconfig.settings.i18n['extract-from'],
+			cmfconfig.settings.i18n['extract-ignore-namespace'],
```
jare-talend: this should be removed

PR #1327, opened 2018-05-03
chore(Containers/List): multi selection on list container

jlame:
Having multiselection using ListContainer is not possible at the moment (only available at component level).
Implement multi selection behaviour for list container


packages/containers/src/List/List.container.js
```diff
@@ -104,12 +114,49 @@ class List extends React.Component {
 		this.props.setState({ displayMode: payload });
 	}
 
+	onToggleMultiSelection(event, data) {
+		const state = this.props.state.toJS();
+		let selectedItems = [].concat(state.selectedItems);
+		if (selectedItems.find(itemKey => itemKey === data[this.props.multiSelectionKey])) {
+			selectedItems = selectedItems.filter(
```
jsomsanith: This is not really efficient. You go over the array to check if the item exists.
If it exists, you go over all elements of the array to filter.

Instead you can : 
* `slice()` to have a new array ref
* get the index of the element you want (`array.indexOf(itemKey)`)
* index > -1, you remove it (`splice()`)
* else you push it

packages/containers/src/List/List.container.js
```diff
@@ -104,12 +114,49 @@ class List extends React.Component {
...
+			selectedItems.push(data[this.props.multiSelectionKey]);
+		}
+		this.props.setState({
+			selectedItems,
+		});
+	}
+
+	onToggleAllMultiSelection() {
+		const state = this.props.state.toJS();
+		const items = this.props.items.toJS();
```
jsomsanith: I think you can do what you want without those `toJS()`

packages/containers/src/List/List.container.js
```diff
@@ -104,12 +114,49 @@ class List extends React.Component {
 		this.props.setState({ displayMode: payload });
 	}
 
+	onToggleMultiSelection(event, data) {
+		const state = this.props.state.toJS();
```
jsomsanith: Can you just convert the state property you need instead of the whole state ?

packages/containers/src/List/List.container.js
```diff
@@ -104,12 +114,49 @@ class List extends React.Component {
...
+	}
+
 	getGenericDispatcher(property) {
 		return (event, data) => {
 			this.props.dispatchActionCreator(property, event, data, this.context);
 		};
 	}
 
+	isSelected(item) {
+		const state = this.props.state.toJS();
```
jsomsanith: Can you convert only `selectedItems`?

packages/containers/src/List/List.container.js
```diff
@@ -104,12 +114,49 @@ class List extends React.Component {
...
+		if (state.selectedItems.find(itemKey => itemKey === item[this.props.multiSelectionKey])) {
+			return true;
+		}
+		return false;
```
jsomsanith: Replace this if/else by `return selectedItems.some(condition)`

packages/containers/src/List/List.container.js
```diff
@@ -180,8 +231,39 @@ class List extends React.Component {
...
+			props.toolbar.actionBar = { actions: {}, multiSelectActions: {} };
+
+			// settings up multi selection
+			if (props.multiSelectActions && props.multiSelectionKey) {
+				props.list.itemProps.onToggle = this.onToggleMultiSelection;
+				props.list.itemProps.onToggleAll = this.onToggleAllMultiSelection;
+				props.list.itemProps.isSelected = this.isSelected;
+				// selectedItems is part of the default state
+				// but host app can override it so
+				if (!state.selectedItems) {
```
jsomsanith: Is that really necessary ? This is a big constrains, that means that for every default state, we have to reset it each time it is set with an unwanted value.
I think we should not do that. Moreover, you set a new state within render, triggering a new render. If app overrides it, it should put a valid value.

packages/containers/src/List/List.test.js
```diff
@@ -295,6 +311,111 @@ describe('Container List', () => {
...
+			multiSelectionSetting.setState = jest.fn();
+			const state = fromJS({ selectedItems: [1, 2, 3] });
+			multiSelectionSetting.state = state;
+
+			// when
+			const wrapper = shallow(<Container {...multiSelectionSetting} items={items} />, {
+				lifecycleExperimental: true,
+			});
+			// then
+			expect(wrapper.props().toolbar.actionBar.selected).toBe(3);
```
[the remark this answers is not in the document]
jlame: renamed the test, in fact i wanted to say "display the multi action bar" which is caused by the selected props > 0

PR #1330, opened 2018-05-04
feat: add is active to list

VolodymyrKovalM:
Currently we add isActive function (which is responsible for changing color of selected list item) in every container that uses List

Add it to List from UI, so we will be able just pass "selectedIdPath" as a prop, in cmf views settings


packages/containers/src/List/List.connect.js
```diff
@@ -32,6 +32,19 @@ export function mapStateToProps(state, ownProps, cmfProps) {
 	};
 	if (ownProps.list) {
 		config.columns = ownProps.list.columns;
+
+		if (ownProps.selectedIdPath) {
+			const isActive = item =>
+				item.id === api.selectors.collections.get(state, ownProps.selectedIdPath);
+			props.list = {
```
jmfrancois: You create a new object on each state mutation, this need to be memoized

acateland:
Is that supposed to replace the `isActive`if yes please add deprecation warnings if `isActive` function is given and update storybook examples with this new usage
If not please explain why we should keep both and how isActive doesn't solve the problem of showing selection.

jmfrancois:
This is the next topic in my todo list: https://jira.talendforge.org/browse/TUI-121
So I will work on that

PR #1333, opened 2018-05-04
feat(components): Table

timbault:
We need a new component to display schema as a table. This component will be used first by the future TDP Mapper component.

New component using the html table tags and custom rendering capabilities.


acateland:
Before going deep in the rabbit hole, please explain why the existing components can't be used ?

timbault:
In the new TDP mapper component we need to display connections between an input and an output schema. A connection is a curve that links an input element to an output element. Connection is displayed as soon as at least one element is visible. So we need a non vistualized list for the schema.

packages/components/src/SimpleTable/ClassNameProvider.js
```diff
@@ -0,0 +1,37 @@
+/**
+ * This class provides classnames for the components which compose a table:
+ * the table itself, the header, the rows and the data.
+ */
+export default class ClassNameProvider {
```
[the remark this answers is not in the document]
timbault: with a classNameProvider, user can provide classnames dynamically, depending on each element of the table.

packages/components/src/SimpleTable/Cell.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+/**
+ * This component displays a cell as a simple label. The data must be a string.
+ */
+export default function Cell({ data, className }) {
+	return <div className={`comp-cell ${className}`}>{data}</div>;
```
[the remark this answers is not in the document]
timbault: It was originally a span, but for layout purpose (text alignment in a cell) i turn it into div. But i am not sure what is the best: div or span and display:block in css.

packages/components/src/SimpleTable/Cell.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+/**
+ * This component displays a cell as a simple label. The data must be a string.
+ */
+export default function Cell({ data, className }) {
```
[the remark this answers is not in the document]
timbault: done

packages/components/src/SimpleTable/Cell.js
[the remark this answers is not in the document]
timbault: done

packages/components/src/SimpleTable/DraggableComponent.js
```diff
@@ -0,0 +1,91 @@
+import React, { Component } from 'react';
+import { DragSource, DropTarget } from 'react-dnd';
+import PropTypes from 'prop-types';
+import flow from 'lodash/flow';
```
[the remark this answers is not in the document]
timbault: done

packages/components/src/SimpleTable/DraggableComponent.js
```diff
@@ -0,0 +1,91 @@
...
+			return props.extra.beginDrag(props.element);
+		},
+		endDrag(props) {
+			props.extra.endDrag();
+		},
+	};
+
+	const elementTarget = {
+		canDrop(props, monitor) {
+			const sourceItem = monitor.getItem();
```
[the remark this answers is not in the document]
timbault: elementTarget is a useless const? No, it is used at the end of the module.

return flow(
		DragSource(DRAGGABLE_ELEMENT_TYPE, elementSource, collectForDragSource),
		DropTarget(DRAGGABLE_ELEMENT_TYPE, elementTarget, collectForDropTarget),
	)(DraggableComponent);

packages/components/src/SimpleTable/Header.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+/**
+ * This component displays a header of a table column as a simple label.
+ * The data must be a string.
+ */
+export default function Header({ data, className }) {
```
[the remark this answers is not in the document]
timbault: done

packages/components/src/SimpleTable/DraggableComponent.test.js
```diff
@@ -0,0 +1,30 @@
...
+import DraggableComponent from './DraggableComponent.js';
+import Cell from './Cell';
+import * as TestData from './TestData';
+
+const draggableCell = DraggableComponent(Cell);
+
+/**
+ * Render a single draggable cell
+ */
+it('draggable-cell', () => {
```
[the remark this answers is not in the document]
timbault: done

packages/components/src/SimpleTable/Header.js
```diff
@@ -0,0 +1,15 @@
...
+	return <div className={`comp-header ${className}`}>{data}</div>;
```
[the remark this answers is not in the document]
timbault: done

packages/components/src/SimpleTable/SimpleTable.scss
```diff
@@ -0,0 +1,135 @@
...
+		padding: 4px;
+	}
+
+	.simple-table-row {
+		height: 25px;
+		background-color: #fff;
+	}
+
+	.simple-table-row-data-type {
+		font-family: Inconsolata, Monaco, Menlo, Consolas, Courier New, mono;
```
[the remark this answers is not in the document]
timbault: i don't understand

packages/components/src/SimpleTable/SimpleTable.scss
```diff
@@ -0,0 +1,135 @@
+:global {
+	#simple-table-container {
+		&,
+		> div,
+		> div > div {
```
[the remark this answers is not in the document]
timbault: done

packages/components/src/SimpleTable/SimpleTable.scss
```diff
@@ -0,0 +1,135 @@
...
+			display: flex;
+			justify-content: center;
+			width: 100vw;
```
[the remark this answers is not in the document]
timbault: do you mean "width: 100vw" is useless?

packages/components/src/SimpleTable/TestData.js
```diff
@@ -0,0 +1,81 @@
...
+
+export const KEYS = {
+	NAME: 'name',
+	TYPE: 'type',
+	DESC: 'desc',
+};
+
+export const columnKeys = [KEYS.NAME, KEYS.TYPE, KEYS.DESC];
+
+export const classNameProvider = {
```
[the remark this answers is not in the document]
timbault: well it is used in many tests.

packages/components/src/SimpleTable/TestData.js
```diff
@@ -0,0 +1,81 @@
...
+};
+
+export const element2 = {
+	id: 'elem_2',
+	name: 'element 2',
+	type: 'string',
+	desc: 'This element is a string',
+};
+
+export const KEYS = {
```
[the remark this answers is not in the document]
timbault: It is used by the tests

packages/components/src/SimpleTable/__snapshots__/Row.test.js.snap
```diff
@@ -0,0 +1,32 @@
...
+  <td>
+    <div
+      className="comp-cell classname-of-row-data-(elem_1,type)"
+    >
+      string
+    </div>
+  </td>
+  <td>
+    <div
+      className="comp-cell classname-of-row-data-(elem_1,desc)"
```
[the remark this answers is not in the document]
timbault: done

packages/components/src/SimpleTable/__snapshots__/DraggableComponent.test.js.snap
```diff
@@ -0,0 +1,13 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`draggable-cell 1`] = `
```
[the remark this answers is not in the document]
timbault: is there something to fix?

packages/components/stories/SimpleTable.js
```diff
@@ -0,0 +1,403 @@
...
+stories
+	.addDecorator(story => (
+		<div id="simple-table-container">
+			<IconsProvider />
+			{story()}
+		</div>
+	))
+	.addWithInfo('Simple Table', () => {
+		return (
+			<div>
```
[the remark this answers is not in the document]
timbault: removed

packages/components/src/SimpleTable/DraggableComponent.js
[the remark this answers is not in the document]
acateland: @frassinier why ?

packages/components/src/SimpleTable/DraggableComponent.js
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	function collectForDropTarget(connect, monitor) {
+		return {
+			connectDropTarget: connect.dropTarget(),
+			dragOver: monitor.isOver(),
+		};
+	}
+
+	class DraggableComponent extends Component {
```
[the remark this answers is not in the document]
acateland: @frassinier PureComponent subclassing does not what you think it does, it have nothing to do with function purity

packages/components/src/SimpleTable/TestData.js
```diff
@@ -0,0 +1,81 @@
...
+		return 'classname-of-simple-table';
+	},
+	getForHeader(columnKey) {
+		return `classname-of-header-${columnKey}`;
+	},
+	getForRow(element) {
+		return `classname-of-row-(${element.id})`;
+	},
+	getForRowData(element, columnKey) {
+		return `classname-of-row-data-(${element.id},${columnKey})`;
```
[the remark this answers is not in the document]
timbault: without this comma, eslint says: "Missing trailing comma"!

packages/components/stories/SimpleTable.js
```diff
@@ -0,0 +1,403 @@
...
+			columnKeys,
+			withHeader,
+			onScroll,
+			onClick,
+			onDoubleClick,
+		} = this.props;
+		return (
+			<SimpleTable
+			  elements={elements}
+	      columnKeys={columnKeys}
```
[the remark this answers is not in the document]
timbault: done

packages/components/src/SimpleTable/Cell.js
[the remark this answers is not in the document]
frassinier: Both are valid, but if you choose div, just add a comment to explain why please

packages/components/src/SimpleTable/ClassNameProvider.js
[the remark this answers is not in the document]
frassinier: I don't get it! Can we discuss it in private?

packages/components/src/SimpleTable/DraggableComponent.js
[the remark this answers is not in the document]
frassinier: Mea culpa https://stackoverflow.com/questions/35250500/correct-way-to-import-lodash

packages/components/src/SimpleTable/DraggableComponent.js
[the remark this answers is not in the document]
frassinier: `sourceItem` is useless IMHO

packages/components/src/SimpleTable/DraggableComponent.js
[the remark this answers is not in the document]
frassinier: `PureComponent` does not manipulate any state, right? 
If Tony wants to write it using class notation, he should prefer PureComponent, but my point was he should prefer pure function notation?

packages/components/src/SimpleTable/SimpleTable.scss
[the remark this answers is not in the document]
frassinier: Prefer relative values like `100%`

packages/components/src/SimpleTable/SimpleTable.scss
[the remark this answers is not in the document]
frassinier: If you want to display monospaced font, use the correct HTML element to do so.

More info: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/pre

packages/components/src/SimpleTable/__snapshots__/DraggableComponent.test.js.snap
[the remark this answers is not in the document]
frassinier: Yes snapshot title :) 

I prefer read this
```javascript 
exports[`display simple draggable-cell content 1`] = `
```

packages/components/src/SimpleTable/DraggableComponent.js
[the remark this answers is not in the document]
frassinier: 2nd mea culpa
https://reactjs.org/docs/react-api.html#reactpurecomponent

Tony, you must prefer fn annotation to render pure component.

packages/components/src/SimpleTable/Cell.js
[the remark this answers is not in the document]
timbault: ok done

packages/components/src/SimpleTable/DraggableComponent.js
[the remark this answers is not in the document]
timbault: ok done

packages/components/src/SimpleTable/DraggableComponent.js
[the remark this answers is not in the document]
timbault: yes sure, sorry.

frassinier:
From demo, I saw 2 points:
1/ Description (div) is larger than its table datacell (td), there is no ellipsis at all
2/ If you increase the first table datacell content in drag n drop demo, it will pass on top of the second (because absolute position)

We need to rework them

packages/components/stories/Table.js
```diff
@@ -0,0 +1,401 @@
...
+		'Email',
+		'URL',
+		'City',
+	],
+};
+
+const salesForceAccountSchema = {
+	id: 'sdc22c25sd1csd',
+	name: 'SALESFORCE.ACCOUNT',
+	elements: [
```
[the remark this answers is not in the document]
timbault: it is not the final schema used by the component. The method finalizeSchema is called to transform this initial column-oriented schema into a row-oriented one.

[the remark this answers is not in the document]
timbault: it is just a convenient way to define quickly a schema for a story.

packages/components/stories/Table.js
```diff
@@ -0,0 +1,401 @@
...
+			elements,
+			columnKeys,
+			withHeader,
+			onScroll,
+		} = this.props;
+		return (
+			<Table
+				elements={elements}
+				columnKeys={columnKeys}
+				classNameProvider={this}
```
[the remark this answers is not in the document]
timbault: The component ConnectedTable has also the role of classNameProvider. It is because some classnames are computed depending on component state.

packages/components/src/Table/Body/TableBody.js
```diff
@@ -0,0 +1,98 @@
...
+	classNameProvider,
+	columnKeys,
+	rowDataGetter,
+	rowRenderer,
+	onEnterRow,
+	onLeaveRow,
+) {
+	return (
+		<TableRow
+			key={getRowId(rowDataGetter, element)}
```
acateland: the function signature doesn't not match with `getRowId` id where the default return is based on the third argument `index` this may result as an undefined error please add index

packages/components/src/Table/Body/Body.scss
```diff
@@ -0,0 +1,9 @@
+$body-height: 300px !default;
```
frassinier: `$tc-table-tbody-height` instead

packages/components/src/Table/Body/TableBody.js
```diff
@@ -0,0 +1,102 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import TableRow, { getRowId } from '../Row/TableRow.js';
+import theme from './Body.scss';
```
frassinier: Rename file

packages/components/src/Table/Table.scss
```diff
@@ -0,0 +1,15 @@
+$tc-table-padding: 0.3rem;
+$tc-table-margin: 0.3rem;
+
+.tc-table {
+	padding: $tc-table-padding;
+	margin: $tc-table-margin;
+	position: relative;
+
+	table {
```
frassinier: Could you prefer a class instead of element here

packages/components/src/Table/Cell/TableClickableCell.js
```diff
@@ -0,0 +1,56 @@
...
+			</div>
+		);
+	}
+}
+
+TableClickableCell.propTypes = {
+	element: PropTypes.object,
+	data: PropTypes.string,
+	className: PropTypes.string,
+	extra: PropTypes.object,
```
acateland: please enhance extra with a shape proptypes

packages/components/src/Table/Header/TableHeader.js
```diff
@@ -0,0 +1,101 @@
...
+
+TableHeader.propTypes = {
+	updateHeadNodeRef: PropTypes.func,
+	classNameProvider: PropTypes.shape({
+		getForTable: PropTypes.func,
+		getForHeader: PropTypes.func,
+		getForRow: PropTypes.func,
+		getForRowData: PropTypes.func,
+	}),
+	columnKeys: PropTypes.array,
```
acateland: please add this props as required or add a default within TableHeader signature

packages/components/src/Table/Row/TableRow.js
```diff
@@ -0,0 +1,174 @@
...
+TableRow.propTypes = {
+	element: PropTypes.object,
+	index: PropTypes.number,
+	classNameProvider: PropTypes.shape({
+		getForTable: PropTypes.func,
+		getForHeader: PropTypes.func,
+		getForRow: PropTypes.func,
+		getForRowData: PropTypes.func,
+	}),
+	columnKeys: PropTypes.array,
```
acateland: please add a isRequired or add a default value to TableRow props

packages/components/src/Table/Table.js
```diff
@@ -0,0 +1,140 @@
...
+export default class Table extends Component {
+	constructor(props) {
+		super(props);
+		this.updateTableNodeRef = this.updateTableNodeRef.bind(this);
+		this.updateContentNodeRef = this.updateContentNodeRef.bind(this);
+		this.updateBodyNodeRef = this.updateBodyNodeRef.bind(this);
+		this.updateHeadNodeRef = this.updateHeadNodeRef.bind(this);
+	}
+
+	getTableNode() {
```
acateland: Please explain the purpose of thos 4 accessors ?

packages/components/src/Table/TableConfiguration.js
```diff
@@ -0,0 +1,140 @@
...
+	 * Return a classname for an element of the table
+	 * @param {object} element - An element of the table.
+	 */
+	getForRow() {
+		return `${rootClassname}-row`;
+	},
+	/**
+	 * Return a classname for a data of the given element.
+	 * @param {string} columnKey - The key identifying a column.
+	 * @param {object} element - An element of the table.
```
acateland: please remove as it does not appear on the function signature

packages/components/src/Table/TableConfiguration.js
```diff
@@ -0,0 +1,140 @@
...
+
+/**
+ * This object provides a cell renderer for a given column key.
+ * It is a default implementation which always return a TableCell component.
+ */
+export const rowRenderer = {
+	/**
+	 * Indicates if a row needs to be updated or not according to the received props.
+	 * Default implementation returns true.
+	 * @param {object} props - The props of the component rendering a row.
```
acateland: please remove as it does not appear on the function signature

packages/components/src/Table/TableConfiguration.js
```diff
@@ -0,0 +1,140 @@
...
+	 */
+	needRowUpdate() {
+		return true;
+	},
+	/**
+	 * Return the component used to render a data for the given column.
+	 * @param {string} columnKey - The key identifying a column.
```
acateland: please remove as it does not appear on the function signature

packages/components/src/Table/TableConfiguration.js
```diff
@@ -0,0 +1,140 @@
...
+	 */
+	getCellComponent() {
+		return TableCell;
+	},
+	/**
+	 * Return custom properties for the component displaying the data of the given column.
+	 * @param {string} columnKey - The key identifying a column.
```
acateland: please remove as it does not appear on the function signature

packages/components/src/Table/TableConfiguration.js
```diff
@@ -0,0 +1,140 @@
...
+};
+
+/**
+ * This object provides a header renderer for a given column key.
+ * It is a default implementation which always return a TableHeaderCell component.
+ */
+export const headerRenderer = {
+	/**
+	 * Return the component used to render a header for the given column.
+	 * @param {string} columnKey - The key identifying a column.
```
acateland: please remove as it does not appear on the function signature

packages/components/src/Table/TableConfiguration.js
```diff
@@ -0,0 +1,140 @@
...
+	 */
+	getHeaderComponent() {
+		return TableHeaderCell;
+	},
+	/**
+	 * Return extra properties for the header component rendering the given column header.
+	 * @param {string} columnKey - The key identifying a column.
```
acateland: please remove as it does not appear on the function signature

packages/components/src/Table/TableConfiguration.js
```diff
@@ -0,0 +1,140 @@
...
+	 */
+	getHeaderData(columnKey) {
+		return columnKey;
+	},
+	/**
+	 * Return the data corresponding to the given element and column.
+	 * @param {object} element - An element of the table.
+	 * @param {string} columnKey - The key identifying a column.
+	 */
+	getRowData(element, columnKey) {
```
acateland: shouldn't element be renamed `row` or `rowData` ?

packages/components/src/Table/Body/TableBody.js
```diff
@@ -0,0 +1,98 @@
...
+					onEnterRow,
+					onLeaveRow,
+				),
+			)}
+		</tbody>
+	);
+}
+
+TableBody.propTypes = {
+	elements: PropTypes.array,
```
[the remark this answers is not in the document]
timbault: done

packages/components/src/Table/Body/TableBody.js
timbault: done

packages/components/src/Table/Header/TableHeader.js
timbault: done

packages/components/src/Table/Row/TableRow.js
timbault: done

packages/components/src/Table/Header/TableHeader.js
```diff
@@ -0,0 +1,101 @@
...
+	rowDataGetter: PropTypes.shape({
+		getElementId: PropTypes.func,
+		getHeaderData: PropTypes.func,
+		getRowData: PropTypes.func,
+	}),
+	headerRenderer: PropTypes.shape({
+		getHeaderComponent: PropTypes.func,
+		getExtraProps: PropTypes.func,
```
jsomsanith: Instead of providing tons of getters, that makes it difficult to use because it is totally customisable.
And if we have to write all parts as getter, wouldn't it be easier to just write a table component instead of using this one ?
I think we should define a more closed object api (need more opinon, ping @frassinier @acateland 
Example : 
```
columns: PropTypes.arrayOf(PropTypes.shape({
    key: PropTypes.string, // column key
    label: PropTypes.string, // label to display
    className: PropTypes.string, // cell classname
    headerClassName: PropTypes.string, // header classname
    cellRenderer: PropTypes.func, // optional renderer override
    // and you can put all the extra props, no need to have a getter for every info
}))
```

packages/components/src/Table/Body/TableBody.js
```diff
@@ -0,0 +1,102 @@
...
+	if (classNameProvider && classNameProvider.getForBody) {
+		return classNameProvider.getForBody();
+	}
+	return '';
+}
+
+/**
+ * This function is responsible for rendering an element in the table.
+ */
+function renderRow(
```
jsomsanith: You don't need this function, just put the resulting jsx in your elements.map()

packages/components/src/Table/Row/TableRow.js
```diff
@@ -0,0 +1,174 @@
...
+		super(props);
+		this.handleMouseEnter = this.handleMouseEnter.bind(this);
+		this.handleMouseLeave = this.handleMouseLeave.bind(this);
+		this.updateRowRef = this.updateRowRef.bind(this);
+	}
+
+	componentDidMount() {
+		if (this.rowRef != null) {
+			this.rowRef.addEventListener('mouseenter', this.handleMouseEnter);
+			this.rowRef.addEventListener('mouseleave', this.handleMouseLeave);
```
jsomsanith: Use refs + DOM manipulation to add the listeners ?
In your jsx : 
```
<tr onMouseEnter={} onMouseLeave={}>
```

packages/components/src/Table/Row/TableRow.js
```diff
@@ -0,0 +1,174 @@
...
+		}
+	}
+
+	shouldComponentUpdate(nextProps) {
+		if (this.props.rowRenderer && this.props.rowRenderer.needRowUpdate) {
+			return this.props.rowRenderer.needRowUpdate(nextProps);
```
jsomsanith: What is the need behind that ?
Why do we let an external function control the rendering of the component ?

packages/components/src/Table/TableConfiguration.js
```diff
@@ -0,0 +1,133 @@
...
+	 * Return the data corresponding to the given row and column.
+	 * @param {object} row - A row of the table.
+	 * @param {string} columnKey - The key identifying a column.
+	 */
+	getRowData(row, columnKey) {
+		return row[columnKey];
+	},
+};
+
+const defaultConfiguration = {
```
jsomsanith: What is that ? It's not used internally, but is exported.

packages/components/src/Table/TableWithDnD.test.js
```diff
@@ -0,0 +1,104 @@
...
+function wrapInTestContext(DecoratedComponent) {
+	return dndContext(TestBackend)(props => <DecoratedComponent {...props} />);
+}
+
+const draggableCell = draggable(TableCell);
+
+/**
+ * This tests the drag and drop of the element1 onto the element2.
+ */
+it('drag-and-drop-on-table', () => {
```
jsomsanith: Put that is Talend.test.js please

packages/components/src/Table/Cell/DraggableComponent.js
```diff
@@ -0,0 +1,88 @@
...
+ * This constant allows to specify which drag sources and drop targets are compatible.
+ */
+export const DRAGGABLE_ELEMENT_TYPE = 'element';
+
+/**
+ * This function adds the 'drag & drop' behaviour on the given component.
+ * Its returns a new component which encaspulates the given one.
+ * The new component is draggable and dropppable.
+ */
+export default function getDraggable(Comp) {
```
jsomsanith: I guess this is a HOC to wrap the celle renderers to inject DnD in it.
2 possibilities : 
* host it in Table/Cell but integrate it through the Table component api. I mean, accept a props.onMove, when present, wrap the renderer with this HOC.
* host it in src/, making it a component on its own, that manage DnD.

I would say number 2

packages/components/src/Table/Cell/TableClickableCell.js
```diff
@@ -0,0 +1,60 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+
+/**
+ * This component displays a cell of a table as a clickable element. The data must be a string.
+ * Callback methods (onClick, onDoubleClick and onKeyPress) must be passed via extra props.
+ */
+export default class TableClickableCell extends Component {
```
jsomsanith: This seems to be a default cell renderer.
For now you have to override the default renderer in order to have a click support. But what about a DefaultCellRenderer that manage the clicks if you have the callbacks in props. And no onClick will be passed to the div if they are undefined.
So that make it a merge of your 2 components TableCell and TableClickableCell.

packages/components/src/Table/Body/TableBody.js
```diff
@@ -0,0 +1,95 @@
...
+					onLeaveRow,
+				),
+			)}
+		</tbody>
+	);
+}
+
+TableBody.propTypes = {
+	elements: PropTypes.array,
+	classNameProvider: PropTypes.shape({
```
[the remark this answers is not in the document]
timbault: yes it should be lighter but the aim of this object is to provide classnames dynamically. For example it can return different classnames depending on the current row element or the current column. This feature will be used in the future data-mapper.

packages/components/src/Table/Table.js
```diff
@@ -0,0 +1,140 @@
...
+			onEnterRow,
+			onLeaveRow,
+		} = this.props;
+		const classnames = classNames(
+			'tc-table',
+			theme['tc-table'],
+			getTableClassName(classNameProvider),
+		);
+		return (
+			<div ref={this.updateContentNodeRef} className={classnames}>
```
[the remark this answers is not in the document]
timbault: Ok, but Table component is composed of two sub-parts (a header and a table). So i need a container. What is the alternative solution?

packages/components/src/Table/Table.js
```diff
@@ -0,0 +1,140 @@
...
+	updateBodyNodeRef(ref) {
+		this.bodyNode = ref;
+	}
+
+	updateHeadNodeRef(ref) {
+		this.headNode = ref;
+	}
+
+	updateContentNodeRef(ref) {
+		this.contentNode = ref;
```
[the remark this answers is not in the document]
timbault: It is removed.

packages/components/src/Table/Header/TableHeader.js
```diff
@@ -0,0 +1,101 @@
...
+	const classnames = classNames(
+		'tc-table-header',
+		theme['tc-table-header'],
+		getHeaderClassName(classNameProvider),
+	);
+	return (
+		<thead ref={updateHeadNodeRef} className={classnames}>
+			<tr className={theme['tc-table-header-row']}>
+				{columnKeys.map(col =>
+					renderHeaderCell(classNameProvider, rowDataGetter, headerRenderer, col),
```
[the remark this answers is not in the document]
timbault: yes, bad name. I propose to rename it "tableDatagetter".

packages/components/src/Table/Header/TableHeader.js
```diff
@@ -0,0 +1,101 @@
...
+function getHeaderData(rowDataGetter, columnKey) {
+	if (rowDataGetter && rowDataGetter.getHeaderData) {
+		return rowDataGetter.getHeaderData(columnKey);
+	}
+	return columnKey;
+}
+
+function getHeaderComponent(headerRenderer, columnKey) {
+	if (headerRenderer && headerRenderer.getHeaderComponent) {
+		return headerRenderer.getHeaderComponent(columnKey);
```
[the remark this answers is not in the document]
timbault: It will be used to display header with sort feature. Some columns may have a sorter, others may not.

packages/components/src/Table/Header/TableHeader.js
timbault: In fact providers are not mandatory. If you don't provide them, the table component use default renderers. Table has only two mandatory props: the array of elements and the array of column keys.

packages/components/src/Table/Body/TableBody.js
timbault: ok

packages/components/src/Table/Table.scss
```diff
@@ -0,0 +1,16 @@
+.table {
+	padding: 0;
+	padding-top: 2px;
+	padding-bottom: 2px;
+	margin: 0;
+	margin-top: 2px;
+	margin-bottom: 2px;
```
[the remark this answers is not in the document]
timbault: ok

packages/components/src/Table/Row/TableRow.js
timbault: We could control the rendering of the rows depending on the parent state for example. This feature will be used by the data-mapper.

packages/components/src/Table/TableConfiguration.js
timbault: this aggregates the different default providers in a single object. An external user can import this and thus have access to all providers by a single object

packages/components/src/Table/TableWithDnD.test.js
timbault: Sorry i don't understand. I have to rename the file? Or i have to move this test elsewhere?

packages/components/src/Table/Table.js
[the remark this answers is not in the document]
jsomsanith: The container is `<table>`

packages/components/src/Table/Header/TableHeader.js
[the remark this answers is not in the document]
jsomsanith: I still don't get it because in the default header component you don"'t use it. What is that ?

packages/components/src/Table/Row/TableRow.js
jsomsanith: I still don't get it. I understand that something from the parent controls if this component should render or not, but I don't understand why. This pattern is a little weird.

packages/components/src/Table/Table.js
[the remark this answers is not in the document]
timbault: sorry you're right. Of course, i remove this div.

packages/components/src/Table/TableConfiguration.js
jsomsanith: But if those are default providers, they should not have to use them.
Default providers are the default props, if none is provided. So I still don't get why this aggregation is created and exported.

packages/components/src/Table/TableWithDnD.test.js
jsomsanith: I suggest to move this test in `Table.test.js` (sorry typo in my first comment).

packages/components/src/Table/Header/TableHeader.js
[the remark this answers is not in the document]
timbault: TableHeader calls the getHeaderData() method in order to get the data corresponding to the column header. This data is then passed to the component which renders the header cell.

packages/components/src/Table/Row/TableRow.js
timbault: well in fact i need it for some performance issues in the future data-mapper. When i display hundreds of rows, i provide some custom optimizations in order to reduce the rendering process.

packages/components/src/Table/TableConfiguration.js
timbault: Well in fact you're right, the table component doesn't use it by default. In fact i have written this module in order to provide in a single place all the specifications of the needed providers. We can remove it.

packages/components/src/Table/TableWithDnD.test.js
timbault: ok

packages/components/src/Table/Cell/TableClickableCell.js
timbault: just to be sure, i merge these two components into only one TableCell component, that's right?

packages/components/src/Table/Header/TableHeader.js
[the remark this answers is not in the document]
jsomsanith: My question is about `rowDataGetter`

packages/components/src/Table/Cell/TableClickableCell.js
jsomsanith: Yes, to have only 1 component that deals with clicks if some props are provided

packages/components/src/Table/Header/TableHeader.js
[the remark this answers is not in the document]
timbault: yes i understand, rowDataGetter provides data for rows and header. So it is why i propose to rename it to "tableDataGetter". It was called like that because in its initial version it didn't provide data for header.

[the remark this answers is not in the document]
jsomsanith: Can you give me an example of what this tableDataGetter would provide ?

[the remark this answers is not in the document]
timbault: For the story, i have defined the following tableDataGetter:

	const tableDataGetter = {
	  getElementId(element) {
    	return element.id;
	  },
	  getHeaderData(columnKey) {
		switch (columnKey) {
			case ColumnKey.NAME:
      			return 'Name';
			case ColumnKey.TYPE:
	  			return 'Type';
			case ColumnKey.DESC:
				return 'Description';
			default:
				return columnKey;
		}
	  },
	  getRowData(element, columnKey) {
    	switch (columnKey) {
			case ColumnKey.NAME:
        		return element.name;
			case ColumnKey.TYPE:
				return element.type;
			case ColumnKey.DESC:
				return element.description;
			default:
				return 'No data available!';
 	  	}
	  },
	};

packages/components/src/Table/Row/TableRow.js
```diff
@@ -0,0 +1,152 @@
...
+import TableCell from '../Cell/TableCell';
+import theme from './TableRow.scss';
+
+export function getRowId(rowDataGetter, element, index) {
+	if (rowDataGetter && rowDataGetter.getElementId) {
+		return rowDataGetter.getElementId(element);
+	} else if (element.id && typeof element.id === 'string') {
+		return element.id;
+	}
+	return index.toString();
```
jsomsanith: Why this `toString()` ?

packages/components/src/Table/Row/TableRow.js
```diff
@@ -0,0 +1,152 @@
...
+}
+
+function getRowClassName(classnames, index) {
```
jsomsanith: This can be easier to write with lodash/get
```
import get from 'lodash/get';

function getRowClassName(classnames, index) {
	return get(classnames, `rows[${index}]`);
}

```

packages/components/src/Table/Row/TableRow.js
```diff
@@ -0,0 +1,152 @@
...
+}
+
+function getRowDataClassName(column) {
+	if (column.cellClassName) {
+		return column.cellClassName;
+	}
+	return `tc-table-row-data-${column.key}`;
+}
+
+function getCellComponent(column) {
```
jsomsanith: This function is not needed.
You can remove it and write `const CellComponent = column.cellRenderer || TableCell;`

packages/components/src/Table/Row/TableRow.js
```diff
@@ -0,0 +1,152 @@
...
+		classnames.rows &&
+		Array.isArray(classnames.rows) &&
+		classnames.rows.length >= index
+	) {
+		return classnames.rows[index];
+	}
+	return null;
+}
+
+function getRowDataClassName(column) {
```
jsomsanith: This function is not necessary. You can directly write
```
column.cellClassName || `tc-table-row-data-${column.key}`
```

But shouldn't you add both classnames ?

packages/components/src/Table/Row/TableRow.js
```diff
@@ -0,0 +1,152 @@
...
+	if (column.cellRenderer) {
+		return column.cellRenderer;
+	}
+	return TableCell;
+}
+
+function getRowData(rowDataGetter, element, columnKey) {
+	if (rowDataGetter && rowDataGetter.getRowData) {
+		return rowDataGetter.getRowData(element, columnKey);
+	} else if (element && element[columnKey]) {
```
jsomsanith: This check is unnecessary. `element` is required so you can write 
```
if (rowDataGetter && rowDataGetter.getRowData) {
	return rowDataGetter.getRowData(element, columnKey);
}
return element[columnKey];
```

packages/components/src/Table/Cell/TableCell.js
```diff
@@ -0,0 +1,82 @@
...
+/**
+ * This component displays a cell of a table as a clickable element. The data must be a string.
+ * Callback methods (onClick, onDoubleClick and onKeyPress) must be passed via extra props.
+ */
+export default class TableCell extends Component {
+	constructor(props) {
+		super(props);
+		this.onClick = this.onClick.bind(this);
+		this.onDoubleClick = this.onDoubleClick.bind(this);
+		this.onDoubleClick = this.onDoubleClick.bind(this);
```
jsomsanith: Remove this line please

packages/components/src/Table/Cell/TableCell.js
```diff
@@ -0,0 +1,82 @@
...
+		}
+		return renderCell(data, className);
+	}
+}
+
+TableCell.propTypes = {
+	element: PropTypes.object,
+	data: PropTypes.string,
+	className: PropTypes.string,
+	extra: PropTypes.shape({
```
jsomsanith: Why do we have this `extra` level in props ? Why don't you put onClick, onDoubleClick, onKeyPress directly in props ?

packages/components/src/Table/Cell/TableCell.js
```diff
@@ -0,0 +1,82 @@
...
+		}
+	}
+
+	onKeyPress(ev) {
+		if (this.props.extra && this.props.extra.onKeyPress) {
+			this.props.extra.onKeyPress(this.props.element, ev);
+		}
+	}
+
+	render() {
```
jsomsanith: Missing unit tests

packages/components/src/Table/Header/TableHeader.js
```diff
@@ -0,0 +1,80 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import theme from './TableHeader.scss';
+import TableHeaderCell from './TableHeaderCell';
+
+function getHeaderClassName(classnames) {
```
jsomsanith: You don't need this function, you can directly write `classnames && classnames.header`

packages/components/src/Table/Header/TableHeader.js
```diff
@@ -0,0 +1,80 @@
...
+}
+
+function getHeaderCellClassName(column) {
+	if (column.headClassName) {
+		return column.headClassName;
+	}
+	return column.key;
+}
+
+function getHeaderComponent(column) {
```
jsomsanith: You don't need this function, you can directly write `column.headRenderer || TableHeaderCell`

packages/components/src/Table/Header/TableHeader.js
```diff
@@ -0,0 +1,80 @@
...
+import TableHeaderCell from './TableHeaderCell';
+
+function getHeaderClassName(classnames) {
+	if (classnames && classnames.header) {
+		return classnames.header;
+	}
+	return null;
+}
+
+function getHeaderCellClassName(column) {
```
jsomsanith: You don't need this function, you can directly write `column.headClassName`. (I removed `column.key`, it's weird adding that as classname)

packages/components/src/Table/Cell/TableCell.js
timbault: Because it is the contract for all the cell component. Whatever the cell component, the props are always { element, data, className, extra }. The content of extra may vary depending on the cell component.

output/components.eslint.txt
```diff
@@ -21,11 +21,14 @@ The react/require-extension rule is deprecated. Please use the import/extensions
   1:1  error  '@storybook/react' should be listed in the project's dependencies, not devDependencies  import/no-extraneous-dependencies
   4:2  error  Unexpected require()                                                                    global-require
 
+/home/travis/build/Talend/ui/packages/components/src/Table/Cell/TableCell.js
+  15:3  error  Visible, non-interactive elements should not have mouse or keyboard event listeners  jsx-a11y/no-static-element-interactions
```
talend-glorieux: Please remove that error.

packages/components/src/Draggable/DraggableComponent.js
```diff
@@ -0,0 +1,88 @@
...
+ * The props 'element' is used as dragged data.
+ * The props 'extra' must provide dnd callback methods (see elementSource and elementTarget).
+ * @param {Component} Comp - The component on which the dragd and drop behavior must be defined.
+ * @param {string} type - Define the type of element which can be dragged and dropped.
+ */
+export default function getDraggable(Comp, type) {
+	/*
+	 * This defines the drag source specifications.
+	 */
+	const elementSource = {
```
talend-glorieux: Please use single line comments on all those one liners.

packages/components/src/Draggable/DraggableComponent.test.js
```diff
@@ -0,0 +1,142 @@
...
+import PropTypes from 'prop-types';
+import TestUtils from 'react-dom/test-utils';
+import draggable from './DraggableComponent.js';
+
+function Element({ element }) {
+	return (
+		<div className="elem-class">
+			<span>{element.symbol}</span>
+			<span>:</span>
+			<span>{element.name}</span>
```
talend-glorieux: Why not use a definition list here?

packages/components/src/Draggable/DraggableComponent.test.js
```diff
@@ -0,0 +1,142 @@
...
+	symbol: 'rh',
+	name: 'Rhodium',
+};
+
+const pd = {
+	symbol: 'pd',
+	name: 'Palladium',
+};
+
+const elements = [ru, rh, pd];
```
talend-glorieux: Please define those inline inside the array.

packages/components/src/Table/Row/TableRow.js
```diff
@@ -0,0 +1,129 @@
...
+			cellRenderer: PropTypes.func, // cell renderer
+			cellExtraProps: PropTypes.object, // cell extra props
+		}),
+	).isRequired,
+	rowDataGetter: PropTypes.shape({
+		/**
+		 * Return an unique identifier for the given element.
+		 * Each element is displayed in a row.
+		 * @param {object} element - An element of the table.
+		 */
```
talend-glorieux: Documentation should be applied to the function definition. Not the propTypes.

packages/components/src/Table/Table.js
```diff
@@ -0,0 +1,69 @@
...
+	columns: PropTypes.arrayOf(
+		PropTypes.shape({
+			key: PropTypes.string.isRequired, // column key
+			label: PropTypes.string, // label to display
+			headClassName: PropTypes.string, // header classname
+			headRenderer: PropTypes.func, // header renderer
+			headExtraProps: PropTypes.object, // header extra props
+			cellClassName: PropTypes.string, // cell classname
+			cellRenderer: PropTypes.func, // cell renderer
+			cellExtraProps: PropTypes.object, // cell extra props
```
talend-glorieux: Please remove useless extra comments.

packages/components/src/Draggable/DraggableComponent.js
```diff
@@ -0,0 +1,88 @@
...
+import PropTypes from 'prop-types';
+import flow from 'lodash/flow';
+
+/**
+ * This function adds the 'drag & drop' behaviour on the given component.
+ * Its returns a new component which encaspulates the given one.
+ * The new component is draggable and dropppable.
+ * The props of the component must have at least: 'element' (object) and 'extra' (object).
+ * The props 'element' is used as dragged data.
+ * The props 'extra' must provide dnd callback methods (see elementSource and elementTarget).
```
talend-glorieux: `extra` doesn't sound like a good name. Anything can be put on it.
If the dnd callback must be provided they should be in there own object.
Or find another name.

output/components.eslint.txt
timbault: Seen with @frassinier , we don't see why this error appears. The related code is:

<div
	className={classNames('tc-table-clickable-cell', className)}
	onClick={onClick}
	onDoubleClick={onDoubleClick}
	onKeyPress={onKeyPress}
	role="button"
	tabIndex="-1"
>
	{data}
</div>

According to this documentation [https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/no-static-element-interactions.md], it should be ok

packages/components/src/Draggable/DraggableComponent.test.js
timbault: i just need a simple component for the test

packages/components/src/Draggable/DraggableComponent.test.js
timbault: But i use the ru and rh objects later in the test.

timbault:
@talend-glorieux where the component's guidelines should be defined?

packages/components/src/Table/Row/TableRow.js
timbault: Yes but the Table module doesn't define any implementation of rowDataGetter object. So i don't see where i could define the contract of this object. PropTypes seems to be the only reasonable place.

output/components.eslint.txt
talend-glorieux: It's because you attach an onclick to a div. 
Just add and `eslint-ignore` comment.

packages/components/src/Draggable/DraggableComponent.test.js
talend-glorieux: OK.

packages/components/src/Table/Table.js
timbault: done

packages/components/src/Draggable/DraggableComponent.js
timbault: done

PR #1341, opened 2018-05-09
fix(cmf): listen the last settings fetched to get the document title

jare-talend:
With the i18n, we call the first time the default settings 'en' and whe we get the preferencedLanguages,w e call again the settings for this localse.
When the router change, he get the documentTitle from the first call to the settings and not the last.

change take by takeLatest


packages/cmf/src/sagas/documentTitle.test.js
```diff
@@ -0,0 +1,66 @@
+import { takeLatest, take } from 'redux-saga/effects';
```
annilsson: There's a test file already in `cmf\__tests__\sagas\documentTitle.test.js`, maybe these tests could be moved there?

PR #1343, opened 2018-05-10
fix: few prop-types

VolodymyrKovalM:
We pass uiSchema as object also, to UIForm, and since in prop-types only array is allowed we have warnings in console

Fix prop-types


VolodymyrKovalM:
@jsomsanith thanks for Your review, so I will close this PR in this case, and we will fix this on our side,
because now there is a place where still an object passed

VolodymyrKovalM:
@jsomsanith thanks for Your review, so I will close this PR in this case, and we will fix this on our side,
because now there is a place where still an object passed

PR #1345, opened 2018-05-14
feat(scripts): cmf plugin

jsomsanith:
Talend/scripts does not use cmf webpack plugin

Integrate it within Talend/scripts
* cmf webpack plugin updated to match webpack 4 api
* Talend/scripts read `cmf.json` and set the destination depending on webpack conf. Convention : `settings.json` at the build folder root
* cmf settings merge accepts cmf conf from input, and fallback to `cmf.json`


packages/scripts/preset/config/webpack.config.js
```diff
@@ -61,7 +61,7 @@ module.exports = ({ getUserConfig }) => {
 	return {
 		entry: ['babel-polyfill', 'whatwg-fetch', `${process.cwd()}/src/app/index.js`],
 		output: {
-			path: `${process.cwd()}/dist`,
+			path: `${process.cwd()}/build`,
```
jmfrancois: why build ?
I don't really like it, build folder make me feel like `tmp`
dist stands for distribution and is the default value in webpack

jsomsanith: 👍 Thanks for the info, I just removed it to use webpack default value and adapt devServer to server static from this default folder

packages/cmf-webpack-plugin/src/index.js
```diff
@@ -32,95 +12,85 @@ function ReactCMFWebpackPlugin(options = {}) {
...
+ReactCMFWebpackPlugin.prototype.apply = function reactCMFWebpackPluginApply(compiler) {
 	this.log('apply');
 
-	compiler.plugin('emit', (compilation, callback) => {
-		this.log('emit', JSON.stringify({ canRun: this.canRun, lastRun: this.lastRun, lastWatch: this.lastWatch }));
+	// adapt cmf settings result to output to /settings.json by default
+	let outputPath = compiler.options.output.path;
+	if (compiler.options.devServer && compiler.options.devServer.outputPath) {
+		outputPath = compiler.options.devServer.outputPath;
+	}
```
nicomaligne: IMHO if we required lodash we can do something a little more sexy, what do you think ?
const outpath = get(compiler.options, 'devServer.outputPath', compiler.options.output.path);

PR #1351, opened 2018-05-17
fix(cmf): set default header on http saga

jare-talend:
I define a default header on the parent app with for example an header with the accept language.
With the "create" way, I need to  define my header all on sub-dependencies which uses saga/http.

```javascript
//project parent
import http from '@talend/react-cmf/lib/sagas/http';
const httpClient = http.create({headers: {"Accept-Language": 'fr'}};
httpClient.get('/foo') // call /foo with the header Accept-Language at fr

// on a dependencie which use cmf
import http from '@talend/react-cmf/lib/sagas/http';
const httpClient = http.create();

httpClient.get('/foo') // call /foo whithout the headers defined in parent project.
```

Add a method to define a defaultConfig
Apply it when you fetch the request
Apply the CRSF token when you fetch the request

```javascript
//project parent
import http from '@talend/react-cmf/lib/sagas/http';
const httpClient = http.create({headers: {"Accept-Language": 'fr'}};
httpClient.get('/foo') // call /foo with the header Accept-Language at fr

// on a dependencie which use cmf
import http from '@talend/react-cmf/lib/sagas/http';
const instanceHTTP = http.create(); 
httpClient.get('/foo') instanceHTTP.get('/foo') // now  call /foo with the header Accept-Language at fr 
http.get('/foo') // now call also with the header Accept-Language at fr 
```

Keep the compatibility with "create" way


packages/cmf/src/sagas/index.md
```diff
@@ -55,9 +55,14 @@ const { data, response } = yield call(configuredHttp.get, `${API['dataset-sample
 importing the saga, allow you to statically call any member function `get, post ...` but also `create` which return an object with the exact same API.
 
 `create` also allow you to provide a configuration object.
+`setDefaultConfig` also allow you to provide a default header object which 'll be use at each http call.
```
jsomsanith: `which 'll` --> `which will`

packages/cmf/src/sagas/http.js
```diff
@@ -11,6 +10,12 @@ import {
 	testHTTPCode,
 } from '../middlewares/http/constants';
 
+let defaultConfig = {};
+
+export function handleCSRFToken(config) {
+	return mergeCSRFToken(config)(config);
```
jsomsanith: This feels so weird :P
So if I'm not wrong you do this because your `CSRFTokenHeaderKey` is in your http config right ?

packages/cmf/src/sagas/http.js
```diff
@@ -213,33 +223,52 @@ export function* httpGet(url, config, options) {
...
 export default {
 	delete: httpDelete,
 	get: httpGet,
 	post: httpPost,
 	put: httpPut,
 	patch: httpPatch,
-	create(defaultConfig = {}) {
-		const configEnhancer = handleDefaultConfiguration(defaultConfig);
+	create(createConfig = {}) {
+		setDefaultConfig(createConfig);
```
jsomsanith: I don't think we should override the default config here.
You exposed and documented `setDefaultConfig()` which is explicit, so if we want to set default config, we should call it. Overriding it here will surely introduce unwanted behavior, where we want to pass custom config in `create()` but end up overriding config for all fetches.

jare-talend: if I don't set setDefaultConfig, i have a breaking changes with data-streams. They use create with a default config.

packages/cmf/__tests__/sagas/http.test.js
```diff
@@ -509,410 +687,201 @@ describe('Http{Method} calls httpFetch with appropriate method', () => {
...
+	// 				httpPatch,
+	// 				url,
+	// 				payload,
+	// 				{
+	// 					'Accept-Language': 'fr',
+	// 				},
+	// 				options,
+	// 			),
+	// 		);
+	// 	});
```
talend-glorieux: That commented code should be removed if not used.

packages/cmf/src/sagas/http.js
```diff
@@ -213,18 +235,64 @@ export function* httpGet(url, config, options) {
...
+
 export default {
 	delete: httpDelete,
 	get: httpGet,
 	post: httpPost,
 	put: httpPut,
 	patch: httpPatch,
-	create(defaultConfig = {}) {
-		const configEnhancer = handleDefaultConfiguration(defaultConfig);
+	create(createConfig = {}) {
```
talend-glorieux: `create` should always return a new instance of `http`. If not how I'm I supposed to get a new instance without the default configuration. It also look like the default configuration can be set from multiple point which can be misleading.

packages/cmf/__tests__/sagas/http.test.js
```diff
@@ -509,410 +687,201 @@ describe('Http{Method} calls httpFetch with appropriate method', () => {
...
-			credentials: 'overloaded non nested config',
-			newConfig: 'newConfig element',
-		};
-		const httpInstance = http.create();
-		// when
-		const gen = httpInstance.put(url, payload, config);
-		// then
-		expect(gen.next().value).toEqual(call(httpPut, url, payload, expectedConfig, {}));
-	});
+	// 	it('check that defaultConfig is defined with headers', () => {
```
jmfrancois: please remove or fix this test

packages/cmf/src/sagas/http.js
```diff
@@ -213,18 +235,64 @@ export function* httpGet(url, config, options) {
...
+ * setDefaultHeader({headers: {
+ *  'Accept-Language': preferredLanguage,
+ * }});
+ */
+export function setDefaultConfig(config) {
+	if (defaultConfig) {
+		// eslint-disable-next-line no-console
+		console.error(
+			'ERROR: setDefaultConfig should not be called twice, if you wish to change the language use setDefaultLanguage api.',
+		);
```
jmfrancois: should we throw an error ?

packages/cmf/src/sagas/http.js
jare-talend: It is wanted by @jsomsanith  and @jmfrancois. So, if you want an new implementation, talk to you & let's free to change it. Now, I wish to merge this f****** PR.

jare-talend: I don't have other solution that use a default configuration if a want use the same headers in many dependencies. So, what do you propose other else ?

packages/cmf/src/sagas/http.js
```diff
@@ -213,18 +235,68 @@ export function* httpGet(url, config, options) {
...
+	}
+
+	defaultConfig = config;
+
+	if (throwError) {
+		throw new Error(
+			'ERROR: setDefaultConfig should not be called twice, if you wish to change the language use setDefaultLanguage api.',
+		);
+	}
+}
```
talend-glorieux: Should be a single `if` statement.
```js
if (defaultConfig) {
    throw new Error(
	'ERROR: setDefaultConfig should not be called twice, if you wish to change the language use setDefaultLanguage api.')
}
	defaultConfig = config;
}
```

packages/cmf/src/sagas/http.js
```diff
@@ -213,18 +235,68 @@ export function* httpGet(url, config, options) {
...
+		throwError = true;
+	}
+
+	defaultConfig = config;
+
+	if (throwError) {
+		throw new Error(
+			'ERROR: setDefaultConfig should not be called twice, if you wish to change the language use setDefaultLanguage api.',
+		);
+	}
```
jmfrancois: Please start with the if default config throw.
You can create an accessor to defaultConfig for testing purpose.

packages/cmf/src/sagas/http.js
```diff
@@ -213,18 +235,68 @@ export function* httpGet(url, config, options) {
...
+ * setDefaultHeader - define a default config to use with the saga http
+ * this default config is stored in this module for the whole application
+ *
+ * @param  {object} config key/value of header to apply
+ * @example
+ * import { setDefaultHeader } from '@talend/react-cmf/sagas/http';
+ * setDefaultHeader({headers: {
+ *  'Accept-Language': preferredLanguage,
+ * }});
+ */
```
talend-glorieux: Please update JSDoc.

PR #1356, opened 2018-05-18
feat(Component/Breadcrumb): improve a11y

jsomsanith:
1. no landmark
2. first items dropdown trigger has no explicit alternative text (only the ellipsis)
3. items links need a more explicit text than its content

1. adding a navigation role is not compatible with our theme. Still no landmark but wrap component with a `<nav>` providing navigation info
2. add an aria-label (i18n)
3. set the title (tooltip) content as aria-label


packages/components/src/Breadcrumbs/Breadcrumbs.component.js
```diff
@@ -71,6 +74,7 @@ function Breadcrumbs(props) {
 						bsStyle="link"
 						role="link"
 						title={title || text}
+						aria-label={title}
```
frassinier: Why adding an aria-label if a label and a title are already provided? 
Which information is added for screen readers?

packages/components/src/Breadcrumbs/Breadcrumbs.component.js
```diff
@@ -109,15 +122,20 @@ function Breadcrumbs(props) {
 	}
 
 	return (
-		<ol id={props.id} className={classNames('breadcrumb', theme['tc-breadcrumb'], 'tc-breadcrumb')}>
+		<ol
+			id={props.id}
+			className={classNames('breadcrumb', theme['tc-breadcrumb'], 'tc-breadcrumb')}
+			role="navigation"
```
frassinier: ⚠️ Be careful with [that](https://github.com/Talend/ui/blob/master/packages/theme/src/theme/variations/_tdp.scss#L12)

packages/components/src/Breadcrumbs/Breadcrumbs.component.js
jsomsanith: I assume that 
* `text` is the label, so the content of the button
* `title` the tooltip

The problem is that the tooltip is more accurate, providing more info than the content.
But `title` is less important than `content` in a11y text order. So screen readers will take the content over the title, so the less accurate/precise text.

So when you have `title` and `content`, we add an `aria-label` with the `title` value to be read by SR over the content.

packages/components/src/Breadcrumbs/Breadcrumbs.component.js
jsomsanith: Oh thank you, I'll remove it on Breadcrumb

packages/components/src/Breadcrumbs/Breadcrumbs.component.js
frassinier: Ok got it thx

packages/components/src/Breadcrumbs/Breadcrumbs.component.js
frassinier: Here is a good [resource](https://www.w3.org/TR/2017/NOTE-wai-aria-practices-1.1-20171214/examples/breadcrumb/index.html)

PR #1357, opened 2018-05-18
feat(Component/loaders): improve a11y

jsomsanith:
Missing alternative texts in loaders

Add them.

### DIFF
--- a/packages/components/src/CircularProgress/CircularProgress.component.js
+++ b/packages/components/src/CircularProgress/CircularProgress.component.js
@@ -1,9 +1,11 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 import classNames from 'classnames';
-import { CIRCULAR_PROGRESS_SIZE as SIZE } from '../constants';
+import { translate } from 'react-i18next';
 
+import I18N_DOMAIN_COMPONENTS, { CIRCULAR_PROGRESS_SIZE as SIZE } from '../constants';
 import theme from './CircularProgress.scss';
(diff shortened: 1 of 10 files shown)


packages/components/src/CircularProgress/CircularProgress.component.js
```diff
@@ -38,8 +40,19 @@ function CircularProgress({ size, light, percent, className }) {
...
 
+	const percentLabel = percent && `${percent}%`;
 	return (
-		<svg focusable="false" className={classes} viewBox={`0 0 ${DIAMETER} ${DIAMETER}`}>
+		<svg
+			focusable="false"
+			className={classes}
+			viewBox={`0 0 ${DIAMETER} ${DIAMETER}`}
+			aria-busy="true"
+			aria-live="polite"
```
frassinier: Why do you force it to be read again for each modification?

jsomsanith: This is not something like errors where we need to break the user flow with an important info.
Here a progress can take some time and be updated `very` regularly.

While the "thing" is loading, if the rest of the application is not blocked, user will be interrupted frequently in his actions with the percentage info, which is disturbing.

In the case where the rest of the application is blocked, it's ok, the user will be idle, and the info will be read at each update.

jsomsanith: Ok, after a little chat, I didn't understand the first message.
`aria-live` removed, thanks

PR #1371, opened 2018-05-25
feat(cmf): add two state static accessors

jmfrancois:

Every time we need to modify a state from the outside (saga, actionCreator) we need to deal with a low level api (actions.component.mergeState).
That means the dev need to know and hard code the name of the component and this is not easy, error prone and not futur proof (add HOC case for example which change the name).

Expose two static functions:

* function getState(reduxState, id='default')
* function setStateAction(componentState, id='default', type='ComponentName.setState')


acateland:
@jmfrancois what happen if i have multiple instance of this component ? how to deal with such situation ?

jmfrancois:
I have updated the doc to show you how

packages/cmf/src/cmfConnect.md
```diff
@@ -221,6 +221,32 @@ If you want to render some component conditionally, just pass "renderIf" prop (t
 You can also use Expression for this and customize this prop like "renderIfExpression" in
 CMF json configuration files
 
+## How to access state and mute from the outside
```
yyanwang: hi, by 'mute' do you mean 'mutate'?

output/cmf.eslint.txt
```diff
@@ -26,5 +32,5 @@ The react/require-extension rule is deprecated. Please use the import/extensions
 /home/travis/build/Talend/ui/packages/cmf/src/sagas/collection.js
   10:1  error  Prefer default export  import/prefer-default-export
 
-✖ 12 problems (9 errors, 3 warnings)
+✖ 18 problems (15 errors, 3 warnings)
```
nicomaligne: Hmm some cleaning maybe ;)

packages/cmf/src/cmfConnect.md
acateland: i think that is what he mean, but it is also misleading since the state is not mutated but updated with a non mutable operation.
`## Read and update component state from outside`

jmfrancois: 👍

packages/cmf/src/cmfConnect.js
```diff
@@ -245,6 +246,21 @@ export default function cmfConnect({
 				router: PropTypes.object,
 			};
 			static WrappedComponent = WrappedComponent;
+			static getState = function getState(state, id = 'default') {
+				return state.cmf.components.getIn([getComponentName(WrappedComponent), id], defaultState);
+			};
+			static setStateAction = function setStateAction(state, id = 'default', type) {
```
acateland: state parameter to be renamed newState i think
also shouldn't those state update operation also support synchronous state update function ? for safe state update operation when relying on previous state ?

packages/cmf/src/cmfConnect.md
```diff
@@ -221,6 +221,32 @@ If you want to render some component conditionally, just pass "renderIf" prop (t
 You can also use Expression for this and customize this prop like "renderIfExpression" in
 CMF json configuration files
 
+## How to access state and mutate from the outside
```
acateland: title may be misleading, the state is not mutated in sense of pointer value mutation, but it is updated
to avoid confusion i would rename this title to
`## Read and update component state from outside`

packages/cmf/src/cmfConnect.js
jmfrancois: newState already exists in the upper declaration
I have added the asyn state update support

PR #1375, opened 2018-05-25
fix(forms): language prop can't be a function

lmaillet:
`props.language` is tested to be a function.
Can't use a function in `Object.assign`.

Remove the `typeof function` test.


acateland:
@lmaillet didn't check the rest of the code but maybe the `language` props was indeed intended to be a function or an object and be assigned to the language variable ?

lmaillet:
@acateland 
You mean that language could be a function **OR** an object ?
Something like that:
```
if (typeof props.language === 'function') {
	Object.assign(language, props.language());
} else if (props.language !== undefined) {
	Object.assign(language, ...props.language);
}
```

packages/forms/src/UIForm/UIForm.component.js
```diff
@@ -36,8 +36,8 @@ export class UIFormComponent extends React.Component {
 		this.onActionClick = this.onActionClick.bind(this);
 		// control the tv4 language here.
 		const language = getLanguage(props.t);
-		if (typeof props.language === 'function') {
-			Object.assign(language, props.language);
+		if (props.language !== undefined) {
```
ncomont: you should use `!= null` to check both `undefined`and `null`values.

packages/forms/src/UIForm/UIForm.component.js
```diff
@@ -36,8 +36,8 @@ export class UIFormComponent extends React.Component {
 		this.onActionClick = this.onActionClick.bind(this);
 		// control the tv4 language here.
 		const language = getLanguage(props.t);
-		if (typeof props.language === 'function') {
-			Object.assign(language, props.language);
+		if (props.language != null) {
+			Object.assign(language, ...props.language);
```
ncomont: Just to be sure... Are you sure you really want to spread `props.language`, here ?

Considering : 

```
const a = {aa: 11, bb: 22}
const b = {cc: 33, bb: 99}
```

Then : 

```
Object.assign({}, a, b)
{aa: 11, bb: 99, cc: 33}
```

Whereas :

```
Object.assign({}, a, ...b)
VM9957:1 Uncaught TypeError: undefined is not a function
```

ncomont: (if I'm right, I think that a test is missing 😄)

PR #1380, opened 2018-05-28
chores(doc): change proposal for ActionDropdown

acateland:
Using for the first time the component and using the doc is stumbled on some comprehension issues
update the doc to take into account this issue, removing the less interesting part and making some other more concise and clear.


jmfrancois:
Nice ! I would add a link to the react-bootstrap dropdown documentation

packages/components/src/Actions/ActionDropdown/ActionDropdown.component.md
```diff
@@ -1,176 +1,115 @@
...
+| divider       | Bool                                                                                       |
+| withMenuItem  | Bool                                                                                       |
+| menuItemProps | [react-bootstrap](https://react-bootstrap.github.io/components/dropdowns/#menu-item-props) |
+| liProps       | Object                                                                                     |
 
-The ```<li>``` and ```<MenuItem>``` can be deeply customized with menuItemProps and liProps.
+Your component will be wrapped into a `MenuItem` if `withMenuItem` property is set to true
+This will give you access to customisation of [react-bootstrap](https://react-bootstrap.github.io/components/dropdowns/#menu-items) `MenuItem` component
+
+The `<li>` and `<MenuItem>` can be deeply customized with menuItemProps and liProps.
```
nicomaligne: With the LI we are loosing some stuff from react bootstrap like the accessibility with the keyboard shortcut for example.
If we decide to keep it like that, you can maybe precise it in the documentation.

nicomaligne:
I will use this doc as a reference for my next one, thanx Axcel ;)

packages/components/src/Actions/ActionDropdown/ActionDropdown.component.md
```diff
@@ -1,176 +1,115 @@
...
+| ----- | ---------------------------------- |
+| items | Required (array OR Immutable.List) |
+
+each composed item being an object composed of
+
+| Properties | type            |
+| ---------- | --------------- |
+| label      | Required String |
+| hidelabel  | Bool            |
+| icons      | String          |
```
jmfrancois: I don't see the divider case, so label is not always required

packages/components/src/Actions/ActionDropdown/ActionDropdown.component.md
```diff
@@ -1,176 +1,115 @@
...
-	beforeItemsDropdown: [] or {},
-	itemsDropdown: [] or {},
-	afterItemsDropdown: [] or {},
-}
-``` 
-They are 3 'placeholders' in the component to inject new stuff.
-'itemsDropdown' is the classic content list of the dropdown, equivalent to where the items props will be rendered.
-And you have a 'beforeItemsDropdown' and 'afterItemsDropdown' key, that will be rendered before or after the itemsDropdown in the list, if you need specific stuff that will give you more flexibility.
 
+Using this you can defined additionnal component that your `ActionDropdown` will show.
```
jmfrancois: Using this **props**
I would add a link to Inject documentation here

packages/components/src/Actions/ActionDropdown/ActionDropdown.component.md
```diff
@@ -185,34 +124,35 @@ components: {
 	],
 }
 ```
+
 #### 4 ) overlay
+
 You can have a overlay / tooltip on the ActionDropdown.
```
jmfrancois: either a link or an example would be nice here.

packages/components/src/Actions/ActionDropdown/ActionDropdown.component.md
```diff
@@ -185,34 +124,35 @@ components: {
...
 
 #### 5 ) select callback
+
 With the onSelect props you can add a callback when an item is selected.
```
jmfrancois: same here

packages/components/src/Actions/ActionDropdown/ActionDropdown.component.md
```diff
@@ -1,176 +1,115 @@
-
-
 # ACTION DROPDOWN COMPONENT
```
jmfrancois: Maybe the real name would be better: ActionDropdown component

PR #1403, opened 2018-05-30
fix(Component): switch to dropup when it's short for dropdown

jsomsanith:
For dropdown in restricted divs, it's possible to have the dropdown menu cropped if it is around the bottom of the div

Make it be a dropup when it's too close from the bottom

### DIFF
--- a/packages/components/src/Actions/ActionDropdown/ActionDropdown.component.js
+++ b/packages/components/src/Actions/ActionDropdown/ActionDropdown.component.js
@@ -2,29 +2,29 @@ import PropTypes from 'prop-types';
 import ImmutablePropTypes from 'react-immutable-proptypes';
 import React from 'react';
 import classNames from 'classnames';
-import { Iterable } from 'immutable';
-import { DropdownButton, MenuItem, OverlayTrigger } from 'react-bootstrap';
+import {Iterable} from 'immutable';
+import {DropdownButton, MenuItem, OverlayTrigger} from 'react-bootstrap';
 import Inject from '../../Inject';
 import theme from './ActionDropdown.scss';
 import TooltipTrigger from '../../TooltipTrigger';
 import Icon from '../../Icon';
-import { wrapOnClick } from '../Action/Action.component';
+import {wrapOnClick} from '../Action/Action.component';
 
 function InjectDropdownMenuItem({
-	getComponent,
-	component,
-	divider,
-	withMenuItem,
-	liProps,
-	menuItemProps,
-	key,
-	onSelect,
-	onKeyDown,
-	...rest
-}) {
-	const Renderers = Inject.getAll(getComponent, { MenuItem });
+									getComponent,
+									component,
+									divider,
+									withMenuItem,
+									liProps,
+									menuItemProps,
+									key,
+									onSelect,
+									onKeyDown,
+									...rest
+								}) {
+	const Renderers = Inject.getAll(getComponent, {MenuItem});
 	if (divider) {
-		return <Renderers.MenuItem key={key} {...menuItemProps} divider />;
+		return <Renderers.MenuItem key={key} {...menuItemProps} divider/>;
 	}
 	if (withMenuItem) {
 		return (
@@ -54,9 +54,9 @@ InjectDropdownMenuItem.propTypes = {
 InjectDropdownMenuItem.displayname = 'InjectDropdownMenuItem';
 
 function renderMutableMenuItem(item, index, getComponent) {
-	const Renderers = Inject.getAll(getComponent, { MenuItem });
+	const Renderers = Inject.getAll(getComponent, {MenuItem});
 	if (item.divider) {
-		return <Renderers.MenuItem key={index} divider />;
+		return <Renderers.MenuItem key={index} divider/>;
 	}
 	return (
 		<Renderers.MenuItem
@@ -67,7 +67,7 @@ function renderMutableMenuItem(item, index, getComponent) {
 			title={item.title || item.label}
 			className={classNames(theme['tc-dropdown-item'], 'tc-dropdown-item')}
 		>
-			{item.icon && <Icon name={item.icon} />}
+			{item.icon && <Icon name={item.icon}/>}
 			{!item.hideLabel && item.label}
 		</Renderers.MenuItem>
 	);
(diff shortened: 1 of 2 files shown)


output/components.eslint.txt
```diff
@@ -6,6 +6,9 @@ The react/require-extension rule is deprecated. Please use the import/extensions
   1:1  error  '@storybook/react' should be listed in the project's dependencies, not devDependencies  import/no-extraneous-dependencies
   4:2  error  Unexpected require()                                                                    global-require
 
+/home/travis/build/Talend/ui/packages/components/src/Actions/ActionDropdown/ActionDropdown.component.js
+  186:5  error  Using string literals in ref attributes is deprecated  react/no-string-refs
```
smouillour: not good

packages/components/src/Actions/ActionDropdown/ActionDropdown.component.js
```diff
@@ -108,67 +108,100 @@ function getMenuItem(item, index, getComponent) {
...
+		const dropdown = (
+			<Renderers.DropdownButton
+				title={title}
+				bsStyle={style}
+				role="button"
+				onSelect={onItemSelect}
+				className={classNames(theme['tc-dropdown-button'], 'tc-dropdown-button', className)}
+				aria-label={tooltipLabel || label}
+				{...rest}
+				ref="trigger"
```
jare-talend: it is used for ?

packages/components/src/VirtualizedList/ListTable/ListTable.component.js
```diff
@@ -54,7 +54,7 @@ function ListTable(props) {
 	return (
 		<VirtualizedTable
 			className={`tc-list-table ${theme['tc-list-table']}`}
-			gridClassName={theme.grid}
+			gridClassName={`${theme.grid} tc-dropdown-container`}
```
jare-talend: use a constant pour target this class tc-dropdown-container and reuse it in https://github.com/Talend/ui/pull/1403/files#diff-f131276a4f6398606943e59f5ea4a345R127

packages/components/src/Actions/ActionDropdown/ActionDropdown.component.js
```diff
@@ -108,67 +108,100 @@ function getMenuItem(item, index, getComponent) {
...
+	}
+
+	onToggle(isOpen, event) {
+		if (!isOpen) {
+			return;
+		}
+		const dropdownTrigger = event.target;
+		const dropdownMenu = event.target.nextSibling;
+		let dropdownContainer = dropdownTrigger.parentElement;
+		while (dropdownContainer && !dropdownContainer.classList.contains('tc-dropdown-container')) {
```
jare-talend: it is specific for the list but if i want use the ActionDropdown in a other context, i can't use it

packages/components/src/Actions/ActionDropdown/ActionDropdown.component.js
jsomsanith: Not used, removed :)

packages/components/src/Actions/ActionDropdown/ActionDropdown.component.js
jare-talend: for example in the ActionDropDown, if i want use this feature to avoid to collapse below the viewport, i have to previously add a class tc-dropdown-container on the element which wraps my ActionDropdown. It is not automatically :

So, if it is this case, can you add an example on this story: http://1403.talend.surge.sh/components/?selectedKind=ActionDropdown&selectedStory=default&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel


image

packages/components/src/Actions/ActionDropdown/ActionDropdown.component.js
```diff
@@ -81,6 +83,26 @@ function getMenuItem(item, index, getComponent) {
...
+	return dropdownTrigger;
+}
+
+function getDropdownContainer(dropdownElement) {
+	let dropdownContainer = dropdownElement;
+	do {
+		dropdownContainer = dropdownContainer.parentElement;
+	} while (
+		dropdownContainer &&
+		dropdownContainer.tagName !== 'BODY' &&
```
jare-talend: I have a doubt if we should check with dropdownContainer.tagName.toUpperCase().

jsomsanith: No need to. From MDN : 

capture d ecran 2018-06-07 a 10 49 55

packages/components/src/Actions/ActionDropdown/ActionDropdown.component.js
```diff
@@ -108,74 +130,111 @@ function getMenuItem(item, index, getComponent) {
...
-			onSelect(event, object);
+		const dropdownTrigger = getDropdownToggleFromInner(event.target);
+		const dropdownMenu = dropdownTrigger.nextSibling;
+		const dropdownContainer = getDropdownContainer(dropdownTrigger);
+
+		if (dropdownContainer) {
+			const dropdownRect = dropdownMenu.getBoundingClientRect();
+			const containerRect = dropdownContainer.getBoundingClientRect();
+
+			let dropup = this.state.dropup;
```
acateland: this state update should be done as a function since it rely on previous dropup state

packages/components/src/Actions/ActionDropdown/ActionDropdown.component.js
```diff
@@ -108,74 +130,111 @@ function getMenuItem(item, index, getComponent) {
...
 
-	function onItemSelect(object, event) {
-		if (onSelect) {
-			onSelect(event, object);
+		const dropdownTrigger = getDropdownToggleFromInner(event.target);
+		const dropdownMenu = dropdownTrigger.nextSibling;
+		const dropdownContainer = getDropdownContainer(dropdownTrigger);
+
+		if (dropdownContainer) {
+			const dropdownRect = dropdownMenu.getBoundingClientRect();
```
acateland: just as a note the may cause layout reflow, we may not have a choice however
https://gist.github.com/paulirish/5d52fb081b3570c81e3a#box-metrics

jsomsanith: Yes it is known but we don't really have a choice, we need to get the positions of the dom element

packages/components/src/Actions/ActionDropdown/ActionDropdown.component.js
```diff
@@ -108,74 +130,111 @@ function getMenuItem(item, index, getComponent) {
...
+			const containerRect = dropdownContainer.getBoundingClientRect();
+
+			let dropup = this.state.dropup;
+			if (!this.state.dropup && dropdownRect.bottom > containerRect.bottom) {
+				dropup = true;
+			} else if (this.state.dropup && dropdownRect.top < containerRect.top) {
+				dropup = false;
+			}
+			this.setState({ dropup });
```
sgendre: maybe one useless setState can be avoided when the value of dropup is not modified by if statements.

jsomsanith: Changed the implementation to use `setState(fn)` to rely on old state.
This is the good way to do it so we don't have a choice to call setState everytime. But in React 16 returning null don't trigger a new render.

PR #1406, opened 2018-05-31
feat(containers/AppLoader): add takeAction support

romainseb:
Currently, we can't wait for some action to be dispatched between two steps
Add a way to do that with the takeAction

### DIFF
--- a/packages/containers/src/AppLoader/AppLoader.saga.js
+++ b/packages/containers/src/AppLoader/AppLoader.saga.js
@@ -22,15 +22,6 @@ export function* waitFor(collectionName, interval = 10) {
 		yield call(delay, interval);
 	}
 }
-
-/**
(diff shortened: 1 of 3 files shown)


packages/containers/src/AppLoader/AppLoader.saga.js
```diff
@@ -33,6 +42,8 @@ export function* handleStep(step) {
 		);
 	} else if (step[WAIT_FOR]) {
 		return yield all(step[WAIT_FOR].map(collectionName => call(waitFor, collectionName)));
+	} else if (step[TAKE_ACTION]) {
+		return yield all(step[TAKE_ACTION].map(actionName => call(takeAction, actionName)));
```
jmfrancois: you can't use take here directly ?
call(take, actionName)

PR #1408, opened 2018-05-31
chore(TUI-142/cmf): remove deprecated getCollection

jmfrancois:

cmf provide a long time deprecated function : getCollection.
This means in the render a component may access to a collection by id and so rely on a part of the state. But it will not be re-render if that collection change. This is bad. We should always use mapStateToProps or expression for that.

Remove it from CMF.

Before

```javascript
function MyComponent(props) {
  return <ul>{props.getCollection('foo').map((item, key) => <li key={key}>item.get('label')</li>}</ul>
}
export default cmfConnect({})(MyComponent);
```

After

```javascript
function MyComponent(props) {
  return <ul>{props.foo.map((item, key) => <li key={key}>item.get('label')</li>}</ul>
}
function mapStateToProps(state) {
  return { foo: state.cmf.collections.get('foo') };
}
export default cmfConnect({ mapStateToProps })(MyComponent);
```


output/cmf.eslint.txt
```diff
@@ -2,6 +2,9 @@
 
 The react/require-extension rule is deprecated. Please use the import/extensions rule from eslint-plugin-import instead.
 
+/home/travis/build/Talend/ui/packages/cmf/src/cmfConnect.js
+  43:8  error  'deprecated' is defined but never used  no-unused-vars
```
ncomont: please fix this error

PR #1414, opened 2018-06-01
fix(component/MultiSelectTag): support titleMap

andrii-holovko-dev:
titleMap option doesn't affect labels inside Core fields / Core-multiSelectTag dropdown

- pass array of objects as Typeahead items prop instead of array of strings
- extract value inside onAddTag function
- update stories-core
- update unit tests


packages/forms/stories-core/json/fields/core-multiSelectTag.json
```diff
@@ -29,16 +29,38 @@
       "description": "This field accepts new items",
       "restricted": false,
       "title": "Tags",
-      "widget": "multiSelectTag"
-    },
-    {
-      "key": "restrictedtags",
```
jmfrancois: please do not remove this show case

PR #1419, opened 2018-06-04
feat(components): Drawer - manage footer according to the tab selected

smouillour:
Currently : 
- the Drawer component permit to have tabs in it but there is no possibility to define specific actions in the footer depending of the tab selected.
- The drawer add cross action in the header when the onCancelAction is defined but without the possibility to choose between the footer action button and the cross action in the header.

- Add a props in the Drawer to share the information with the tabBar of the selected tab.
- Add the attribute "position" to define the position of the cancelAction (footer or header). Default is footer

See the change in action : 

image


packages/components/src/Drawer/Drawer.component.js
```diff
@@ -159,15 +164,33 @@ DrawerFooter.propTypes = {
...
+			...get(enhancedFooterActions, 'actions.center', []),
+			...get(activeTabItem, 'actions.center', []),
+		],
+		right: [
+			...get(enhancedFooterActions, 'actions.right', []),
+			...get(activeTabItem, 'actions.right', []),
+		],
 	};
+
+	if (onCancelItem) {
```
ncomont: Is it normal that if I add a `onCancelAction` whereas some `actions.left` are already defined, the `cancel` one sticks on the right of the actuals `actions.left` ?

image

packages/components/src/Drawer/Drawer.component.js
```diff
@@ -11,6 +11,8 @@ import Inject from '../Inject';
 import theme from './Drawer.scss';
 
 const DEFAULT_TRANSITION_DURATION = 350;
+const ON_CANCEL_ACTION_POSITION_HEADER = 'header';
```
ncomont: Could these `consts` be exported ?

packages/components/src/Drawer/Drawer.component.js
```diff
@@ -182,12 +205,27 @@ function Drawer({
...
 	const TabBarComponent = Inject.get(getComponent, 'TabBar', TabBar);
+
+	let activeTab = {};
+	let activeTabItem = [];
+	let customTabs;
+	if (tabs && tabs.items.length > 0) {
+		customTabs = Object.assign({}, tabs);
+
+		if (selectedTabKey) {
+			Object.assign(customTabs, { selectedKey: selectedTabKey });
```
ncomont: Why didn't you do something like `customTabs.selectedKey = selectedTabKey;` ?

packages/components/src/Drawer/Drawer.component.js
```diff
@@ -182,12 +205,27 @@ function Drawer({
...
+			activeTab = tabs.items.find(tab => tab.key === selectedTabKey);
+		}
+		activeTabItem = get(activeTab, 'footerActions', {});
+	}
```
ncomont: To be honest, I don't really understand why this huge if instead of passing undefined values to functions that already has default parameters. For example, I think that something like that could do the same job with less code and complexity. What do you think ? : 

```
	const activeTab = tabs.items.find(tab => tab.key === selectedTabKey);
	const activeTabItem = get(activeTab, 'footerActions', {});

	[...]

	<TabBarComponent
		{...tabs}
		selectedKey=selectedTabKey
		className={classnames('tc-drawer-tabs', theme['tc-drawer-tabs'])}
	/>

	[...]

	<ActionBar
		{...combinedFooterActions(onCancelAction, footerActions, activeTabItem)}
		className={classnames('tc-drawer-actionbar', theme['tc-drawer-actionbar'])}
	/>
```

packages/components/src/Drawer/Drawer.component.js
```diff
@@ -159,15 +164,33 @@ DrawerFooter.propTypes = {
...
+			...get(activeTabItem, 'actions.left', []),
+		],
+		center: [
+			...get(enhancedFooterActions, 'actions.center', []),
+			...get(activeTabItem, 'actions.center', []),
+		],
+		right: [
+			...get(enhancedFooterActions, 'actions.right', []),
+			...get(activeTabItem, 'actions.right', []),
+		],
```
ncomont: Could it be achieved by looping on `actions` keys ?

packages/components/src/Drawer/Drawer.component.js
```diff
@@ -159,15 +164,33 @@ DrawerFooter.propTypes = {
...
-export function combinedFooterActions(onCancelAction, footerActions) {
-	if (!onCancelAction) {
-		return footerActions;
-	}
+export function combinedFooterActions(onCancelAction, footerActions, activeTabItem = {}) {
+	const onCancelItem =
+		(onCancelAction && onCancelAction.position === ON_CANCEL_ACTION_POSITION_FOOTER) ||
+		(onCancelAction && !onCancelAction.position)
+			? onCancelAction
+			: null;
```
ncomont: Pretty hard to read and debug.

packages/components/src/Drawer/Drawer.component.js
smouillour: i think it is because there are lot of actions and the space dedicated for each button is not sufficient.
Normally, in each part, there is a space between actions but this space desappears when there are lot of action

packages/components/src/Drawer/Drawer.component.js
smouillour: ok

packages/components/src/Drawer/Drawer.component.js
smouillour: i changed it

packages/components/src/Drawer/Drawer.test.js
```diff
@@ -208,6 +285,8 @@ describe('Drawer', () => {
 						actionId: 'drawer:closeDrawer',
 					},
 				],
+				center: [],
+				right: [],
```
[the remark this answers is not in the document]
smouillour: we manage all position for the tabs so we have a default value with an empty array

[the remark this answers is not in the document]
smouillour: this is default values because we manage all position for the tabs

PR #1427, opened 2018-06-06
fix(cmf/sagas): http response could not have headers

frassinier:
If we get a response without headers it will fails

Just check if http response has headers before dealing with them.


packages/cmf/src/sagas/http.js
```diff
@@ -49,11 +49,14 @@ export class HTTPError extends Error {
 export function handleBody(response) {
 	let methodBody = 'text';
 
-	const contentType = response.headers.get('Content-Type');
+	if (response.headers) {
```
sgendre: get content-type with lodash get, provide a default value and avoid the unnecessary if statements

ncomont:
@sgendre 

A _simple_ lodash `get` implies :

- lodash/get.js
2 ternary
1 call to `baseGet` function

- lodash/.internal/baseGet.js
1 call to `castPath` function
1 while loop based on the path's length
n calls to the `toKey` function
N ternaries

- lodash/.internal/castPath.js
1 call to `Array.isArray`
1 call to `isKey` function
1 call to the `stringToPath` function
1 ternary


- lodash/.internal/toKey.js
1 call to the `isSymbol` function
n ternaries
n arithmetical operations

etc, etc, etc.

I could continue again and again.

Why should we use complex libraries when our needs are so simple that a regular if could do the job ?

acateland:
@ncomont the idea is to use the same pattern around access to unsafe value.

sgendre:
the point is to avoid 'undefined is not a function' and also having a lot of if statements all along

my 2 cents
``export function handleBody(response) {
	if (get(response, 'headers', {}).get('Content-Type', []).includes('application/json')) {
		return response['json']().then(data => ({ data, response }));
	}

	return response['text']().then(data => ({ data, response }));
}``

packages/cmf/src/sagas/http.js
```diff
@@ -2,6 +2,7 @@ import { call, put } from 'redux-saga/effects';
 import merge from 'lodash/merge';
 import get from 'lodash/get';
 import curry from 'lodash/curry';
+import { Headers } from 'node-fetch';
```
sgendre: it's frontend code, what is the point of a node library ?

PR #1433, opened 2018-06-08
feat(components): Filtered and Sorted Table

timbault:
The new component Table must offer filtering and sorting features.

Add a title bar on top of the table to display a title and a set of filters.
Sorting feature is handled in columns header.

demo_table_lot2


packages/components/src/Table/DataAccessor/DataAccessorWithSorterAndFilter.js
```diff
@@ -0,0 +1,455 @@
...
+	}
+
+	/**
+	 * Returns the number of elements.
+	 * @param {boolean} withFiltersAndSorter - if true then the result takes into account the filters.
+	 */
+	getSize(withFiltersAndSorter) {
+		if (withFiltersAndSorter) {
+			if (this.hasSorter()) {
+				return this.sorterInfo.result.length;
```
jmfrancois: use this.getSorter ?

packages/components/src/Table/Filters/BooleanFilter.js
```diff
@@ -0,0 +1,28 @@
+import Filter from './Filter';
+
+/**
+ * This is used to filter boolean values.
+ * Default behaviour is: true value is not filtered, false value is filtered.
+ * You can reverse this behaviour with the 'reverse' attribute set to true.
+ */
+export default class BooleanFilter extends Filter {
```
acateland: Please avoid using class hierarchies, use composition, a class for a filter is total overkill, and local state make it hard to test and maintain.

packages/components/src/Table/Filters/Filter.js
```diff
@@ -0,0 +1,61 @@
+/**
+ * This class represents a filter for the elements of a table.
+ * A filter is identified by a unique string identifier.
+ * A filter can be activated or deactivated.
+ */
+export default class Filter {
```
acateland: Please avoid using class hierarchies, use composition, a class for a filter is total overkill, and local state make it hard to test and maintain.

The only bit with value here being `select` instance method

packages/components/src/Table/Filters/RegexpFilter.js
```diff
@@ -0,0 +1,56 @@
...
+
+function lowerCaseMatch(value, regexp) {
+	return value.toLowerCase().match(regexp);
+}
+
+/**
+ * This is used to filter string values based on regular expression.
+ * Default behaviour is: if a string value matches the regexp, then it is not filtered.
+ */
+export default class RegexpFilter extends Filter {
```
acateland: Please avoid using class hierarchies, use composition, a class for a filter is total overkill, and local state make it hard to test and maintain.

packages/components/src/Table/Sorters/Sorter.js
```diff
@@ -0,0 +1,85 @@
...
+	none: 0,
+	ascending: 1,
+	descending: -1,
+};
+
+/**
+ * This class represents a sorter for the elements of a table.
+ * A sorter is identified by a unique string identifier.
+ */
+export default class Sorter {
```
acateland: you already have a global state (prefered way in redux application) a local state in react component, i don't see why we need to spread this state and mutate it in uncontrolled ways outside those, more complicated to test, more error prone, more complex to maintain, use pure function and apply application state (preferred) or component state to them.

packages/components/src/Table/DataAccessor/DataAccessorWithSorterAndFilter.js
```diff
@@ -0,0 +1,455 @@
...
+			mergedResult = mergedResult.concat(result2[i]);
+		}
+	}
+	return mergedResult;
+}
+
+/**
+ * Internal class used to compare elements with a sorter.
+ */
+class Comparator {
```
acateland: you don't need a class for this, pure function will do

packages/components/src/index.js
```diff
@@ -39,7 +39,16 @@ import ObjectViewer from './ObjectViewer';
...
+import {
+	DataAccessorWithSorterAndFilter,
+	FilterFactory,
+	Sorter,
+	SortOrder,
+	Table,
+	TableCell,
+	TableHeaderCell,
+	TableSortHeader,
+} from './Table';
```
jmfrancois: components index is supposed to only contains components, please remove all the things that are not components

packages/components/src/Table/Filters/Filter.js
```diff
@@ -0,0 +1,61 @@
...
+		return this.active;
+	}
+
+	/**
+	 * This methods indicates if an element of a list is visible or not.
+	 * When returns true, the element is visible (not filtered).
+	 * @param {object} dataAccessor - element data getter
+	 * @param {object} element - an element of the table
+	 */
+	select(dataAccessor, element) {
```
jmfrancois: I would go to an other wording here. like `match`

packages/components/.storybook/preview-head.html
```diff
@@ -1,143 +1,164 @@
...
-    }
-
-    [draggable='true'] .tc-table-clickable-cell {
-      position: absolute;
-      top: 2px;
-      left: 0;
-      bottom: 0;
-      width: 100vw;
-    }
-
```
jmfrancois: sorry my fault I have formatted the file

packages/components/src/Table/index.js
```diff
@@ -1,5 +1,15 @@
...
 import Table from './Table';
 import TableCell from './Cell/TableCell';
 import TableHeaderCell from './Header/TableHeaderCell';
+import TableSortHeader from './Header/TableSortHeader';
+import ToggleFilterComponent from './Filters/ToggleFilterComponent';
 
-export { Table, TableCell, TableHeaderCell };
+export {
+	StringFilterComponent,
+	Table,
```
frassinier: Only export Table here

packages/components/src/Table/Filters/StringFilterComponent.js
```diff
@@ -0,0 +1,51 @@
...
+export default class StringFilterComponent extends Component {
+	constructor(props) {
+		super(props);
+		this.state = { docked: true };
+		this.onToggle = this.onToggle.bind(this);
+		this.onFilter = this.onFilter.bind(this);
+	}
+
+	onToggle() {
+		this.setState(prevState => ({
```
frassinier: Components should not manage state, do you really need it?

packages/components/src/Table/index.js
timbault: but i need this components outside Table module.

packages/components/stories/Table.js
```diff
@@ -347,10 +420,257 @@ ConnectedTable.propTypes = {
...
+		match: matchName,
+		renderer: StringFilterComponent,
+		rendererProps: {
+			placeHolder: 'Filter...',
+			dockable: true,
+			navbar: true,
+		},
+		className: nameFilterId,
+	};
+}
```
jmfrancois: Those data look very similar to TestData.js
Can you import them instead ?

packages/components/src/Table/index.js
jmfrancois: you can attach them to Table into Table/index.js

```javascript
import Table ...
import StringFilterComponent ...
Table.StringFilterComponent = StringFilterComponent

export default Table;
```

packages/components/stories/Table.js
timbault: yes but TestData is not intended to be exported, it is an internal module for test purpose.

packages/components/src/Table/TableComp/TableComp.js
```diff
@@ -0,0 +1,60 @@
...
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+import TableHeader from '../Header/TableHeader';
+import TableBody from '../Body/TableBody';
+import theme from './TableComp.scss';
+
+/**
+ * This component is responsible for rendering the table (i.e. head and body)
+ */
+export default function TableComp({
```
frassinier: Place this content into Table component, this layer is useless IMO

packages/components/src/Table/TestData.js
```diff
@@ -40,29 +78,78 @@ function getColumn(col) {
...
+
+export function addSortExtraProps(column) {
+	return {
+		key: column.key,
+		label: column.label,
+		headExtraProps: {
+			iconPosition: 'right',
+			link: true,
+		},
+		headClassName: column.key,
```
frassinier: It could be computed directly in TableHead

packages/components/src/Table/Filters/StringFilterComponent.js
```diff
@@ -0,0 +1,55 @@
...
+			docked: this.props.filter.params.docked,
+		};
+		this.props.onFilterChange(this.props.filter.id, Boolean(value), params);
+	}
+
+	render() {
+		const { className, placeHolder, dockable, navbar } = this.props;
+		return (
+			<FilterBar
+				className={classnames('tc-string-filter', className)}
```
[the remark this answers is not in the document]
timbault: you're right, it is useless

PR #1293, opened 2018-04-20
feat(cmf/TUI-113): make cmf a framework

jmfrancois:
**What is the problem this PR is trying to solve?**

CMF today is more a tool than a framework. Worst you have to know how it works to setup it and add so many dependencies. The getting started is way too much complex

**What is the chosen solution to this problem?**

Need:
* https://github.com/Talend/ui/pull/1332/
* https://github.com/Talend/ui/pull/1331

Simplify the setup and just require '@talend/react-cmf' to start a project.
So move all peerDependencies to dependencies
Add `api.bootstrap` to start your app

so a start app look like this:

```javascript
import '@talend/bootstrap-theme/src/theme/theme.scss';
import 'focus-outline-manager';
import cmf from '@talend/react-cmf';
import configure from './configure';
import sagas from './sagas';

configure.initialize();
cmf.bootstrap({
	middlewares: [],
	settingsURL: '/settings.json',
	saga: sagas.appSaga, // function* appSaga() {...}
	sagaRouterConfig: sagas.routes, // { '/foo/bar': function* bar() {...}, ... },
        ...manyMoreOptions
});
```

**Please check if the PR fulfills these requirements**

* [x] The PR commit message follows our [guidelines](https://github.com/talend/tools/blob/master/tools-root-github/CONTRIBUTING.md)
* [x] Tests for the changes have been added (for bug fixes / features) And [non reg](./screenshots.md) done before need review
* [x] Docs have been added / updated (for bug fixes / features)
* [ ] Related design / discussions / pages (not in jira), if any, are all linked or available in the PR

**[x] This PR introduces a breaking change**

Before :
you were able to import many internal stuff, which has been removed:

* App
* RegistryProvider
* UIRouter
* history
* store
* reducers
* httpMiddleware
* getErrorMiddleware

They have all been removed because now you just have to use bootstrap api to start your project

PR #1440, opened 2018-06-11
feat(cmf): add saga to handle http actions

jmfrancois:

We have redux middleware to handle http action which is doing fetch
We have http saga which are doing fetch.

code duplication with some tiny difference:

middleware handle onError by default and propose to override it
middleware handle onResponse + transform options

Need #1293

To migrate that we will first add a saga to handle http actions
We will let the middleware to do some warning if you are in some corner case

### DIFF
--- a/packages/cmf/src/sagas/http.js
+++ b/packages/cmf/src/sagas/http.js
@@ -1,8 +1,9 @@
-import { call, put } from 'redux-saga/effects';
+import { call, put, take, takeEvery } from 'redux-saga/effects';
 import merge from 'lodash/merge';
 import get from 'lodash/get';
 import curry from 'lodash/curry';
 
+import actions from '../actions';
 import { mergeCSRFToken } from '../middlewares/http/csrfHandling';
 import {
 	ACTION_TYPE_HTTP_ERRORS,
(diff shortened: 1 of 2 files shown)


packages/cmf/src/sagas/http.js
```diff
@@ -286,6 +287,78 @@ export function getDefaultConfig() {
...
+ * saga forked at bootstrap to take all actions.http signal
+ */
+function* takeAllHTTPAction() {
+	yield takeEvery(HTTP_METHODS.DELETE, handleHTTPAction);
+	yield takeEvery(HTTP_METHODS.GET, handleHTTPAction);
+	yield takeEvery(HTTP_METHODS.POST, handleHTTPAction);
+	yield takeEvery(HTTP_METHODS.PUT, handleHTTPAction);
+	yield takeEvery(HTTP_METHODS.PATCH, handleHTTPAction);
+	yield take('DO_NOT_QUIT');
+}
```
ncomont: Could be replaced by : 

```
function* takeAllHTTPAction() {
	yield takeEvery(
		[
			HTTP_METHODS.DELETE,
			HTTP_METHODS.GET,
			HTTP_METHODS.POST,
			HTTP_METHODS.PUT,
			HTTP_METHODS.PATCH
		],
		handleHTTPAction
	);
	yield take('DO_NOT_QUIT');
}
```

https://github.com/redux-saga/redux-saga/blob/f85674b3a319c1f9c209e62bc509a0e2c0b4bdeb/packages/core/src/internal/proc.js#L448

https://github.com/redux-saga/redux-saga/blob/d98344e4fd6a7be561bd26f58d5c831e3b14c1b8/packages/core/src/internal/matcher.js#L14



... Even if I'm not sure to understand the goal of the last `take` in this case.

packages/cmf/src/sagas/http.js
```diff
@@ -286,6 +287,78 @@ export function getDefaultConfig() {
...
+		response = yield call(httpDelete, url, config, options);
+	} else if (type === HTTP_METHODS.GET) {
+		response = yield call(httpGet, url, config, options);
+	} else if (type === HTTP_METHODS.POST) {
+		response = yield call(httpPost, url, body, config, options);
+	} else if (type === HTTP_METHODS.PUT) {
+		response = yield call(httpPut, url, body, config, options);
+	} else if (type === HTTP_METHODS.PATCH) {
+		response = yield call(httpPatch, url, body, config, options);
+	}
```
ncomont: Maybe this could be achieved by looping on every HTTP methods types ?

packages/cmf/src/sagas/http.js
```diff
@@ -286,6 +287,78 @@ export function getDefaultConfig() {
...
+	if (!response.response.ok) {
```
ncomont: I'd like to talk with you about what should be an HTTP error for the HTTP middleware. I'm not sure to agree with the actual implementation, and I'd like to understand why this has been done like that.

packages/cmf/src/sagas/http.js
```diff
@@ -286,6 +287,78 @@ export function getDefaultConfig() {
...
+			yield put(onResponse(data));
+		} else if (typeof onResponse === 'string') {
+			yield put({
+				type: onResponse,
+				response: response.data,
+			});
+		}
+		if (cmf) {
+			if (cmf.collectionId) {
+				yield put(actions.collections.addOrReplace(cmf.collectionId, data));
```
ncomont: Too deeply imbricated in my opinion.

PR #1441, opened 2018-06-11
fix(versions): update redux to 3.7 and fix script behavior

jmfrancois:

Current version script is no nore usable in UI because we have a mix of react versions
It s impossible to fix version in deps, it make sens in devdeps but not for dependencies.

We have two redux versions: 3.6 and 3.7

First we add ˆ in version if we are under "dependencies""

Ensure react is not in dependencies of a library in UI and add a warning in that case. It s ok to add react under "dependencies" only for a project. (same for react-dom)

So we want the stack to be tested on react 15 and 16.

react / react-dom:
* 16 in devDeps
* 15||16 in peerDeps
* warning in Deps up to you

The script now read into env variable REACT_VERSION and use 16 by default.

Warning: Enumeration snapshot is crazy ... @jsomsanith is working on it in an other PR


version.js
```diff
@@ -189,13 +189,21 @@ if (program.debug) {
...
+function check(source, dep, version, category = 'dep') {
+	let safeVersion = version;
+	if (category === 'dep' && !version.startsWith('^')) {
+		safeVersion = `^${version}`;
+	}
 	let modified = false;
-	if (source && source[dep] && source[dep] !== version) {
+	if (source && source[dep] && source[dep] !== safeVersion) {
+		if (dep === 'react' && category === 'dep') {
+			console.warn(`WARNING: ${dep} should always be added as peer dependencies in library`);
```
jsomsanith: You don't need the string interpolation, just write `WARNING: react should ...`

version.js
```diff
@@ -35,7 +35,8 @@ if (program.debug) {
 	console.log(`use stack version ${stack_version}`);
 }
 
-const REACT_VERSION = '^16.0.0';
+const REACT_VERSION = process.env.REACT_VERSION || '^16.0.0';
```
jsomsanith: We should set a default value.

packages/logging/package.json
```diff
@@ -52,12 +52,12 @@
...
     "eslint-plugin-import": "^1.16.0",
-    "eslint-plugin-jsx-a11y": "^2.2.2",
+    "eslint-plugin-jsx-a11y": "^6.0.0",
     "eslint-plugin-react": "^6.3.0",
     "jest": "20.0.3",
     "jest-cli": "20.0.3",
-    "react": "^15.6.2",
-    "react-dom": "^15.6.2",
+    "react": "^15.6.2 || ^16.0.0",
+    "react-dom": "^15.6.2 || ^16.0.0",
```
jsomsanith: For logging, it is not compatible with 16.
Please revert it.

Moreover, when we adapt for 16, it will drop the support for 15

packages/components/package.json
```diff
@@ -128,7 +128,7 @@
     "invariant": "2.2.2",
     "keycode": "2.2.0",
     "prop-types": "15.5.10",
-    "react": "^15.6.2 || ^16.0.0",
+    "react": "^16.0.0",
```
jsomsanith: The peer dependency here should include 15.
Plus, it doesn't fit the script, that introduce `|| 16`

output/cmf-webpack-plugin.eslint.txt
```diff
@@ -1,3 +1,24 @@
...
+	Value "" is the wrong type.
+
+Referenced from: airbnb
+Referenced from: ../../.eslintrc
+Error: /home/travis/build/Talend/ui/node_modules/eslint-config-airbnb/rules/react-a11y.js:
+	Configuration for rule "jsx-a11y/anchor-has-content" is invalid:
+	Value "" is the wrong type.
+
+Referenced from: airbnb
+Referenced from: ../../.eslintrc
```
jmfrancois: I have downgrade a11y to fix that

test-setup.js
```diff
@@ -4,7 +4,7 @@ import { configure } from 'enzyme';
 import AdapterReact15 from 'enzyme-adapter-react-15';
 import AdapterReact16 from 'enzyme-adapter-react-16';
 
-if (process.env.REACT_VERSION === 15) {
+if (process.env.REACT_VERSION.startsWith('^15.')) {
```
jsomsanith: Is it a choice to have the wildcard (so no fixed version) ? 
Just a question, it can be a choice here.

packages/logging/package.json
```diff
@@ -52,7 +52,7 @@
     "eslint": "^3.6.1",
     "eslint-config-airbnb": "^11.1.0",
     "eslint-plugin-import": "^1.16.0",
-    "eslint-plugin-jsx-a11y": "^2.2.2",
+    "eslint-plugin-jsx-a11y": "^6.0.0",
```
jsomsanith: Why do we still have 6 here ?

packages/components/package.json
```diff
@@ -45,17 +45,17 @@
...
+    "rc-slider": "^8.4.1",
+    "rc-tooltip": "^3.7.0",
+    "react-autowhatever": "^10.1.0",
+    "react-debounce-input": "^3.1.0",
     "react-dnd": "^2.5.4",
     "react-dnd-html5-backend": "^2.5.4",
-    "react-immutable-proptypes": "2.1.0",
-    "react-virtualized": "9.10.1"
+    "react-immutable-proptypes": "^2.1.0",
+    "react-virtualized": "^9.10.1"
```
jmfrancois: "react-virtualized": "^9.10.1"

version.js
```diff
@@ -190,14 +191,28 @@ if (program.debug) {
 	console.log(`will update ${files}`);
 }
 
-function check(source, dep, version) {
+function check(source, dep, version, category = 'dep') {
```
acateland: this does not only check, it also mutate the source parameter, can you please split check and transform in two different function ? can you please add a bit of code doc

version.js
```diff
@@ -208,9 +223,9 @@ function checkAll(versions, source, dep) {
 	const devDeps = source.devDependencies;
 	const deps = source.dependencies;
 	const peer = source.peerDependencies;
-	const isModDevDeps = check(devDeps, dep, version);
+	const isModDevDeps = check(devDeps, dep, version, 'dev');
```
acateland: i don't get the name isModDevDeps is not a function shouldn't it be updatedDependencies or something like that

also check seem to only return a boolean. it is confusing

version.js
```diff
@@ -36,7 +36,9 @@ if (program.debug) {
 	console.log(`use stack version ${stackVersion}`);
 }
 
-const REACT_VERSION = '^16.0.0';
+const REACT_VERSION = process.env.REACT_VERSION || '^16.0.0';
+console.log('REACT_VERSION: ', REACT_VERSION);
```
acateland: should be removed i guess

jmfrancois: No please keep it for debug in the CI
Because we need to handle both and it comes from ENV, it nice to have it logged here

version.js
jmfrancois: I think Rename it could be nice in fact ! true

acateland:
will transform the script into a beautiful pipeline later

PR #1442, opened 2018-06-11
feat(Enumeration): improve a11y

jsomsanith:
Accessibility in Enumeration component

* Add labels on inputs, with translation
* Inputs are linked to the error message
* Set back outline
* Fix role hierarchy : some elements have roles, but we expect some roles to have children with certain roles. Example : `grid` role expect `gridcell` (or other related grid roles) in its children.
* Fix html hierarchy : we had `<ul><div><div><li></></></></>` which is not correct

Bonus 
* all snapshot tests are in enumeration tests only. No more duplication
* snapshot tests are more relevant. Each one test one situation or a specific part of the component
* snapshots are (waaaaaaaay) lighter, taking a photo of only the part that is tested
* fixed classnames usage
* Toggle component pass extra components to input. We use it to passe `aria-describedby` in Enumeration.


packages/components/src/Enumeration/Enumeration.component.js
```diff
@@ -52,6 +55,7 @@ Enumeration.propTypes = {
 	headerDefault: PropTypes.arrayOf(PropTypes.shape(headerPropTypes)).isRequired,
 	headerInput: PropTypes.arrayOf(PropTypes.shape(headerPropTypes)),
 	headerSelected: PropTypes.arrayOf(PropTypes.shape(headerPropTypes)),
+	id: PropTypes.string,
```
annilsson: Using multiple EnumrationComponent would break if there's no id right (duplicate id's in DOM)? Should we not make it required? 
(I'm not sure if there's a risk there would be more than one)

output/components.eslint.txt
```diff
@@ -6,6 +6,9 @@ The react/require-extension rule is deprecated. Please use the import/extensions
   1:1  error  '@storybook/react' should be listed in the project's dependencies, not devDependencies  import/no-extraneous-dependencies
   4:2  error  Unexpected require()                                                                    global-require
 
+/home/travis/build/Talend/ui/packages/components/src/Enumeration/Enumeration.snapshot.test.js
+  2:10  error  'shallow' is defined but never used  no-unused-vars
```
annilsson: Should fix this one

packages/components/src/Enumeration/Enumeration.component.js
jsomsanith: Yes, and worst, there were already ids, generated from the index of the item in the list. So currently, the ids unicity breaks if you have multiple Enumeration in the page :/
I added this prop so all the ids are now based on the one passed by user.

I didn't want to make it required, because in the case you have only 1 instance, it isn't required. In fact like the List.

packages/components/src/Toggle/Toggle.component.js
```diff
@@ -33,6 +34,7 @@ function Toggle({ id, onChange, onBlur, label, checked, autoFocus, disabled, cla
 					checked={checked}
 					autoFocus={autoFocus}
 					disabled={disabled}
+					{...omit(rest, 'data-feature')}
```
jmfrancois: in that case please remove useless destructured from props:
* onChange
* onBlur,
* autoFocus

they will be in the rest  which I prefer call props
> {id, checked, className, disabled, ...props}

packages/components/src/Enumeration/Enumeration.component.js
jare-talend: I would have made the same remark with id. So, to avoid duplication id, should we use id generated by an uuid ?

PR #1443, opened 2018-06-11
fix(forms/ColumnsWidget): change how onColumnChange scope is handled

acateland:
only the first handleChange generated with its closure would be taken into account by children component (probably have something to do with 
https://github.com/mozilla-services/react-jsonschema-form/blob/master/src/utils.js#L77
since on each render a new closure is generated and function attached to this scope passed down

What problem does it create ? formData is kept at initial value meaning that two consecutives widgets edition in a columns would result in formData is not usable.
change how the closure is handled by removing formData, formData is going to be given by this.props on each onChange execution, meaning that formData is in the right shape

**Change on yarn.lock are due to the fact taht some previous update to package.json were not reflected and commited**


acateland:
@glorieux im pretty sure it has been done because it wasn't updated along the package.json changes

packages/forms/src/widgets/ColumnsWidget/ColumnsWidget.js
```diff
@@ -44,32 +44,42 @@ if (process.env.NODE_ENV !== 'production') {
...
-							onBlur={onBlur}
-							className={`tf-column-${key}`}
-						/>
-					))
-				: null}
-		</div>
-	);
+	onColumnChange(key) {
+		return change => {
+			this.props.onChange(Object.assign({}, this.props.formData, { [key]: change }));
```
jmfrancois: please rename change -> value

yarn.lock
```diff
@@ -523,6 +523,13 @@ argparse@^1.0.6, argparse@^1.0.7:
   dependencies:
```
jmfrancois: please revert this file

output/forms.eslint.txt
```diff
@@ -21,7 +21,8 @@ The react/require-extension rule is deprecated. Please use the import/extensions
   1:8  error  Using exported name 'DefaultArrayTemplate' as identifier for default export  import/no-named-as-default
 
 /home/travis/build/Talend/ui/packages/forms/src/widgets/ColumnsWidget/ColumnsWidget.js
-  59:7  error  Expected indentation of 5 tab characters but found 6  react/jsx-indent
+  68:8  error  Expected indentation of 6 tab characters but found 7  react/jsx-indent
+  78:2  error  Mixed spaces and tabs                                 no-mixed-spaces-and-tabs
```
jmfrancois: please fix this

packages/forms/src/widgets/ColumnsWidget/ColumnsWidget.js
```diff
@@ -44,32 +44,42 @@ if (process.env.NODE_ENV !== 'production') {
...
 
-function onColumnChange(key, onChange, formData) {
-	return function handleChange(change) {
-		onChange(Object.assign({}, formData, { [key]: change }));
-	};
-}
+export default class ColumnsWidget extends React.Component {
+	constructor(props) {
+		super(props);
+		this.onColumnChange = this.onColumnChange.bind(this);
```
jare-talend: you can do that class ColumnsWidget extends React.Component { onColumnChange = () => { ... your body function } }

jare-talend: i do a study on that :)

acateland: for the sake of code base homogeneity we won't.

jsomsanith: And the sake of debug stacktraces

PR #1455, opened 2018-06-14
chore: fix versions

jsomsanith:
Some libs version have been changed, adding a `^`.
This introduced some DOM changes. But QAs has tests that fail because of that.
We need to have more control on the versions.

* Fix the versions of components libs.
* Set the wildcard to dev/peer/non-component libs.

The libs that introduce changes in DOM must be updated regularly, making sure that there are no structural changes.
If there are, it must be explicitly said, and communicated to the QAs


version.js
```diff
@@ -57,65 +57,64 @@ const STACK_VERSION = {
...
-	'whatwg-fetch': '2.0.3',
+	'path-to-regexp': '^2.0.0',
+	'react-addons-perf': '^15.4.2',
+	'react-tap-event-plugin': '^2.0.0',
+	'whatwg-fetch': '^2.0.3',
 };
 
 const VERSIONS = Object.assign({}, ADDONS, {
-	// deps
+	// deps: js libs
```
jmfrancois: what is the meaning of that ? a dependency is a js lib aka package ...

jsomsanith: I'll rename it to `non component libs`

jmfrancois:
FYI this PR fix https://nodesecurity.io/advisories?search=lodash&version=4.17.4

PR #1458, opened 2018-06-15
feat(Checkbox): improve a11y with outline

jsomsanith:
There is no outline on Checkboxes and Radios, especially when they are checked.

Add an outiline from checkbox guideline


packages/theme/src/theme/_forms.scss
```diff
@@ -6,6 +6,11 @@
 $form-shadow: inset 0 -1px 0 $gallery;
 $form-focus-shadow: inset 0 -2px 0 $scooter;
 
+@mixin checkOutline() {
+	outline: 2px auto -webkit-focus-ring-color;
```
annilsson: What about non-webkit browsers? Could we have a fallback color for them on the line before maybe?

packages/theme/src/theme/_forms.scss
```diff
@@ -6,6 +6,11 @@
 $form-shadow: inset 0 -1px 0 $gallery;
 $form-focus-shadow: inset 0 -2px 0 $scooter;
 
+@mixin checkOutline() {
```
annilsson: The linter complained about camelCase, but while at it, why not checkbox-outline?

packages/theme/src/theme/_forms.scss
jmfrancois: we still use autoprefix for compatibility

jsomsanith: Indeed, that doesn't fix for non webkit browsers :/
I changed it to use `Highlight` color that takes the default highlight color.

packages/theme/src/theme/_forms.scss
jsomsanith: Done

jsomsanith:
Oh you're right, that's weird that a search dooesn't find anything :/
I found it here : https://ghinda.net/article/mimic-native-focus-css/

PR #1459, opened 2018-06-16
fix(forms): prevent error when quickly add values

romainseb:
if you're fast at add values in enumeration, it could end with an error that block the enumeration state because we try to access properties that are not here at this moment. ( there is only one header input action -> the loading spinner )
return null or previous state ( as it's better ) to prevent move the state 


packages/forms/src/widgets/EnumerationWidget/EnumerationWidget.js
```diff
@@ -854,8 +854,12 @@ class EnumerationForm extends React.Component {
 			// checking if the value already exist
 			const valueExist = this.valueAlreadyExist(value, prevState);
 			const [validateAndAddAction, validateAction, abortAction] = prevState.headerInput;
+			if (!validateAction && !abortAction) {
+				return prevState;
```
jsomsanith: You're in a `setState(fn)`. You should return null instead of current state to indicate that there is no change.
If I'm not wrong, in React 16, returning null doesn't trigger a new render.

packages/forms/src/widgets/EnumerationWidget/EnumerationWidget.js
```diff
@@ -854,8 +854,14 @@ class EnumerationForm extends React.Component {
 			// checking if the value already exist
 			const valueExist = this.valueAlreadyExist(value, prevState);
 			const [validateAndAddAction, validateAction, abortAction] = prevState.headerInput;
+			// in this case, we could have the loading state that implied we have just one icon
+			if (!validateAction && !abortAction) {
+				// returning null in setState prevent re-rendering
+				return null;
```
acateland: not explicitly described in the documentation, make hard for future work on this to find reference on how this work.

acateland: or please add a link to https://reactjs.org/blog/2017/09/26/react-v16.0.html#breaking-changes
wich explain how and why this work this way

jsomsanith:
@romainseb we discussed with @acateland, can you add a comment in the code with the url to the blog post explaining the behavior of react 16 setState(null) ?
(https://reactjs.org/blog/2017/09/26/react-v16.0.html#breaking-changes)

PR #1461, opened 2018-06-18
chore(Enumeration): disabling enumeration

jlame:
there's no disable state for enumeration widget

adding what's necessary to make it work


packages/components/src/Enumeration/Header/HeaderInput.component.js
```diff
@@ -25,7 +25,7 @@ function getAction(action, index, getInternalInputRef) {
 		}
 	}
 
-	return (
+	return !action.disabled ? (
```
jare-talend: you can add a test on disabled and if it is the case return null.
add if(action.disabled) { return null }

packages/components/src/Enumeration/Header/Header.component.js
```diff
@@ -20,7 +20,7 @@ function getAction(action, index) {
 	}
```
jare-talend: you can add a test on disabled and if it is the case return null.
add if(action.disabled) { return null }

packages/components/src/Enumeration/Header/Header.component.js
```diff
@@ -20,7 +20,7 @@ function getAction(action, index) {
...
 
 	if (action.displayMode === 'dropdown') {
-		return (
+		return !action.disabled ? (
```
jare-talend: with the previous test it is not necessary.

packages/components/src/Enumeration/Header/Header.component.js
```diff
@@ -31,9 +31,9 @@ function getAction(action, index) {
 				pullRight
 				link
 			/>
-		);
+		) : null;
 	}
-	return (
+	return !action.disabled ? (
```
jare-talend: idem

packages/components/src/Enumeration/Header/HeaderSelected.component.js
```diff
@@ -18,7 +18,7 @@ function getAction(action, index) {
 		}
 	}
 
-	return (
+	return !action.disabled ? (
```
jare-talend: you can add a test on disabled and if it is the case return null.
add if(action.disabled) { return null }

packages/components/src/Enumeration/Items/Item/Item.component.js
```diff
@@ -48,7 +48,7 @@ function Item({ id, item, searchCriteria, showCheckboxes, style, t }) {
 			}
 		}
 
-		return (
+		return !action.disabled ? (
```
jare-talend: you can add a test on disabled and if it is the case return null.
add if(action.disabled) { return null }

packages/components/src/Enumeration/Header/HeaderInput.component.js
jlame: test already done

packages/components/src/Enumeration/Header/Header.component.js
jlame: test already done

packages/components/src/Enumeration/Header/HeaderSelected.component.js
jlame: test already done

packages/forms/stories/json/enumerationDisabled.json
```diff
@@ -0,0 +1,60 @@
...
+  "jsonSchema": {
+    "title": "A enumeration form",
+    "description": "A enumeration form example.",
+    "type": "object",
+    "properties": {
+      "documents": {
+        "type": "array",
+        "required": true,
+        "allowDuplicates": true,
+        "disabled": true,
```
jsomsanith: This should be in uiSchema

packages/forms/src/widgets/EnumerationWidget/EnumerationWidget.js
```diff
@@ -70,6 +70,7 @@ class EnumerationForm extends React.Component {
 		this.timerSearch = null;
 		this.allowDuplicate = false;
 		this.allowImport = false;
+		this.disabledAction = props.schema ? props.schema.disabled : false;
```
jsomsanith: No need to store it in the instance. You only use it in constructor. Use a const.
```
const 
disabled = props.schema && props.schema.disabled;
```

packages/components/src/Enumeration/Header/Header.component.js
```diff
@@ -18,7 +18,10 @@ function getAction(action, index) {
 			action.onClick(event, { value: event.target.value });
 		}
 	}
-
+	if (action.disabled) {
```
jsomsanith: You should filter the actions in the Header `return`, before the map
```
{
    headerDefault
        .filter(...)
        .map(getAction);
}
```

packages/components/src/Enumeration/Items/Item/Item.test.js
```diff
@@ -64,6 +64,24 @@ describe('Item', () => {
 		expect(props.item.itemProps.onSelectItem).toBeCalled();
 	});
 
+	it('should display value with only button which are not disabled', () => {
+		// given
+		const itemWithDisabled = { ...item };
+		itemWithDisabled.itemProps.actions[0].disabled = true;
```
jsomsanith: You change the action from the same reference as all the tests.
Please don't mutate it as it can impact all the following tests. --> deepClone.

packages/components/src/Enumeration/Items/Item/Item.test.js
```diff
@@ -64,6 +64,24 @@ describe('Item', () => {
...
+		};
+
+		const itemInstance = <Item {...props} />;
+
+		// when
+		const wrapper = mount(itemInstance);
+		const buttons = wrapper.find(Button);
+
+		// then
+		expect(buttons.length).toBe(2);
```
jsomsanith: I don't understand the assertion. You have 2 actions, 1 disabled, tou should have only 1 button

packages/components/src/Enumeration/Header/HeaderInput.component.js
```diff
@@ -24,7 +24,9 @@ function getAction(action, index, getInternalInputRef) {
 			});
 		}
 	}
-
+	if (action.disabled) {
+		return null;
+	}
```
jsomsanith: Same remark  : You should filter the actions in the main component `return`, before the map()

packages/components/src/Enumeration/Header/HeaderSelected.component.js
```diff
@@ -17,7 +17,9 @@ function getAction(action, index) {
 			action.onClick(event);
 		}
 	}
-
+	if (action.disabled) {
+		return null;
+	}
```
jsomsanith: Same remark : You should filter the actions in the main component return, before the map()

packages/components/src/Enumeration/Items/Item/Item.component.js
```diff
@@ -47,7 +47,9 @@ function Item({ id, item, searchCriteria, showCheckboxes, style, t }) {
 				});
 			}
 		}
-
+		if (action.disabled) {
+			return null;
+		}
```
jsomsanith: Same remark : You should filter the actions in the main component return, before the map()

packages/components/src/Enumeration/Items/Item/Item.test.js
jlame: the selection also includes the title which is clickable and is indeed a button, i've narrow the selection so we only check the one we're interested in.

PR #1463, opened 2018-06-18
feat(notification): notification list handled even if Notification component is not mounted

acateland:
Notifications can't be pushed into notification list if the component is not mounted
Notification state slot not existing
This prevent for example to have notification dispatched trought the application bootstrap from being displayed to the user when the app is finally loaded
Allow to create an Immutable.List and feed the Notification component state slot when it is not yet mounted

This also follow the implementation of the Notification component who can register and display notification available before he was mounted


packages/containers/src/Notification/pushNotification.js
```diff
@@ -1,13 +1,21 @@
...
+ * @param {*} notification
+ * @returns {state}
+ */
 export default function pushNotification(state, notification) {
 	if (!get(notification, 'message')) {
 		return state;
 	}
 	const path = ['Container(Notification)', 'Notification', 'notifications'];
-	let notifs = state.cmf.components.getIn(path);
+	let notifs = state.cmf.components.getIn(path, new Immutable.List());
```
jmfrancois:
```javascript
import Notification from './Notification.connect';
...
let naffs = Notification.getState(state, 'Notification').get('notifications');
```

packages/containers/src/Notification/pushNotification.js
```diff
@@ -1,13 +1,21 @@
...
 	if (!notifs) {
 		invariant(true, 'Notifications are not yet available.', notification);
 		return state;
```
jmfrancois: remove that if please

PR #1465, opened 2018-06-18
feat(components/Layout): add inject support

jmfrancois:
**What is the problem this PR is trying to solve?**

Layout is nice but wait for real components
So it doesn t play well with CMF settings

**What is the chosen solution to this problem?**

header, subheader and footer props now support the follwing types:
* string the component to use
* object { component: "TheComponent", ...props }
* React Element (actual type)
* Array of the three previous types

cleanup: remove 'react-test-renderer' from Layout tests

**Please check if the PR fulfills these requirements**

* [x] The PR commit message follows our [guidelines](https://github.com/talend/tools/blob/master/tools-root-github/CONTRIBUTING.md)
* [x] Tests for the changes have been added (for bug fixes / features) And [non reg](./screenshots.md) done before need review
* [x] Docs have been added / updated (for bug fixes / features)
* [ ] Related design / discussions / pages (not in jira), if any, are all linked or available in the PR

**[ ] This PR introduces a breaking change**

PR #1466, opened 2018-06-18
feat(TUI-162): add component kit

jmfrancois:

We have UIForm which is nice
We have a backend (component-runtime)

Setup can be hard and will lead to bad code and hard to maintain.

Provide a component which deals with the backend. Just provide the URL to query and it will do the job.

The component need two required URL:
* definitionURL to GET the definition
* triggerURL to call component kit actions

We provide a mock so in dev you will have one component with many triggers in it and you will do not need to setup backend

Note: it rely on #1465 (embeded here)
 

package.json
```diff
@@ -12,7 +12,7 @@
     "tmp-promise": "^1.0.4"
   },
   "scripts": {
-    "postinstall": "lerna clean --yes && lerna exec -- yarn --pure-lockfile --no-progress --mutex network",
+    "postinstall": "lerna exec -- yarn --pure-lockfile --no-progress --mutex network",
```
jmfrancois: 20 seconds better. It s an old fix when we were using the lerna bootstrap.
Now we are all on yarn workspace so let remove the clean

packages/kit/src/components/ComponentForm/ComponentForm.component.js
```diff
@@ -0,0 +1,186 @@
...
+			});
+		});
+	}
+
+	onSubmit(event, data) {
+		event.persist();
+		if (this.props.onSubmit) {
+			this.props.onSubmit(event, data);
+		}
+		this.props.dispatch({
```
jsomsanith: Why do we have the onSubmit and dispatch ?
Shouldn't we allow only one of them to avoid the possibility to make the same thing in 2 different ways ? (that was one of the pain points of cmf for some people, feeling lost in too many possibilities)

packages/kit/src/components/ComponentForm/ComponentForm.component.js
```diff
@@ -0,0 +1,186 @@
...
+
+	onChange(event, data) {
+		event.persist();
+		if (!this.props.state.get('dirty')) {
+			this.props.setState({ dirty: true });
+		}
+		const properties = data.properties;
+		if (data.schema.titleMap && data.value) {
+			// Here we add a field side by side with the value
+			// to keep the title associated to the value
```
jsomsanith: wat :)
Can you explain the need ?

packages/kit/src/components/ComponentForm/ComponentForm.component.js
```diff
@@ -0,0 +1,186 @@
...
+			this.props.dispatch({
+				type: ComponentForm.ON_DEFINITION_URL_CHANGED,
+				...this.props,
+			});
+		}
+	}
+
+	onChange(event, data) {
+		event.persist();
+		if (!this.props.state.get('dirty')) {
```
jsomsanith: You don't need this `if`, just set `dirty: true` when you set the properties in state

packages/kit/src/components/ComponentForm/ComponentForm.component.js
```diff
@@ -0,0 +1,186 @@
...
+				properties,
+				uiSpec: this.getUISpec(),
+			});
+		}
+	}
+
+	onTrigger(event, payload) {
+		return this.trigger(event, payload).then(data => {
+			if (data.properties) {
+				this.setState({ properties: data.properties });
```
jsomsanith: I assume we always get the entire properties (and schemas for next lines) from triggers ?

packages/kit/src/components/ComponentForm/ComponentForm.component.js
```diff
@@ -0,0 +1,186 @@
...
+			}
+			if (data.errors || data.jsonSchema || data.uiSchema) {
+				this.props.setState(data);
+			}
+			this.props.dispatch({
```
jsomsanith: Do we need that dispatch ?
Triggers are supposed to be managed from the inside.

packages/kit/src/components/ComponentForm/ComponentForm.component.js
```diff
@@ -0,0 +1,186 @@
...
+	setupTrigger(props) {
+		this.trigger = kit.createTriggers({
+			url: props.triggerURL,
+			customRegistry: this.getTriggers(),
+		});
+	}
+
+	getTriggers() {
+		return {
+			reloadForm: ({ body }) => ({
```
jsomsanith: Can you explain this `reloadForm` trigger ? I am not sure to understand.
I saw that you have it in your example in datalist onFocus event. Why do we reload all the form ? That was how tcomp v0 trigger works and we wanted to get out of this. Shouldn't we have a trigger that only fetch the suggestions and reconcile the schema ? 

Plus the supported triggers should be in `kit` instead of the form

packages/kit/src/components/ComponentForm/ComponentForm.md
```diff
@@ -0,0 +1,38 @@
...
+        type: ComponentForm.ON_SUBMIT,
+        event,
+        jsonSchema,
+        uiSchema,
+        properties,  // your data
+    }
+}
+```
+
+definitionURL: url to fetch to get the form uiSpec
```
jsomsanith: `url to fetch to get`

packages/kit/src/components/ComponentForm/ComponentForm.sagas.js
```diff
@@ -0,0 +1,60 @@
+import { call, put, select, take, takeEvery } from 'redux-saga/effects';
+import cmf from '@talend/react-cmf';
+import get from 'lodash/get';
+import Component from './ComponentForm.component';
+
+function* fecthDefinition({ definitionURL, componentId, uiSpecPath }) {
```
jsomsanith: Typo `fecthDefinition`

packages/kit/src/components/ComponentForm/ComponentForm.component.js
jmfrancois: We would like to not have to load all values available on a remote service just to display the name of the entity.
So the idea is to save it.
Our backend filter all attribute which start by a `$` and to make it more readable I add a `_name`

But here I forgot to add the reading part of that during the build of the titleMap (which should live in the datalist field)

packages/kit/src/components/ComponentForm/ComponentForm.component.js
```diff
@@ -0,0 +1,186 @@
...
+			}),
+		};
+	}
+
+	getUISpec() {
+		const spec = { properties: this.state.properties };
+		const jsonSchema = this.props.state.get('jsonSchema');
+		const uiSchema = this.props.state.get('uiSchema');
+		if (jsonSchema) {
+			spec.jsonSchema = jsonSchema.toJS();
```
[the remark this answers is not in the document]
jsomsanith: memoize-one

packages/kit/src/components/ComponentForm/ComponentForm.component.js
```diff
@@ -0,0 +1,186 @@
...
+			if (response) {
+				return <p className="danger">{response.get('statusText')}</p>;
+			}
+			return <CircularProgress />;
+		}
+		if (this.state.properties) {
+			props.properties = this.state.properties;
+		}
+		if (this.props.state.get('errors')) {
+			props.errors = this.props.state.get('errors').toJS();
```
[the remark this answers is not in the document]
jsomsanith: memoize-one

packages/kit/src/components/ComponentForm/ComponentForm.component.js
jsomsanith: Remove onSubmit

packages/kit/src/components/ComponentForm/ComponentForm.component.js
jsomsanith: Remove it

packages/components-kit/src/components/ComponentForm/ComponentForm.component.js
```diff
@@ -0,0 +1,191 @@
...
+					state: this.state,
+				},
+				data,
+				uiSpec: this.getUISpec(),
+			});
+		});
+	}
+
+	onSubmit(event, data) {
+		event.persist();
```
acateland: what happen if a submit is trigerred before a running trigger is not yet resolved.

packages/components-kit/src/components/ComponentForm/ComponentForm.component.js
```diff
@@ -0,0 +1,191 @@
...
+			}),
+		};
+	}
+
+	getUISpec() {
+		const spec = { properties: this.state.properties };
+		const jsonSchema = this.props.state.get('jsonSchema');
+		const uiSchema = this.props.state.get('uiSchema');
+		if (jsonSchema) {
+			spec.jsonSchema = jsonSchema.toJS();
```
acateland: why to JS ? can't we take this opportunity to avoid those ?

packages/components-kit/webpack.config.dev.js
```diff
@@ -0,0 +1,103 @@
...
+	return {
+		...req.query,
+		args: req.body,
+	};
+}
+
+function basicAuth(args) {
+	let comment;
+	let status;
+	if (!args['basicAuth.password']) {
```
acateland: this handle only one error information even if in fact they are many

packages/kit/src/components/ComponentForm/ComponentForm.component.js
jsomsanith: Remove the dispatch

packages/components-kit/src/components/ComponentForm/ComponentForm.component.js
jsomsanith: UIForm use json-schema-form-core (jsfc) that process the json and ui schemas, as js objects.
We need to pass js objects, not immutablejs objects.
I just changed that to memoize the last result for jsonSchema, uiSchema and errors

packages/components-kit/src/components/ComponentForm/ComponentForm.component.js
jsomsanith: Ok I talked to @jmfrancois, seems that even with the `event.persist()`, we are not sure to have the same event during saga execution.
The plan is to extract useful infos from event to add it in action to dispatch. This point has been added to the component form TODO list #1525 (let's try not to make this PR bigger, but be sure that the list is all the points that must be fixed to be usable).

PR #1478, opened 2018-06-21
feat(components): hide tooltip between mouseDown & mouseUp

romainseb:
On some behaviour, we could have to update a label on a tooltip on click. 
In this case, the label could be a different size & i could be missplaced. 
Another issue is that it show instantly the new label ( if the label is dynamic in props ) and it could be nice to wait some time to do so
bad

Hide the tooltip between mouseDown & mouseUp.
But it's not all pretty because i have to trigger the "click" event on the mouseDown to make the click available as the react element are changed

good


packages/components/src/TooltipTrigger/TooltipTrigger.component.js
```diff
@@ -80,8 +133,21 @@ class TooltipTrigger extends React.Component {
...
+				ref={ref => {
+					this.overlay = ref;
+				}}
+				placement={this.props.tooltipPlacement}
+				overlay={tooltip}
+				delayShow={400}
+				animation={false}
+			>
+				{cloneElement(child, {
+					onMouseDown: createChainedFunction(this.onMouseDown, childrenProps.onMouseDown),
```
jsomsanith: Everytime there is a render, a new function props is created. (x the number of callback).
And `createChangedFunction()` is quite complicated for chaining (at most) 2 functions.

Just set the calls in the class methods if the props exist.

packages/components/src/TooltipTrigger/TooltipTrigger.component.js
```diff
@@ -25,8 +26,6 @@ class TooltipTrigger extends React.Component {
 
 	constructor(props) {
 		super(props);
-		this.onMouseOver = this.onMouseOver.bind(this);
-		this.onFocus = this.onFocus.bind(this);
 
 		this.state = {
 			hovered: false,
```
nicomaligne: with babel class properties, u can declare state out the constructor it will transpile to this.state

PR #1481, opened 2018-06-21
fix(theme): overflow wrap the tooltip

jare-talend:
When the text in the tooltip overflow the max-width, the text is truncated.
image

fix that with an overflow-wrap: break-word.
image


packages/theme/src/theme/_tooltip.scss
```diff
@@ -0,0 +1,4 @@
+.tooltip-inner {
+	overflow-wrap: break-word;
+	word-wrap: break-word;
```
jare-talend: handle IE11 with his proprietary css

PR #1482, opened 2018-06-21
chore: eslint config module

jsomsanith:
Talend/script is not ready for all the needs, and will not be in a short time.
Moreover, some teams need to share the eslint config.

Make a shareable eslint config module.

I updated the dependencies to the last, and disabled the rules that produced new violations to stay almost with the same rules as before.

Next step : integration in Talend/ui and Talend/scripts


packages/eslint-config/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+```json
+{
+  "extends": "@talend/eslint-config"
+}
+```
+
+## Profiles
+
+There are multiple profiles available for now. The idea is to merge then eventually.
```
quilicicf: *them

packages/eslint-config/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+This module is a set of shared eslint configuration.
+
+## How to use
+
+Create an `.eslintrc` file in the folder you need to lint.
+
+```json
+{
+  "extends": "@talend/eslint-config"
```
quilicicf: That would require an `npm install` too, right ?

packages/eslint-config/package.json
```diff
@@ -0,0 +1,26 @@
...
+    "url": "https://github.com/Talend/ui/issues"
+  },
+  "license": "Apache-2.0",
+  "dependencies": {
+    "babel-eslint": "8.2.3",
+    "eslint": "4.19.1",
+    "eslint-config-airbnb": "16.1.0",
+    "eslint-plugin-import": "2.12.0",
+    "eslint-plugin-jsx-a11y": "6.0.3",
+    "eslint-plugin-react": "7.9.1"
```
quilicicf: I'm puzzled that there's no mention of `jest` here Oo

packages/eslint-config/package.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "@talend/eslint-config",
+  "version": "0.1.0",
```
quilicicf: Semver has deprecated versions 0.X.Y, versions should start with at least 1 :)
More on this here https://github.com/semver/semver/issues/221
(I don't give a damn, especially for a tooling project but it's interesting nonetheless)

packages/eslint-config/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+| Profile | Extends | Description |
+|---------|---------|-------------|
+| default | `@talend/eslint-config` | Default Talend configuration |
+| default | `@talend/eslint-config/api-app.yaml` | Api team, javascript base configuration |
+| default | `@talend/eslint-config/api-app-react.yaml` | Api team, react applications configuration |
+| default | `@talend/eslint-config/api-app-angular.yaml` | Api team, Angular applications configuration |
+| default | `@talend/eslint-config/api-test.yaml` | Api team, test configuration |
+| default | `@talend/eslint-config/api-test-e2e.yaml` | Api team, e2e test configuration |
+| default | `@talend/eslint-config/packaging.yaml` | Api team, packaging script configuration |
```
quilicicf: Why is the profile `default` everywhere ? Oo

packages/eslint-config/index.js
```diff
@@ -0,0 +1 @@
+module.exports = require('./.eslintrc.json');
```
quilicicf: Is this `require` mandatory ? 
I find it weird that there is one for the JSON files but not for the YAML ones ?

packages/eslint-config/package.json
jsomsanith: I don't understand why we should have. It's only about the linter and no usage of eslint jest plugin.

packages/eslint-config/index.js
jsomsanith: This wrapper allows use tu write `@talend/eslint-config` instead of `@talend/eslint-config/.eslintrc.json`.
The others are yaml, so no possible js wrapper (especially with the extends of other yaml files). Moreover, they are named profiles, so having them directly or with a wrapper doesn't change anything int your app eslintrc `extends` value.

packages/eslint-config/README.md
jsomsanith: My bad, I'll remove the first column, unnecessary

packages/eslint-config/package.json
jsomsanith: Ok let's set 1.0.0

packages/eslint-config/README.md
jsomsanith: I add it

packages/eslint-config/package.json
jsomsanith: The only reference to jest that we have is in the `env` section to indicate that jest is global

PR #1483, opened 2018-06-21
chore: use @talend/eslint-config

jsomsanith:
We have a new module @talend/eslint-config, containing the shareable eslint configuration.
* We have a duplicated version in Talend/ui.
* This module comes with defined versions of eslint and its plugins

* Remove the duplication from Talend/ui.
* Remove the eslint + plugin dependencies. They come with @talend/esint-config
* Adapt some code to remove some violations. Those violations come from current rules updates, including fixes)
* Update version.js with the defined versions.


.eslintrc
```diff
@@ -1,34 +1,6 @@
...
-    "jest": true
-  },
-  "globals": {},
-  "parserOptions": {
-    "ecmaVersion": 6,
-    "sourceType": "module"
-  }
+  "extends": [
+    "@talend/eslint-config",
+    "prettier"
```
jsomsanith: Disable prettier rules, so it doesn't conflict with CI that run prettier

jdescottes-talend: Why not defined the prettier extends override inside the new package ?
This will remove the need for each package to have a direct dependency on a eslint plugin.
Plus the apps to have the plugin on devDependencies and also to have this extra line copied all over the .eslintrc apps.
If the reason is the need for an extra step, having a base one + the overriden one (like this file) inside the new package and export the overriden one by default, I think we should.

jmfrancois:
no activity during two months

PR #1486, opened 2018-06-22
fix(cmf):handleDefaultHttpConfiguration

acateland:
if you override one time the default config doing 
`yield call(http.get, `${API.streams}/${streamId}/logs`, {
		headers: {
			Accept: 'text/plain',
			'Content-Type': 'text/plain',
		},
	});`
instead of having this applied only for this call it will be applied to all subsequent call using `call(http.get)`

lodash merge is mutating the target object reference given to it, before returning it, crass but no big deal most of the time.
But here we where using it inside a curryed function that first take the default config and on subsequent call will try to merge temp override, issue being that the default config being mutated what was a one time application, was a whole app lifecycle application.

Solution ?
cloneDeep target object.

### DIFF
--- a/packages/cmf/src/sagas/http.js
+++ b/packages/cmf/src/sagas/http.js
@@ -2,7 +2,6 @@ import { call, put } from 'redux-saga/effects';
 import merge from 'lodash/merge';
 import get from 'lodash/get';
 import curry from 'lodash/curry';
-import cloneDeep from 'lodash/cloneDeep';
 
 import { mergeCSRFToken } from '../middlewares/http/csrfHandling';
 import {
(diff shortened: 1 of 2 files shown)


packages/cmf/src/sagas/http.js
```diff
@@ -274,7 +274,17 @@ export function setDefaultLanguage(language) {
 }
 
 export const handleDefaultHttpConfiguration = curry((defaultHttpConfig, httpConfig) =>
-	merge(defaultHttpConfig, httpConfig),
+	/**
+	 * Wall of explain
```
jsomsanith: `explaination`

PR #1487, opened 2018-06-22
feat: add memoized selector for components

VolodymyrKovalM:
Curently we have memorized selector only for "collections", when getting form data for example (which is stored in "components.Container(Form)"), sometimes we are forced to use toJS directly (in some cases even in mapStateToProps), which affects performance
 
Add memoized selector functionality for "components" also


packages/cmf/src/selectors/components.js
```diff
@@ -0,0 +1,41 @@
...
+const selectors = {};
+
+export function toJS(state, path) {
+	const joinedPath = Array.isArray(path) ? path.join('.') : path;
+	if (!selectors[joinedPath]) {
+		selectors[joinedPath] = getToJSMemoized(calledState => get(calledState, path));
+	}
+	return selectors[joinedPath](state);
+}
+
```
VolodymyrKovalM: Functions "get" and "toJS" can be extracted as common, and can be used by collections and components, as they do pretty mutch the same

could be good to hear opinions, whether need to do this, or better leave like this

packages/cmf/src/selectors/components.js
```diff
@@ -0,0 +1,41 @@
...
+	if (typeof collectionPath === 'string') {
+		path = collectionPath.split('.');
+	} else if (Array.isArray(collectionPath)) {
+		path = collectionPath;
+	}
+	if (path) {
+		return state.cmf.components.getIn(path, defaultValue);
+	}
+	throw Error(`Type mismatch: collectionPath should be a string or an array of string
+got ${collectionPath}`);
```
talend-glorieux: Should use a switch case without the extra `path` variable and with the default case throwing the error.

packages/cmf/src/selectors/components.js
```diff
@@ -0,0 +1,41 @@
...
+}
+
+const selectors = {};
+
+export function toJS(state, path) {
```
talend-glorieux: The fact that it is memoized should be reflected in the name.

packages/cmf/__tests__/selectors/components.test.js
```diff
@@ -0,0 +1,62 @@
...
+				'Container(Form)': {
+					MY_FORM_ID: formData,
+				},
+				'Container(List)': {
+					items: [],
+				},
+			}),
+		},
+	};
+	describe('toJS', () => {
```
jsomsanith: `describe` execute a test ?
Same for next tests.

I don't know if that executes well, but try to follow the purpose of jest functions
* `describe` : group tests under a descriptive label. Ex: `describe('toJS')`
* `it`: unit test, try to make it english readable. Ex: `it('should do that when we do this')`

In this case `it('should get requested state and memoize it')`

Why ?
Here the test report will output `selectors.components toJS` which is not a good description of the test.
With the previous suggestion, the test report will output `selector.components toJS should do that when we do this` --> en readable.

packages/cmf/__tests__/selectors/components.test.js
```diff
@@ -0,0 +1,62 @@
...
+		const result2 = selectors.components.toJS(state, 'Container(Form).MY_FORM_ID');
+		expect(result1).toEqual(formData);
+		expect(result1).toBe(result2);
+	});
+
+	describe('get', () => {
+		const collection = fromJS({ items: [] });
+
+		it('try to find the collection if collectionPath is a string', () => {
+			expect(selectors.components.get(state, 'Container(List)')).toEqual(collection);
```
jsomsanith: Can you make it a dot-separated key please ?

packages/cmf/__tests__/selectors/components.test.js
```diff
@@ -0,0 +1,62 @@
...
+		const result2 = selectors.components.getFormDataToJS(state, 'MY_FORM_ID');
+		expect(result1).toEqual(formData.data);
+		expect(result1).toBe(result2);
+	});
+
+	describe('getFormData', () => {
+		const result1 = selectors.components.getFormData(state, 'MY_FORM_ID');
+		const result2 = selectors.components.getFormData(state, 'MY_FORM_ID');
+		expect(result1).toEqual(fromJS(formData.data));
+		expect(result1).toBe(result2);
```
jsomsanith: This returns the immutable state, no memoization, don't need to test 2 results equality, just the get is sufficient

packages/cmf/src/selectors/components.js
VolodymyrKovalM: Hello @talend-glorieux , sorry, seems I didn't quite get what you mean

We use two different expressions in both if cases (typeof and Array.isArray checking), the switch expression is evaluated once, so what is the point to use switch here?

packages/cmf/src/selectors/components.js
```diff
@@ -0,0 +1,41 @@
...
+ * @param {String or Array<String>} collectionPath
+ * @example
+ *  get('foo.bar', true) === state.cmf.components.getIn(['foo', 'bar'], true)
+ */
+export function get(state, collectionPath, defaultValue) {
+	let path;
+	if (typeof collectionPath === 'string') {
+		path = collectionPath.split('.');
+	} else if (Array.isArray(collectionPath)) {
+		path = collectionPath;
```
jmfrancois: please replace collectionPath by componentPath

packages/cmf/src/selectors/components.md
```diff
@@ -0,0 +1,62 @@
...
+| state| redux state |
+| path | path to collections which is stored in cmf.components (support dotted accessor like lodash get) |
+| default value | the default value (optional) |
+
+The selector works the same as selectors.collections.get, just it returns data that is stored in cmf.components,
+not in cmf.collections
+
+## cmf.selectors.components.toJSMemoized
+
+Unlike cmf.selectors.components.get, cmf.selectors.components.toJSMemoized gets collection from state,
```
jmfrancois: this is not a collection here but a component state.
We will have two kind of state. Hard to find a better name here.

packages/cmf/src/selectors/components.md
```diff
@@ -0,0 +1,62 @@
...
+    }
+}
+```
+
+This selector accept the following arguments:
+
+| name | description |
+| -- | -- |
+| state| redux state |
+| path | path to collections which is stored in cmf.components (support dotted accessor like lodash get) |
```
jmfrancois: same here:
`path to component state which is stored in cmf.components...`

PR #1489, opened 2018-06-25
feat(Containers): HeaderBar container

sgangnant:
HeaderBar component requires products props to fill the menu, it can't load data from backend

Wrap the HeaderBar component with a container so that it is able to perform this loading operations.

This PR was originally started on https://github.com/Talend/ui/pull/1485 but I prefer pushing 2 distincts branches to split separate purposes.


packages/containers/src/HeaderBar/__snapshots__/HeaderBar.test.js.snap
```diff
@@ -0,0 +1,34 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Container HeaderBar should render HeaderBar container 1`] = `
+<Inject
+  component="HeaderBar"
+/>
```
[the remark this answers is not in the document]
sgangnant: Generator got me wrong here, it should be changed in `packages/generator` I think then

packages/containers/src/HeaderBar/HeaderBar.sagas.test.js
```diff
@@ -0,0 +1,58 @@
...
+			const httpResponse = { response: { ok: true }, data: 'foo' };
+			const action = { payload: { url: 'url', lang: 'lang', env: 'env' } };
+
+			const gen = fetchProducts(action);
+
+			gen.next(); // Toggle fetching flag (enable)
+			const effect = gen.next().value; // HTTP call
+			expect(effect.CALL).toBeDefined();
+			gen.next(httpResponse); // Toggle fetching flag (disable)
+			gen.next(); // Update CMF collections
```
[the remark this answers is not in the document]
sgangnant: Working on improving assertions right now, but do we really need to ensure that no store update action has been dispatched since we're already verifying every yield of the generator and checking that it has reached its end (`expect(done).toBe(true)`) ?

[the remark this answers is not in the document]
jsomsanith: You're right no need to if you have the check on `done`

packages/containers/src/HeaderBar/HeaderBar.sagas.js
```diff
@@ -0,0 +1,47 @@
...
+	const productsUrl = `${url}?lang=${lang}`;
+
+	yield put(Connected.setStateAction({ fetchingProducts: true }));
+
+	const { response, data } = yield call(cmf.sagas.http.get, productsUrl);
+
+	yield put(Connected.setStateAction({ fetchingProducts: false }));
+
+	if (!response.ok) {
+		return; // Loading products failed, do nothing
```
acateland: i think the error state should be stated clearly on the component state.
Guessing the status and why it happen is only harder when the state
`not_asked`, `fetching`, `error`, `success data` are not clearly stated

packages/containers/src/HeaderBar/HeaderBar.container.js
```diff
@@ -0,0 +1,50 @@
...
+		// Add onClickDispatch event to items
+		products.items = productsItems.map(product => ({
+			onClickDispatch: { type: Constants.HEADER_BAR_OPEN_PRODUCT, payload: product },
+			...product,
+		}));
+
+		componentProps.products = products;
+	} else if (!props.state.get('fetchingProducts')) {
+		// Trigger fetch if not already fetching
+		props.dispatch({
```
acateland: please never dispatch in a render function !

packages/containers/src/HeaderBar/HeaderBar.sagas.js
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	yield put(cmf.actions.collections.addOrReplace(Constants.COLLECTION_ID, data));
+}
+
+/**
+ * Handle the opening of a product from the HeaderBar
+ * @param {Object} action
+ */
+export function handleOpenProduct(action) {
```
acateland: this should live in a component, also this should be handled by a simple html link with a target blank

packages/containers/src/HeaderBar/HeaderBar.container.js
sgangnant: What do you suggest for this ?
We couldn't make it work with that process in componentWillMount

packages/containers/src/HeaderBar/HeaderBar.sagas.js
jmfrancois: no we have some case where we need more (check if a browser extension exists, open the extension or the market place)
That code will goes here later

acateland: i think this info should land on a code comment for future readings and understandings

packages/containers/src/HeaderBar/HeaderBar.sagas.js
```diff
@@ -0,0 +1,51 @@
...
+ */
+export function* fetchProducts(action) {
+	const { url, lang } = action.payload;
+	const productsUrl = `${url}?lang=${lang}`;
+
+	yield put(Connected.setStateAction({ productsFetchState: Constants.FETCHING_PRODUCTS }));
+
+	const { response, data } = yield call(cmf.sagas.http.get, productsUrl);
+
+	if (!response.ok) {
```
acateland: you can switch how the error is handled, this to avoid a negative test expressions and a return inside the if statement

PR #1493, opened 2018-06-26
feat(component/step3): calendar management

jdescottes-talend:
Fully manage the calendar display dynamically.

Do some stuff and things with lots of date-fns magic.
- Throw the ingredients (props) needed to display the right calendar based on month and year previously cooked (selected)
- Achieve some black magic to invoke the right weeks
- Give back the selected/picked date as wanted

### DIFF
--- a/packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js
+++ b/packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js
@@ -62,8 +62,7 @@ class DateTimePicker extends React.Component {
 		if (this.state.isDateTimeView) {
 			viewElement = (<DateTimeView
 				onClickTitle={this.setMonthYearView}
-				selectedMonthIndex={this.state.calendar.monthIndex}
-				selectedYear={this.state.calendar.year}
+				selectedCalendar={this.state.calendar}
 				onSelectMonthYear={this.onSelectCalendarMonthYear}
 				onSelectDate={this.onSelectDate}
 				selectedDate={this.state.selectedDate}
--- a/packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js
+++ b/packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js
@@ -17,6 +17,7 @@ class DateTimePicker extends React.Component {
 				monthIndex: now.getMonth(),
 				year: now.getFullYear(),
 			},
+			selectedDate: undefined,
 		};
 
 		this.setDateTimeView = this.setView.bind(this, true);
@@ -24,6 +25,11 @@ class DateTimePicker extends React.Component {
 		this.onSelectMonth = this.onSelectMonth.bind(this);
 		this.onSelectYear = this.onSelectYear.bind(this);
 		this.onSelectMonthYear = this.onSelectMonthYear.bind(this);
+		this.onSelectDate = this.onSelectDate.bind(this);
+	}
+
+	onSelectDate(selectedDate) {
+		this.setState({ selectedDate });
 	}
 
 	onSelectMonthYear(newCalendar) {
@@ -56,6 +62,8 @@ class DateTimePicker extends React.Component {
 				selectedMonthIndex={this.state.currentCalendar.monthIndex}
 				selectedYear={this.state.currentCalendar.year}
 				onSelectMonthYear={this.onSelectMonthYear}
+				onSelectDate={this.onSelectDate}
+				selectedDate={this.state.selectedDate}
 			/>);
 		} else {
 			viewElement = (<MonthYearView
--- a/packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js
+++ b/packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js
@@ -40,16 +40,12 @@ class DateTimePicker extends React.Component {
 		}));
 	}
 
-	onSelectCalendarMonth(index) {
-		this.onSelectCalendarMonthYear({
-			monthIndex: index,
-		});
+	onSelectCalendarMonth(monthIndex) {
+		this.onSelectCalendarMonthYear({ monthIndex });
 	}
 
-	onSelectCalendarYear(calendarYear) {
-		this.onSelectCalendarMonthYear({
-			year: calendarYear,
-		});
+	onSelectCalendarYear(year) {
+		this.onSelectCalendarMonthYear({ year });
 	}
 
 	setView(isDateTimeView) {
--- a/packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js
+++ b/packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js
@@ -13,7 +13,7 @@ class DateTimePicker extends React.Component {
 
 		this.state = {
 			isDateTimeView: true,
-			currentCalendar: {
+			calendar: {
 				monthIndex: now.getMonth(),
 				year: now.getFullYear(),
 			},
@@ -22,9 +22,9 @@ class DateTimePicker extends React.Component {
 
 		this.setDateTimeView = this.setView.bind(this, true);
 		this.setMonthYearView = this.setView.bind(this, false);
-		this.onSelectMonth = this.onSelectMonth.bind(this);
-		this.onSelectYear = this.onSelectYear.bind(this);
-		this.onSelectMonthYear = this.onSelectMonthYear.bind(this);
+		this.onSelectCalendarMonth = this.onSelectCalendarMonth.bind(this);
+		this.onSelectCalendarYear = this.onSelectCalendarYear.bind(this);
+		this.onSelectCalendarMonthYear = this.onSelectCalendarMonthYear.bind(this);
 		this.onSelectDate = this.onSelectDate.bind(this);
 	}
 
@@ -32,21 +32,25 @@ class DateTimePicker extends React.Component {
 		this.setState({ selectedDate });
 	}
 
-	onSelectMonthYear(newCalendar) {
+	onSelectCalendarMonthYear(newCalendar) {
 		this.setState(previousState => ({
-			currentCalendar: {
-				...previousState.currentCalendar,
+			calendar: {
+				...previousState.calendar,
 				...newCalendar,
 			},
 		}));
 	}
 
-	onSelectMonth(monthIndex) {
-		this.onSelectMonthYear({ monthIndex });
+	onSelectCalendarMonth(calendarMonthIndex) {
+		this.onSelectCalendarMonthYear({
+			monthIndex: calendarMonthIndex,
+		});
 	}
 
-	onSelectYear(year) {
-		this.onSelectMonthYear({ year });
+	onSelectCalendarYear(calendarYear) {
+		this.onSelectCalendarMonthYear({
+			year: calendarYear,
+		});
 	}
 
 	setView(isDateTimeView) {
@@ -59,19 +63,19 @@ class DateTimePicker extends React.Component {
 		if (this.state.isDateTimeView) {
 			viewElement = (<DateTimeView
 				onClickTitle={this.setMonthYearView}
-				selectedMonthIndex={this.state.currentCalendar.monthIndex}
-				selectedYear={this.state.currentCalendar.year}
-				onSelectMonthYear={this.onSelectMonthYear}
+				selectedMonthIndex={this.state.calendar.monthIndex}
+				selectedYear={this.state.calendar.year}
+				onSelectMonthYear={this.onSelectCalendarMonthYear}
 				onSelectDate={this.onSelectDate}
 				selectedDate={this.state.selectedDate}
 			/>);
 		} else {
 			viewElement = (<MonthYearView
 				onClickBack={this.setDateTimeView}
-				selectedMonthIndex={this.state.currentCalendar.monthIndex}
-				selectedYear={this.state.currentCalendar.year}
-				onSelectMonth={this.onSelectMonth}
-				onSelectYear={this.onSelectYear}
+				selectedMonthIndex={this.state.calendar.monthIndex}
+				selectedYear={this.state.calendar.year}
+				onSelectMonth={this.onSelectCalendarMonth}
+				onSelectYear={this.onSelectCalendarYear}
 			/>);
 		}
 
--- a/packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js
+++ b/packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js
@@ -17,7 +17,6 @@ class DateTimePicker extends React.Component {
 				monthIndex: now.getMonth(),
 				year: now.getFullYear(),
 			},
-			selectedDate: undefined,
 		};
 
 		this.setDateTimeView = this.setView.bind(this, true);
@@ -41,9 +40,9 @@ class DateTimePicker extends React.Component {
 		}));
 	}
 
-	onSelectCalendarMonth(calendarMonthIndex) {
+	onSelectCalendarMonth(index) {
 		this.onSelectCalendarMonthYear({
-			monthIndex: calendarMonthIndex,
+			monthIndex: index,
 		});
 	}
 
--- a/packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js
+++ b/packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js
@@ -58,7 +58,7 @@ class DateTimePicker extends React.Component {
 		if (this.state.isDateTimeView) {
 			viewElement = (<DateTimeView
 				onClickTitle={this.setMonthYearView}
-				selectedCalendar={this.state.calendar}
+				calendar={this.state.calendar}
 				onSelectMonthYear={this.onSelectCalendarMonthYear}
 				onSelectDate={this.onSelectDate}
 				selectedDate={this.state.selectedDate}
--- a/packages/components/src/DateTimePickers/pickers/DatePicker/DatePicker.component.js
+++ b/packages/components/src/DateTimePickers/pickers/DatePicker/DatePicker.component.js
@@ -55,7 +55,6 @@ class DatePicker extends React.Component {
 
 		this.getWeeks = memoize(DatePicker.buildWeeks, (year, monthIndex, firstDayOfWeek) => `${year}-${monthIndex}|${firstDayOfWeek}`);
 
-		this.selectedDay = new Date(2018, 5, 12);
 		this.disabledDays = [
 			new Date(2018, 5, 6),
 			new Date(2018, 5, 15),
@@ -63,7 +62,8 @@ class DatePicker extends React.Component {
 	}
 
 	isSelectedDay(date) {
-		return isSameDay(this.selectedDay, date);
+		return this.props.selectedDate !== undefined
+			&& isSameDay(this.props.selectedDate, date);
 	}
 
 	isDisabledDay(date) {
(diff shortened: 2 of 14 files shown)


packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js
```diff
@@ -13,34 +13,44 @@ class DateTimePicker extends React.Component {
 
 		this.state = {
 			isDateTimeView: true,
-			currentCalendar: {
+			calendar: {
 				monthIndex: now.getMonth(),
 				year: now.getFullYear(),
 			},
+			selectedDate: undefined,
```
jmfrancois: useless, you can remote it

packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js
```diff
@@ -13,34 +13,44 @@ class DateTimePicker extends React.Component {
...
 				...newCalendar,
 			},
 		}));
 	}
 
-	onSelectMonth(monthIndex) {
-		this.onSelectMonthYear({ monthIndex });
+	onSelectCalendarMonth(calendarMonthIndex) {
+		this.onSelectCalendarMonthYear({
+			monthIndex: calendarMonthIndex,
```
jmfrancois: please rename calendarMonthIndex -> index
by the way index is 0->11 ?
if 1_>12 it s not an index is the the month number

packages/components/src/DateTimePickers/pickers/DatePicker/DatePicker.component.js
```diff
@@ -1,83 +1,170 @@
...
 import theme from './DatePicker.scss';
 import DayPickerAction from './DayPickerAction';
 
-function DatePicker(props) {
-	const dayNames = ['Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday', 'Sunday'];
-	const days = (new Array(7))
-					.fill(0)
-					.map((_, i) => i + 1);
+const FIRST_DAY_OF_WEEK = 1;
+const BASE_DATE = new Date(0);
```
jmfrancois: If I let open my app at 23h50 and I have a re-render at 00h01 ?

packages/components/src/DateTimePickers/pickers/DatePicker/DatePicker.scss
```diff
@@ -19,20 +18,34 @@
...
+
+    .calendar-body {
+        display: flex;
+        flex-direction: column;
+        justify-content: space-evenly;
+        flex: 1;
+
+        .calendar-row {
+            flex: 1;
+        }
```
jmfrancois: with css module you really don t need cascading, one level is enough, two is ok, three is too much deep.

packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js
jdescottes-talend: it's really the index [0-11] like in Date API

packages/components/src/DateTimePickers/pickers/DatePicker/DatePicker.component.js
jdescottes-talend: Yes i forgot to ask the question, it was in my head too.
But it's not this line the problem it's in the DateTimeView.
(new Date(0) => 0ms from Date 1970-01-01T00:00:00.000Z, so not relative to current time)

```javascript
class DateTimeView extends React.Component {

	constructor(props) {
		super(props);

		this.state = {
			today: new Date(),
		};
```

Do I need to recreate the today date at each render ?
Or i set a timeout to update the today date in state when we are changing day to keep the render pure ?

packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js

packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js

packages/components/src/DateTimePickers/pickers/DatePicker/DatePicker.scss
jdescottes-talend: ✔

packages/components/src/DateTimePickers/pickers/DatePicker/DatePicker.component.js
jdescottes-talend: Check if today directly in the picker which is no more pure.
Need a re-render to update the visual today day, not re-render if staying in front of the calendar while there is a date changed. (23:59 - 00:00)

jdescottes-talend: ✔

packages/components/src/DateTimePickers/views/DateTimeView/DateTimeView.test.js
```diff
@@ -10,8 +10,13 @@ describe('DateTimeView', () => {
 			selectedYear={2006}
 			onClickTitle={() => {}}
 			onSelectMonthYear={() => {}}
+			onSelectDate={() => {}}
 		/>);
 
+		wrapper.setState({
+			today: new Date(2018, 5, 20),
+		});
```
jmfrancois: not more needed

jdescottes-talend: ✔

packages/components/src/DateTimePickers/pickers/DatePicker/DatePicker.component.js
```diff
@@ -1,83 +1,166 @@
...
-					<abbr
-						className={theme['calendar-item']}
-						key={i}
-						title={dayName}
-					>
-						{dayName.charAt(0)}
-					</abbr>
-				)}
-			</div>
+	isDisabledDate(dateToCheck) {
```
jsomsanith: This seems is quite a complicated api + code to handle that.

You should accept an array of functions, or an array of functions or dates. 
But not an array of arrays :/
Moreover, the condition of the top level array is an OR, and the conditions between each element in the nested arrays are OR too. So let's flatten.

BTW, What is the use case of having a specific date disabled ?

packages/components/src/DateTimePickers/pickers/DatePicker/DatePicker.component.js
```diff
@@ -1,83 +1,166 @@
...
+				}
+			});
+	}
+
+	isCurrentMonth(date) {
+		return getMonth(date) === this.props.calendar.monthIndex;
+	}
+
+	render() {
+		const { year, monthIndex } = this.props.calendar;
```
jsomsanith: I would like to raise an api manipulation issue here.
In `DateTimePicker`, you have `state.calendar`. You destructure it to pass month and year to `DateTimeView`.
In `DateTimeView`, you gather them again to pass a unique prop `calendar`, and pass it to `DatePicker`.
In `DatePicker`, you destructure it again to have month and year.

Can you uniformise please ? Either pass down calendar everywhere, or pass down month and year everywhere.

packages/components/src/DateTimePickers/pickers/DatePicker/DatePicker.component.js
```diff
@@ -1,83 +1,166 @@
...
+								>
+									{
+										this.isCurrentMonth(date) &&
+										<DayPickerAction
+											label={getDate(date).toString()}
+											isSelected={this.isSelectedDate(date)}
+											isDisabled={this.isDisabledDate(date)}
+											isToday={isToday(date)}
+											aria-label={this.isDisabledDate(date)
+												? 'Unselectable date'
```
jsomsanith: Remove that since you don't render it if it is disabled.
And it avoid testing the disable status twice

packages/components/src/DateTimePickers/pickers/DatePicker/DatePicker.component.js
```diff
@@ -1,83 +1,166 @@
...
-							/>
+							{week.map((date, j) =>
+								<div
+									className={theme['calendar-item']}
+									key={j}
+								>
+									{
+										this.isCurrentMonth(date) &&
+										<DayPickerAction
+											label={getDate(date).toString()}
```
jsomsanith: the `toString()` is unnecessary.
You can change the propTypes in `DatePickerAction` too

packages/components/src/DateTimePickers/pickers/DatePicker/DatePicker.test.js
```diff
@@ -1,14 +1,572 @@
...
+
+				const wrapper = shallow(<DatePicker
+					calendar={calendar}
+					onSelect={() => {}}
+				/>);
+
+				expect(getFirstRowCellText(wrapper, 1)).toBe('1');
+			});
+
+			it('should have thursday in the 4th column for march 2018 ', () => {
```
jsomsanith: Hmm not sure what you test in the 3 tests here
DayOfWeek position does not depends on the month/year.

packages/components/src/DateTimePickers/pickers/DatePicker/DatePicker.test.js
```diff
@@ -1,14 +1,572 @@
...
+				return wrapper
+					.find('.theme-calendar-body-row')
+					.last()
+					.find('.theme-calendar-item')
+					.at(column - 1)
+					.find('DayPickerAction')
+					.prop('label');
+			}
+
+			it('should have monday in the 1st column for april 2018', () => {
```
jsomsanith: Same here Oo

packages/components/src/DateTimePickers/pickers/DatePicker/DatePicker.test.js
```diff
@@ -1,14 +1,572 @@
...
+				const wrapper = shallow(<DatePicker
+					calendar={calendar}
+					onSelect={() => {}}
+				/>);
+
+				expect(nbOfWeeksRendered(wrapper)).toBe(6);
+			});
+		});
+
+		describe('contiguous ordered numbers', () => {
```
jsomsanith: This section is implicitly tested in the snapshot of the 1st test. You can remove that (and don't need to test that for multiple days)

packages/components/src/DateTimePickers/pickers/DatePicker/DatePicker.test.js
```diff
@@ -1,14 +1,572 @@
...
+					onSelect={() => {}}
+				/>);
+
+				expect(getLastRowCellText(wrapper, 7)).toBe('30');
+			});
+		});
+
+		describe('number of weeks displayed', () => {
+			function nbOfWeeksRendered(wrapper) {
+				return wrapper.find('.theme-calendar-body-row').length;
```
jsomsanith: Never rely on `theme-my-classname` classNames. Those are generated with some config.
Add the `my-classname` in addition to the css module in you code. So your test depends in your code, and not the test configuration

packages/components/src/DateTimePickers/pickers/DatePicker/DatePicker.test.js
```diff
@@ -1,14 +1,572 @@
...
+				/>);
+
+				const currentDayItems = wrapper
+					.find('.theme-calendar-body-row .theme-calendar-item DayPickerAction')
+					.filterWhere(item => item.prop('isToday'));
+
+				expect(currentDayItems).toHaveLength(0);
+			});
+
+			it.only('should have updated the current today day if has changed between two renders', () => {
```
jsomsanith: You tested that :D
but what is this `it.only` ?

packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js
```diff
@@ -13,34 +13,43 @@ class DateTimePicker extends React.Component {
...
 				...newCalendar,
 			},
 		}));
 	}
 
-	onSelectMonth(monthIndex) {
-		this.onSelectMonthYear({ monthIndex });
+	onSelectCalendarMonth(index) {
+		this.onSelectCalendarMonthYear({
+			monthIndex: index,
```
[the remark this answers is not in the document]
jdescottes-talend: because @jmfrancois ...

packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js
```diff
@@ -13,34 +13,43 @@ class DateTimePicker extends React.Component {
...
+		});
 	}
 
-	onSelectYear(year) {
-		this.onSelectMonthYear({ year });
+	onSelectCalendarYear(calendarYear) {
+		this.onSelectCalendarMonthYear({
+			year: calendarYear,
```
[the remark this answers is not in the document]
jdescottes-talend: because i renamed to fast some variables with calendar to split the infos for calendar and the final date

packages/components/src/DateTimePickers/pickers/DatePicker/DatePicker.component.js
jdescottes-talend: i suppose disabled for :
- date min/max allowed
- range white or black list
- specific dates white or black list

I don't know the real constraints

But an array of array have some sens.
The first one is just for a set of rules, it can actually be one single date, a list of single date, and later eventually a range, a min/max, a custom function, etc. So the second array is just a specific rule.

But yes we don't know the needs actually, moreover, the month and year picker will need to have some disabled actions too. So maybe handle that at this level is too low, maybe handle that at a higher component level and have just function as prop to check dates (months and years) is fine.

packages/components/src/DateTimePickers/pickers/DatePicker/DatePicker.test.js
jdescottes-talend: oups, already commit a fix to remove the '.only'.
✔

packages/components/src/DateTimePickers/pickers/DatePicker/DatePicker.component.js
jdescottes-talend: When I think of `label`, I think string that's why.
If we give directly a number, we need to change the name and if we do that we can change to a full date, and delegate the extraction of day to the `DayPickerAction` component...
Lot's of thought for a `toString()` oO

packages/components/src/DateTimePickers/pickers/DatePicker/DatePicker.component.js
jdescottes-talend: ❓

packages/components/src/DateTimePickers/pickers/DatePicker/DatePicker.component.js
jdescottes-talend: The action/button is always render, just `disabled`.
But i can memoize or put a variable for re-usability or just have the same aria-label but it seems misleading ❓

packages/components/src/DateTimePickers/pickers/DatePicker/DatePicker.test.js
jdescottes-talend: ✔

packages/components/src/DateTimePickers/pickers/DatePicker/DatePicker.component.js
jdescottes-talend: ✔

packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js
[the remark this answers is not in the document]

packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js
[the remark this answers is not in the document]

packages/components/src/DateTimePickers/pickers/DatePicker/DatePicker.test.js
jdescottes-talend: I renamed some tests.
What I test here is the column position of the first day of month accordingly to its day of week.
=> If first day is a friday, it must be in 5th column

jdescottes-talend: ✔

packages/components/src/DateTimePickers/pickers/DatePicker/DatePicker.test.js
jdescottes-talend: ✔

packages/components/src/DateTimePickers/pickers/DatePicker/DatePicker.test.js
jdescottes-talend: ✔

packages/components/src/DateTimePickers/pickers/DatePicker/DatePicker.component.js
jsomsanith: You can do that with an array of (Date || function)
If you have a function, you run it and take the result as boolean.
If you have a Date, you test `isSameDay`

No need to take an array of arrays.
* date min/max allowed: function
* range white or black list: function 
* specific dates black list: Dates
* specific dates white list: function

packages/components/src/DateTimePickers/pickers/DatePicker/DatePicker.component.js
jsomsanith: you still have a name mismatch
`DateTimePicker`: state.calendar
`DateTimeView`: props.selectedCalendar
`DatePicker`: props.calendar

packages/components/src/DateTimePickers/pickers/DatePicker/DatePicker.component.js
jsomsanith: Oh you're right, I misread this code.
Yes please, check it once and use the result twice

packages/components/src/DateTimePickers/pickers/DatePicker/DatePicker.component.js
jsomsanith: or you can accept `string || number`. And no need to call this toString()

packages/components/src/DateTimePickers/pickers/DatePicker/DatePicker.test.js
jsomsanith: The only differences between your cases are the number of day.
You can remove the `continuous` part in description and assertion. This is tested in the snapshot. Here you change the month/year to change the number of days, just assert that.

packages/components/src/DateTimePickers/pickers/DatePicker/DatePicker.test.js
jsomsanith: We have a sort of convention to have the classnames be like
`tc-${component-name}-${classname-qualifiers}`

tc: Talend-Component
(If forms package we have tf)

packages/components/src/DateTimePickers/pickers/DatePicker/DatePicker.component.js
jdescottes-talend: Removed `selected` wording in `DateTimeView`
✔

packages/components/src/DateTimePickers/pickers/DatePicker/DatePicker.component.js
jdescottes-talend: ✔

packages/components/src/DateTimePickers/pickers/DatePicker/DatePicker.component.js
jdescottes-talend: ✔

packages/components/src/DateTimePickers/pickers/DatePicker/DatePicker.component.js
jdescottes-talend: I don't really understand what's wrong with `toString()` but whatever...
✔

packages/components/src/DateTimePickers/pickers/DatePicker/DatePicker.test.js
jdescottes-talend: ✔

packages/components/src/DateTimePickers/pickers/DatePicker/DatePicker.test.js
jdescottes-talend: ✔

PR #1495, opened 2018-06-27
feat(ListView): improve a11y

jsomsanith:
* no alternative texts on icon buttons
* checkboxes label not clear (only the entry label)
* no outline

* Add alternative texts
* set back the outline
* Bonus: fix style


packages/components/.storybook/preview-head.html
```diff
@@ -1,12 +1,6 @@
 <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/font-awesome/4.6.3/css/font-awesome.min.css">
 <!-- Storybook should have visible overflow since we have Action Logger -->
 <style>
-	body {
-		overflow-y: auto !important;
-		max-height: none !important;
-		min-height: 100vh;
```
jsomsanith: Seems to do nothing (perhaps with the upgrade of storybook).
Even worse, it introduce a double scroll

jdescottes-talend: You could have commit that in an other commit, even in an another PR :s
Same for the package.json on this commit

packages/components/src/ListView/Items/Item/Item.component.js
```diff
@@ -42,6 +44,17 @@ class Item extends Component {
 			{ 'switch-nested': children },
 			{ switch: isSwitchBox },
 		);
+		const checked = !!item.checked;
+		const ariaLabel = checked
+			? t('TC_LISTVIEW_DESELECT', { defaultValue: `Deselect {{ value }}`, value: item.label })
```
annilsson: Should probably use single quotes in `defaultValue` to avoid confusion here

packages/components/src/ListView/Items/Item/Item.component.js
```diff
@@ -42,6 +44,17 @@ class Item extends Component {
...
+			: t('TC_LISTVIEW_SELECT', { defaultValue: `Select {{ value }}`, value: item.label });
```
annilsson: Should probably use single quotes in `defaultValue` to avoid confusion here

packages/components/src/ListView/Items/Item/Item.component.js
```diff
@@ -42,6 +44,17 @@ class Item extends Component {
...
+
+		let expandLabel;
+		if (children) {
+			expandLabel = item.expanded
+				? t('TC_LISTVIEW_COLLAPSE', { defaultValue: `Collapse {{ value }}`, value: item.label })
```
annilsson: Should probably use single quotes in `defaultValue` to avoid confusion here

packages/components/src/ListView/Items/Item/Item.component.js
```diff
@@ -42,6 +44,17 @@ class Item extends Component {
...
+				: t('TC_LISTVIEW_EXPAND', { defaultValue: `Expand {{ value }}`, value: item.label });
```
annilsson: Should probably use single quotes in `defaultValue` to avoid confusion here

packages/components/src/ListView/Items/Item/Item.component.js
```diff
@@ -1,14 +1,16 @@
...
 import ItemPropTypes from './Item.propTypes';
 import Action from '../../../Actions/Action';
+import I18N_DOMAIN_COMPONENTS from '../../../constants';
 
 function itemLabelClasses() {
 	return classNames(theme['tc-listview-item-label'], 'tc-listview-item-label');
 }
 
-class Item extends Component {
+export class ItemComponent extends Component {
```
annilsson: Just a heads-up, Was this added to be imported by the test file? It's still importing the default

jdescottes-talend: What's the convention for the components suffix when wrapped or will be wrapped ?

packages/components/src/ListView/Items/Items.component.js
```diff
@@ -110,18 +110,18 @@ export class ItemsComponent extends React.PureComponent {
 
 	renderToggleAll() {
 		const { id, isSwitchBox, toggleAllChecked, onToggleAll, t } = this.props;
+		const checked = !!toggleAllChecked;
```
jdescottes-talend: Default props could be used to avoid bang operator

packages/components/src/ListView/Items/Item/Item.component.js
```diff
@@ -42,6 +44,17 @@ class Item extends Component {
 			{ 'switch-nested': children },
 			{ switch: isSwitchBox },
 		);
+		const checked = !!item.checked;
```
jdescottes-talend: Default props could be used to avoid bang operator

packages/components/src/ListView/Items/__snapshots__/items.test.js.snap
```diff
@@ -280,28 +280,65 @@ exports[`Items should render 1`] = `
                       }
                       measure={[Function]}
                     >
-                      <div>
-                        <div
-                          className="checkbox-container"
+                      <I18n
```
jdescottes-talend: Like @andreasmattes said in an other way, I don't think the actual snapshots should have been updated (not so much though) if keeping the underlying component for most of the cases, and just adding some tests for translation.

jsomsanith: It should. This is on `Items` that import `Item`. But `Item` default export has changed, it's not the component but the i18n wrapped component. That's why we have this difference

packages/components/src/ListView/Items/Item/Item.component.js
jsomsanith: In fact we don't even need the default props. I just extracted the checked value that is passed in the input props, but even with undefined it's ok. I remove that

packages/components/src/ListView/Items/Items.component.js
jsomsanith: In fact we don't even need the default props. Let's test/pass it directly

packages/components/src/ListView/Items/Item/Item.component.js
jsomsanith: @jdescottes-talend there is no real convention but in practice, to stay in line with the existing code, when you need to export a component that is not the default export you add the `Component` suffix. Changing the name helps user to avoid mistake in their imports.

PR #1503, opened 2018-06-29
fix(cmf): cancel one saga

jare-talend:
when i launch a component with multiple instance in the same time, when one instance is unmounted, WILL_UMOUNT is listened by all and cancel all tasks

fix that


packages/cmf/package.json
```diff
@@ -93,7 +93,8 @@
     "redux-storage-decorator-filter": "^1.1.8",
     "redux-storage-decorator-immutablejs": "^1.0.4",
     "redux-storage-engine-localstorage": "^1.1.4",
-    "redux-thunk": "^2.2.0"
+    "redux-thunk": "^2.2.0",
+    "uuid": "^3.3.0"
```
jmfrancois: When you add a lib please check versions.js first 

```javascript
uuid: '^3.0.1', // prefer bson-objectid
```

So please replace it with bson-objectid

packages/cmf/src/cmfConnect.js
```diff
@@ -255,12 +256,19 @@ export default function cmfConnect({
 				super(props, context);
 				this.dispatchActionCreator = this.dispatchActionCreator.bind(this);
 				this.getOnEventProps = this.getOnEventProps.bind(this);
+				this.state = {
+					componentUuid: uuid.v4(),
+				};
```
jmfrancois: I would prefer this.uuid = bsonObjectid().toString().
I don t think we need to re-render it, it s very internal and defined in the constructor.

packages/cmf/src/sagas/component.js
```diff
@@ -33,8 +33,13 @@ export function* onSagaStart(action) {
 		);
 	} else {
 		const task = yield fork(saga, action.props);
-		yield take(`${CONST.WILL_UNMOUNT_SAGA_STOP}_${action.saga}`);
-		yield cancel(task);
+		// eslint-disable-next-line no-constant-condition
+		while (true) {
+			const stopAction = yield take(`${CONST.WILL_UNMOUNT_SAGA_STOP}_${action.saga}`);
```
jdescottes-talend: You can use the [`take(pattern)`](https://redux-saga.js.org/docs/api/index.html#takepattern) where pattern is a function matching. It can then be simplify something like :

```javascript
yield take(a =>
    a.type === `${CONST.WILL_UNMOUNT_SAGA_STOP}_${action.saga}`
    && a.event.componentId === action.event.componentId
);
yield cancel(task);
```
Without any loop and if statements

PR #1516, opened 2018-07-05
feat(component): overlay always visible

jare-talend:

- The overlay is fixed with the body. If the overlay is in a list and when the user scrolls, the overflow don't follow the scroll.

if i open an overlay in the bottom of the viewport, the overlay isn't fully visible. Like the ActionDropDown, if the overlay is in tc-dropdown-container, we determinate automatically the placement to get the overlay always visible.

add property preventScrolling to set the overlay in the ActionButton


packages/components/src/Actions/ActionButton/ActionButtonOverlay.component.js
```diff
@@ -0,0 +1,104 @@
...
+		const overlayRect = overlayElement.getBoundingClientRect();
+		const triggerRect = this.triggerElement.getBoundingClientRect();
+
+		this.setState(previousState => {
+			const adaptedPlacement = getAdaptedPlacement(
+				triggerRect,
+				overlayRect,
+				containerRect,
+				previousState.placement,
+			);
```
jmfrancois: I think overlay should provide a more high level API, so getPlacement(overlayElement,   previousPlacement)

packages/components/src/Actions/ActionButton/ActionButton.component.js
```diff
@@ -1,7 +1,7 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 import classNames from 'classnames';
-import { Button, OverlayTrigger, Popover } from 'react-bootstrap';
+import { Button, OverlayTrigger } from 'react-bootstrap';
```
jmfrancois: OverlayTrigger seems not used

packages/components/src/Actions/ActionButton/ActionButtonOverlay.component.js
```diff
@@ -0,0 +1,104 @@
...
+			if (!adaptedPlacement) {
+				return null;
```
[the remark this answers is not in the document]
jare-talend: when setState return null, react no re-renderer the component

packages/components/src/Actions/ActionButton/ActionButtonOverlay.component.js
```diff
@@ -0,0 +1,112 @@
...
+			ref: this.props.overlayRef,
+			rootClose: true,
+			trigger: 'click',
+		};
+
+		if (this.props.preventScrolling) {
+			props = {
+				...props,
+				container: this,
+			};
```
jmfrancois: `props.container = this`

packages/components/src/Actions/ActionButton/ActionButton.component.js
```diff
@@ -212,14 +208,11 @@ ActionButton.propTypes = {
...
-	overlayId: PropTypes.string,
-	overlayComponent: PropTypes.element,
-	overlayPlacement: OverlayTrigger.propTypes.placement,
-	overlayRef: PropTypes.func,
-	tooltipPlacement: OverlayTrigger.propTypes.placement,
+	tooltipPlacement: BaseOverlayTrigger.propTypes.placement,
 	t: PropTypes.func,
 	tooltip: PropTypes.bool,
 	tooltipLabel: PropTypes.string,
+	...overlayPropTypes,
```
jdescottes-talend: @jmfrancois Could you explain me some time why flat component props is a good thing ? I can't find any good reason other than a false simplicity reason because of not nesting.

In the other hand this generate a potential props override if conflict of name :
   - So need to check carefully mentally that's not the case across multiple components/files (cerebral useless pain)
   - Or need to name things super specifically, which break the encapsulation scope by including parent name in children things (component name in props prefix) => which is actually more or less the case in ` packages/components/src/OverlayTrigger/OverlayTrigger.component.js`. Plus add more verbosity and useless informations to mentally process/truncate

And it also in my opinion turn the props interface to a more complicated approach because when we use it we need to check all props one at a time. Rather than in the other case, we can check if we need a feature and discard all informations of this context quickly if no needed (example here: tooltip, overlay, are two specific feature of the action button, it's not directly related to). And if we need a feature, **only** at this time, we can follow the rabbit hole.

packages/components/src/OverlayTrigger/overlay.js
```diff
@@ -0,0 +1,109 @@
...
+const OVERLAY_CONTAINER_CLASS_NAME = 'tc-dropdown-container';
+const REACT_BOOSTRAP_OVERLAY_CLASS_NAME = 'popover';
+
+/**
+ * getOverlayElement - return the overlay element
+ *
+ * @param  {Element} innerElement the root element of the action
+ * @return {Element}              return the overlay component
+ */
+export function getOverlayElement(innerElement) {
```
jdescottes-talend: Native API is always better code because less code :)
https://developer.mozilla.org/en-US/docs/Web/API/Element/closest

Can totally delete the function and directly use something of this smell.
```javascript
innerElement.closest(`.${REACT_BOOSTRAP_OVERLAY_CLASS_NAME}`); 
```
Need to check if compliant with browsers range supported in our projects (I don't know the real exacts versions), there is still a polyfill for some oldest version (not before IE9, etc. however).

packages/components/src/OverlayTrigger/overlay.js
```diff
@@ -0,0 +1,109 @@
...
+ * @return {Element}                 return the container if exists
+ */
+export function getContainerElement(overlayElement) {
+	let containerElement = overlayElement;
+	do {
+		containerElement = containerElement.parentElement;
+	} while (
+		containerElement &&
+		containerElement.tagName !== BODY &&
+		!containerElement.classList.contains(OVERLAY_CONTAINER_CLASS_NAME)
```
jdescottes-talend: Seems doable the same way (if I understand well the goal)

```javascript
overlayElement.closest(`${BODY} .${OVERLAY_CONTAINER_CLASS_NAME}`);
```

packages/components/src/OverlayTrigger/overlay.js
jare-talend: nice but no supprt on IE :)

jdescottes-talend: > For browsers that do not support Element.closest(), but carry support for element.matches() (or a prefixed equivalent, meaning IE9+), a polyfill exists

packages/components/src/OverlayTrigger/overlay.js
```diff
@@ -0,0 +1,109 @@
...
+	};
+}
+
+/**
+ * getAdaptedPlacement - return the adapted placement for the overlay
+ * @param  {DOMRect} triggerRect   bound of the trigger element
+ * @param  {DOMRect} overlayRect   bound of the overlay element
+ * @param  {DOMRect} containerRect    bound of the container element
+ * @param  {string} currentPlacement current placement of the overlay
+ * @return {string|false}                  return the adapted placement
```
jdescottes-talend: It would be more easy to reason about if the return value was of the same type all the time.
Currently the doc say return null and the code return false.
But in all cases, i think this function should return a placement if the overlay has to be visible. Even if no space to the top or the bottom because the initial placement is kept.

packages/components/src/OverlayTrigger/overlay.js
jare-talend: need to implement babel-polyfill to implement that.

PR #1517, opened 2018-07-05
feat(cmf): pass componentId in forked saga

jmfrancois:

when component on didMount fork a saga it s to handle business.
At the moment we pass a subset of props received by the component.

This approche as multiple issues:
* props are not supposed to be constant during the life of the component. So you should not rely on it
* we often need to read or update the redux internal state of the corresponding component but for that we need it s componentId which is omited by the current implementation
* the props are passed using an redux action which are not supposed to be functions and so on.

Pass the componentId, because its a well known props of a component that is *supposed* to not change (or you will have other issues)
Keep the props passed with it but its DEPRECATED

So some usecases where we use props:
* read the router
* take some params (the saga is generic)

Now CMF use react-router-redux so you have to read the state to get info from the router.
CMF provides API for that (cmf.selectors.router.getPath + cmf.router.matchPath).

For parameters we have added props.saga support to be an object:

```javascript
<MyComponent saga={{ id: 'mysaga', args:['datasets'] }} />
```

### DIFF
--- a/packages/cmf/src/cmfConnect.js
+++ b/packages/cmf/src/cmfConnect.js
@@ -246,7 +246,7 @@ export default function cmfConnect({
 						'cmf.saga.start',
 						{ type: 'DID_MOUNT', componentId: this.id },
 						{
-							...this.props,  // DEPRECATED
+							...this.props, // DEPRECATED
 							componentId: getComponentId(componentId, this.props),
 						},
 					);
--- a/packages/cmf/src/cmfConnect.js
+++ b/packages/cmf/src/cmfConnect.js
@@ -40,12 +40,9 @@ import actionCreator from './actionCreator';
 import component from './component';
 import CONST from './constant';
 import expression from './expression';
-import deprecated from './deprecated';
 import onEvent from './onEvent';
 import { initState, getStateAccessors, getStateProps } from './componentState';
 import { mapStateToViewProps } from './settings';
-
-let newState;
 
 export function getComponentName(WrappedComponent) {
 	return WrappedComponent.displayName || WrappedComponent.name || 'Component';
@@ -62,19 +59,6 @@ export function getComponentId(componentId, props) {
 	return 'default';
 }
 
-function oldGetCollection(id) {
-	return newState.cmf.collections.get(id);
-}
-
-const getCollection = deprecated(
-	oldGetCollection,
-	`This function will be deprecated,
-	since it permit store access outside cmfConnect mapStateToProps function
-	and maybe not executed if cmf connect do not detect ref change to props
-	given to the component using this function
-	Please bind your collection update to your component using mapStateToProps`,
-);
-
 export function getStateToProps({
 	defaultProps,
 	componentId,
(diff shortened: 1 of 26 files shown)


packages/cmf/src/cmfConnect.js
```diff
@@ -245,7 +245,10 @@ export default function cmfConnect({
 					this.dispatchActionCreator(
 						'cmf.saga.start',
 						{ type: 'DID_MOUNT', componentId: this.id },
-						this.props,
+						{
+							componentId: this.props.componentId || 'default',
```
jare-talend: can you reuse the function getComponentId in CmfConnect to get the componentId.

jare-talend: How my custom saga can use the props ?

jmfrancois: I have updated the doc

jmfrancois:
OK So I remove the breaking change to keep compat and I add the saga with args feature so we just deprecate the use of the props

packages/cmf/src/cmfConnect.md
```diff
@@ -242,7 +242,16 @@ export default function* myDeLorean() {
 If you have multiple instance of the same component those api support `id` as a second argument.
 
 ```javascript
-const componentState = Clock.getState(state, 'a-component-id');
+import Clock from './Clock.connect';
```
acateland: can you provide a concrete example on why we need this feature and how to use it

packages/cmf/src/sagas/component.js
```diff
@@ -32,14 +31,27 @@ export const isActionCancelable = curry(
...
+		throw new Error(`no saga found in action: ${JSON.stringify(action)}`);
+	}
+	const saga = get(sagaId);
 	if (!saga) {
-		invariant(
-			process.env.NODE_ENV === 'production',
-			`You cannot register undefined as saga for id "${action.saga}"`,
-		);
+		throw new Error(`saga not found: ${sagaId}`);
 	} else {
```
jdescottes-talend: Is `else` really needed ?

jdescottes-talend:
I was thinking, more than passing the componentId, we could give some sagas with the componentId already in the context.
For exemple a `getComponentState() : state` and a `setComponentState(state)` ish
In this way we could give to all sub business sagas under the DID_MOUNT one, a first object/context/util thing with a way to manipulate the state (and other CMF things i've not in mind) without knowing the structure of the store neither the actual componentId.
In a similiar way the `dispatch` function is given in component props.

Sorry if it's not clear ;)

packages/cmf/src/cmfConnect.md
jdescottes-talend: If i understand, the feature already exists, having access to componentId in sagas but by the way of component props.
```javascript
myDeLorean({ componentId, props }) {
```
`props.componentId` should be equal to the new `componentId` in most cases unless intentional mind fuck...

And now we can access and rely on a more stable information.

But what of the `default` componentId ? which is `undefined` via component props if not explicitly set. Is it a real 'default' string in the new version or undefined ? i.e. componentId in CMF would be super great to have a consistant string type, at least in the app point of view (external interface of CMF)

jmfrancois:
yes @jdescottes-talend 
It would be even better to give accessors

I will try different approch around that.

At the moment the static accessors + componentId works well if needed.

packages/cmf/src/cmfConnect.md
jmfrancois: about the componentId I would like it to be only a props, but it can be set as a function in cmfConnect.
Yet another breaking change...

PR #1518, opened 2018-07-06
fix(containers/DeleteResource): fix redirect saga, etc

yyanwang:

resourceType is used for two purpose:
* it is the collection id where all resources of this type is stored in redux
* it is the fragment of URL to call to delete

Most of the time it can be different.

Change several things about DeleteResource:
* add `redirectUrl` to action model data
* add prop `collectionId` in which stores resources. If not provided then use `resourceType` as collection name.
* add prop `resourceUri` which is used as backend api to delete resource. if not provided then use `${uri}/${resourceType}/${resourceId}`;
* rename actions from `Yes/No` to `DELETE/CANCEL`(according to guidelines).


packages/containers/src/DeleteResource/DeleteResource.container.js
```diff
@@ -46,7 +48,9 @@ export class DeleteResource extends React.Component {
 	 */
 	getLabel() {
 		return {
-			label: this.props.resource ? this.props.resource.get('label', '') : '',
+			label: this.props.resource
```
acateland: could be extracted in its own function

packages/containers/src/DeleteResource/README.md
```diff
@@ -82,11 +82,14 @@ Required :
 * **redirectUrl** : is the url to redirect when delete is complete or cancel action is triggered
 
 Optional :
+
```
jare-talend: Could you add a example to use it

packages/containers/src/DeleteResource/DeleteResource.container.js
yyanwang: done

PR #1520, opened 2018-07-09
fix(components): support ActionButton with overlay in the list

jare-talend:

* Action in the virtualized list doesn't use the containerized Action.
To use a containerized Action, I have to pass a getComponent in the Action props. https://github.com/Talend/ui/blob/master/packages/components/src/Actions/Action/Action.component.js#L66

If I use the list in a container, the Action use the register to load the containerized Action.
 * I have added onClick on the overlay to send an action when the user click to open the overlay.
 * give onClick on the container only if props has onClick or actionCreator or payload to avoid error on the middleware if the action hasn't type action.

implement that


packages/containers/src/ActionButton/ActionButton.connect.js
```diff
@@ -28,14 +28,6 @@ export function mergeProps(stateProps, dispatchProps, ownProps) {
 export function ContainerActionButton(props) {
 	const newProps = Object.assign({}, props);
 
-	if (typeof props.overlayComponent === 'string' && props.overlayComponent) {
```
jare-talend: Now, the inject is in the component OverlayTrigger

packages/components/src/Actions/ActionButton/ActionButton.component.js
```diff
@@ -172,6 +172,7 @@ export function ActionButton(props) {
 	if (!inProgress && overlayComponent) {
 		btn = (
 			<OverlayTrigger
+				onClick={rClick}
```
romainseb: i don't get this part, where would be used the button onClick in the overlay ?

jare-talend: Before to open the overlay, i need to save the current index of the line which is open.

packages/components/src/List/ListToVirtualizedList/ListToVirtualizedList.component.js
```diff
@@ -59,6 +59,7 @@ export function ListToVirtualizedList(props) {
 		<VirtualizedList
 			id={props.id}
 			collection={props.items}
+			getComponent={props.getComponent}
```
jsomsanith: It's a pain to write/follow/maintain as this is passed through 
List > ListToVirtualizedList > VirtualizedList > RendererSelector > ListTable/ListGrid > gridRow > CellTitle > CellTitleActions

So 7 levels that don't need it.

What I propose : 
* remove all that
* make List accept customCellDictionary. List container will inject some customCells (more info in following points)
* ListToVirtualizedList : takes custom cell first (allow to override default ones)
* List container will connect CellTitle and pass the connected version in the custom cell dictionary. It is connected so the component will receive the getComponent.
* Make CellTitle component support getComponent, passing it to CellTitleActions that will use it.

As a general rule, do not connect a top level component and pass data through too many levels. Connect what need to be connected.

packages/containers/src/ActionButton/ActionButton.connect.js
jsomsanith: I don't see the code for that

packages/containers/src/ActionButton/ActionButton.test.js
```diff
@@ -33,24 +33,56 @@ describe('CMF(Container(ActionButton))', () => {
...
+		const props = {
+			actionId: 'menu:article',
+			extra: 'foo',
+			payload: 'FOO',
+		};
+		const context = mock.context();
+		const wrapper = shallow(<ContainerActionButton {...props} />, {
+			context,
+		});
+		expect(wrapper.getElement().props).toEqual({ ...props, onClick: jasmine.any(Function) });
```
jsomsanith: You need to test the onClick. Same for next test

packages/components/src/List/ListToVirtualizedList/ListToVirtualizedList.component.js
```diff
@@ -51,6 +52,12 @@ export function ListToVirtualizedList(props) {
 				supposedActions[key] = true;
 			});
 	}
+
+	// allow to connect the CellTitle
+	const InjectedCellTitle = Inject.get(props.getComponent, 'CellTitle', CellTitle.cellRenderer);
```
jsomsanith: This doesn't scale.
If a project need to connect another cellRenderer, to inject a new one, or to replace an existing one, thay will have to add it here :/

In fact I just saw that we already have the system for that.
* List accept cellDictionary props
* ListToVirtualizedList merge the dictionaries `const listCellDictionary = { ...cellDictionary, ...props.cellDictionary };`

So what needs to be done : 
* Containers connect the CellTitle and export it (instead of exporting it in Components)
* List container Inject the connected cellTitle and pass it in the props.cellDictionary to the List component
* ListToVirtualizedList must use `listCellDictionary['title']` instead of `CellTitle` directly. It doesn't need the inject or a getComponent as props anymore.

packages/components/src/VirtualizedList/__snapshots__/RendererSelector.test.js.snap
```diff
@@ -16,6 +16,7 @@ exports[`RendererSelector should render grid list 1`] = `
       },
     ]
   }
+  getComponent={undefined}
```
jsomsanith: This snapshot is out of date. Please update

packages/containers/src/List/List.container.js
```diff
@@ -204,28 +216,37 @@ class List extends React.Component {
...
-				acc[key] = {
-					...omit(current, ['component']),
-					cellRenderer: props.getComponent(current.component),
-				};
-				return acc;
-			}, {});
+			props.list.cellDictionary = {
+				...props.list.cellDictionary,
+				...Object.keys(props.cellDictionary).reduce(
+					(accumulator, key) => ({
```
jsomsanith: You don't need to create a new object at each iteration. We are dealing with object that we just created, so we can modify them.

```
const cellDictionary = { ...connectedCellDictionary };
if (props.cellDictionary) {
	Object.keys(props.cellDictionary).forEach(key => {
		const current = props.cellDictionary[key];
		// eslint-disable-next-line no-param-reassign
		cellDictionary[key] = {
			...omit(current, ['component']),
			cellRenderer: props.getComponent(current.component),
		};
	});
}

props.list.cellDictionary = cellDictionary;
```

packages/containers/src/List/List.container.js
```diff
@@ -204,28 +216,37 @@ class List extends React.Component {
...
-				// eslint-disable-next-line no-param-reassign
-				acc[key] = {
-					...omit(current, ['component']),
-					headerRenderer: props.getComponent(current.component),
-				};
-				return acc;
-			}, {});
+			props.list.headerDictionary = Object.keys(props.headerDictionary).reduce(
+				(accumulator, key) => ({
+					...accumulator,
```
jsomsanith: Same here

packages/components/src/VirtualizedList/RowLarge/__snapshots__/RowLarge.test.js.snap
```diff
@@ -19,7 +19,67 @@ exports[`RowLarge should render large row 1`] = `
     <div
       className="theme-header"
     >
-      
+      <div
```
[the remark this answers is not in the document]
jare-talend: before the test was bad, the title was in the content instead in the header

PR #1529, opened 2018-07-10
feat(containers/selectobject): allow to filter on everything

ncomont:

We use this container in TDP for the copy/move feature, so we need to allow the user to search for any folder in any sub folder.

Introduce a new `filterMode` property to allow to filter the whole folder tree.


packages/containers/src/SelectObject/SelectObject.container.js
```diff
@@ -88,6 +92,38 @@ export function filter(
...
+			const children = item.get('children', null);
+			if (name.toLowerCase().includes(query.toLowerCase())) {
+				onMatch(item);
+				results = results.push(item.set('currentPosition', currentPosition));
+			}
+			if (children) {
+				results = results.concat(
+					filterAll(children, query, { nameAttr }, `${currentPosition} > ${name}`),
+				);
+			}
```
nicomaligne: Can you cut this block  to be more readable ? More functionnal oriented ? What do u think about it ?

packages/containers/src/SelectObject/SelectObject.container.js
```diff
@@ -88,6 +92,38 @@ export function filter(
...
+		});
+	}
+
+	return results;
```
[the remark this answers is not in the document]
ncomont: I don't see any difference. The `result` list must exists to be returned if a query is present. Maybe I'm missing something ?

packages/containers/src/SelectObject/SelectObject.container.js
ncomont: Something like that ?

```javascript
const filtered = filterAll(children, query, { nameAttr }, `${currentPosition} > ${name}`);
results = results.concat(filtered);
```

packages/containers/examples/ExampleSelectObject.js
```diff
@@ -50,5 +50,11 @@ const ExampleSelectObject = {
 			<SelectObject {...props} schema={schema} />
 		</div>
 	),
+	'tree with filter mode set to ALL': () => (
+		<div>
+			<IconsProvider />
+			<SelectObject {...props} filterMode={'ALL'} />
```
jmfrancois: `<SelectObject {...props} filterMode={SelectObject.ALL} />`

PR #1537, opened 2018-07-12
feat(): Add mainSrc field in package.json

annilsson:
**TLDR**: *This is necessary to allow apps using ui to auto reload on file changes in ui, and skip prepublish & copy libs part to preview changes.* 

When we develop on ui, each code change in ui we have to:

1. Prepublish library
2. Copy the ui/packages/components/lib folder into node_modules/@talend/react-components
3. Wait for webpack to refresh page

This takes some time.

On top of that: If ui has a new dependency that the app doesn't have, we can't use the libs this way, because the app can't find the new module.

On ui all we need is a new field "srcMain" which points to the src entry file. The rest you do in the app, explained below:

1. Make sure that the talend/ui libraries resolve to your local repo instead of node_modules (this means no need to copy libs)
2. Make sure react and react-dom is resolved to the app, and never ui (or we will get duplicate instances of react running)
3. Tell webpack to use "mainSrc" entry file instead of "main", if it exists. (This means we don't have to prepublish after each change)

This is *development only*. Because this means that the *app* will build ui code instead of ui itself, so you can't be 100% sure the result is the same. However we move towards reusable webpack config so most likely it will not be an issue. 

When you do your final tests you should let ui build.   

Example command:
`webpack --env.ui`

Example webpack config:
```js
function getUiPath(folderName) {
  return path.join(__dirname, `../../ui/packages/${folderName}`);
}
module.exports = env => {
  return {
    // other config
    resolve: {
      alias: {
        ...env.ui ? {
          react: path.join(__dirname, 'node_modules/react'),
          'react-dom': path.join(__dirname, 'node_modules/react-dom'),
          '@talend/react-components': getUiPath('components'),
          '@talend/react-containers': getUiPath('containers'),
          '@talend/react-sagas': getUiPath('sagas'),
          '@talend/react-cmf': getUiPath('cmf'),
          '@talend/react-cmf-cqrs': getUiPath('cmf-cqrs'),
          '@talend/icons': getUiPath('icons'),
          '@talend/bootstrap-theme': getUiPath('theme'),
          '@talend/react-forms': getUiPath('forms'),
        } : {},
      },
      ...env.ui ? {
        // except for "mainSrc" these are default mainFields unless you specify "target" in your config
        mainFields: ['mainSrc', 'browser', 'module', 'main'],
      } : {},
    },
  }
};
```


annilsson:
duplicated and already merged in https://github.com/Talend/ui/pull/1745

PR #1538, opened 2018-07-13
feat(component/step6): input integration with picker

jdescottes-talend:
Integrate the picker with the input and synchronize the two.

Implement a new component for wrapping the picker and adding the input logic


packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js
```diff
@@ -28,12 +35,35 @@ class DateTimePicker extends React.Component {
...
+	componentWillReceiveProps(nextProps) {
+		const isSelectedDateUpdated = this.props.selectedDate !== nextProps.selectedDate;
+		const isSelectedTimeUpdated = this.props.selectedTime !== nextProps.selectedTime;
+
+		if (isSelectedDateUpdated || isSelectedTimeUpdated) {
+			const canUpdateCalendar = isSelectedDateUpdated && nextProps.selectedDate !== undefined;
+			this.setState({
+				...(isSelectedDateUpdated && { selectedDate: nextProps.selectedDate }),
+				...(isSelectedTimeUpdated && { selectedTime: nextProps.selectedTime }),
+				...(canUpdateCalendar && {
```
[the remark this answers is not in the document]
jsomsanith: In fact shouldn't we consider that if a props date or time change, it means that the whole selection should be updated ?
In storybook testing, I bumped on weird things : 
- empty the input
- select the hour, it doesn't trigger the onChange, that's ok
- type a date, the selected date is updated in the picker
I end up with date and hour selected in the picker, but the input doesn't reflect that.

If you update selected date and time in state everytime that one of them change, you'll have the behavior where user time, it automatically invalidate previous selection. Everything stay up to date.
In summary, any date-picker submit updates all the input, and any input change updates all the selection

packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.test.js
```diff
@@ -11,32 +11,273 @@ describe('DateTimePicker', () => {
...
+		dateTimeView.prop('onSelectTime')(t2);
+		expect(wrapper.state('selectedTime')).toBe(t2);
+
+		wrapper.setProps({
+			selectedTime: t3,
+		});
+		expect(wrapper.state('selectedTime')).toBe(t3);
+	});
+
+	// TODO: Enable when the submit button will be implemented
```
jsomsanith: ??

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -0,0 +1,221 @@
...
+class InputDateTimePicker extends React.Component {
+	static propTypes = {
+		selectedDateTime: PropTypes.instanceOf(Date),
+		onChange: PropTypes.func.isRequired,
+	};
+
+	constructor(props) {
+		super(props);
+
+		this.state = (() => {
```
jsomsanith: (() => {})()
Remove the auto invoked arrow function please

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -0,0 +1,221 @@
...
+	const timeText = `, ${twoDigits(hours)}:${twoDigits(minutes)}`;
+
+	return dateText + timeText;
+}
+
+
+class InputDateTimePicker extends React.Component {
+	static propTypes = {
+		selectedDateTime: PropTypes.instanceOf(Date),
+		onChange: PropTypes.func.isRequired,
```
jsomsanith: This is an input, so it should follow the same propTypes.
* `value`: string
* `onChange`: func no required

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -0,0 +1,221 @@
...
+			textInput,
+		} = this.state;
+
+		return (
+			<div>
+				<input
+					type="text"
+					placeholder="DD/MM/YYYY, hh:mm"
+					value={textInput}
+					onChange={this.onChangeInput}
```
jsomsanith: Spread all rest props except value and onChange to the input

packages/components/src/DateTimePickers/pickers/TimePicker/index.js
```diff
@@ -1,3 +1,6 @@
-import TimePicker from './TimePicker.component';
+import TimePicker, { twoDigits } from './TimePicker.component';
 
 export default TimePicker;
+export {
+    twoDigits,
```
jsomsanith: This is used for Time and date formatting in InputDateTimePicker. You should extract that somewhere, expose it, and import that when needed.
It feels weird to import something from `blabla/TimePicker` to format a date

packages/components/stories/DateTimePicker.js
```diff
@@ -2,11 +2,24 @@ import React from 'react';
...
-	.add('Full DateTimePicker', () => (
+	.add('InputDateTimePicker', () => (
+		<div>
+			<h1>InputDateTimePicker</h1>
+			<IconsProvider />
+
+			<div>
+				<InputDateTimePicker
+					selectedDateTime={new Date(2018, 4, 13, 12, 30)}
+					onChange={console.log}
```
jsomsanith: use storybook actions to make it appear in the action logger panel

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -0,0 +1,221 @@
...
+		this.updateDateTime(date, time);
+	}
+
+	onChangeInput(event) {
+		const text = event.target.value;
+
+		const splitMatches = text.match(splitDateAndTimeRegex);
+
+		if (!splitMatches) {
+			console.warn('DATETIME - INCORRECT FORMAT:', text);
```
jsomsanith: Remove all debug console log

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -0,0 +1,221 @@
...
+	}
+
+	onSubmitPicker({ date, time }) {
+		this.updateDateTime(date, time);
+	}
+
+	onChangeInput(event) {
+		const text = event.target.value;
+
+		const splitMatches = text.match(splitDateAndTimeRegex);
```
jsomsanith: When you enter `01/01/2015` it fails, but `01/01/2015,` passes

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -0,0 +1,221 @@
...
+			console.warn('DATETIME - INCORRECT FORMAT:', text);
+			this.updateDateTime(undefined, undefined, text);
+			return;
+		}
+
+		const [
+			,
+			dateText,
+			timeText,
+		] = splitMatches;
```
jsomsanith: Feels so weird to write `,` to skip the first one. Please write
```
const dateText = splitMatches[1];
const timeText = splitMatches[2];
```

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -0,0 +1,221 @@
...
+
+		const date = (() => {
+			const dateMatches = dateText.match(dateRegex);
+			if (!dateMatches) {
+				console.warn('DATE - INCORRECT FORMAT:', dateText);
+				return undefined;
+			}
+
+			const [
+				,
```
jsomsanith: Same here for not using destructuration

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -0,0 +1,221 @@
...
+				dayString,
+				monthString,
+				yearString,
+			] = dateMatches;
+
+			const day = parseInt(dayString, 10);
+			const month = parseInt(monthString, 10);
+			const monthIndex = month - 1;
```
jsomsanith: You can get rid of `month`

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -0,0 +1,221 @@
...
+
+			const monthDate = new Date(year, monthIndex);
+			const lastDateOfMonth = lastDayOfMonth(monthDate);
+
+			if (day > getDate(lastDateOfMonth)) {
+				console.warn('DATE - INCORRECT DAY NUMBER RELATIVE TO MONTH:', dateText);
+				return undefined;
+			}
+
+			const dateValidated = setDate(monthDate, day);
```
jsomsanith: Instead of performing all the checks yourself, is there a way with Date or date-fns to try to create a date from `dateText`, and get undefined if it's not valid ?
The less we write code, the better, and this seems to be something quite basic in date manipulation libs.

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -0,0 +1,221 @@
...
+
+		this.onChangeInput = this.onChangeInput.bind(this);
+		this.onSubmitPicker = this.onSubmitPicker.bind(this);
+	}
+
+	onSubmitPicker({ date, time }) {
+		this.updateDateTime(date, time);
+	}
+
+	onChangeInput(event) {
```
jsomsanith: Can you debounce to avoid useless computation when user is typing ?

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -0,0 +1,221 @@
...
+				return undefined;
+			}
+
+			const dateValidated = setDate(monthDate, day);
+
+			return dateValidated;
+		})();
+
+
+		const time = (() => {
```
jsomsanith: Same remarks as for date

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -0,0 +1,221 @@
...
+			};
+		})();
+
+		this.onChangeInput = this.onChangeInput.bind(this);
+		this.onSubmitPicker = this.onSubmitPicker.bind(this);
+	}
+
+	onSubmitPicker({ date, time }) {
+		this.updateDateTime(date, time);
+	}
```
jsomsanith: Can you try to align your 2 apis to just pass `this.updateDateTime` in the render, and remove this almost useless method ?

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -0,0 +1,221 @@
...
+		});
+
+		const fullDate = (() => {
+			if (date === undefined || time === undefined) {
+				return undefined;
+			}
+
+			return setMinutes(date, time);
+		})();
+		this.props.onChange(fullDate);
```
jsomsanith: As said in the proptypes, onChange should be called with text.
The exchange format with the outside world is text as you implement a form input. Dates are just internal formats to implement the logic.

+ it will remove the (() => {})()

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
jdescottes-talend: I'll handle that after integrating into form when i'll be ready for error handling.
The console will not stay ;)

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
jdescottes-talend: yeah ... I'll change the format to ISO "YYYY-MM-DD hh:mm" (for preventing unexpected month position : MM-DD-YYYY vs DD-MM-YYYY) and this unexpected behavior with it

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
jdescottes-talend: Nop I cannot :P
Even if it was exactly the same function signature, things always change. Signatures change, code is added/modified, and stick exactly one behavior for one event with the same exact signature will more than likely when changes occurs, give extra params not wanted or less or whatelse not beeing seen easily because it's all implicit.

And in this case for future PR, I've got already added some more behavior when this event occurs, so I definitely don't change it :)

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
jdescottes-talend: I was not thinking of it as an input but more like a thing that take a selected date and give a date.
The date is the final value wanted, not really the string.
We need to discuss about it, it depends on how we reason and manage date in our frontend programs and eventually how we convert it from/to the server API.
But a date object is more useful to do manipulations and I don't think server API should dictate the frontend for the format (we'll just eventually need a date parser/stringifier).

packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js
```diff
@@ -57,6 +87,20 @@ class DateTimePicker extends React.Component {
...
 
+	trySubmit() {
+		if (this.state.selectedDate !== undefined &&
+			this.state.selectedTime !== undefined) {
+			this.submit();
+		}
+	}
+
+	submit() {
+		this.props.onSubmit({
```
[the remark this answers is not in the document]
jdescottes-talend: Why ?
I don't named it this way, because the **values can change when picked internally without be emitted outside**. They are emitted only when `submitted`/`validated` or an other word.
I understand submit can be misunderstood with form submit but if I set `onChange`, it will not be really trigger when the date or time changes :s.

packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js
```diff
@@ -57,6 +87,20 @@ class DateTimePicker extends React.Component {
 		this.setState({ isDateTimeView });
 	}
 
+	trySubmit() {
```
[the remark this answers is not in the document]
jdescottes-talend: Or maybe the other way, remove the other method and take it's code back here ?!
The `try` has it's importance, `submit()` don't tell us it could not submit on some cases

packages/components/src/DateTimePickers/pickers/TimePicker/index.js
jdescottes-talend: yeah I have think the same of it, it feel wrong, but @jmfrancois is utils-epileptic.
I will do it, it's our little secret ^^

packages/components/stories/DateTimePicker.js
jdescottes-talend: great, I didn't know those things :)

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
jdescottes-talend: oki

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
jdescottes-talend: I can get rid of so many things... but does it gain something apart from one line, to do all in one line ?
I find it much simpler with one instruction at a time and more meaningful why we are decrementing of one.

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
jdescottes-talend: I didn't think to much of it, I confess.
I was mostly thinking, this part will need to be rework because of the internationalization thingy.
And have a parser which take every custom local format is not a common thing (not impossible though).
The `date-fns` parser take only exact ISO string format not visual custom one like ours. And also, like we did it for user experience, we need to extract date or time only when valid before the full date is.

I was thinking, when we'll do all the configuration format, to extract the input part into a specific component and deal with a lib or a bunch of custom parsers/formatters.

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
jdescottes-talend: ✔

packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js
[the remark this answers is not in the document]
jsomsanith: ` when picked internally without be emitted outside`
But here we are talking about the `external` api. Form the outside world, it's a value change. Don't name exposed api things with internal concepts.

packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js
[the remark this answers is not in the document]
jsomsanith: yes, one or the other, the problem here is to have those 2 methods

packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js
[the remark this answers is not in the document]
jsomsanith: I think we are aligned. 
But don't add any date/time as props. You must have a clean api, don't duplicate the Date props with string equivalent for change detection. Just consider that if the Date prop ref change, you have a new Date. Or if you really want to avoid refresh, implement `shouldComponentUpdate()` to compare old and new Date.

packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.test.js
jsomsanith: Remove it, this TODO is not useful, when you want to implement the button, you'll just do what is needed

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
jsomsanith: I don't understand all you wrote but we already discussed about that. It is a bad practice: 
* harder to read
* create functions that you don't need, just write the code in the constructor

And you seem to say that it solve problems, but I think you can solve those problems removing the immediatly invoked arrow function :/

packages/components/src/DateTimePickers/pickers/TimePicker/index.js
jsomsanith: Just don't name the things `utils` but with meaningful names related with the methods you expose.
Something like `formatter.js`

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
jsomsanith: You can implement the `onError` we discussed. So in storybook you pass a storybook action to make it appear in action logger.
But ok you can do it in another PR to avoid making this one heavier

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
jsomsanith: Making it inline is fully understandable without any problem. So creating a new var for something that simple doesn't help.

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
jsomsanith: But if you find something that takes the string and the `format` it could be good. And we don't necessarily need to know what part precisely is wrong, but telling the user, the input doesn't follow the wanted format (and giving him the format) is totally fine.

packages/components/src/DateTimePickers/pickers/TimePicker/index.js
jdescottes-talend: ✔

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
jsomsanith: I don't understand your argument here. Even with your solution, changing any signature implies a code adaptation.
If you want to be flexible on that, just align all the signatures to take an object so you can add any property you want.

packages/components/stories/DateTimePicker.js
jdescottes-talend: ✔

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
jdescottes-talend: If `onSubmit` output a second argument, this will break `this.updateDateTime`.
If we need to do different things on submit, this implies breaking it down to this function, so lots of lines changes in diff for nothing, here, i could just add one behavior/action with just a one more line change.
`this.updateDateTime` is not an event handler anyway and directly passing a 'action' function to an handler lead to errors in present or future because we implicitly pass args between two interfaces that are not the same.

packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js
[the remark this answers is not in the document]
jdescottes-talend: The behavior of changing the picking value then submitted/validated it, is clear from the outside. It's the functional way the picker work, it's not technical.
=> I picked a date and a time, eventually change my picking multiple times, then I validate my choice (event without explicit user action it's a special case of validation)

If I use a input component with a 'done' button next to.
A `onChange`would be confusing : is it triggered when I change the input or when I submit it with the button or even both ?
A `onSubmit` is clear that it's not when the input change, but only when the button is used.

And by definition, all events/callbacks are emitted when a value change, it's not particularly inherent to `onChange`. There is just no need to precise for a single input because only text input can change, but for a complex widget, `onChange` is really confusing.

In the current `DateTimePicker` component context :
- `onChange` say to me "I will give you a DateTime"
- `onSubmit` say to me "I will give you a DateTime when I'm submitted"

packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js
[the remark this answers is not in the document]
jdescottes-talend: I don't see for what reason it's a problem ?! But I can do the change.

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
jdescottes-talend: ✘

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
jdescottes-talend: ✔

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
jdescottes-talend: With a few improvement to parse date and time alone and pick them while it can still be a global invalid datetime.
✔

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
jdescottes-talend: But it doesn't hurt too ...
It's much easier to reason globally when each value is hold by one variable. Otherwise it makes it totally subjective when to break it in multiple lines or keep it in one line. Where are the objective limits in both way (too long and too short) ?

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
jdescottes-talend: ✔

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
jdescottes-talend: ✔

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
jdescottes-talend: I use it later so I keep it like this.
✘

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
jdescottes-talend: We need to check specifically the parts because of relation between day and month/year (29/02 could be valid or not, depending on year ; 31 is valid for some month but not all, etc.).
How to you constrain format and allow '14/01/2018', '01/01/2018', '10/01/2018' but not '00/01/2018' ?
I don't know you can do it with a simple regexp and just anyway with a regexp.
The same for month before 1 and after 12.

After that, like I did, we extract the date or the time even if the entire string don't match the format. Allowing a quick real time feedback in the picker.

We'll see to update/rework that an other time, with l10n.
✘

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
jdescottes-talend: ✔

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
jdescottes-talend: ✘

packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js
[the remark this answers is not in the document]
jdescottes-talend: ✘

packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js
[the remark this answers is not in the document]

packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js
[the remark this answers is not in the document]

packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.test.js
jdescottes-talend: ✔

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
jdescottes-talend: For later, it's have no impact right now
✘

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
jdescottes-talend: For later, it's have no impact right now
✘

PR #1540, opened 2018-07-16
feat(components/HttpError):  add go back link

Innes333:
https://jira.talendforge.org/browse/TPSVC-2681
http://guidelines.talend.com/document/92132#/navigation-layout/error-pages
Redirect action is missing by new guidelines of NotFound page 

HttpError component is extended with new props -> onLinkClick, linkLabel


packages/components/src/HttpError/HttpError.component.js
```diff
@@ -21,6 +23,9 @@ function HttpError(props) {
 			>
 				<h1>{title}</h1>
 				<p>{message}</p>
+				{backAction && (
+					<Action icon="talend-arrow-left" label={backLabel} link onClick={backAction} />
```
jsomsanith: I think you should align the api with action api, instead of taking onclick + label.
2 choices : 
* you allow to put any action, you just take a props.action and spread it to `<Action {...props.action} />`
or
* you only allow a back action, you take a props.backAction and just enforce the link style + arrow icon `<Action {...props.backAction} icon="talend-arrow-left" link />`

packages/components/src/HttpError/HttpError.scss
```diff
@@ -38,5 +38,12 @@ $tc-http-error-message-color: #888 !default;
 			font-size: 3rem;
 			font-weight: 600;
 		}
+
+		button {
+			padding: 0;
+			font-weight: 600;
```
jsomsanith: I don't see any spec about font in the guideline. Where does that come from ?

jsomsanith: The color doesn't seem to be right neither. I think you should see with UX to get those info

packages/components/src/HttpError/HttpError.component.js
Innes333: Good point! I chose first way

packages/components/src/HttpError/HttpError.snapshot.test.js
```diff
@@ -9,6 +9,14 @@ const forbiddenProps = {
 	message: 'You are not allowed to access this page',
 };
 
+const notFoundProps = {
+	status: 404,
+	title: 'Oops...',
+	message: 'The page you are looking for cannot be found',
+	backAction: jest.fn(),
+	backLabel: 'Start over',
```
jsomsanith: This is not the api anymore

PR #1541, opened 2018-07-16
chore: take in account changes requested

jdescottes-talend:
Fix/adapt to match changes requested

Some tweaks and adding a wrapper component which handle actions and selections, around the incrementable scroll list to mutualize more common code between Time et Year pickers without overloading the actual incrementable scroll list component.


packages/components/src/DateTimePickers/pickers/IncrementableScrollActionList/IncrementableScrollList/IncrementableScrollList.component.js
```diff
@@ -35,7 +47,17 @@ class IncrementableScrollList extends React.Component {
...
+		const lastIndex = this.props.items.length - NB_ITEMS_DISPLAYED;
+		const needToScroll = increment < 0 && this.state.startIndex > firstIndex ||
+							increment > 0 && this.state.startIndex < lastIndex;
+
+		if (!needToScroll) {
+			return;
+		}
+
+		const incrementedIndex = this.state.startIndex + increment;
+		const newRowIndex = keepInBoundaries(incrementedIndex, firstIndex, lastIndex);
```
jsomsanith: This is always equals to `incrementedIndex` because of your `if`.
But look at the previous comment.

packages/components/src/DateTimePickers/pickers/IncrementableScrollActionList/IncrementableScrollActionList.component.js
```diff
@@ -0,0 +1,65 @@
...
+
+	constructor(props) {
+		super(props);
+
+		const middleInitialIndex = this.props.items.findIndex(
+			item => item.id === this.props.initialMiddleVisibleItemId);
+
+		this.initialIndex = middleInitialIndex === -1
+			? 0
+			: Math.max(middleInitialIndex - 2, 0);
```
jsomsanith: You can remove the ternary and write `this.initialIndex = Math.max(middleInitialIndex - 2, 0)`

packages/components/src/DateTimePickers/pickers/IncrementableScrollActionList/IncrementableScrollActionList.component.js
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	isSelected(id) {
+		return id === this.props.selectedItemId;
+	}
```
jsomsanith: This method is useless, just inline it in the `render()` method

packages/components/src/DateTimePickers/pickers/IncrementableScrollActionList/IncrementableScrollActionList.component.js
```diff
@@ -0,0 +1,65 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import IncrementableScrollList from './IncrementableScrollList';
+import PickerAction from '../../PickerAction';
+
+class IncrementableScrollActionList extends React.Component {
```
jsomsanith: Why did you create a new component to inject an item renderer instead of setting it in IncrementalScrollList ?
You add code to maintain a generic IncrementalScrollList.props.itemRenderer that is used only once, by this new extra component.

packages/components/src/DateTimePickers/pickers/IncrementableScrollActionList/IncrementableScrollList/IncrementableScrollList.component.js
```diff
@@ -35,7 +47,17 @@ class IncrementableScrollList extends React.Component {
 	}
 
 	scrollRows(increment) {
-		const newRowIndex = this.state.startIndex + increment;
+		const firstIndex = 0;
+		const lastIndex = this.props.items.length - NB_ITEMS_DISPLAYED;
+		const needToScroll = increment < 0 && this.state.startIndex > firstIndex ||
```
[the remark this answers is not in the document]
jdescottes-talend: The 2 comments were tempting at first, but after a few headache, there is some problems :(.

Here : in the case we are positioned on firstIndex, we can't scroll before, ok, but we can't scroll after neither, not ok (We will have infinite sticky border, min max).

In the next comment : This is not always equal, in the case of positioning on index 3 (min to index 0). we ask for the 5 before so index -2 (3-5). But in that case we want don't want to go to an inexistant index, ok, but we still want to go to 0 and not stay where we at, not ok. Because we cannot see index 0, 1 and 2. It's why the keepInBoundaries.

packages/components/src/DateTimePickers/pickers/IncrementableScrollActionList/IncrementableScrollActionList.component.js
jdescottes-talend: Yeah we could, but i find it less obvious that the 'not find item' case is taken care. It sort of mix 2 things

packages/components/src/DateTimePickers/pickers/IncrementableScrollActionList/IncrementableScrollActionList.component.js
jdescottes-talend: it decouple the scrolling mechanic from the action specific things. It makes more smaller and focus unit that do one thing well :).
The generic scrolling component already had a lot of complexity.

packages/components/src/DateTimePickers/pickers/IncrementableScrollActionList/IncrementableScrollActionList.component.js
jdescottes-talend: It avoid creating new anonymous function for each row render call (when rerender component and also each scrolling)

jsomsanith: inlining `id === this.props.selectedItemId` doesn't create a function

packages/components/src/DateTimePickers/pickers/IncrementableScrollActionList/IncrementableScrollList/IncrementableScrollList.component.js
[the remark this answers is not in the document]
jsomsanith: Oh you're right, I misread the method.

packages/components/src/DateTimePickers/pickers/IncrementableScrollActionList/IncrementableScrollActionList.component.js
jdescottes-talend: I wasn't thinking of this possibility, that's greater !

PR #1494, opened 2018-06-27
feat(Datalist): add loading and trigger support

jmfrancois:
**What is the problem this PR is trying to solve?**

* Datalist component doesn t support loading feeback in the input field
* Datalist component doesn t support onFocus props
* Datalist doesn t use the talend-caret-down icon
* The UIForm trigger onFocus is not supported in the datalist widget

**What is the chosen solution to this problem?**

Typeahead the component:
* add isLoading props to give user a feedback
* add caret props to render talend-caret-down and remove span class="caret"

Datalist the component:
* add loading feedback at the input level
* add onFocus props support

Datalist the UIForm widget
* onFocus the widget will try to find a trigger onFocus

**Please check if the PR fulfills these requirements**

* [x] The PR commit message follows our [guidelines](https://github.com/talend/tools/blob/master/tools-root-github/CONTRIBUTING.md)
* [ ] Tests for the changes have been added (for bug fixes / features) And [non reg](./screenshots.md) done before need review
* [ ] Docs have been added / updated (for bug fixes / features)
* [ ] Related design / discussions / pages (not in jira), if any, are all linked or available in the PR

**[ ] This PR introduces a breaking change**

PR #1542, opened 2018-07-16
feat(Components-kit): add async datalist case

jsomsanith:
No async datalist case in the components-kit app.

Add it

Don't merge it before #1494, this PR depends on it


packages/components-kit/src/components/ComponentForm/ComponentForm.component.js
```diff
@@ -180,6 +188,7 @@ export class TCompForm extends React.Component {
 			onTrigger: this.onTrigger,
 			onChange: this.onChange,
 			onSubmit: this.onSubmit,
+			widgets: { ...this.props.widgets, ...tcompFieldsWidgets },
```
jmfrancois: don t you want props widget to take precedence on tcompFields ?

packages/components-kit/src/components/fields/Datalist/Datalist.component.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import DatalistWidget from '@talend/react-forms/lib/UIForm/fields/Datalist';
+import { getValue } from '@talend/react-forms/lib/UIForm//utils/properties';
```
jmfrancois: typo //

jmfrancois: //

packages/components-kit/src/components/fields/Datalist/Datalist.component.js
```diff
@@ -0,0 +1,24 @@
...
+	function resolveName(value) {
+		// create schema to get entry name from internal properties
+		const key = Array.from(props.schema.key);
+		key[key.length - 1] = `$${key[key.length - 1]}_name`;
+
+		const nameSchema = { ...props.schema, key };
+		return getValue(props.properties, nameSchema) || value;
+	}
+
+	return <DatalistWidget {...props} resolveName={resolveName} />;
```
jmfrancois: it looks weird to me, props here are available when it called by the Datalist.
So here you create a new function on each render, why not using it this way:
```javascript
function resolveName(value, props) {
...same content
}
function Datalist(props) {
  return <DatalistWidget {...props} resolveName={resolveName} />;
}
```
So you change that it Datalist.component.js

```javascript
.map(value => ({ name: this.props.resolveName(value), value }))
// by
.map(value => ({ name: this.props.resolveName(value, this.props), value }))
```

packages/components-kit/src/components/ComponentForm/ComponentForm.component.js
```diff
@@ -67,7 +68,8 @@ export function keepOnlyDatasetMetadataProperties({ body, properties }) {
 export class TCompForm extends React.Component {
 	constructor(props) {
 		super(props);
-		this.state = {};
+		const properties = props.state.get('properties');
+		this.state = { properties: properties && properties.toJS() };
```
jmfrancois: I don t see how state properties could exists at this stage.
So let s remove it

packages/components-kit/src/components/ComponentForm/ComponentForm.component.js
```diff
@@ -80,6 +82,12 @@ export class TCompForm extends React.Component {
 		this.getMemoizedErrors = memoizeOne(toJS);
 	}
 
+	componentWillReceiveProps(nextProps) {
+		if (this.props.state.get('properties') !== nextProps.state.get('properties')) {
+			this.setState({ properties: nextProps.state.get('properties').toJS() });
```
jmfrancois: The current use case is not to support a mutation in the state of the properties but more to support edit form.
So i would go to render method with 
`properties: this.state.properties || this.props.state.get('properties').toJS(),`

jsomsanith: It was to avoid re-computation because of the `.toJS()`. But do you prefer using memoize-one instead of component state ?

PR #1544, opened 2018-07-17
fix(cmf): cancel a saga on a component unmount

romainseb:
This is not working, the id returned is a random number when there is no "id" param but componentId
Do the same thing as the didmount to pass the compoentId.
Remove a curry in a take that is not working

### DIFF
--- a/packages/cmf/src/sagas/component.js
+++ b/packages/cmf/src/sagas/component.js
@@ -1,5 +1,4 @@
 import { fork, cancel, take, takeEvery } from 'redux-saga/effects';
-import curry from 'lodash/curry';
 import CONST from '../constant';
 import registry from '../registry';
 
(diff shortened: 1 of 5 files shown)


packages/cmf/src/sagas/component.js
```diff
@@ -24,11 +23,9 @@ export function get(id, context) {
 
 export const registerMany = registry.getRegisterMany(register);
 
-export const isActionCancelable = curry(
-	(startAction, action) =>
-		action.type === `${CONST.WILL_UNMOUNT_SAGA_STOP}_${startAction.saga}` &&
-		startAction.event.componentId === action.event.componentId,
-);
+export const isActionCancelable = startAction => action =>
```
jare-talend: just remove the curry. It seems that there is a issue with the TAKE to check the random componentId and not the componentId from the settings.

replace the take call by yield take(actionTAKE => isActionCancelable(action, actionTAKE));

PR #1545, opened 2018-07-17
feat(List): add support for hierarchical data

jmfrancois:

if your data is not flat, you can t display it.

Add use Virutalized.Column cellDataGetter accessor props to use lodash get.

Note: column.key is used to create rowData so we need to add a new option to get the data.


output/components.eslint.txt
```diff
@@ -10,6 +10,9 @@ The react/require-extension rule is deprecated. Please use the import/extensions
   1:1  error  '@storybook/react' should be listed in the project's dependencies, not devDependencies  import/no-extraneous-dependencies
   4:2  error  Unexpected require()                                                                    global-require
 
+/home/travis/build/Talend/ui/packages/components/src/List/ListToVirtualizedList/ListToVirtualizedList.component.js
+  115:1  error  Line 115 exceeds the maximum line length of 100  max-len
```
jsomsanith: Please fix the 2 new linter errors

packages/components/src/List/ListToVirtualizedList/ListToVirtualizedList.component.js
```diff
@@ -102,7 +103,16 @@ export function ListToVirtualizedList(props) {
...
+					if (column.selector) {
+						if (cProps.columnData) {
+							cProps.columnData.selector = column.selector;
+						} else {
+							cProps.columnData = {
+								selector: column.selector,
+							};
+						}
+					}
+					return <VirtualizedList.Content key={index} cellDataGetter={cellDataGetter} {...cProps} />;
```
jsomsanith: The VirtualizedList is a component that we expose, that everyone can use instead of List if we don't want toolbar for example.
Putting this `cellDataGetter` here makes it available for List but not for VirtualizedList. And I don't think it is the responsibility of the ListToVirtualizedList adapter to handle, but more a VirtualizedList feature that is available for everyone.

What I propose : 
* `VirtualizedList.Content` is just the react-virtualized Column component. Instead, you can wrap it with your HOC that inject your cellDataGetter only if there is a selector.
* expose this wrapped Column as `VirtualizedList.Content`.
* in ListToVirtualizedList, you just keep the selector props. But instead of passing it in props.columnData, it can become a top level props of the HOC.

packages/components/stories/List.js
```diff
@@ -162,59 +162,68 @@ const props = {
 	list: {
 		columns: [
 			{ key: 'id', label: 'Id', order: 0 },
-			{ key: 'name', label: 'Name', order: 1 },
-			{ key: 'author', label: 'Author', order: 3 },
-			{ key: 'created', label: 'Created', order: 2 },
+			{ key: 'name', selector: 'meta.name', label: 'Name', order: 1 },
+			{ key: 'author', selector: 'meta.author', label: 'Author', order: 3 },
+			{ key: 'created', selector: 'meta.created', label: 'Created', order: 2 },
```
jsomsanith: I think devs that are used to simple data without selector can be confused if you pass almost all the example with a selector.
IMO you should only pass 1 column (or a new one) into nested data with selector. Or perhaps, if you want to have coherent data, you can group the created/modified date into a nested `meta` property. That would be 2 columns with selector.

PR #1547, opened 2018-07-17
feat(components): implement RichTooltip

jare-talend:
Implement the wrapper to the RichTooltip with the component Header, Body, Footer.
We can inject custom component like BodyList in this component.

apply this specs: 
https://company-57688.frontify.com/document/92132#/ui-controls/tooltips
https://company-57688.frontify.com/document/92132#/messaging-communication/loading-feedback

demo:
http://1547.talend.surge.sh/components/?selectedKind=RichTooltip&selectedStory=default&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

implement that


packages/components/src/RichTooltip/RichTooltip.component.js
```diff
@@ -0,0 +1,123 @@
...
+
+export function RichTooltipFooter(props) {
+	return (
+		<footer className={classNames(theme['tc-tooltip-footer'], 'tc-tooltip-footer')}>
+			<Actions actions={props.left} />
+			<Actions
+				className={classNames(theme['tc-tooltip-right-actions'], 'tc-tooltip-right-actions')}
+				actions={props.right}
+			/>
+		</footer>
```
nicomaligne: Why still using actions ? The guidlines shows that there are 3 kind footers, can we make a dedicated components and inject a custom one if we need.
The props will be simpler.

packages/components/src/RichTooltip/RichTooltip.component.js
```diff
@@ -0,0 +1,123 @@
...
+	);
+}
+
+RichTooltipBody.propTypes = {
+	className: PropTypes.string,
+	Content: RichTooltipContent.propTypes.Content,
+};
+
+export function RichTooltipHeader(props) {
+	return (
```
nicomaligne: Same as below

packages/components/src/RichTooltip/RichTooltip.component.js
```diff
@@ -0,0 +1,123 @@
...
+				<div className={theme['tc-error-body']}>
+					<h1>Whoops !</h1>
+					<div>{props.error}</div>
+				</div>
+			</div>
+		);
+	}
+
+	if (typeof props.Content === 'string' && !(props.error || props.loading)) {
+		content = props.Content;
```
nicomaligne: Here if i'm not mistaking Content is a string but with an Uppercase, that's not semanticaaly correct for props.

packages/components/src/RichTooltip/RichTooltip.component.js
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	if (content) {
+		return <RichTooltip.Body className={props.className} Content={content} />;
```
nicomaligne: same here why content is with an uppercase ?
maybe renderContent will be better nop ?
Also it's my opinion but too much mutation possible on the content variable.

packages/components/src/RichTooltip/RichTooltip.component.js
```diff
@@ -0,0 +1,129 @@
...
+	error: PropTypes.string,
+};
+
+export function RichTooltipBody(props) {
+	// fix ie11 flexbox align-items: center with min-height https://stackoverflow.com/a/47180276
+	return (
+		<div
+			className={classNames(props.className, {
+				[theme['tc-tooltip-body']]: true,
+				'tc-tooltip-body': true,
```
jmfrancois: if you don t have condition on it you can just use the args syntax instead of object.

packages/components/src/RichTooltip/RichTooltip.component.js
```diff
@@ -0,0 +1,129 @@
...
+				<Icon name="talend-warning" className={theme['tc-center-icon']} />
+				<div className={theme['tc-error-body']}>
+					<h1>Whoops !</h1>
+					<div>{props.error}</div>
+				</div>
+			</div>
+		);
+	}
+
+	return content;
```
jmfrancois: I prefer to use else here the priority is reversed, so error is the most important.
if (props.error){}
else if (props.loading) {}
else if (typeof === string) {}

with return in it will make it more clear

packages/components/src/RichTooltip/RichTooltip.component.js
```diff
@@ -0,0 +1,129 @@
...
+			</div>
+		);
+	}
+
+	if (props.error) {
+		content = (
+			<div className={classNames(theme['tc-center-content'], theme['tc-error-wrapper'])}>
+				<Icon name="talend-warning" className={theme['tc-center-icon']} />
+				<div className={theme['tc-error-body']}>
+					<h1>Whoops !</h1>
```
jmfrancois: question: is this woops really hard coded ?

packages/components/src/RichTooltip/RichTooltip.scss
```diff
@@ -0,0 +1,110 @@
...
+.tc-tooltip-header {
+	min-width: $tc-popover-max-width;
+	border-bottom: $tc-popover-border-width solid $tc-popover-header-border-color;
+	padding: 0 $padding-large;
+	flex: 1;
+	height: $tc-popover-header-footer-height;
+	display: flex;
+	align-items: center;
+
+	h1 {
```
jmfrancois: h1 is for a title of a page.
For example bootstrap3 modal title use h4 and bootstrap 4 use h5.

packages/components/src/RichTooltip/RichTooltip.component.js
```diff
@@ -0,0 +1,129 @@
...
+export function RichTooltipBody(props) {
+	// fix ie11 flexbox align-items: center with min-height https://stackoverflow.com/a/47180276
+	return (
+		<div
+			className={classNames(props.className, {
+				[theme['tc-tooltip-body']]: true,
+				'tc-tooltip-body': true,
+			})}
+		>
+			<div className={classNames(theme['tc-tooltip-fix-ie'], 'tc-tooltip-fix-ie')}>
```
jmfrancois: please find a better name

packages/components/src/RichTooltip/RichTooltip.scss
```diff
@@ -0,0 +1,110 @@
...
+	.tc-error-wrapper {
+		display: flex;
+		align-items: center;
+
+		.tc-center-icon {
+			width: $tc-popover-icon-error-size;
+			height: $tc-popover-icon-error-size;
+		}
+
+		.tc-error-body {
```
jmfrancois: avoir the most possible deep selectors.
We are already on css module and those components will not be used outside of this use case

packages/components/src/index.js
```diff
@@ -157,6 +158,10 @@ export {
 	Notification,
 	ObjectViewer,
 	Progress,
+	RichTooltip,
+	RichTooltipHeader,
+	RichTooltipBody,
+	RichTooltipFooter,
```
jmfrancois: please only export RichTooltip

packages/components/src/RichTooltip/index.js
```diff
@@ -0,0 +1,8 @@
+import Component, {
+	RichTooltipHeader,
+	RichTooltipBody,
+	RichTooltipFooter,
+} from './RichTooltip.component';
+
+export { RichTooltipHeader, RichTooltipBody, RichTooltipFooter };
```
jmfrancois: you don t need to export all that

packages/components/src/index.js
jare-talend: i need to export RichTooltipHeader, RichTooltipBody, RichTooltipFooter to allow cmf to connect them.

packages/components/src/RichTooltip/RichTooltip.component.js
```diff
@@ -0,0 +1,126 @@
...
+	right: PropTypes.array,
+};
+
+export default function RichTooltip(props) {
+	return (
+		<span>
+			{Inject.getReactElement(props.getComponent, props.Header)}
+			<RichTooltipContent {...props} />
+			{Inject.getReactElement(props.getComponent, props.Footer)}
+		</span>
```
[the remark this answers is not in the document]
jmfrancois: next to that that means we need to have componetn that play wells in the three area.
I can gess a Title component that could be shared with the subheaderbar for example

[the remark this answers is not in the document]
nicomaligne: I kind of agree with Jeanmich, he make a good resume of my comments.
I'm still thinking that we can have default header footer but i think the richtooltip must be a very simple layout, with slot for inject | render props | children style pattern.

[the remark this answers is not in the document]
nicomaligne: Just one thing let's call the props renderHeader, renderFooter no ?
Header with uppercase directly really suggest that he only accep a class that extends React.

packages/components/src/RichTooltip/RichTooltip.component.js
```diff
@@ -0,0 +1,43 @@
...
+
+export default function RichTooltip(props) {
+	return (
+		<span>
+			<header className={classNames(theme.header, 'tc-tooltip-header')}>
+				{Inject.getReactElement(props.getComponent, props.Header)}
+			</header>
+			<div className={classNames(props.className, theme.body, 'tc-tooltip-body')}>
+				<div className={classNames(theme.content, 'tc-tooltip-content')}>
+					<RichTooltipContent {...props} />
```
jmfrancois: what do you think if we remove the RichTooltipContent and move its render inside RichToolTip ?

jare-talend: yes i can

packages/components/src/HeaderTitle/HeaderTitle.component.js
```diff
@@ -0,0 +1,19 @@
...
+
+import theme from './HeaderTitle.scss';
+
+export default function HeaderTitle(props) {
+	return (
+		<div className={classNames(theme.container, 'tc-header-title')}>
+			<h4 className={classNames(theme.title, 'tc-header-title-text')} title={props.title}>
+				{props.title}
+			</h4>
+		</div>
```
jmfrancois: FYI this is a first iteration to implement HeaderTitle guideline which will be next used in the subheaderbar

PR #1550, opened 2018-07-17
feat(forms): add loading skeleton

jmfrancois:

the guideline specify to use skeleton when the component display but we need to load remote data before.
ComponentForm is using forms and load the schema on component kit.
Currently it displays a circular progress.

http://recordit.co/c1IFBs4Ngx

This PR:
* add FormSkeleton
* add and expose by default FormSwitcher which handle the `loading` props and switch the library to use
* remove the switch to UIForm in the Form lib.


packages/forms/src/FormSkeleton.scss
```diff
@@ -0,0 +1,60 @@
...
+		flex-direction: column;
+		flex-grow: 1;
+		min-height: 0;
+		margin: 0;
+	}
+
+	.form-content {
+		flex-grow: 1;
+		min-height: 0;
+		padding: 10px;
```
annilsson: This should probably be `$padding-small`?

packages/forms/src/FormSkeleton.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import { Skeleton } from '@talend/react-components';
+import theme from './FormSkeleton.scss';
+
+export default function FormSkeleton() {
+	return (
+		<div className={`${theme.container}`}>
```
annilsson: Can be `className={theme.container}`

packages/forms/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@talend/react-forms",
   "description": "React forms library based on json schema form.",
-  "main": "lib/Form.js",
+  "main": "lib/index.js",
```
annilsson: I like this change, but I found a few imports that might need to be updated(?). Could you please check if they still should import Form.js, I'm not sure? 

- `import Form from '../Form';` in "ui\packages\forms\src\fields\CollapsibleFieldset.test.js"
- `import Form, { renderActionIcon, renderActions } from './Form';` in "ui\packages\forms\src\Form.test.js"
- `import Form from '../src/Form';` in "ui\packages\forms\stories\index.js"

jmfrancois: Those are internals and for this `Form.js` implementation so no need to be changed.
I have exposed some of the Form.js to not break all the things but yes, I should expose all of them in index.js to not add a breaking change.

The goal here is to enforce the import of UIForm from index and not from lib/UIForm

packages/forms/src/FormSkeleton.scss
```diff
@@ -0,0 +1,62 @@
...
+	:global(.stacked) :global(.tc-drawer-container) {
+		.form-content {
+			width: $tc-drawer-content-max-width;
+			margin: 0 auto;
+		}
+		.submit-wrapper {
+			max-width: $tc-drawer-content-max-width;
+		}
+		:global(.form-actions) {
+			max-width: 650px;
```
jare-talend: use varaible $tc-drawer-content-max-width === 650px;

jmfrancois:
no blink on my side: http://recordit.co/4SE40akTo4

PR #1552, opened 2018-07-18
fix(components): fix nested items positioning for components/listview

andrii-holovko-dev:
after row expanding or filtering nested rows positioning become incorrect

use measure callback provided by CellMeasurer component to update row height on expand and on filtering


jsomsanith:
I think what you need is that : https://github.com/bvaughn/react-virtualized/blob/master/docs/CellMeasurer.md

packages/components/src/ListView/Items/Items.component.js
```diff
@@ -37,6 +37,11 @@ export class ItemsComponent extends React.PureComponent {
 		this.renderToggleAll = this.renderToggleAll.bind(this);
 
 		this.hasToggleAll = this.props.showToggleAll && this.props.items.length > 1;
+		this.cache = new CellMeasurerCache({
+			defaultHeight: props.getItemHeight(),
+			fixedWidth: true,
+			keyMapper: () => 1,
```
jsomsanith: With that all rows will have the same height everytime. You trigger a measure(), that will calculate and cache the height with the key `1`. All other rows will then be mapped with `1` too, and have the same size.
Is that what you want ?

packages/components/src/ListView/Items/Item/Item.component.js
```diff
@@ -8,58 +8,71 @@ function itemLabelClasses() {
...
+							</span>
+						</label>
+						{children && (
+							<div className={classNames('checkbox-nested-expand', { expanded: item.expanded })}>
+								<Action
+									bsStyle="link"
+									icon="talend-caret-down"
+									onClick={event => {
+										item.onExpandToggle(event, item);
+										measure();
```
jsomsanith: I don't think this will work as the Action will probably have some async steps (React render).
Moreover, you added a `componenWillReceiveProps`. I think you can remove this `measure()` on action trigger.

packages/components/src/ListView/Items/Item/Item.component.js
```diff
@@ -8,58 +8,71 @@ function itemLabelClasses() {
...
-	 * @param label the current label to parse & to render
-	 */
-	function getSearchedLabel(label) {
-		const splitRegex = new RegExp(`(${searchCriteria})`, 'gi');
-		return label.split(splitRegex).map((part, index) => {
-			const higlighted = part.toLowerCase() === searchCriteria.toLowerCase() ? <b>{part}</b> : part;
-			return <span key={index}>{higlighted}</span>;
-		});
+class Item extends Component {
+	componentWillReceiveProps(nextProps) {
```
jsomsanith: Please avoid using componentWillReceiveProps (will be deprecated in future versions of React).
Instead use `componentDidUpdate(prevProps)`. You then compare prevProps with current props to call `measure()`

output/components.eslint.txt
```diff
@@ -10,6 +10,9 @@ The react/require-extension rule is deprecated. Please use the import/extensions
   1:1  error  '@storybook/react' should be listed in the project's dependencies, not devDependencies  import/no-extraneous-dependencies
   4:2  error  Unexpected require()                                                                    global-require
 
+/home/travis/build/Talend/ui/packages/components/src/ListView/Items/Item/Item.component.js
+  24:72  error  'measure' is assigned a value but never used  no-unused-vars
```
jsomsanith: Please fix lint error

PR #1566, opened 2018-07-23
fix(AppLoader): wrong a11y attributes

jsomsanith:
AppLoader has `aria-busy=true` and `aria-atomic=true`.
The first one is wrong because the component/container is not loading, it's the app that is loading, and will replace the current component/container.
The second one is wrong because nothing changes in it, so screen readers don't need it. So in fact it is useless.

Remove them, and add `role=status` instead. Coupled with the current aria-label, it's understandable that the app is loading.

### DIFF
--- a/packages/components/src/AppLoader/constant.js
+++ b/packages/components/src/AppLoader/constant.js
@@ -170,7 +170,7 @@ body {
   }
 }`;
 
-const APP_LOADER = `<div class="tc-app-loader-container">
+const APP_LOADER = `<div class="tc-app-loader-container" aria-label="Loading application" role="status">
 	<div class="tc-app-loader-icon">
 		<div class="tc-app-loader" >
 			<div class="spinner-wrapper">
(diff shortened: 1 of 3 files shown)


packages/components/src/AppLoader/constant.js
```diff
@@ -170,7 +170,7 @@ body {
   }
 }`;
 
-const APP_LOADER = `<div class="tc-app-loader-container">
+const APP_LOADER = `<div class="tc-app-loader-container" aria-label="Loading application" role="status">
```
frassinier: aria-label should be i18n ready

sgangnant: Shouldn't this label be translated ?

jsomsanith: Not the constant one. The app is not loaded :/

jsomsanith: Not the constant one. The app is not loaded :/

PR #1568, opened 2018-07-24
feat(Dialog): improve a11y

jsomsanith:
Dialog missing
* role
* aria-modal
* aria-labelledby
* aria-describedby

* role=dialog
* aria-modal=true
* if we have a header: aria-labelledby use the header id. We suppose we have only 1 modal at a time, so no id generation needed
* for `aria-labelledby` (in case of no header) and `aria-describedby`, we can't guess user input, but all rest props are spread to the modal. So the user must manage that in its project

### DIFF
--- a/packages/components/src/Dialog/Dialog.component.js
+++ b/packages/components/src/Dialog/Dialog.component.js
@@ -31,12 +31,23 @@ function Dialog({
 	});
 	const injected = Inject.all(getComponent, components);
 
+	const headerId = 'tc-dialog-header';
+	const a11yProps = {
+		role: 'modal',
+		'aria-modal': true,
+	};
+	if (header) {
+		a11yProps['aria-labelledby'] = headerId;
+	}
+
 	return (
-		<Modal bsSize={size} {...props}>
+		// we disable jsx-a11y/aria-props because the version we use does not consider it valid (bug)
+		// eslint-disable-next-line jsx-a11y/aria-props
+		<Modal bsSize={size} {...a11yProps} {...props}>
 			{injected('before-modal-header')}
 			{header && (
 				<Modal.Header closeButton={closeButton}>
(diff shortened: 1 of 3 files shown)


packages/components/src/Dialog/Dialog.component.js
```diff
@@ -31,12 +31,23 @@ function Dialog({
 	});
 	const injected = Inject.all(getComponent, components);
 
+	const headerId = 'tc-dialog-header';
```
jmfrancois: please move it directly to the jsx

packages/components/src/Dialog/Dialog.component.js
```diff
@@ -31,12 +31,23 @@ function Dialog({
...
+	const a11yProps = {
+		role: 'dialog',
+		'aria-modal': true,
+	};
```
jmfrancois: same here

packages/components/src/Dialog/Dialog.component.js
```diff
@@ -31,12 +31,23 @@ function Dialog({
...
+		// we disable jsx-a11y/aria-props because the version we use does not consider it valid (bug)
+		// eslint-disable-next-line jsx-a11y/aria-props
+		<Modal bsSize={size} {...a11yProps} {...props}>
 			{injected('before-modal-header')}
 			{header && (
 				<Modal.Header closeButton={closeButton}>
-					<Modal.Title>{header}</Modal.Title>
+					<Modal.Title id={headerId}>{header}</Modal.Title>
 				</Modal.Header>
 			)}
```
jmfrancois:
```
<Modal bsSize={size} role="dialog" aria-modal aria-labelledby={header ? 'tc-dialog-header' : null} {...props}>

</Modal>
```

PR #1570, opened 2018-07-24
feat(Progress): improve a11y

jsomsanith:
Progress component doesn't implement aria.

* aria attributes on infinite progress (role=status + aria-busy)
* aria attributes on definite progress (role=progressbar + aria-(valuenow, valuemin, valuemax))
* make tooltip required to add it in aria-label
* ConfirmDialog accepts a props.progressLabel to be passed as tooltip

### DIFF
--- a/packages/components/src/Progress/Progress.component.js
+++ b/packages/components/src/Progress/Progress.component.js
@@ -40,8 +40,23 @@ function Progress({ id, percent, tooltip, infinite, contained }) {
 		);
 	}
 
+	let a11yProps;
+	if (!infinite) {
+		a11yProps = {
+			role: 'progressbar',
+			'aria-valuenow': normalizedPercent,
+			'aria-valuemin': 0,
+			'aria-valuemax': 100,
+		};
+	} else {
+		a11yProps = {
+			role: 'status',
+			'aria-busy': true,
+		};
+	}
+
 	return (
-		<div id={id} className={rootClassNames}>
+		<div id={id} className={rootClassNames} aria-label={tooltip} {...a11yProps}>
 			{progress}
 		</div>
 	);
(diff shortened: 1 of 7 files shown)


packages/components/src/Progress/Progress.component.js
```diff
@@ -52,7 +67,7 @@ Progress.displayName = 'Progress';
 Progress.propTypes = {
 	id: PropTypes.string,
 	percent: PropTypes.number,
-	tooltip: PropTypes.string,
+	tooltip: PropTypes.string.isRequired,
```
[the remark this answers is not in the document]
jsomsanith: Nope, it still work as before but you'll have a warning not passing it.

packages/components/src/Progress/Progress.component.js
```diff
@@ -40,8 +40,23 @@ function Progress({ id, percent, tooltip, infinite, contained }) {
...
+	} else {
+		a11yProps = {
+			role: 'status',
+			'aria-busy': true,
+		};
+	}
+
 	return (
-		<div id={id} className={rootClassNames}>
+		<div id={id} className={rootClassNames} aria-label={tooltip} {...a11yProps}>
```
jmfrancois: why not adding aria-label in a11yProps ?

jsomsanith: Because I would have to put it twice, of to add some uglier code to insert it :)
In fact only the aria-label is constant here. The other a11 attributes are conditioned.

jsomsanith:
@sgangnant I think you misunderstand a little the quote.
What is described is that the component that is loading should have the `aria-describedby` attribute on it, that point to the loader id. The loader is then the description of the component.
Progress is the loader here and accept an id. It's this id that should be on the loading component `aria-describedby`.

So the Progress is ok.

sgangnant:
You're right, my bad

PR #1572, opened 2018-07-24
fix(datagrid): fix redrawsRows when the datagrid is not showed

jare-talend:
- javascript error when props loading is true because ag-grid is not loaded
- custom function to force the redrawRows

fix that

after
```javascript
<DataGrid forceRedrawRows={
     (props, prevProps) => 
        // add your condition to force the redrawsRows when componentDidUpdate is called
       return true;
   } 
/>
```

### DIFF
--- a/packages/datagrid/src/components/DataGrid/DataGrid.component.js
+++ b/packages/datagrid/src/components/DataGrid/DataGrid.component.js
@@ -53,57 +53,6 @@ function getAvroRenderer(avroRenderer) {
 	};
 }
 
-/**
- * this function return information about how many row of each type are loaded
- * @param {array} rowData the data of the datagrid
- */
-export function getRowDataInfos(rowData) {
-	return rowData.reduce(
-		(acc, item) => {
-			if (item.loading === false && Object.keys(item).length === 2) {
-				// eslint-disable-next-line no-param-reassign
-				acc.notLoaded += 1;
-			} else if (item.loading === true) {
-				// eslint-disable-next-line no-param-reassign
-				acc.loading += 1;
-			} else {
-				// eslint-disable-next-line no-param-reassign
-				acc.loaded += 1;
-			}
-			return acc;
-		},
-		{
-			loaded: 0,
-			loading: 0,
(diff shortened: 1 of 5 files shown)


packages/datagrid/src/components/DataGrid/DataGrid.component.js
```diff
@@ -80,6 +80,30 @@ export function getRowDataInfos(rowData) {
...
+export function redrawRows(props, gridAPI, prevProps) {
+	if (props.loading || !gridAPI) {
+		return;
+	}
+
+	const prevInfos = getRowDataInfos(prevProps.rowData);
+	const currentInfos = getRowDataInfos(props.rowData);
+
+	if (
+		prevInfos.loaded !== currentInfos.loaded ||
```
acateland: not a fan of using many boolean for loading status, this can lead to impossible state easily, like having none of them true, or many of them true like, loaded and notLoaded, i rather have a `status` and `statusList` with the four status and a checker when updating `status` value

packages/datagrid/src/components/DataGrid/DataGrid.component.js
```diff
@@ -80,6 +80,30 @@ export function getRowDataInfos(rowData) {
...
 
+/**
+ * redrawRows - call redrawRows only if necessary. (improve ag-grid performance)
+ *
+ * @param  {object} props     component props
+ * @param  {object} gridAPI   ag-grid api
+ * @param  {object} prevProps previous component prop
+ */
+export function redrawRows(props, gridAPI, prevProps) {
+	if (props.loading || !gridAPI) {
```
acateland: you can revert the conditional and put the code in this if statement instead of using an early return

jdescottes-talend: What's problematic with early returns ? here or in general.

jare-talend: i prefer this rather that a condition pyramid or huge block.

PR #1573, opened 2018-07-25
fix(components): Align subheaderbar items

annilsson:
The subheaderbar items are not aligned properly (introduced in #1506 ):
- Center items are right-aligned
- It's difficult to align and size items equally when containers are listed like "left, left, center, right, right, right". 
- inputtitle takes over the subheaderbar completely if too big.
subheadertitle-margin

- Use 3 containers only: "left, center, right" that can contain multiple items. That allows for more accurate spacing,sizing, and alignment. 
- Subheaderbar center items are now centered in the middle between "left" and "right" containers' edges. 
subheaderbar-center

These 3 dummy examples might help demonstrating how align and grow/shrink should work now:
headerbar-examples

Only time inputtitle should need margin-right now would be if you add more "left"-items, but I don't think we do anywhere? If we do, we should keep some margin I guess. 


jmfrancois:
We will split the three case to have three different behavior, so let s close this PR for the moment.
Thanks for the effort

PR #1574, opened 2018-07-26
fix(Navbar) Dropdown width on IE11

sgangnant:
There are issues with navbar's dropdown list items width under IE11
image

Remove `display:flex` from navbar > li CSS rules

### DIFF
--- a/packages/theme/src/theme/_navbar.scss
+++ b/packages/theme/src/theme/_navbar.scss
@@ -30,7 +30,6 @@
 		display: flex;
 
 		li {
-			display: flex;
 			align-items: center;
 		}
 	}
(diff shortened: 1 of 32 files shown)


packages/theme/src/theme/_navbar.scss
```diff
@@ -25,7 +25,6 @@
 		display: flex;
 
 		li {
-			display: flex;
 			align-items: center;
```
annilsson: I don't think this does anything without flex? Can we not remove the whole `li` block?

sgangnant: You're probably right, although I wanted to make as less moves  as possible on this since I can't test storybook, updating PR

PR #1580, opened 2018-07-27
fix(List): placement and outline shape

jsomsanith:
The List outline weird shape was fixed, but a regression was added with the automatic dropup/dropdown.
capture d ecran 2018-08-06 a 13 52 33

The reason: we need dropdown menus to be invisible by default, rendered and displayed, so we can get its position to decide if we have an overflow.
If it is displayed, and takes its place, the outline is weird.

* Set back the non rendering of the menu (display: none)
* In dropdown, keep its open state and on component update, we add the `dropup` classname if needed


packages/components/src/Actions/ActionDropdown/ActionDropdown.component.js
```diff
@@ -134,36 +127,42 @@ class ActionDropdown extends React.Component {
...
+			const dropdownMenu = dropdownTrigger.nextSibling;
+			const dropdownContainer = getDropdownContainer(dropdownTrigger);
 
-			this.setState(oldState => {
-				if (!oldState.dropup && dropdownRect.bottom > containerRect.bottom) {
-					return { dropup: true };
-				} else if (oldState.dropup && dropdownRect.top < containerRect.top) {
-					return { dropup: false };
+			if (dropdownContainer) {
+				console.log('container found', dropdownContainer.tagName);
```
sgangnant: Did you leave those on purpose ? Plus, it adds error to the linting report (above)

jsomsanith: Oh no, thanks

jsomsanith: + I'll suppress the other linter errors as I haven't found any other way to do that

packages/components/src/Actions/ActionDropdown/ActionDropdown.test.js
```diff
@@ -110,79 +111,72 @@ describe('InjectDropdownMenuItem', () => {
...
+		const isDropup = true;
+		const isDropdown = false;
 
-		const wrapper = shallow(<ActionDropdown items={[{ label: 'item 1' }, { label: 'item 2' }]} />);
-		expect(wrapper.state().dropup).toBeFalsy();
+		function testSwitch({
+			containerPosition,
+			menuPosition,
+			isInitialDropup,
+			isDropupOrIsDropdown,
```
jdescottes-talend: Either the naming or the data stored is weird. I don't understand what's the meaning. When seeing it in standalone, It seems to be always true, and when seeing with the usage, I just don't understand neither ^^'

packages/components/src/Actions/ActionDropdown/ActionDropdown.component.js
```diff
@@ -134,36 +128,40 @@ class ActionDropdown extends React.Component {
 	constructor(props) {
 		super(props);
 		this.onToggle = this.onToggle.bind(this);
-		this.state = {
-			dropup: props.dropup,
-		};
+		this.state = {};
```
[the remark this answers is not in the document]
jsomsanith: By default a dropdown is not opened. So isOpen is falsy.

packages/components/src/Actions/ActionDropdown/ActionDropdown.test.js
```diff
@@ -110,79 +111,67 @@ describe('InjectDropdownMenuItem', () => {
...
-		// dropdown that is under the bottom of container
-		const bottomOverflowEvent = getActionDropdownEvent(20, 40, divContainerElement);
-		// dropdown that is under the bottom of body element
-		const bodyBottomOverflowEvent = getActionDropdownEvent(20, 40, bodyElement);
+		const isDropup = true;
+		const isDropdown = false;
 
-		const wrapper = shallow(<ActionDropdown items={[{ label: 'item 1' }, { label: 'item 2' }]} />);
-		expect(wrapper.state().dropup).toBeFalsy();
+		function testSwitch({ containerPosition, menuPosition, isInitialDropup, expectedDropupState }) {
```
jdescottes-talend: If I'm not mislead, this naming matching the data held : `isDropupExpected`

packages/components/src/Actions/ActionDropdown/ActionDropdown.test.js
```diff
@@ -110,79 +111,67 @@ describe('InjectDropdownMenuItem', () => {
...
+			// given
+			const container = document.createElement('div');
```
jdescottes-talend: When checking the tests in local, I can't find a way to go through this cases function.

packages/components/src/Actions/ActionDropdown/ActionDropdown.test.js
```diff
@@ -110,79 +111,67 @@ describe('InjectDropdownMenuItem', () => {
...
-		// when / then
-		wrapper.props().onToggle(true, bodyBottomOverflowEvent);
-		expect(wrapper.state().dropup).toBe(true);
+		cases('dropdown/dropup switch', testSwitch, [
+			{
+				name: 'should dropup on dropdown bottom overflow',
+				containerPosition: { top: 0, bottom: 35 },
+				menuPosition: { top: 20, bottom: 40 },
+				isInitialDropup: false,
+				isDropupOrIsDropdown: isDropup,
```
jdescottes-talend: The props cases need to be updated too.
And the data corresponding can be directly given ; `isDropup` and `isDropdown`, even if I see now why they are, are not obvious, and misleading. Simple boolean without variables would avoid thinking :p

packages/components/src/Actions/ActionDropdown/ActionDropdown.test.js
jsomsanith: done

packages/components/src/Actions/ActionDropdown/ActionDropdown.test.js
jsomsanith: fixed

packages/components/src/Actions/ActionDropdown/ActionDropdown.component.js
```diff
@@ -1,6 +1,8 @@
+/* eslint-disable-next-line import/no-extraneous-dependencies */
```
romainseb: Could it be better to add react-dom in peer-dependency of @talend/react-component as we use it this way ?

jsomsanith: You're totally right, I made the change, thanks

PR #1581, opened 2018-07-30
feat(components): date picker draft

jdescottes-talend:
Implement a draft of the date picker defined in this [guideline](http://guidelines.talend.com/document/92132#/ui-controls/date-picker-1513013325).

The actual work has globally been reviewed by @jsomsanith in sub PRs.
This is not a final implementation, some features will follow in next PRs for easy reviews.
Future features :
- dropdown
- form integration
- date only without time
- picker submit with an explicit button
- i18n
- l10n
- accessibility with keyboard navigation

Create a set of custom components to ease the implementation of the guideline (rather than using a lib which require a lot of override and unnecessary stuff).


packages/components/src/DateTimePickers/pickers/IncrementableScrollList/IncrementableScrollList.component.js
```diff
@@ -0,0 +1,127 @@
...
+
+		const startIndex = Math.max(this.props.initialIndex - Math.floor(NB_ITEMS_DISPLAYED / 2), 0);
+		this.scrollToIndex = keepInBoundaries(startIndex, firstIndex, lastIndex);
+
+		this.state = {
+			startIndex: this.scrollToIndex,
+		};
+
+		this.scrollUp = this.scrollRows.bind(this, -5);
+		this.scrollDown = this.scrollRows.bind(this, 5);
```
talend-glorieux: Is it the same `5` as `NB_ITEMS_DISPLAYED`.

packages/components/src/DateTimePickers/pickers/IncrementableScrollList/IncrementableScrollList.component.js
```diff
@@ -0,0 +1,127 @@
...
+		this.onRowsRendered = this.onRowsRendered.bind(this);
+		this.setListRef = this.setListRef.bind(this);
+	}
+
+	onRowsRendered(data) {
+		const { startIndex } = data;
```
talend-glorieux: Useless constant.

packages/components/src/DateTimePickers/pickers/MonthPicker/MonthPicker.component.js
```diff
@@ -0,0 +1,56 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { chunk } from 'lodash';
+import addMonths from 'date-fns/add_months';
+import format from 'date-fns/format';
+import theme from './MonthPicker.scss';
+import PickerAction from '../../shared/components/PickerAction';
+
+const baseDate = new Date(0);
```
talend-glorieux: This is repeated on another file.

packages/components/src/DateTimePickers/pickers/MonthPicker/MonthPicker.scss
```diff
@@ -0,0 +1,27 @@
+
```
talend-glorieux: Extra newline.

packages/components/src/DateTimePickers/shared/components/IconButton/IconButton.component.js
```diff
@@ -0,0 +1,45 @@
...
+		...buttonPropsRest
+	} = props;
+
+	const {
+		className: iconClassNameProp,
+		...iconPropsRest
+	} = iconProps;
+
+	const buttonClassName = classNames(buttonClassNameProp, theme.btn);
+	const iconClassName = classNames(iconClassNameProp, theme.icon);
```
talend-glorieux: Useless constants. classNames can be used directly in the return.

packages/components/src/DateTimePickers/views/HeaderTitle/HeaderTitle.component.js
```diff
@@ -0,0 +1,40 @@
...
+		className,
+		...(isButton ? props.button : {}),
+	};
+
+	const date = setYear(setMonth(new Date(0), props.monthIndex), props.year);
+	const label = format(date, 'MMMM YYYY');
+
+	return isButton
+		? <button type="button" {... propsToSpread}>{label}</button>
+		: <span {...propsToSpread}>{label}</span>;
```
talend-glorieux: Should be a regular `if/else`.

packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js
```diff
@@ -0,0 +1,160 @@
...
+
+		// eslint-disable-next-line
+		console.warn(`UNSTABLE WARNING: The 'DateTimePicker' and all the sub components aren't ready to be used in Apps. Code can (will) change outside the release process until it's ready.`);
+
+		const selectedDate = props.selection && props.selection.date;
+		const selectedTime = props.selection && props.selection.time;
+
+		const initialCalendarDate = selectedDate === undefined
+			? new Date()
+			: selectedDate;
```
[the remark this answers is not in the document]
jdescottes-talend: It's not the prop which is initialized, it's the initial date of the calendar.

packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js
```diff
@@ -0,0 +1,160 @@
...
+		this.onSelectCalendarYear = this.onSelectCalendarYear.bind(this);
+		this.onSelectCalendarMonthYear = this.onSelectCalendarMonthYear.bind(this);
+		this.onSelectDate = this.onSelectDate.bind(this);
+		this.onSelectTime = this.onSelectTime.bind(this);
+	}
+
+	componentWillReceiveProps(nextProps) {
+		const propsUpdated = this.props.selection !== nextProps.selection;
+
+		if (!propsUpdated) {
```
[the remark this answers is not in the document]
jdescottes-talend: Why 'It should' ? :D
I often declare variables to speak for the code/value it holds even if it's used only once.
I don't think of variable like it's there just for re-usability, but like it's hold the result of a computation even if it's intermediary.
In this way, it's not necessary to read the code for knowing what's done, just read the variables. I think it's ease the read with natural language.

packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js
```diff
@@ -0,0 +1,160 @@
...
+				onSelectMonth={this.onSelectCalendarMonth}
+				onSelectYear={this.onSelectCalendarYear}
+			/>);
+		}
+
+		return (
+			<div className={theme.container}>
+				{viewElement}
+			</div>
+		);
```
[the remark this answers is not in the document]
jdescottes-talend: How to you achieve it ? With duplication of the `div` container inside each `if/else` block ? ?

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -0,0 +1,250 @@
...
+import isSameMinute from 'date-fns/is_same_minute';
+import startOfDay from 'date-fns/start_of_day';
+import startOfMinute from 'date-fns/start_of_minute';
+import twoDigits from '../shared/utils/format/twoDigits';
+import DateTimePicker from '../DateTimePicker';
+import theme from './InputDateTimePicker.scss';
+
+const splitPartsRegex = new RegExp(/^\s*([^\s]+?)\s+([^\s]+?)\s*$/);
+const datePartRegex = new RegExp(/^\s*([0-9]{4})-([0-9]{1,2})-([0-9]{1,2})\s*$/);
+const timePartRegex = new RegExp(/^\s*([0-9]{1,2}):([0-9]{2})\s*$/);
```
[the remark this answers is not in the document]
jdescottes-talend: The variable are not self explaining ? I could change the first one to `splitDateAndTimePartsRegex` though.
But I don't see what I could tell more specific than the variables in a way that don't duplicate the exact RegExp code :/

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -0,0 +1,250 @@
...
+		return dateText;
+	}
+
+	const hours = Math.floor(time / 60);
+	const minutes = time % 60;
+
+	const timeText = `${twoDigits(hours)}:${twoDigits(minutes)}`;
+
+	return `${dateText} ${timeText}`;
+}
```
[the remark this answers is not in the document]
jdescottes-talend: `parse` is not possible because it only parse a full date, and here we can parse a date or time. There is other problems : 
```
// Seems to have no check for month boundaries
parse('2014-24-11T11:30:30') // => 2015-12-11T10:30:30.000Z
// Neither for day
parse('2014-01-42T11:30:30') // => 2014-02-11T10:30:30.000Z
```
And with l10n, there will be plenty of format, and I don't think any lib could handle all of that with a real error handling.

For format, I'll use it.

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -0,0 +1,250 @@
...
+
+		const fullDateUpdated = fullDate !== this.state.lastFullDate
+								&& !isSameMinute(fullDate, this.state.lastFullDate);
+
+		if (fullDateUpdated && this.props.onChange) {
+			this.props.onChange(fullDate);
+		}
+
+		const errorUpdated = errorMsg !== this.state.lastErrMsg;
+		if (errorUpdated && this.props.onError) {
```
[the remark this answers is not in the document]
jdescottes-talend: Conditions with all the checks inside are particularly unreadable. It needs more cerebral activity to understand the checks and deduce the meaning than just reading the variable that hold the meaning in it's semantic.

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -0,0 +1,250 @@
...
+		} = this.state;
+
+		return (
+			<div>
+				<DebounceInput
+					{...this.props.inputProps}
+					type="text"
+					placeholder="YYYY-MM-DD hh:mm"
+					value={textInput}
+					debounceTimeout={300}
```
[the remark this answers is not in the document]
jdescottes-talend: ✎

packages/components/src/DateTimePickers/pickers/DatePicker/DatePicker.component.js
```diff
@@ -0,0 +1,150 @@
...
+import theme from './DatePicker.scss';
+import DayPickerAction from './DayPickerAction';
+
+const FIRST_DAY_OF_WEEK = 1;
+const BASE_DATE = new Date(0);
+
+function buildDayNames(firstDayOfweek) {
+	return (new Array(7))
+		.fill(0)
+		.map((_, i) => (i + firstDayOfweek) % 7)
```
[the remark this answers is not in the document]
jdescottes-talend: ✎

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -0,0 +1,250 @@
...
+				return [undefined, errMsg];
+			}
+
+			const dateValidated = setDate(monthDate, day);
+
+			return [dateValidated];
+		})();
+
+
+		const [time, errMsgTime] = (() => {
```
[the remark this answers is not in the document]
jdescottes-talend: ✎

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -0,0 +1,250 @@
...
+	}
+
+	onChangeInput(event) {
+		const fullString = event.target.value;
+
+		const splitMatches = fullString.match(splitPartsRegex);
+
+		const canParseFullString = splitMatches !== null;
+
+		const [date, errMsgDate] = (() => {
```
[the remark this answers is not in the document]
jdescottes-talend: ✎

packages/components/src/DateTimePickers/pickers/DatePicker/DatePicker.component.js
```diff
@@ -0,0 +1,150 @@
...
+			weekStartsOn: firstDayOfWeek,
+		});
+		const nbWeeksToRender = diffWeeks + 1;
+
+		const dates = (new Array(7 * nbWeeksToRender))
+						.fill(0)
+						.map((_, i) => addDays(firstDateOfCalendar, i));
+
+		return chunk(dates, 7);
+	}
```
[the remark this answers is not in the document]
jdescottes-talend: ✎

packages/components/src/DateTimePickers/pickers/MonthPicker/MonthPicker.component.js
jdescottes-talend: ✎

packages/components/src/DateTimePickers/pickers/MonthPicker/MonthPicker.scss
jdescottes-talend: ✎

packages/components/src/DateTimePickers/shared/components/IconButton/IconButton.component.js
jdescottes-talend: More or less like for the others variables remarks.
Here extracting those informations from the JSX, give the possibility to think of the main structure and props without needing to dive directly in the exact data.

packages/components/src/DateTimePickers/views/HeaderTitle/HeaderTitle.component.js
jdescottes-talend: ✎

packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js
```diff
@@ -0,0 +1,160 @@
...
+			return;
+		}
+
+		const newSelectedDate = nextProps.selection && nextProps.selection.date;
+		const newSelectedTime = nextProps.selection && nextProps.selection.time;
+		const needToUpdateDate = newSelectedDate !== this.state.selectedDate;
+		const needToUpdateTime = newSelectedTime !== this.state.selectedTime;
+		const needToUpdateState = needToUpdateDate || needToUpdateTime;
+
+		if (!needToUpdateState) {
```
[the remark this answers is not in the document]
jdescottes-talend: Same response as for other unnecessary variables.

packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js
```diff
@@ -0,0 +1,160 @@
...
+		const newState = {
+			...(needToUpdateDate && { selectedDate: newSelectedDate }),
+			...(needToUpdateTime && { selectedTime: newSelectedTime }),
+			...(needToUpdateCalendar && {
+				calendar: {
+					monthIndex: getMonth(newSelectedDate),
+					year: getYear(newSelectedDate),
+				},
+			}),
+		};
```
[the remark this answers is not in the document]
jdescottes-talend:
```
const newState = {
    ...(needToUpdateDate && { selectedDate: newSelectedDate }),
    ...(needToUpdateTime && { selectedTime: newSelectedTime }),
    ...(needToUpdateCalendar && {
        calendar: {
            monthIndex: getMonth(newSelectedDate),
            year: getYear(newSelectedDate),
        },
    }),
};
```
is equivalent in the actual result with
```
let newState = {};

if (needToUpdateDate) {
    newState.selectedDate = newSelectedDate;
}

if (needToUpdateTime) {
    newState.selectedTime = newSelectedTime;
}

if (needToUpdateCalendar) {
    newState.calendar = {
        monthIndex: getMonth(newSelectedDate),
        year: getYear(newSelectedDate),
    };
}
```
It's conditional object prop assignment... , which allow creating state all in once, less risk to have additional code inserted between conditions or to reassign completely the variable.

packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js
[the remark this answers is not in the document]
talend-glorieux: It is true that using proper constant names improve comprehension. 
But in this case you're making extra allocations that are not reused. And on top of it the constant is negated in the conditional invalidating the naming improvement.

packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js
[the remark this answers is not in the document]
talend-glorieux: OK. I'd rather see the second option being used. So much clearer.

packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js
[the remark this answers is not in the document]
talend-glorieux: Yes. It's not an issue to duplicate one line of JSX for comprehension.

packages/components/src/DateTimePickers/pickers/IncrementableScrollList/IncrementableScrollList.component.js
talend-glorieux: I'd rather see the last one being used.
It removes an allocation.

packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js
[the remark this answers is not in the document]
jdescottes-talend: I have inverse the double negation for easy reading, I didn't notice it.
For the allocation and performance, yes it's a thing, but need to be relativized. We're not in a context with a billion record loop. It will be called only when user change the date via input or picker ; input is debounce and even if spam clicking the picker was possible (which is not, due to the submit feature), the impact will not be sensitive.

packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js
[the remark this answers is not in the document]
jdescottes-talend: I've change it.
It still think it's seems clearer because we arn't used to see the other pattern. It's a vicious circle :'(

packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js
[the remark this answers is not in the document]
jdescottes-talend: Having two lines that needs to be identical is a code synchronization maintenance issue.
I don't see any reason why it's actually complex or not understandable.
=> We decide what view is active and how it's define and then use it in the render. Here there is 2 views but it could have been more.

packages/components/src/DateTimePickers/pickers/IncrementableScrollList/IncrementableScrollList.component.js
jdescottes-talend: done it

packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.md
```diff
@@ -0,0 +1,12 @@
+# DateTimePicker
+
+This component display a datetime picker
+
+## props
+
+- **selection** : The datas used to control the current selection
```
talend-glorieux: data is already plural.

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -0,0 +1,250 @@
...
+	const timeText = `${twoDigits(hours)}:${twoDigits(minutes)}`;
+
+	return `${dateText} ${timeText}`;
+}
+
+function extractDate(strToParse) {
+	const dateMatches = strToParse.match(datePartRegex);
+
+	if (!dateMatches) {
+		const errMsg = 'DATE - INCORRECT FORMAT';
```
talend-glorieux: Useless constant.

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -0,0 +1,250 @@
...
+	const monthString = dateMatches[2];
+	const dayString = dateMatches[3];
+
+	const day = parseInt(dayString, 10);
+	const month = parseInt(monthString, 10);
+	const monthIndex = month - 1;
+	const year = parseInt(yearString, 10);
+
+	if (month === 0 || month > 12) {
+		const errMsg = 'DATE - INCORRECT MONTH NUMBER';
```
talend-glorieux: Useless constant.

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -0,0 +1,250 @@
...
+		return [undefined, errMsg];
+	}
+
+	if (day === 0) {
+		const errMsg = 'DATE - INCORRECT DAY NUMBER';
```
talend-glorieux: Useless constant.

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -0,0 +1,250 @@
...
+		return [undefined, errMsg];
+	}
+
+	const monthDate = new Date(year, monthIndex);
+	const lastDateOfMonth = lastDayOfMonth(monthDate);
+
+	if (day > getDate(lastDateOfMonth)) {
+		const errMsg = 'DATE - INCORRECT DAY NUMBER RELATIVE TO MONTH';
```
talend-glorieux: Useless constant.

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -0,0 +1,250 @@
...
+		return [undefined, errMsg];
+	}
+
+	const dateValidated = setDate(monthDate, day);
```
talend-glorieux: Useless constant.

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -0,0 +1,250 @@
...
+		return [undefined, errMsg];
+	}
+
+	const hoursString = timeMatches[1];
+	const minutesString = timeMatches[2];
+
+	const hours = parseInt(hoursString, 10);
+
+	if (hours >= 24) {
+		const errMsg = 'TIME - INCORRECT HOUR NUMBER';
```
talend-glorieux: Useless constant.

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -0,0 +1,250 @@
...
+		return [undefined, errMsg];
+	}
+
+	const minutes = parseInt(minutesString, 10);
+
+	if (minutes >= 60) {
+		const errMsg = 'TIME - INCORRECT MINUTES NUMBER';
```
talend-glorieux: Useless constant.

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -0,0 +1,250 @@
...
+		return [undefined, errMsg];
+	}
+
+	const timeValidated = hoursAndMinutesToTime(hours, minutes);
```
talend-glorieux: Useless constant.

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -0,0 +1,250 @@
...
+			lastErrMsg: errorMsg,
+		});
+	}
+
+	render() {
+		const {
+			date,
+			time,
+			textInput,
+		} = this.state;
```
talend-glorieux: No need to spread that value if it's used directly.
The `this.state.date` form is fine.

talend-glorieux:
Second round of reviews.

packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js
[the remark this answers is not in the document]
jdescottes-talend: no no no : there is no initial `selectedDate`, it should remain undefined if not set.
We only define the date for the month calendar to display here. When a date is selected, we initialize the calendar to show the selection and when no date is selected, we display by default the calendar corresponding to the current date : we don't touch the selection.

packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js
```diff
@@ -0,0 +1,160 @@
...
+import DateTimeView from '../views/DateTimeView';
+import MonthYearView from '../views/MonthYearView';
+
+class DateTimePicker extends React.Component {
+
+	constructor(props) {
+		super(props);
+
+		// eslint-disable-next-line
+		console.warn(`UNSTABLE WARNING: The 'DateTimePicker' and all the sub components aren't ready to be used in Apps. Code can (will) change outside the release process until it's ready.`);
```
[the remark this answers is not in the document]

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -0,0 +1,250 @@
...
+		onChange: PropTypes.func,
+		onError: PropTypes.func,
+		inputProps: PropTypes.object,
+	};
+
+	constructor(props) {
+		super(props);
+
+		// eslint-disable-next-line
+		console.warn(`UNSTABLE WARNING: The 'InputDateTimePicker' and all the sub components aren't ready to be used in Apps. Code can (will) change outside the release process until it's ready.`);
```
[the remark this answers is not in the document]
jdescottes-talend: ✔

packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js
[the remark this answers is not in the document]
jdescottes-talend: no no no : there is no initial `selection.date`, it should remain undefined if not set.

Here , we only define the date for the month calendar to display. When a date is selected, we initialize the calendar to show the selection and when no date is selected, we display by default the calendar corresponding to the current date : we don't touch the selection.

There is two things : the date selected (or no date selected) AND the calendar display at the current time (corresponding to a month-year).
The calendar can differ from the selection date : when searching for a date in an other month or just when there is no date selected (there is still a calendar displayed, the default current date at construct time or the last one showing if resetting the date via the input after the construct time).
=> We can discuss it if you want

packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js
```diff
@@ -0,0 +1,160 @@
...
+			}),
+		};
+
+		this.setState(newState);
+	}
+
+	onSelectDate(selectedDate) {
+		this.setState({ selectedDate }, () => {
+			this.trySubmit();
+		});
```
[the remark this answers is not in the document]
jdescottes-talend: I prefer keep it separate : the handler and the action(s). Even if there is only one action which didn't take any args, it could change in the future. And when it could, it will.
Furthermore, the callback is called with two args that will be spread to the action which don't want it.
Implicit spreading args like that all over the place is hard to reason about.
I don't think reduce the call-stack of 1 and removing some couple of parentheses and curly brackets is more important than explicit function calls and args definition.

packages/components/src/DateTimePickers/shared/utils/format/twoDigits/twoDigits.js
```diff
@@ -0,0 +1,5 @@
+export default function twoDigits(number) {
```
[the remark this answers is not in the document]
jdescottes-talend: Maybe, but it's far outside the current scope (will be for i18n or l10n).
It's not used right now, we'll do it when we need to. Just keep it simple for now :)

packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.md
jdescottes-talend: And when there is more than more ? :p
✔

packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js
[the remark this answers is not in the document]
jdescottes-talend: ✘

packages/components/src/DateTimePickers/DateTimePicker/DateTimePicker.component.js
```diff
@@ -0,0 +1,160 @@
...
+	}
+
+	onSelectTime(selectedTime) {
+		this.setState({ selectedTime }, () => {
+			this.trySubmit();
+		});
```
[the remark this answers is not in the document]
jdescottes-talend: Same response as above.
✘

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
jdescottes-talend: Not necessary but far to be useless.
It improves readability by keeping the return statement simple, add semantic of what's returning as second value.
Then for i18n, we'll need to add some more stuff, this will definitely don't feat with cleanliness.

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
jdescottes-talend: Same as above.
✘

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
jdescottes-talend: Same as above.
✘

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
jdescottes-talend: Same as above.
✘

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
jdescottes-talend: Same as above.
✘

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
jdescottes-talend: Same as above.
✘

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
jdescottes-talend: Same as above.
✘

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
jdescottes-talend: Same as above.
✘

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
jdescottes-talend: ✔

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -0,0 +1,239 @@
...
+
+		const [date, errMsgDate] = extractDate(dateStrToParse);
+
+		const timeStrToParse = canParseFullString ? splitMatches[2] : fullString;
+
+		const [time, errMsgTime] = extractTime(timeStrToParse);
+
+		const errMsg = canParseFullString ? errMsgDate || errMsgTime : 'DATETIME - INCORRECT FORMAT';
+
+		this.updateDateTime(date, time, fullString, errMsg);
```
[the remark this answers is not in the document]
jdescottes-talend: Huuum, the CI went through there....
It doesn't seem prettier now ^^'.

I'll do the changes.

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -0,0 +1,239 @@
...
+			lastFullDate: fullDate,
+			lastErrMsg: errorMsg,
+		});
+	}
+
+	render() {
+		return (
+			<div>
+				<DebounceInput
+					{...this.props.inputProps}
```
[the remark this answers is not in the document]
jdescottes-talend: The component `InputDateTimePicker` use/hold/render an input, but it's not an input. It's a composite of an input AND a picker for returning/giving back a `Date`. The input data are purely internal.
Moreover, at the component props level, mix the component ones and the input subcomponent ones disallow re-usage of the same prop name. Like a `className` for the top container and the input ; or an `onChange` for the component and for the input (I know here some props like `onChange` are overriden but we could merge them with a precedence rule or in a composition way for functions and there is the same deal to do with spreading the remaining).

In fact, I don't see one objective reason to spread all the component remaining props (besides having a flat props interface but even here I just see a story of taste).

If it was the implementation of the `DebounceInput` or some input wrapper like that, yeah, I would have been agree to say it's an input and the props interface is before all, the one of an input, but here, heeee, I see a date picker implementation with an input :P.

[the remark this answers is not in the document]
jmfrancois: > The component InputDateTimePicker use/hold/render an input, but it's not an input.
So what it is ?
In that case find a better name.

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -0,0 +1,236 @@
...
+
+	constructor(props) {
+		super(props);
+
+		// eslint-disable-next-line
+		console.warn(
+			"UNSTABLE WARNING: The 'InputDateTimePicker' and all the sub components aren't ready to be used in Apps. Code can (will) change outside the release process until it's ready.",
+		);
+
+		this.state = (() => {
```
jsomsanith: Please remove the auto invoked arrow function
```
if () {
    this.state = {...};
} else {
    this.state = {...};
}
```

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -0,0 +1,236 @@
...
+			};
+		})();
+
+		this.onChangeInput = this.onChangeInput.bind(this);
+		this.onSubmitPicker = this.onSubmitPicker.bind(this);
+	}
+
+	onSubmitPicker({ date, time }) {
+		this.updateDateTime(date, time);
+	}
```
jsomsanith: This function is not useful, and introduce more complexity as sometimes we have the date and time as params, and sometimes we have an object containing both.
Remove that and use `updateDateTime`. Change the api of `updateDateTime` to fit what is needed.

jdescottes-talend: We already discussed it, there is no complexity at all, even it's the reverse. Each piece is divided from the other, there is no strong coupling between components and between handler/function actions.
- `onSubmitPicker` is the handler, `updateDateTime` is the action.
    => An handler can trigger multiple actions, in fact that's what'll happen
```javascript
onSubmitPicker({ date, time }) {
	this.updateDateTime(date, time);
	this.switchDropdownVisibility(false);
}
```
- The signature of the two are not the same : date and time VS date and time and input string and error message.
- Explicit definition is far more clear about what happens and also what's not happening
    => If I do the following `onSubmit={this.updateDateTime}`, we need to go check in the other component what's the interface of submit, to verify with what args `updateDateTime` will be called.
    => Even worse, if we decide to add an argument to the submit callback, from the other component point of view, this is legitimate, it's an adding, it's not a breaking change, but for this component, it will break because it assumes the interface is immutable.
- The interface of the other component `DateTimePicker` is it's own. It have a coherent input/output interface with an identical object format. Transforming all interface to have exactly the same format value is just a step closer to strong coupling and lead always to : "change one little thing, break everything".
    => In the actual component, date and time variables/args, are always separate, it's totally coherent. And when we use the interface of an another component yes it can be different, but it's not one time like this one time like that, it's always coherent in a component scope point of view (internal method signature (ie not handlers from others components/libs/etc.), always have date and time separate)

In the other hand : the only benefit I see for tying functions together, is to remove some characters. And it's not at all a benefit in itself. It is when it's concise, not when it's obscure.

And I will not have any word about performance for a single function called added, the performance reasons, are always the worst one.

> Programs must be written for people to read, and only incidentally for machines to execute.
> -Abelson & Sussman, Structure and Interpretation of Computer Programs

PR #1590, opened 2018-08-06
feat(containers/List): default saga for filters and sorting

ybaskaran:
Custom functions support not available for filter and sorting in case required for server side filter or sorting with pagination.

Added default saga for onToggle, onFilter and onSorting for ListContainer.
The custom saga can be defined in the application that will override the default action defined in the ListContainer.


packages/containers/src/List/List.container.js
```diff
@@ -259,15 +264,24 @@ class List extends React.Component {
...
+						props.toolbar.sort.onChangeSortOrder(event, data);
+					}
 					this.onSelectSortBy(event, data);
 				};
 			}
 
 			if (props.toolbar.filter) {
 				props.toolbar.filter.onToggle = (event, data) => {
+					if (props.toolbar.filter.onToggleFilter) {
+						props.toolbar.filter.onToggleFilter();
```
jdescottes-talend: Is there a reason for not sending the `event` and `data` ?

packages/containers/src/List/List.container.js
```diff
@@ -191,7 +191,12 @@ class List extends React.Component {
 		props.list.sort = {
 			field: state.sortOn,
 			isDescending: !state.sortAsc,
-			onChange: this.onSelectSortBy,
+			onChange: (event, data) => {
+				if (props.toolbar.sort.onChangeSortOrder) {
+					props.toolbar.sort.onChangeSortOrder(event, data);
```
jdescottes-talend: The props are not defined as component propTypes.

packages/containers/src/List/List.test.js
```diff
@@ -99,8 +102,11 @@ describe('Container List', () => {
 		expect(typeof props.list.titleProps.onEditCancel).toBe('function');
 		expect(props.toolbar.filter.placeholder).toBe('find an object');
 		expect(typeof props.toolbar.filter.onFilter).toBe('function');
+		expect(typeof props.toolbar.filter.onToggleFilter).toBe('function');
```
jdescottes-talend: I don't think those checks need to be made here. It seems to be for validating the default props, when no one are given.
Here the only test I can see, for the adding expectations and some which where already there before is : verifying the above setting object 'settings' in this file which doesn't test something of the component :/ . I could swap the component rendered with whatever other component and those expectations are still true ^^.

packages/containers/src/List/List.container.js
ybaskaran: fixed

packages/containers/src/List/List.container.js
ybaskaran: no specific reason for excluding the attributes, its just that we are not using those values. However I agree that we might need these info in some other scenarios so Its added now.

packages/containers/src/List/List.test.js
ybaskaran: added unit tests for those callbacks separately

jdescottes-talend:
I don't understand why the code has diverged as much since the first reviews.
The code doesn't seem to fit the PR description anymore. Why the sagas and all the actions stuff are needed ? I don't understand that part. Why it shouldn't be in our apps ?

ybaskaran:
@jdescottes-talend the callbacks wont work because the filter or sorting logic is implemented in ListContainers via selectors in "List/selectors.js" hence i removed the selectors and included default saga to achieve it.
This is only for Apps which has server side pagination for Lists so the apps that is already using the ListContainer for client side operations will work with the default saga.

packages/containers/src/List/List.sagas.js
```diff
@@ -0,0 +1,32 @@
...
+
+function* defaultHandler() {
+	yield takeEvery(Constants.LIST_TOGGLE_FILTER, onToggleFilter);
+	yield takeEvery(Constants.LIST_FILTER_CHANGE, onFilterChange);
+	yield takeEvery(Constants.LIST_CHANGE_SORT_ORDER, onChangeSortChange);
+	yield take('DO_NOT_QUIT');
+}
+
+export default {
+	'List#default': defaultHandler,
```
jdescottes-talend: `default` strongly looks like the default componendId. I suggest a thing more like `'List#root'` (I really like this one :p) or `'List#startup'` (or main, boot, etc.)

packages/containers/src/List/List.sagas.js
```diff
@@ -0,0 +1,32 @@
...
+		sortOn: data.payload.field,
+		sortAsc: !data.payload.isDescending,
+	}));
+}
+
+function* defaultHandler() {
+	yield takeEvery(Constants.LIST_TOGGLE_FILTER, onToggleFilter);
+	yield takeEvery(Constants.LIST_FILTER_CHANGE, onFilterChange);
+	yield takeEvery(Constants.LIST_CHANGE_SORT_ORDER, onChangeSortChange);
+	yield take('DO_NOT_QUIT');
```
jdescottes-talend: I don't think this instruction is needed anymore, I don't use it in my app through. Normally, the root saga is dismantled when the component is unmount, as the sub sagas, but not before.

packages/containers/src/List/List.sagas.js
```diff
@@ -0,0 +1,32 @@
+import cmf from '@talend/react-cmf';
+import { put, takeEvery, take } from 'redux-saga/effects';
+import Constants from './List.constant';
+
+export function* onToggleFilter(data) {
+	yield put(cmf.actions.components.mergeState('Container(List)', 'default', {
```
jdescottes-talend: Hard coding the `componentId` `'default'` can lead to several issues :
- If we instantiate a List with a specific `componentId`, it will try to update the state of someone else.
    - So the state will just not been updated for all others List components than 'default'.
    - If the 'default' one exists, side effect will happen
    - If no 'default' exists, I think it will try to update a non existing/initialized state which is not possible and log an error.
- We can't instantiate 2 or more List connected components without knowing the fact that we need to override the default saga. It breaks the general rule of thumb that a component should be reusable.

packages/containers/src/List/List.connect.js
```diff
@@ -2,20 +2,18 @@ import get from 'lodash/get';
...
 function componentId(ownProps) {
 	return ownProps.collectionId;
 }
 
-function getItems(state, config) {
-	const items = configureGetFilteredItems(config)(state);
-
-	return items || new List();
+function getItems(componentState, config) {
+	let items = config.items;
```
jdescottes-talend: It would be easier to reason about if the variable is not declared.
Just return the value conditionally.

packages/containers/src/List/List.sagas.js
```diff
@@ -0,0 +1,76 @@
...
+				searchQuery: data.payload.query,
+			},
+			data.props.collectionId || 'default',
+		),
+	);
+	yield call(updateList, data);
+}
+
+export function* onToggleFilter(data) {
+	const filterData = data;
```
jdescottes-talend: you can change the arg name or remove this variable

packages/containers/src/List/List.sagas.js
```diff
@@ -0,0 +1,76 @@
...
+
+export function* updateList(data) {
+	const collectionId = data.props.collectionId || 'default';
+	const state = data.context.store.getState();
+	const componentState = state.cmf.components.getIn(['Container(List)', collectionId]);
+	const items = getCollectionItems(state, collectionId);
+
+	let results = items || data.props.config.items;
+	results = sortList(componentState, results, data.props.config);
+	results = filterList(componentState, results, data.props.config);
```
jdescottes-talend: Reassign a variable is hard to reason about. Plus the two selectors are not really selectors anymore, they do a holly bunch of thing, I don't even want to look inside x).

I suggest the following for the top part (here): 
```javascript
const stateItems = getCollectionItems(state, collectionId);
const propItems = data.props.config.items;

// Get data from state with simple selectors (sortBy, sortAsc, searchQuery, etc.)
// Doesn't even need a redux selector, just need to extract the lodash get

// Maybe extract before the starting List with checking if is an Immutable List if really needed.
const items = (stateItems || propItems || new List())
    .sort(sorting_fn)
    .filter(filtering_fn);
    // the sorting && filtering fn can be created with currying,
    // allowing config data injection reachable inside the final fn. 
```
In this way, the sorting and filter function doesn't deal with where the configs are, it's more portable and testable and more importantly readable :).

packages/containers/src/List/selector.js
```diff
@@ -41,91 +41,74 @@ export function configureGetPagination(state, { collectionId }) {
...
-
-	const getFilteredList = createSelector(
-		[getCollectionData(localConfig.collectionId), getComponentState(localConfig.collectionId)],
-		(items, componentState) => {
-			let results = items || localConfig.items;
-			if (componentState) {
-				const searchQuery = componentState.get('searchQuery');
-				if (searchQuery && results) {
-					results = results.filter(item => contains(item, searchQuery, localConfig.columns));
+export function sortList(componentState, items, localConfig) {
```
jdescottes-talend: We'll need to do something with this function, it deals with too many things (too complex). And should be elsewhere because it's not a redux/cmf state selector.

packages/containers/src/List/selector.js
```diff
@@ -41,91 +41,74 @@ export function configureGetPagination(state, { collectionId }) {
...
+		results = results.sort(compare);
+		if (!sortAsc) {
+			results = results.reverse();
+		}
+	}
+	return results;
+}
 
-	return createSelector([getSortedList, getComponentState], items => items);
+export function filterList(componentState, items, localConfig) {
```
jdescottes-talend: same as for the `sortList` fn

packages/containers/src/List/selector.js
```diff
@@ -41,91 +41,74 @@ export function configureGetPagination(state, { collectionId }) {
...
-		[getSortedList, getComponentState(configure.collectionId)],
-		(items, componentState) => {
-			let results = items;
+		[getComponentState(configure.collectionId)],
+		componentState => {
+			let results = configure.items;
 			if (componentState) {
+				if (componentState.get('items')) {
+					results = componentState && componentState.get('items');
+				}
```
jdescottes-talend: The `configureGetPagedItems` could have been refactor but whatever...
However you could have done the following :
```javascript
const stateItems = componentState.get('items');
if (stateItems) {
    results = stateItems;
}
```
and even
```javascript
results = componentState.get('items', results);
```

packages/containers/src/List/List.connect.js
ybaskaran: added condition to config.items to return empty List if items are not available initially

packages/containers/src/List/List.container.js
```diff
@@ -191,7 +192,15 @@ class List extends React.Component {
 		props.list.sort = {
 			field: state.sortOn,
 			isDescending: !state.sortAsc,
-			onChange: this.onSelectSortBy,
```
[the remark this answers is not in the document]
ybaskaran: removed

packages/containers/src/List/List.sagas.js
ybaskaran: the data needs to be updated to invoke filter change when the filterBar is closed and reassignment is not allowed here so using the variable

packages/containers/src/List/List.connect.js
jdescottes-talend: What I was telling is about the variable reassignment which is not necessary, and come with more thinking about how it's manipulated et so what's really inside, where what really interest us is what the function return.

```javascript
if (componentState) {
    return componentState.get('items', config.items);
}
return config.items;
```

and if we know we have a state which would be a better idea :
```javascript
    return componentState.get('items', config.items);
```
which would have removed the need for a function at all ;)

packages/containers/src/List/List.sagas.js
jdescottes-talend: All objects are given by reference so when you're assigning `data` to `filterData`, you just getting the reference, not a copy.
So below, writing `filterData.payload.searchQuery = '';` or `data.payload.searchQuery = '';` is equivalent.

packages/containers/src/List/utils.js
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	return null;
+}
+
+export function getComponentState(collectionId) {
+	return state => state.cmf.components.getIn(['Container(List)', collectionId || 'default']);
+}
+
+export function sortList(componentState, localConfig, items) {
```
jdescottes-talend: Same as filter fn, we need a simple function interface

```javascript
export function sortList(items, sorBy, sortAsc, columnsConfig) {
```
We need also to expect the args are properly given, or can be specifically undefined (with a default value or a specific logic if not defined)

packages/containers/src/List/utils.js
```diff
@@ -0,0 +1,101 @@
...
+	if (!List.isList(items)) {
```
jdescottes-talend: We need to assume `items` to be a `List` in this fn. It's the job of the caller to give the good args.
This remove the need the check and early return

packages/containers/src/List/utils.js
```diff
@@ -0,0 +1,101 @@
...
+		return new List();
+	}
+	let results = items;
+	if (componentState) {
+		const sortBy = componentState.get('sortOn');
+		const sortAsc = componentState.get('sortAsc');
```
jdescottes-talend: Those values will be given in specifics args.
And the exist check of componentState will not be needed anymore

packages/containers/src/List/utils.js
```diff
@@ -0,0 +1,101 @@
...
+
+		const sortedColumn = get(localConfig, 'columns', [])
+			.filter(column => column.key === sortBy)
+			.pop();
+		if (sortedColumn && sortedColumn.sortFunction) {
+			compare = cmf.registry.getFromRegistry(sortedColumn.sortFunction)(sortBy);
+		}
+		results = results.sort(compare);
+		if (!sortAsc) {
+			results = results.reverse();
```
jdescottes-talend: the `asc` or `desc` should be done directly in the sort method.
You can use currying for creating the compare fn like this :

```javascript
const compareBy = sortAsc => (aIn, bIn) => {
   // you can inverse 'a' and 'b' if in descending with some other variables
   const a = sortAsc ? aIn : bIn;
   const b = sortAsc ? bIn : aIn;
   ...
}

------

results.sort(compareBy(sortAsc));
```

jdescottes-talend: After that the ascending check is not more needed in the top level of the function. We can now just return the result of the sorting. And BTW, we can directly use the `items` parameter without assign it to an other variable so :

```javascript
return items.sort(compareBy(sortAsc));
```

packages/containers/src/List/utils.js
```diff
@@ -0,0 +1,101 @@
...
+
+export function filterList(componentState, localConfig, items) {
+	let results = items;
+	if (componentState && componentState.get('searchQuery')) {
+		results = items.filter(item =>
+			contains(item, componentState.get('searchQuery'), localConfig.columns),
+		);
+	}
+	return results;
+}
```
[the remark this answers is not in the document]
jdescottes-talend: What I was telling by using currying in the first review is for using the filter that way :
```javascript
// Saga file

// Do the filter only if searchQuery exists
const items = ...
const filteredItem = searchQuery
    ? items.filter(containsQuery(searchQuery, columnConfig))
    : items;

-----------

// Here (and don't really need to be in the utils file now it's lighter).

export function containsQuery(searchQuery, columnsConfig) {
    return item =>
        contains(item, searchQuery, columnsConfig),
    );
}
```

packages/containers/src/List/List.sagas.js
```diff
@@ -0,0 +1,80 @@
...
+	const filterItemsCurried = sortCurried(componentState, config);
+	return filterItemsCurried(filterListFn(componentState, config, results));
+};
+
+export function* updateList({ props, context }) {
+	const collectionId = props.collectionId || 'default';
+	const state = context.store.getState();
+	const stateItems = getCollectionItems(state, collectionId);
+	const componentState = state.cmf.components.getIn(['Container(List)', collectionId]);
+
```
jdescottes-talend: Extract here in `const` the `isAsc`, `sortBy`, `searchQuery` and `columnsConfig`.
It's here you need to deal if the componentState doesn't exists or for those values with default or something else.
I don't see a case where component state don't exists other that selecting a bad part of the state, so it should always exists.

packages/containers/src/List/List.sagas.js
```diff
@@ -0,0 +1,80 @@
...
+	const items = stateItems || props.config.items || new List();
+	const results = getFilteredItems(filterList)(componentState, props.config, items);
```
jdescottes-talend: the currying was for the functions for the sort and filter directly not new top one :p
like this :

```javascript
// Here I assume 'stateItems' and 'props.config.items' are Immutable List, if not always the case, will need to check that before, an make sure we have at this point a strong consistent 'items' Immutable List define (empty or not but defined).
const items = stateItems || props.config.items || new List();
const filteredItem = searchQuery
    ? items.filter(containsQuery(searchQuery, columnConfig))
    : items;
// Here we do the same as for the filter, the 'sortCurriedToRename' need to return a function with the normal interface : (item1, item2) => compute boolean with access to sortBy, etc.
const sortedItems = filteredItem.sort(sortCurriedToRename(sortBy, sortAsc, columnsConfig));
```

I inverse the filter before the sort, I don't know if it's possible for the current case, but it make less computation to do this in this order. It should be possible unless if there is side effects somewhere.

packages/containers/src/List/List.sagas.js
```diff
@@ -0,0 +1,49 @@
...
+			{
+				searchQuery: data.payload.query,
+			},
+			data.props.collectionId || 'default',
+		),
+	);
+}
+
+export function* onToggleFilter(data) {
+	const filterData = data;
```
jdescottes-talend: Now it can be removed :)

packages/containers/src/List/List.sagas.js
```diff
@@ -0,0 +1,49 @@
...
+	yield put(
+		Connected.setStateAction(
+			{
+				filterDocked: !filterData.payload.filterDocked,
+				searchQuery: '',
+			},
+			filterData.props.collectionId,
```
jdescottes-talend: Add ` || 'default'` for consistency

packages/containers/src/List/List.connect.js
```diff
@@ -26,20 +21,26 @@ function getPagedItems(state, config) {
...
+	props.config = {
 		collectionId: ownProps.collectionId,
-		items: ownProps.items,
+		items: collectionItems || ownProps.items || new List(),
 	};
 	if (ownProps.list) {
-		config.columns = ownProps.list.columns;
+		props.config.columns = ownProps.list.columns;
+	}
+	if (ownProps.defaultSaga !== false) {
```
romainseb: i'm not sure we want that, it mean that the saga would be loaded by default on the list ?
Why the props `defaultSaga` is checked to set a `saga` attribute then ? 
Shouldn't it test the `saga` attribute instead ?

ybaskaran: The saga attribute shouldn't be loaded when the defaultSaga flag is set to false, this enables the target application to load the custom saga (i.e., the default saga will not be called).
Existing applications that use ListContainer will function normally since the defaultSaga prop is newly introduced for server side list actions.

packages/containers/src/List/List.connect.js
```diff
@@ -4,42 +4,43 @@ import { cmfConnect } from '@talend/react-cmf';
...
 	return items || new List();
 }
 
 export function mapStateToProps(state, ownProps, cmfProps) {
 	const props = {};
-	const config = {
+	let filteredItems;
+	const collectionItems = getCollectionItems(state, ownProps.collectionId);
+
+	props.config = {
```
jsomsanith: You attach the config to props but you don't use it except in this function.
Please remove it, this will introduce perf issues, as it's always a new object at each render.

packages/containers/src/List/List.connect.js
```diff
@@ -4,42 +4,43 @@ import { cmfConnect } from '@talend/react-cmf';
...
 		collectionId: ownProps.collectionId,
-		items: ownProps.items,
+		items: collectionItems || ownProps.items || new List(),
```
jsomsanith: If there is no items, to attach a new `List`. It means that it's a new ref at each render.
Can you remove that and replace it by a default props in container please ?

packages/containers/src/List/List.connect.js
jsomsanith: But that's not how the saga mount feature works in CMF.
`props.data` should be a boolean to indicate if the attached saga should be mounted or not.

So if you don't want to mount the default saga, you just don't set the saga props in your cmf settings. No need to introduce a new props `defaultSaga`.

And I don't get the fact that you filter only if you want to use the default sagas. That sound weird.

packages/containers/src/List/List.connect.js
```diff
@@ -6,29 +6,40 @@ import {
...
 
 export function mapStateToProps(state, ownProps, cmfProps) {
 	const props = {};
+	const collectionItems = getCollectionItems(state, ownProps.collectionId);
 	const config = {
 		collectionId: ownProps.collectionId,
-		items: ownProps.items,
+		items: collectionItems || ownProps.items || ownProps.listItems,
+		defaultFilter: get(ownProps, ['toolbar', 'filter', 'defaultFilter']),
+		defaultSorting: get(ownProps, ['sort', 'defaultSorting']),
```
jsomsanith: Name the props `defaultSort` to use the same ton as `defaultFilter` (or defaultSorting/defaultFiltering)

packages/containers/src/List/List.connect.js
```diff
@@ -6,29 +6,40 @@ import {
...
+	}
+	return items;
 }
 
-function getPagedItems(state, config) {
-	const items = configureGetPagedItems(config)(state);
-
-	return items || new List();
+function getPagedItems(state, config, items) {
+	if (config.defaultFilter !== false && config.defaultSorting !== false) {
```
jsomsanith: Why don't you add a `props.defaultPage(ing)`. It's the same issue as before, this is too implicit to have default pagging is defaultFilter and defaultSorting is  activated :/

packages/containers/src/List/List.connect.js
```diff
@@ -66,6 +77,10 @@ export function mergeProps(stateProps, dispatchProps, ownProps) {
 
 export default cmfConnect({
 	defaultState: DEFAULT_STATE,
+	defaultProps: {
+		saga: 'List#root',
```
jsomsanith: `saga: true`

packages/containers/src/List/List.connect.js
```diff
@@ -6,29 +6,40 @@ import {
...
 
-function getPagedItems(state, config) {
-	const items = configureGetPagedItems(config)(state);
-
-	return items || new List();
+function getPagedItems(state, config, items) {
+	if (config.defaultFilter !== false && config.defaultSorting !== false) {
+		return configureGetPagedItems(config, items)(state);
+	}
+	return config.items;
```
jsomsanith: Isn't that `items` you should return ?

packages/containers/src/List/List.connect.js
ybaskaran: we should provide saga Id to mount the saga

packages/containers/src/List/List.connect.js
ybaskaran: fixed

packages/containers/src/List/List.connect.js
ybaskaran: fixed

packages/containers/src/List/List.connect.js
ybaskaran: this function returns the paged items except when the defaultPaging is set to false

ybaskaran: and yes the function should return the items, its fixed now

packages/containers/src/List/List.connect.js
ybaskaran: fixed, enabled saga under default props and disabled filter and sorting conditionally if the target app needs to override the default list operations

PR #1454, opened 2018-06-14
feat(cmf): add bootstrap modules

jmfrancois:
**What is the problem this PR is trying to solve?**

We have more and more webapps based on CMF and some parts are shared (like containers + some other privates).
When you build a project you include those shared and may provide some customization around it).

**What is the chosen solution to this problem?**

Lets call @talend/react-containers a cmf module.

A CMF Module leave in a package and provide a configuration object which is understood by cmf.bootstrap.

So we can do the following:

```javascript
import cmf from '@talend/react-cmf';
import containers from '@talend/react-containers/lib/cmfModule';
import myPrivateLib from '@talend/my-private-lib';
import components from './components';

cmf.bootsrap({
  modules: [containers, myPrivateLib]
  components,
});
```

So with this PR each module can expose:

* components
* expressions
* actionCreators
* saga
* sagas
* sagaRouterConfig

**Please check if the PR fulfills these requirements**

* [x] The PR commit message follows our [guidelines](https://github.com/talend/tools/blob/master/tools-root-github/CONTRIBUTING.md)
* [ ] Tests for the changes have been added (for bug fixes / features) And [non reg](./screenshots.md) done before need review
* [ ] Docs have been added / updated (for bug fixes / features)
* [ ] Related design / discussions / pages (not in jira), if any, are all linked or available in the PR

**[ ] This PR introduces a breaking change**

PR #1596, opened 2018-08-07
feat(cmf/TUI-279): add cmf.bootstrap modules support

jmfrancois:

This PR is an alternative implementation to #1454 

We have more and more web apps based on CMF and some parts are shared (like containers + some other privates).
When you build a project you include those shared and may provide some customization around it.

Add support to an array of modules in the configuration which will itself contains configuration.
Those configurations can themselves have an array of modules.

All that will be merged by cmf.bootstrap.

update containers and cqrs to export default `cmfModule`.

So in your project, you can create a `cmfModule` file with

```javascript
import containersModule from '@talend/react-containers';
import components from './components';
import saga from './saga';

export default {
   components,
   saga,
   modules: [containersModule],
};
```

Then in your `index.js` file

```javascript
import cmf from '@talend/react-cmf';
import app from './cmfModule';

cmf.bootstrap(app);
```


jmfrancois:
> I don't see a scoping feature
There are no scope.

> This allows interferences naming between modules
More than that CMF exists just for that. Allow overrides.

> I don't see how to tell the merge mechanism it's component from module A or module B to keep.
The order.

> And there will remain the problem of private components
This notion doesn't exists.

> I don't understand the principle of flat components you mentioned, but I definitely think, mixing all modules stuff (components, action creators, etc.) go against the modularity principle and make modules integrations hard to reason.

What a conclusion !
Please define `modularity principle`.
> make modules integrations hard to reason
This is the opposite, at least from the current situation.

jmfrancois:
about your first review:

> You need to explain totally the concept of the modules in this first PR.

No it s already done in the PR I mentioned in the first line #1454.

packages/cmf/src/cmfModule.js
```diff
@@ -0,0 +1,151 @@
+import compose from 'redux';
+import { fork } from 'redux-saga/effects';
+import { assertTypeOf } from './assert';
+
+/* eslint-disable no-param-reassign */
```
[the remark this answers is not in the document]
jmfrancois: this rule should be avoid every time you do a reduce

packages/cmf-cqrs/src/cmfModule.js
```diff
@@ -0,0 +1,11 @@
+import actions from './actions';
```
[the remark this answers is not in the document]
jdescottes-talend: To identify easily what's what :)
The suffix convention help identify it's a specific convention so a specific file with a specific format.

Example with Angular ecosystem :
```typescript
// Imports examples
import {NgModule} from '@angular/core';
import {ScrollingModule} from '@angular/cdk/scrolling';
import {A11yModule} from '@angular/cdk/a11y';
import {CdkTableModule} from '@angular/cdk/table';
import {CdkTreeModule} from '@angular/cdk/tree';
import {DragDropModule} from '@angular/cdk/drag-drop';

```
File naming real examples :
`angular/packages/http/src/http_module.ts`

[Angular guidelines](https://angular.io/guide/styleguide#style-02-02) are really super great, even if they just don't follow they own rules (file name for they own modules...)

packages/containers/src/cmfModule.js
```diff
@@ -0,0 +1,16 @@
...
+import * as allComponents from '@talend/react-components';
+import * as containers from './index';
+
+const components = Object.keys(allComponents).reduce((acc, key) => {
+	if (!acc[key] && typeof allComponents[key] === 'function') {
+		// eslint-disable-next-line no-param-reassign
+		acc[key] = cmfConnect({})(allComponents[key]);
+	}
+	return acc;
+}, omit(containers, ['actionAPI']));
```
jdescottes-talend: I know it's not part of this PR, but I don't think this utility have something to do in the current package. It seems to belong to cmf package. Maybe it could be fix in an other PR or I open a JIRA issue ?

packages/cmf/src/mergeModules.js
```diff
@@ -0,0 +1,188 @@
...
+			return mergeObjects(reducer, newReducer, 'reducer');
+		}
+		if (typeof reducer === 'object' && typeof newReducer === 'function') {
+			return mergeObjects(reducer, { app: newReducer }, 'reducer');
+		}
+		if (typeof newReducer === 'object' && typeof reducer === 'function') {
+			return mergeObjects({ app: reducer }, newReducer, 'reducer');
+		}
+		// both are functions
+		return mergeFns(reducer, newReducer);
```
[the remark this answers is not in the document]
jmfrancois: This is quite hard. Reducer function are done to work on the state. The question

packages/containers/src/cmfModule.js
jmfrancois: There two action API, one for the CMF already which I would like to remove.
In Container you have multiple Action(s/dropdown/...) containers which have in some way the same mapStateToProps, so we have moved that code here. For Me it should not be exposed at all but it s too late.

So now I remove this object from the import, and this is why I would like to not put the cmfModule at the index.js

So we have one other solution it s to put cmfModule as a default export of containers

packages/cmf/src/mergeModules.js
```diff
@@ -0,0 +1,188 @@
...
+	storeCallback: mergeStoreCallback,
+	reducer: mergeReducer,
+	preloadedState: getFirstOf,
+	settingsURL: getFirstOf,
+	registry: getReduceRegistry('registry', 'object'),
+	sagas: getReduceRegistry('sagas', 'object'),
+	components: getReduceRegistry('components', 'object'),
+	expressions: getReduceRegistry('expressions', 'object'),
+	actionCreators: getReduceRegistry('actionCreators', 'object'),
+	sagaRouterConfig: getReduceRegistry('sagaRouterConfig', 'object'),
```
jsomsanith: You don't need this `getReduceRegistry`. It's only objects.

In general, for each argument, you don't need all those. You have only thos cases : 
* getFirstOf --> perhaps rename to `getUnique()`
* mergeObject : if the result should be an object
* mergeArray : if the result should be an array. Use [], concat to avoid too many if to check the type (array, function, ...)
* mergeFunction : if the result should be a function that calls all the functions you pass.
* mergeSaga : special case for saga

packages/cmf/src/bootstrap.js
```diff
@@ -27,7 +28,11 @@ export function bootstrapRegistry(options) {
 	assertTypeOf(options, 'components', 'object');
 	assertTypeOf(options, 'expressions', 'object');
 	assertTypeOf(options, 'actionCreators', 'object');
+	assertTypeOf(options, 'registry', 'object');
 	registerInternals();
+	if (options.registry) {
```
jsomsanith: Why do you add a `registry` option ?
It's another alternative for `components/expressions/actionCreators` :/ . Let's narrow the options instead of duplicate them

packages/cmf/src/bootstrap.md
```diff
@@ -37,3 +37,18 @@ cmf.bootstrap({
...
+
+For that CMF expose an API to let you merge your settings:
+
+```javascript
+import cmf from '@talend/react-cmf';
+
+import containersModule from '@talend/react-containers/lib/cmfModule';
+import myAppModule from './cmfModule';
+
+cmf.boostrap(cmf.modules.merge(containersModule, myAppModule));
```
jsomsanith: I would see another api instead : 
```
import dataset from '@talend/dataset';
import sharing from '@talend/sharing';

cmf.boostrap({
    ...myAppModule,
    modules: [dataset, sharing]
});
```

And it's cmf.boostrap that merges : 1 less call, 1 less complexity

packages/containers/src/cmfModule.js
```diff
@@ -0,0 +1,16 @@
+import { cmfConnect } from '@talend/react-cmf';
+import omit from 'lodash/omit';
+import * as allComponents from '@talend/react-components';
+import * as containers from './index';
+
+const components = Object.keys(allComponents).reduce((acc, key) => {
+	if (!acc[key] && typeof allComponents[key] === 'function') {
+		// eslint-disable-next-line no-param-reassign
+		acc[key] = cmfConnect({})(allComponents[key]);
```
jsomsanith: This doesn't work for class components.
And why is this duplicated in `index.js` and here `cmfMudle.js`

packages/cmf/src/mergeModules.js
```diff
@@ -0,0 +1,188 @@
...
+function mergeReducer(reducer, newReducer) {
+	assertTypeOf({ reducer }, 'reducer', ['object', 'function']);
+	assertTypeOf({ reducer: newReducer }, 'reducer', ['object', 'function']);
+
+	if (reducer && newReducer) {
+		if (typeof reducer === typeof newReducer && typeof reducer === 'object') {
+			return mergeObjects(reducer, newReducer, 'reducer');
+		}
+		if (typeof reducer === 'object' && typeof newReducer === 'function') {
+			return mergeObjects(reducer, { app: newReducer }, 'reducer');
```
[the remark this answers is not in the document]
jmfrancois: this is a breaking change, so it can be done now

packages/cmf/src/bootstrap.js
jmfrancois: Its not an alternative. The registry is the only way to register somethings outside of the notions of `components/expressions/actionCreators/sagas`.
It lets the application use DI for specific needs.

packages/cmf/src/bootstrap.md
jmfrancois: I really like that, no more things exported, lets make cmf a set of tools + bootstrap and it s ok.

output/cmf.eslint.txt
```diff
@@ -2,6 +2,11 @@
 
 The react/require-extension rule is deprecated. Please use the import/extensions rule from eslint-plugin-import instead.
 
+/home/travis/build/Talend/ui/packages/cmf/src/cmfModule.merge.js
+  3:1   error  'reassemble' should be listed in the project's dependencies. Run 'npm i -S reassemble' to add it  import/no-extraneous-dependencies
+  3:10  error  'noOp' is defined but never used                                                                  no-unused-vars
+  3:22  error  Absolute imports should come before relative imports                                              import/imports-first
```
jsomsanith: Please fix the linter errors

packages/cmf/__tests__/cmfModule.merge.test.js
```diff
@@ -0,0 +1,192 @@
...
+		const b = {
+			components: {
+				foo: function foo() {},
+			},
+		};
+		const config = mergeModules(a, b);
+		expect(config.components.foo).toBe(b.components.foo);
+		expect(config.components.bar).toBe(a.components.bar);
+		expect(config.components.foo).toBe(b.components.foo);
+		expect(config.components.bar).toBe(a.components.bar);
```
jsomsanith: You check the 2 attributes twice :)
Same for lots of tests in this file

packages/cmf/__tests__/cmfModule.merge.test.js
```diff
@@ -0,0 +1,192 @@
...
+	});
+	it('should throw an exeption on unknown keys', () => {
+		const toThrow = () => mergeModules({ foo: {} });
+		expect(toThrow).toThrow();
+	});
+	it('should throw an exeption if two config has appId', () => {
+		const toThrow = () => mergeModules({ appId: 'foo' }, { appId: 'bar' });
+		expect(toThrow).toThrow();
+	});
+	it('should get first appId in config', () => {
```
jsomsanith: I don't think this test is really necessary. Or just test 1 case, where there is an app id in one of the configs and check that it is taken in the merged config.
The description is disturbing, as we only allow a unique app id

packages/cmf/__tests__/cmfModule.merge.test.js
```diff
@@ -0,0 +1,192 @@
...
+				foo: function* foo() {},
+			},
+		};
+		const config = mergeModules(a, b);
+		expect(config.actionCreators.foo).toBe(b.actionCreators.foo);
+		expect(config.actionCreators.bar).toBe(a.actionCreators.bar);
+		expect(config.actionCreators.foo).toBe(b.actionCreators.foo);
+		expect(config.actionCreators.bar).toBe(a.actionCreators.bar);
+	});
+	it('should throw an exeption on unknown keys', () => {
```
jsomsanith: `exeption` --> `exception`
(in the whole file)

packages/cmf/__tests__/cmfModule.merge.test.js
```diff
@@ -0,0 +1,192 @@
...
+		const left = mergeModules({ appId: 'foo' }, {}, {});
+		const right = mergeModules({}, {}, { appId: 'bar' });
+		expect(left.appId).toBe('foo');
+		expect(right.appId).toBe('bar');
+	});
+	it('should throw an exeption if two config has history', () => {
+		const toThrow = () => mergeModules({ history: 'foo' }, { history: 'bar' });
+		expect(toThrow).toThrow();
+	});
+	it('should throw an exeption if two config has history', () => {
```
jsomsanith: Exact same test as above

packages/cmf/__tests__/cmfModule.merge.test.js
```diff
@@ -0,0 +1,192 @@
...
+		const toThrow = () => mergeModules({ history: 'foo' }, { history: 'bar' });
+		expect(toThrow).toThrow();
+	});
+	it('should merge enahncer functions', () => {
```
jsomsanith: `enahncer` --> `enhancer`

packages/cmf/src/bootstrap.md
jsomsanith: Now that you added that in bootstrap, please update this example

packages/cmf/src/cmfModule.merge.js
```diff
@@ -0,0 +1,122 @@
...
+import { noOp } from 'reassemble';
+
+function mergeObjects(obj1, obj2) {
+	if (!obj2 && obj1) {
+		return obj1;
+	}
+	if (!obj1 && obj2) {
+		return obj2;
+	}
+	return Object.keys(obj2).reduce((acc, key) => {
```
jsomsanith: If obj1 AND obj2 are undefined, you'll execute that and have an error.
Please change that to
```javascript
if (!obj2) {
	return obj1;
}
if (!obj1) {
	return obj2;
}
return Object/keys(obj2).reduce(....); // here obj1 and obj2 are defined
```

packages/cmf/src/cmfModule.merge.js
```diff
@@ -0,0 +1,122 @@
...
+		if (obj2[key] && obj1[key] !== obj2[key]) {
+			// eslint-disable-next-line no-console
+			console.warn(`override detected ${key}`);
+		}
+		return {
+			...acc,
+			[key]: obj2[key],
```
jsomsanith: - If `obj2[key]` is set to undefined, the previous value is still overridden and don't have the warning
- If `obj2[key]` is falsy, we don't have the warning

packages/cmf/src/cmfModule.merge.js
```diff
@@ -0,0 +1,122 @@
...
+}
+
+function mergeFns(fn1, fn2) {
+	if (!fn2 && fn1) {
+		return fn1;
+	}
+	if (!fn1 && fn2) {
+		return fn2;
+	}
+	return function mergedFn(...args) {
```
jsomsanith: Same remarks as merge object fn

packages/cmf/src/assert.js
```diff
@@ -1,16 +1,19 @@
-// eslint-disable-next-line import/prefer-default-export
-export function assertTypeOf(options, attr, type, toThrow = true) {
-	let isTypeCorrect;
-	if (options[attr] === undefined) {
+export function assertValueTypeOf(value, type) {
```
jsomsanith: This function does not throw. So any call of it is useless.
Please move the error here, instead of `assertTypeOf`

packages/cmf/src/cmfModule.merge.js
```diff
@@ -0,0 +1,122 @@
...
+		return [].concat(preReducer).concat(newPreReducer);
+	}
+	if (newPreReducer) {
+		return newPreReducer;
+	}
+	return preReducer;
+}
+
+const MERGE_FNS = {
+	id: () => undefined,
```
jsomsanith: what is options.id ?

packages/containers/sandbox/src/app/index.js
```diff
@@ -5,16 +5,19 @@
  */
 
 import '@talend/bootstrap-theme/src/theme/theme.scss';
-import { registerAllContainers } from '@talend/react-containers/lib/register';
+import comtainersModule from '@talend/react-containers';
```
jsomsanith: `comtainersModule` --> `containersModule`

packages/cmf/src/cmfModule.merge.js
jmfrancois: I prefer throw an exception if you have set a config key to undefined.

packages/cmf/src/cmfModule.merge.js
jmfrancois: its to identify module.

PR #1602, opened 2018-08-09
fix(components): highlight the list when the user hover the main title

jare-talend:
update list with the guidines to hightlight the title when the user can interact with it.
https://company-57688.frontify.com/document/92132#/navigation-layout/list

add the filter talend-grayscale in UI to apply to the icon. This filter transform the image to balck and white. Then, we apply a alpha 0.9.

The constrast is bad and don't respect AA . The UX know this issue. No solution for the moment to replace it by another color.

filter doesn't work on IE/Edge. We chosen a degrades solution with the PM/UX.

fix that


packages/components/src/VirtualizedList/CellTitle/CellTitle.component.js
```diff
@@ -42,7 +42,9 @@ function CellTitle({ cellData, columnData, getComponent, rowData, rowIndex, type
 			<CellTitleSelector
 				id={titleId}
 				cellData={cellData}
-				className={theme['main-title']}
+				className={classNames(theme['main-title'], {
```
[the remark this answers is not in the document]
jare-talend: the color in the guidelines is the color of the icon. We haven't nothing to do for that.
By default: filter black and when the user hovers the icon, if the icon has a color we keeps this color.
I need to check if one filter gray is applied by default.

Ok, i will add a plain text class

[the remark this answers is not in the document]
jare-talend: we have issue with IE/Edge to apply the color on the icon. we waits a arbitrary on this with PM/UX

packages/components/src/IconsProvider/IconsProvider.component.js
```diff
@@ -14,14 +19,14 @@ import talendIcons from '@talend/icons/dist/react';
...
-	const style = { display: 'none' };
 	return (
-		<svg xmlns="http://www.w3.org/2000/svg" focusable="false" style={style}>
+		<svg xmlns="http://www.w3.org/2000/svg" focusable="false" className={theme.hide}>
 			{ids.map((id, index) => (
 				<symbol key={index} id={id}>
 					{iconset[id]}
 				</symbol>
 			))}
+			{Object.keys(filters).map((id, index) => addKeyValue(filters[id], index))}
```
jsomsanith: Why don't you use the same thing as icons ?
```
{Object.keys(filters).map((id, index) => (
				<symbol key={`filter-${index}`} id={id}>
					{filters[id]}
				</symbol>
			))}
```

packages/components/src/IconsProvider/IconsProvider.scss
```diff
@@ -1 +1,5 @@
-// IconsProvider styles
+.hide {
+	position: absolute;
+	top: -9999px;
+	left: -9999px;
```
jsomsanith: You don't need that, you have `sr-only` class in bootstrap

packages/components/src/VirtualizedList/CellTitle/CellTitle.component.js
```diff
@@ -49,12 +49,24 @@ class CellTitle extends React.Component {
...
 		return (
 			<div id={titleId} className={classNames('tc-list-title', theme['tc-list-title'])}>
-				{iconKey && rowData[iconKey] && <Icon name={rowData[iconKey]} className={theme.icon} />}
+				{iconKey &&
+					rowData[iconKey] && (
+						<Icon
+							name={rowData[iconKey]}
+							className={classNames(theme.icon, {
+								[theme['tc-main-title-icon-active']]: onClick,
+								'tc-main-title-icon-active': onClick,
```
jsomsanith: You don't need to add specific conditional classnames on each element. Just put one on the main div, and in the scss, you can target existing classnames on each part
```
.my-class {
	...

	&:hover,
	&:focus {
		.icon: {}
		.main-title {}
	}
}
```

Let's not name it `active` but something with filter-on-hover for example, something more meaningful. Active means something else in css.

packages/components/src/IconsProvider/IconsProvider.component.js
jare-talend: filter doesn't use symbol. We have just to add the filter in the svg container.

packages/components/src/VirtualizedList/CellTitle/CellTitle.component.js
jare-talend: OK i change the name but i need to add some condition. this css is only active when onClick is defined.

output/components.sasslint.txt
```diff
@@ -7,5 +7,8 @@ src/Drawer/Drawer.scss
 src/FilterBar/FilterBar.scss
   47:6  warning  Vendor prefixes should not be used  no-vendor-prefixes
 
-✖ 2 problems (0 errors, 2 warnings)
+src/VirtualizedList/CellTitle/CellTitle.scss
+  12:15  error  Quotes around URLs are required  url-quotes
```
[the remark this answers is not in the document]
jare-talend: i waiting travis to resolve that

packages/components/src/VirtualizedList/CellTitle/CellTitle.component.js
jare-talend: In you example, the hover, focus is always applied when onClick exists or not.

packages/components/src/IconsProvider/IconsProvider.component.js
jsomsanith: Right, with symbol it doesn't work on FF.

Without `symbol` : 
```
{Object.keys(filters).map((id, index) => (
	<svg key={`svg-filter-${index}`} id={id}>
		{filters[id]}
	</svg>
))}
```

This works on Chrome, FF, Safari, and it avoid the `addKeyValue` function that is quite hard to understand

packages/components/src/VirtualizedList/CellTitle/CellTitle.component.js
jsomsanith: No because you scope it with `my-class`. This className is on the component if you have the onClick

PR #1603, opened 2018-08-09
fix(forms/rjsf): add support for modal and drawers

jmfrancois:

Old form docs is still the first documentation in surge.
Old form are broken in drawer.
Apps use custom css to make it look nice (but they don t follow the guideline).

Move `Form` doc in surge to the last one. Make core stories first. 
Make old form follow the guideline in modal and drawer

### DIFF
--- a/packages/forms/stories-core/layouts.js
+++ b/packages/forms/stories-core/layouts.js
@@ -1,9 +1,30 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { HeaderBar, Layout, Dialog, Drawer } from '@talend/react-components';
+import { action } from '@storybook/addon-actions';
+
 import Form from '../src/index';
 
 const simple = require('./json/concepts/core-simple.json');
+const old = require('../stories/json/simple.json');
+
+const actions = [
+	{
+		name: 'cancel',
+		style: 'link',
+		onClick: action('CANCEL'),
+		type: 'button',
+		label: 'CANCEL',
+	},
+	{
+		style: 'primary',
+		type: 'submit',
+		label: 'SUBMIT',
+		disabled: true,
+		'data-feature': 'form.feature',
+		onClick: action('SUBMIT'),
+	},
+];
 
 function LayoutDrawer({ title, stacked, ...props }) {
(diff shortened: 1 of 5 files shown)


packages/forms/stories-core/layouts.js
```diff
@@ -1,9 +1,30 @@
...
+
 import Form from '../src/index';
 
 const simple = require('./json/concepts/core-simple.json');
+const old = require('../stories/json/simple.json');
+
+const actions = [
+	{
+		name: 'cancel',
+		style: 'link',
```
annilsson: I think this should be `inverse` & `default`? (it affects alignment if we use link style)

packages/forms/stories-core/layouts.js
```diff
@@ -61,4 +82,26 @@ export default [
...
+		story: () => (
+			<div>
+				<h1>Form (old) by default take 100% with of the container</h1>
+				<Form data={old} />
+			</div>
+		),
+	},
+	{ name: 'drawer-old', story: () => <LayoutDrawer title="Form (old) in a drawer" data={old} /> },
+	{
+		name: 'drawer-stacked-old',
```
annilsson: Cancel is not aligned with the form fields on the left side

packages/forms/stories-core/layouts.js
```diff
@@ -1,9 +1,30 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { HeaderBar, Layout, Dialog, Drawer } from '@talend/react-components';
+import { action } from '@storybook/addon-actions';
+
 import Form from '../src/index';
 
 const simple = require('./json/concepts/core-simple.json');
+const old = require('../stories/json/simple.json');
```
annilsson: I can't figure out why (maybe coincidence), but the stories that use `old` data have the buttons left-aligned. Should they not be right-aligned?

jmfrancois: if you take the OLD stories you can see the submit btn is always left aligned ...

jmfrancois:
rage quit

PR #1605, opened 2018-08-09
feat(containers/ComponentForm): add csrf token

jmfrancois:

`ComponentForm` doesn t use saga http for trigger calls.
It uses fetch so that means trigger calls do not have csrf header.

re-use the handleCSRF from cmf where its needed, and open configuration throw ComponentForm props.
So it merge needed headers with csrf token before each fetch calls.


packages/containers/src/ComponentForm/kit/createTriggers.js
```diff
@@ -30,6 +30,7 @@ const DEFAULT_HEADERS = {
 function noOpTrigger({ error, trigger }) {
 	// eslint-disable-next-line no-console
 	console.error(`${JSON.stringify(trigger)} failed with error ${error || '-'}`);
+	return {};
```
jmfrancois: this fix a bug if a trigger is called but do not exist, we should return empty data

packages/containers/src/ComponentForm/ComponentForm.component.js
```diff
@@ -165,6 +165,10 @@ export class TCompForm extends React.Component {
 			customRegistry: props.customTriggers,
 			headers: config.headers,
 			lang: props.lang,
+			security: {
+				CSRFTokenCookieKey: this.props.CSRFTokenCookieKey,
```
annilsson: Is there a reason to use `this.props` instead of `props` only on these 2? They should be the same, but it looks like they are not meant to be.

output/containers.eslint.txt
```diff
@@ -2,9 +2,6 @@
 
 The react/require-extension rule is deprecated. Please use the import/extensions rule from eslint-plugin-import instead.
 
-/home/travis/build/Talend/ui/packages/containers/src/ComponentForm/kit/createTriggers.js
-  18:8  error  'get' is defined but never used  no-unused-vars
```
ncomont: could you fix that please ?

PR #1608, opened 2018-08-10
feat(components/DatePicker): Add dropdown

jdescottes-talend:
The picker is actually displayed all the time, it needs to be displayed only when interacting with the input or in the dropdown.

[Date picker guidelines](https://company-57688.frontify.com/document/92132#/ui-controls/date-picker-1513013325)

Implement it in the following way :
- The dropdown appears on input focus (clicking trigger the focus for the input case)
- The dropdown disappears on click or focus outside the whole component (input + picker)


output/components.sasslint.txt
```diff
@@ -1,11 +1,14 @@
 > sass-lint -v -q
 
 
+src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.scss
+  1:2  error  Class '.dropdownWrapper' should be written in lowercase with hyphens  class-name-format
```
jsomsanith: Linter error

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -187,6 +199,52 @@ class InputDateTimePicker extends React.Component {
...
+		this.containerRef = ref;
+	}
+
+	setDropdownWrapperRef(ref) {
+		this.dropdownWrapperRef = ref;
+	}
+
+	mountDocumentHandler() {
+		document.addEventListener('click', this.documentHandler);
+		document.addEventListener('focus', this.documentHandler);
```
jsomsanith: I think the focus event on document is useless

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -187,6 +199,52 @@ class InputDateTimePicker extends React.Component {
...
+			this.switchDropdownVisibility(false);
+		}
+	}
+
+	switchDropdownVisibility(isShown) {
+		if (this.state.isDropdownShown === isShown) {
+			return;
+		}
+
+		if (isShown) {
```
jsomsanith: You should set this as `setState` callback

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.scss
```diff
@@ -1,3 +1,22 @@
-.dropdown {
+.dropdownWrapper {
+	position: relative;
+}
+
+.popover {
 	width: 320px; // TODO: Remove when width sizing management has been covered
+	max-width: none;
+	padding: 0;
```
jsomsanith: Those (max-width and padding) don't seem to be used.

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.test.js
```diff
@@ -691,4 +691,117 @@ describe('InputDateTimePicker', () => {
...
+			const overlayWrapper = wrapper.find('Overlay');
+			expect(overlayWrapper.prop('show')).toBe(false);
+		});
+
+		it('should open the dropdown on input focus', () => {
+			const wrapper = shallow(<InputDateTimePicker />);
+
+			const inputWrapper = wrapper.find('DebounceInput');
+
+			inputWrapper.prop('onFocus')();
```
jsomsanith: You mount it, shouldn't you simulate the focus on input ? (I mean instead of getting a props and calling it)

output/components.sasslint.txt
jdescottes-talend: Can the linters (eslint and sasslint) checks be in requirements for the PR to be merged ?
It would avoid reviewers to check that, and prevent omissions :)

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
jdescottes-talend: If we change focus with keyboard by navigating outside.
Or if we use a special keyboard combination for an other part of the page and change focus programmatically.

But whatever the use case, just by semantic : if the focus is done outside, it means we are no more focused on the picker :p

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.scss
jdescottes-talend: Those 2 properties has been globally overridden in a component style :D

/packages/components/src/RichTooltip/RichTooltip.scss
```javascript
:global(.popover) {
	padding: 0;
	border: $tc-popover-border-width solid $tc-popover-border-color;
	border-radius: $border-radius-base;
	background-color: $white;
	box-shadow: $tc-popover-box-shadow;
	max-width: $tc-popover-max-width;
}
```

So I think it's useful if the other component don't override globally the theme. I have discussed with @jare-talend, We should discuss it now with the UX/designers to eventually find a way to have a common style base for the dropdown/popover. But it's far away my priorities.

output/components.sasslint.txt
jdescottes-talend: done

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
jdescottes-talend: done

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.test.js
jdescottes-talend: Yes I can and I did it !! Even if under the hood, it's the same. Now it's more semantic +1. :p

I realized I have forgotten some cases with focus outside, so I added them too.

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
jdescottes-talend: Good catch

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -187,6 +210,46 @@ class InputDateTimePicker extends React.Component {
...
+		document.addEventListener('focus', this.documentHandler);
+	}
+
+	unmountDocumentHandler() {
+		document.removeEventListener('click', this.documentHandler);
+		document.removeEventListener('focus', this.documentHandler);
+	}
+
+	documentHandler(e) {
+		const isActionOutOfComponent = e.path.every(node => node !== this.containerRef);
```
jare-talend: path is compatible for all browsers ?

jdescottes-talend: Effectively it's not !
After some reading, for info :
`event.path` was introduced by chrome and then an `event.composedPath()` has been specified and implemented by the latest Chrome and FF versions but not yet for IE and Edge.

It seems there is no working polyfill, surely because it seems not doable : the path has to be calculated at the time the event is fired, because when it's handled by handlers, the dom structure can have changed and some elements could have been removed/added.
I tried a naive `event.composedPath()` polyfill based on [this repo](https://github.com/DieterHolvoet/event-propagation-path) without success.

Finally I had to do a workaround with checking if the event bubble up in the current component by hand with a second set of handlers.

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.scss
jsomsanith: Ok but that doesn't change the fact those lines are not applied so you can remove them.

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
jdescottes-talend: Normally, all DOM events should go down in capture phase then up in bubble phase.
document --> target --> document
_If there propagation has not been stopped by a ninja._

I'm checking why the focus is not handled in document.

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.scss
jdescottes-talend: > Ok but that doesn't change the fact those lines are not applied

Yes ! (In the current time for the current build)

> so you can remove them.

I can bu I don't want ! :p


This is no about _what actually is_ but about _what should be done_. I want my dropdown like that and no common style is done for that so I do it.
I will not rely on an other component for setting the style I want for multiples reasons :
- When it'll be patched, the current component will broke
- I'm relying on an other component without any explicit mention, in any of the two components. So if the other component decide to tweak a little some padding or whatever (without patching the global style), the current component will be impacted without knowing it.
- If an App decide to not include/load the other component, the style will be different than expected in the storybook.

Those are style properties **my component needs to have** unless we decide to have a common dropdown/popover style in theme for it.
I'll definitely not rely on an other component with an implicit totally invisible pernicious dependency.

PR #1610, opened 2018-08-10
fix(components): highlight the main title on the list 

jare-talend:
update list with the guidines to hightlight the title when the user can interact with it.
https://company-57688.frontify.com/document/92132#/navigation-layout/list

no on the icon. we have issue https://github.com/Talend/ui/pull/1602

### DIFF
--- a/packages/components/src/VirtualizedList/CellTitle/CellTitle.test.js
+++ b/packages/components/src/VirtualizedList/CellTitle/CellTitle.test.js
@@ -38,6 +38,36 @@ describe('CellTitle', () => {
 		expect(wrapper.getElement()).toMatchSnapshot();
 	});
 
+	it('should render without active class if no onClick on the title', () => {
+		// given
+		const columnData = {
+			id: 'my-title',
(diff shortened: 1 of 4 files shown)


packages/components/src/VirtualizedList/CellTitle/CellTitle.test.js
```diff
@@ -38,6 +38,36 @@ describe('CellTitle', () => {
...
+				cellData={'my awesome title'}
+				columnData={columnData}
+				getComponent={jest.fn()}
+				rowData={rowData}
+				rowIndex={1}
+			/>,
+		);
+
+		// then
+		expect(wrapper.getElement()).toMatchSnapshot();
```
jdescottes-talend: Can you check with a selector and a real expectation ?
Because the snapshots are lot's of text hard to read (hard to say what we looking for) when this is not the shape which is tested. And without a snapshot, when there is an other change in the component, we don't have to check all the changed snapshots, so less risk of regressions.

Even I think there should be 2 tests : with and without onClick/active.

PR #1613, opened 2018-08-13
feat(Notification): improve a11y

jsomsanith:
On Notification component 
- no alternative text on close button
- timer cancelable only with mouse event
- errors has not the right live attribute

- add a text to the close button
- make the notification focusable and cancel the timer on focus
- errors have role `alert`


output/components.eslint.txt
```diff
@@ -17,6 +17,13 @@ The react/require-extension rule is deprecated. Please use the import/extensions
 /home/travis/build/Talend/ui/packages/components/src/HeaderBar/HeaderBar.component.js
   105:37  error  Elements with ARIA roles must use a valid, non-abstract ARIA role  jsx-a11y/aria-role
 
+/home/travis/build/Talend/ui/packages/components/src/Notification/Notification.component.js
+   9:27  error  Absolute imports should come before relative imports                                 import/imports-first
```
sgangnant: Most of those errors seem to be easily fixed, can you have a look at it ?

packages/components/src/Notification/Notification.component.js
```diff
@@ -18,9 +20,11 @@ export function CloseButton({ notification, leaveFn }) {
 				theme['tc-notification-close'],
 				'.tc-notification-close',
 			)}
+			aria-label={t('NOTIFICATION_CLOSE', { defaultValue: 'Close notification' })}
```
sgangnant: Are we sure `t` will always be defined, and that this will not end with `undefined is not a function` ? Especially since `t` is a new prop

jsomsanith: I connect it to the translation in the lines just after `export const CloseButton = translate(I18N_DOMAIN_COMPONENTS)(CloseButtonComponent);`

output/components.eslint.txt
jsomsanith: Thanks, fixed

packages/components/src/Notification/Notification.component.js
annilsson: Can we not use defaultProps and `getDefaultT` like many other functions? If we test the component without the HOC (with `WrappedComponent`) it would be an example of t as undefined.

packages/components/src/Notification/Notification.component.js
```diff
@@ -2,11 +2,13 @@ import PropTypes from 'prop-types';
...
 import { TransitionGroup, CSSTransition } from 'react-transition-group';
 import classNames from 'classnames';
+import { translate } from 'react-i18next';
 
 import { Action } from '../Actions';
 import theme from './Notification.scss';
+import I18N_DOMAIN_COMPONENTS from '../constants';
 
-export function CloseButton({ notification, leaveFn }) {
+function CloseButtonComponent({ notification, leaveFn, t }) {
```
annilsson: eslint complaining about missing propTypes, I think we should add them.

packages/components/src/Notification/Notification.component.js
```diff
@@ -67,27 +71,29 @@ export function TimerBar({ type, autoLeaveError }) {
...
-	};
-	const classes = classNames(notificationClasses);
+		[theme['tc-notification-error']]: isError,
+		'tc-notification-error': isError,
+	});
+	const enterAction = event => props.onMouseEnter(event, notification);
+	const leaveAction = event => props.onMouseOut(event, notification);
+
 	return (
-		<div // eslint-disable-line jsx-a11y/no-static-element-interactions
```
annilsson: I think we actually need this eslint-disable?

packages/components/src/Notification/Notification.component.js
jsomsanith: Yeah you're right, I have an eslint issue on my IDE :/
I add them

packages/components/src/Notification/Notification.component.js
jsomsanith: Same here, I revert

packages/components/src/Notification/Notification.component.js
jsomsanith: Ok, added the default props

PR #1614, opened 2018-08-13
feat(Enumeration): add the id on header actions

smouillour:
to facilitate the job for the QA. we need to add an id on the actions in the enumeration's header component

add the id in the related header component


packages/components/src/Enumeration/Header/Header.component.js
```diff
@@ -34,6 +34,7 @@ function getAction(action, index) {
 	}
 	return (
 		<Action
+			id={action.id}
```
jsomsanith: If I look at the propTypes, I see that `headerDefault` is an array of `<Action>` props. Why not spreading those props in `<Action>` and then add our specific ones ?
Furthermore that's what is done when the action is a dropdown, and that will allow to pass anything we want (I'm thinking about the aria attributes for example)
```
<Action
	{...action}
	key={`${index}-enum-header-action`}
	onClick={onClick}
	tooltipPlacement="bottom" // there was a typo here (btooltipPlacement), let's fix it
	hideLabel
	link
/>
```

packages/components/src/Enumeration/Header/Header.component.js
```diff
@@ -34,6 +34,7 @@ function getAction(action, index) {
 	}
 	return (
 		<Action
+			{...action}
 			key={`${index}-enum-header-action`}
 			label={action.label}
 			icon={action.icon}
```
jsomsanith: You can now remove all the props that you pass through as they are already spread

PR #1615, opened 2018-08-13
feat(UIForm): NestedListView widget

sgangnant:
ListView can only be used as a form field with a "flat" data structure. It is not possible to use it with a nested datastructure.

Implement a new field component that works with 2 level of depth in the data structure, like :
```
{
    vegetables: ["carrot", "lettuce"],
    meat: ["beaf", "pork", chicken"]
}
```
You can test the component under the forms storybook > "Migration" > NestedListView.

screenshot

I choosed to write a new component because even if it looks similar to ListView on some points, I don't think it would ease maintenance to have everything in a single component.


packages/forms/src/UIForm/fields/NestedListView/NestedListView.component.js
```diff
@@ -0,0 +1,254 @@
...
+				onClick: this.switchToDefaultMode.bind(this),
+			},
+		];
+
+		const toggledChildren = [];
+
+		this.state = {
+			...initItems(schema, value, null, toggledChildren, this.getInitItemsCallbacks()),
+			toggledChildren,
+			value,
```
jdescottes-talend: `value` don't seem to be used in render so should not be in state.

packages/forms/src/UIForm/fields/NestedListView/NestedListView.component.js
```diff
@@ -0,0 +1,254 @@
...
+				label: t('LISTVIEW_WIDGET_ABORT', { defaultValue: 'Abort' }),
+				onClick: this.switchToDefaultMode.bind(this),
+			},
+		];
+
+		const toggledChildren = [];
+
+		this.state = {
+			...initItems(schema, value, null, toggledChildren, this.getInitItemsCallbacks()),
+			toggledChildren,
```
jdescottes-talend: same here

packages/forms/src/UIForm/fields/NestedListView/NestedListView.component.js
sgangnant: It is used downwards in the .utils file's method

packages/forms/src/UIForm/fields/NestedListView/NestedListView.utils.js
```diff
@@ -0,0 +1,87 @@
...
+		}
+
+		return finalItems;
+	}, []);
+
+	return {
+		displayedItems,
+		items,
+		searchCriteria,
+		toggledChildren,
```
jdescottes-talend: the last 3 object properties are not calculated in the function. They shouldn't be returned. In fact this function seems to only calculate the `displayedItems` so should return only this value and have a more meaningful name according to that unique calculated value.

packages/forms/src/UIForm/fields/NestedListView/NestedListView.utils.js
```diff
@@ -0,0 +1,87 @@
+/**
+ * Build the ListView component props to match current state's values
+ * @param {Array} items
+ * @param {Object} value
+ * @param {String} searchCriteria
+ * @param {Array} toggledChildren
+ * @returns {Object}
+ */
+export function getItemsProps(items, value, searchCriteria, toggledChildren) {
+	const displayedItems = items.reduce((finalItems, item) => {
```
jdescottes-talend: Reducing an array to an array seems the operations could have been done with some filters/maps chained.
It will isolate more clearly what's the filtering conditions and what's the calculated values returned and also remove all the manual array declaration and pushing call.

Example :
```javascript
return items
    .map() // Transform to an object with `item` and an other prop `children` based on the `searchCriteria`
    .filter() // Filter on children length
    .map() // Only the final item
```

packages/forms/src/UIForm/fields/NestedListView/NestedListView.utils.js
jdescottes-talend: ps: The other values can be given to the state or the instance attributes directly without hiding the changes.

packages/forms/src/UIForm/fields/NestedListView/NestedListView.component.js
sgangnant: Same as below
`value` is used to properly render elements (checked states, etc)

packages/forms/src/UIForm/fields/NestedListView/NestedListView.component.js
```diff
@@ -0,0 +1,254 @@
...
+		this.defaultSearchHeaderActions = [
+			{
+				id: 'abort',
+				icon: 'talend-cross',
+				label: t('LISTVIEW_WIDGET_ABORT', { defaultValue: 'Abort' }),
+				onClick: this.switchToDefaultMode.bind(this),
+			},
+		];
+
+		const toggledChildren = [];
```
jdescottes-talend: I think it could store directly the item ref and would avoid the need to create a `toggleId`.
And I don't understand the variable `children` name suffix (children of who ?)

packages/forms/src/UIForm/fields/NestedListView/NestedListView.component.js
```diff
@@ -0,0 +1,254 @@
...
+		}, 400);
+	}
+
+	/**
+	 * Search input ENTER/ESC management
+	 * @param { Object } event The keydown event
+	 */
+	onInputKeyDown(event) {
+		if (event.keyCode === keycode('enter')) {
+			event.stopPropagation();
```
jdescottes-talend: Why the propagation need to be stopped ?

sgangnant: Didn't test without it, I kept it as is from what has been done in `ListView`

packages/forms/src/UIForm/fields/NestedListView/NestedListView.component.js
```diff
@@ -0,0 +1,254 @@
...
+		});
+	}
+
+	/**
+	 * Handle checked item (child) in the ListView change
+	 * @param { Object } event The event that triggered the change
+	 * @param { Object } checked
+	 * @param { Object } parent
+	 */
+	onCheck(event, checked, parent) {
```
jdescottes-talend: `checked` could be `checkedItem` (if I understand) otherwise, we expect just a boolean

packages/forms/src/UIForm/fields/NestedListView/NestedListView.component.js
```diff
@@ -0,0 +1,254 @@
...
+	onChange(event, newValue) {
+		const payload = { schema: this.props.schema, value: newValue };
+		this.props.onChange(event, payload);
+		this.props.onFinish(event, payload);
+	}
+
+	/**
+	 * Search input change
+	 * @param { Object } event The change event
+	 * @param { Array } value
```
jsomsanith: Fix jsDoc

packages/forms/src/UIForm/fields/NestedListView/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+| Property | Description |
+|---|---|
+| type | `object`
+| properties | `object` Every property of the value object (the key used in `properties` being the value's key)
+| properties[key] | `object`
+| properties[key].type | `array`
+| properties[key].items | `object`
+| properties[key].items.enum | `array` The list of available values for this key of the value
+| properties[key].items.enumNames | `array` The list of labels display for each value
```
jsomsanith: This is not supported in new UIForms, use titleMap in uiSchema

packages/forms/src/UIForm/fields/NestedListView/README.md
```diff
@@ -0,0 +1,72 @@
+# NestedListView
+
+This widget allows you to render a nested ListView to select multiple items with a 2 levels hierarchy.
+More than 2 levels is currently not supported.
+
+![NestedListview](screenshot.gif)
+
+| Property | Description |
```
jsomsanith: Add a `type` column please

packages/forms/src/UIForm/fields/NestedListView/README.md
```diff
@@ -0,0 +1,72 @@
...
+|---|---|---|
+| key |  |  |
+| title | The title to display in the ListView header |  |
+| description | A description to display below the ListView |  |
+| emptyLabel | The label to display when there is no item | `This list is empty.` |
+| noResultLabel | The label to display when filter result is empty | `No result found.` |
+| placeholder | The search input placeholder |  |
+| widget | The widget to use | `nestedListView` |
+| onChange | Value change callback  |
+| onFinish | Value change finish callback (triggered after `onChange`)  |
```
jsomsanith: onChange and onFinish are internals. And the content of uiSchema must be serializable. You can remove those.

packages/forms/src/UIForm/fields/NestedListView/README.md
```diff
@@ -0,0 +1,72 @@
+# NestedListView
+
+This widget allows you to render a nested ListView to select multiple items with a 2 levels hierarchy.
+More than 2 levels is currently not supported.
+
+![NestedListview](screenshot.gif)
```
jsomsanith: Please move that at the bottom to fit with the rest of the documentation.

packages/forms/src/UIForm/fields/NestedListView/NestedListView.component.js
jdescottes-talend: But it's not used for the render so it's not a react state, it's just a class attribute 'internal state'. It will avoid checking for re-render.

sgangnant: It is used to calculate the props of the rendered elements (displayedItems)

packages/forms/src/UIForm/fields/NestedListView/NestedListView.utils.js
```diff
@@ -0,0 +1,71 @@
+/**
+ * Build the ListView component props to match current state's values
```
jdescottes-talend: `component props to match current state's values` => `component props accordingly to state's values` ?

packages/forms/src/UIForm/fields/NestedListView/NestedListView.utils.js
```diff
@@ -0,0 +1,71 @@
...
+	let textFilter;
+	if (searchCriteria) {
+		textFilter = ({ label }) => label.toLowerCase().includes(searchCriteria.toLowerCase());
+	}
+
+	return (
+		items
+			.map(item => {
+				// Add boolean "checked" to checked children
+				const checkedChildren = item.children.map(child => ({
```
jdescottes-talend: This var name isn't what's really inside. There could be unchecked children

sgangnant: Hence the comment. Plus, 100 char per line limit is easily reached

packages/forms/src/UIForm/fields/NestedListView/NestedListView.utils.js
sgangnant: Looks already clear to me

packages/forms/src/UIForm/fields/NestedListView/NestedListView.utils.js
```diff
@@ -0,0 +1,71 @@
...
+export function getDisplayedItems(items, value, searchCriteria) {
+	let textFilter;
+	if (searchCriteria) {
+		textFilter = ({ label }) => label.toLowerCase().includes(searchCriteria.toLowerCase());
+	}
+
+	return (
+		items
+			.map(item => {
+				// Add boolean "checked" to checked children
```
jdescottes-talend: Please avoid all code commentaries they are :
- for 50%, useless (when duplicate the code)
- for 25%, pernicious (when one or the other change without the two)
- for 24%, just an indicator that the code is too complex or implicit or is  and need to be more simple and/or explicit.
- for 1%, an indication the code shouldn't be written like this, so the code mustn't exist like this and the comment not too :p

Here you even don't exactly do what the comment say, you don't add the value to the child, you create a new Array of wrappers with the child and an other value.

For adding semantic to code, breaking in named functions and variables are a good way. It adds an other benefit : the possibility to read the code and understand the _what_ is happening without the need to check exactly _how_

jdescottes-talend: So here it's a bit the 2)

packages/forms/src/UIForm/fields/NestedListView/NestedListView.utils.js
```diff
@@ -0,0 +1,71 @@
...
+				const checkedChildren = item.children.map(child => ({
+					...child,
+					checked: (value[item.key] || []).includes(child.value),
+				}));
+
+				return {
+					...item,
+					// Section checked or not (at least 1 item checked)
```
jdescottes-talend: Here totally the 1)

packages/forms/src/UIForm/fields/NestedListView/NestedListView.utils.js
```diff
@@ -0,0 +1,71 @@
...
+					checked: checkedChildren.some(child => child.checked),
+					// Filtered and "checked" or not children
```
jdescottes-talend: Here I just don't understand the comment :s

packages/forms/src/UIForm/fields/NestedListView/NestedListView.utils.js
```diff
@@ -0,0 +1,71 @@
...
+					children: textFilter ? checkedChildren.filter(textFilter) : checkedChildren,
+				};
+			})
+			// Remove empty sections
+			.filter(item => item.children.length > 0)
```
jdescottes-talend: Here the anonymous function can be extract to a named one
```javascript
const containsChildren = item => item.children.length > 0;

---- 

.filter(containsChildren)

```

packages/forms/src/UIForm/fields/NestedListView/NestedListView.utils.js
sgangnant: Source of your statistics ?
I don't think breaking down this code chunk into severals methods would be any better for understanding than commenting each part, that's completely subjective.
Those comments are giving an overview of what's done in the method without entering the details on how javascript would make his transformations.

packages/forms/src/UIForm/fields/NestedListView/NestedListView.utils.js
```diff
@@ -0,0 +1,71 @@
...
+ * @param { Function } callbacks.onExpandToggle
+ * @param { Function } callbacks.onParentChange
+ * @param { Function } callbacks.onCheck
+ * @returns { Array }
+ */
+export function initItems(schema, value, searchCriteria, callbacks) {
+	const { onExpandToggle, onParentChange, onCheck } = callbacks;
+
+	const items = schema.items.map(item => {
+		const key = item.key[item.key.length - 1]; // This is ugly (get a parent "id")
```
jdescottes-talend: Is the comment still necessary ?
I don't understand if the comment tell that the code 'get the parent "id"' or that the code should get an access to a parent id not implemented.

packages/forms/src/UIForm/fields/NestedListView/NestedListView.component.js
jdescottes-talend: @jsomsanith : Correct me if i'm wrong, but I think the react state must be used only for values used for the renders, not for the one used to calculate the values rendered.

It make things more simpler, example : in the current way, a lot's of internal base or pre-computed data are sent to the rendered components. => With default spreading props to the children in the majority of components, some values can easily get through the whole component tree for nothing unless adding potential side effects.
Even worse, some values like `items` are overridden between the state spreading and the custom defined props after ; which makes the code less declarative and more imperative.

packages/forms/src/UIForm/fields/NestedListView/NestedListView.utils.js
```diff
@@ -0,0 +1,71 @@
...
+				value: option.value,
+				onChange: onCheck,
+			})),
+		};
+	});
+
+	return {
+		items,
+		displayedItems: getDisplayedItems(items, value, searchCriteria),
+	};
```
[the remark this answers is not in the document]
sgangnant: Nope, `items` is the list of items available in the widget, `displayedItems` can be text filtered so we need `items` as reference for the list of all available all items. It is stored since there are some computations from `schema` prop that are done.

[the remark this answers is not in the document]
jdescottes-talend: ps: This will remove the need of `value` and `searchCriteria` parameters. Simple function that does well one thing :D

packages/forms/src/UIForm/fields/NestedListView/NestedListView.component.js
```diff
@@ -0,0 +1,258 @@
...
+		const searchCriteria = null;
+
+		const callbacks = {
+			onExpandToggle: this.onExpandToggle.bind(this),
+			onParentChange: this.onParentChange.bind(this),
+			onCheck: this.onCheck.bind(this),
+		};
+
+		this.state = {
+			...initItems(schema, value, searchCriteria, callbacks),
```
jdescottes-talend: Here the items initialized should be stored in the instance, and then the `displayedItems` are calculated based on them for the state.

```javascript
this.items = initItems(schema, callbacks);

this.state = {
   displayedItems: getDisplayedItems(this.items, value, searchCriteria),
   ...
}
```

packages/forms/src/UIForm/fields/NestedListView/NestedListView.utils.js
jdescottes-talend: After re-reading, I just said something totally wrong about the first map, the comment was right with the code. my bad :/
(So it's no more the 2 because I'm a derp, but more the 1 :P, whatever...)

packages/forms/src/UIForm/fields/NestedListView/NestedListView.component.js
```diff
@@ -0,0 +1,259 @@
...
+	onExpandToggle(event, item) {
+		this.setState(({ searchCriteria }) => {
+			this.items = this.items.map(fieldItem => {
+				if (fieldItem.key === item.key) {
+					return { ...fieldItem, expanded: !fieldItem.expanded };
+				}
+				return fieldItem;
+			});
+
+			return {
```
jdescottes-talend: Only the `return` can be kept inside the `setState`, all the above can be extracted before it.

packages/forms/src/UIForm/fields/NestedListView/NestedListView.component.js
```diff
@@ -0,0 +1,259 @@
...
+			({ searchCriteria }) => {
+				const { enum: availableOptions } = this.props.schema.schema.properties[item.key].items;
+
+				// Toggle all values
+				this.value = {
+					...this.value,
+					[item.key]: (this.value[item.key] || []).length === 0 ? availableOptions : [],
+				};
+
+				return {
```
jdescottes-talend: Same here for extracting the above code before the `setState`

packages/forms/src/UIForm/fields/NestedListView/NestedListView.component.js
```diff
@@ -0,0 +1,259 @@
...
+
+				if (!(key in this.value)) {
+					this.value[key] = [];
+				}
+
+				this.value[key] = this.value[key].includes(item.value)
+					? this.value[key].filter(storedValue => storedValue !== item.value) // Unselect
+					: this.value[key].concat(item.value); // Select
+
+				return {
```
jdescottes-talend: Same here for extracting the above code before the `setState`

packages/forms/src/UIForm/fields/NestedListView/NestedListView.component.js
```diff
@@ -0,0 +1,259 @@
...
+	/**
+	 * Search input change
+	 * @param { Object } event
+	 * @param { Object } item
+	 */
+	onInputChange(event, item) {
+		clearTimeout(this.timerSearch);
+		this.timerSearch = setTimeout(() => {
+			const { value: searchCriteria } = item;
+			this.setState(() => ({
```
jdescottes-talend: This don't need to be a callback, it should be directly an object argument

packages/forms/src/UIForm/fields/NestedListView/NestedListView.component.js
```diff
@@ -0,0 +1,259 @@
...
+	}
+
+	/**
+	 * Switch header to default mode.
+	 * Reset display to no filter
+	 */
+	switchToDefaultMode() {
+		const searchCriteria = null;
+
+		this.setState(() => ({
```
jdescottes-talend: Same here for callback to simple object

jdescottes-talend:
@jsomsanith : Is the value provided by the form need to be handled for dynamic changes (programmatically update the form value) ? Because here, it's stored/cached in the instance. It will need an update on prop change right ?

packages/forms/src/UIForm/fields/NestedListView/NestedListView.component.js
```diff
@@ -0,0 +1,253 @@
...
+				icon: 'talend-cross',
+				label: t('LISTVIEW_WIDGET_ABORT', { defaultValue: 'Abort' }),
+				onClick: this.switchToDefaultMode.bind(this),
+			},
+		];
+
+		this.onInputKeyDown = this.onInputKeyDown.bind(this);
+		this.onInputChange = this.onInputChange.bind(this);
+
+		const searchCriteria = null;
```
jsomsanith: Remove that, it's null.

Then 
```
this.state = {
			searchCriteria: null,
			displayedItems: getDisplayedItems(this.items, this.value), 
		};
```

Remove `searchCriteria` from `getDisplayedItems()` args (call and definition)

packages/forms/src/UIForm/fields/NestedListView/NestedListView.component.js
```diff
@@ -0,0 +1,253 @@
...
+
+		const callbacks = {
+			onExpandToggle: this.onExpandToggle.bind(this),
+			onParentChange: this.onParentChange.bind(this),
+			onCheck: this.onCheck.bind(this),
+		};
+
+		this.items = prepareItemsFromSchema(schema, callbacks);
+		this.value = value;
```
jsomsanith: Why don't you store them in state anymore ?

packages/forms/src/UIForm/fields/NestedListView/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+**UI Schema**
+
+| Property | Description | Default |
+|---|---|---|
+| key |  |  |
+| title | The title to display in the ListView header |  |
+| description | A description to display below the ListView |  |
+| placeholder | The search input placeholder |  |
+| widget | The widget to use | `nestedListView` |
```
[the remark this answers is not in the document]
jsomsanith: Still KO

packages/forms/src/UIForm/fields/NestedListView/NestedListView.component.js
sgangnant: `searchCriteria` is needed in `getDisplayedItems`, but I will remove it from constructor

packages/forms/src/UIForm/fields/NestedListView/NestedListView.component.js
jsomsanith: Not necessarily. It really represent the state of your component, and it can have internal state. It depends on the case.
Usually, you should not store things that can be calculated from props. But here it can be heavy, and the state is used for some kind of memoization. Nothing wrong IMO

What you say about items props that is overridden, it's more the state spreading that is wrong. We control all the state and know exactly what we want to pass. We should not spread it.

packages/forms/src/UIForm/fields/NestedListView/README.md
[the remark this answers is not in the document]
sgangnant: Should be good now

packages/forms/src/UIForm/fields/NestedListView/NestedListView.component.js
jdescottes-talend: because they are not used directly in the rendering

jdescottes-talend: But in fact what could have been done, is store all the `value`, `items`, `searchCriteria` and so on in the state, and have the `displayedItems` computed in the render with memoization. But things have already changed 42 times so whatever :p

PR #1618, opened 2018-08-15
feat(Datalist): add live change callback

jsomsanith:
Datalist onChange callback is only triggered when user has finished to edit the value : 
- on blur
- on suggestion selection
This allows for example to revert the value by pressing esc.

But we don't cover usecases like async suggestion loading, where we need to have a callback on user input change.

Add an onLiveChange that is triggered at each input change.

### DIFF
--- a/packages/components/src/Datalist/Datalist.component.test.js
+++ b/packages/components/src/Datalist/Datalist.component.test.js
@@ -522,7 +522,7 @@ describe('Datalist component', () => {
 		expect(instance.updateSelectedIndexes).toHaveBeenCalled();
 	});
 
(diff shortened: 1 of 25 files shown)


packages/components/src/Datalist/Datalist.component.test.js
```diff
@@ -524,9 +524,9 @@ describe('Datalist component', () => {
 
 	it('should call onTokenChange when user type in the input', () => {
```
jdescottes-talend: onTokenChange => onLiveChange

jsomsanith: Done, thanks

PR #1620, opened 2018-08-15
chore(datalist/liveupdate): Adding onLiveUpdate on Datalist

rmannibucau:

No way to implement dynamic suggestions with current Datalist API.

- Add a onLiveChange which is like onChange but called even on not data update (volatile)
- Use that in the form datalist field and call triggers (which are filtered by "onEvent" so user has to set onEvent=change to have this feature)


jsomsanith:
Duplicate #1618

PR #1625, opened 2018-08-17
feat(Forms): text display mode

jsomsanith:
With old forms, to display the info in text format we had JSONSchemaRenderer component.
Adapting JSONSchemaRenderer to the new format is not a good solution as fieldsets widgets (tabs, arrays, etc) would have to be implemented again, even if we have common styles with the form.

The approach here is to introduce display mode.
- to provide specific implementations of each widgets for a `text` display mode.
- they are registered in the widget dictionary with the key `{widgetId}_{displayMode)`. Example `checkbox_text` for text display mode.
- UIForm will select the right widget based on a `props.displayMode` (form global mode) or `uiSchema.displayMode` (property specific mode)
- storybook has now a checkbox to switch from classical and text display mode

For info : 
- [dl MDN page](https://developer.mozilla.org/fr/docs/Web/HTML/Element/dl)
- [dt MDN page](https://developer.mozilla.org/fr/docs/Web/HTML/Element/dt)
- [dd MDN page](https://developer.mozilla.org/fr/docs/Web/HTML/Element/dd)

Our structure is correct, event if the storybook a11y check returns an error. Plus, I tested with a screen reader
```
<dl>
    <div>
        <dt></dt>
        <dd></dd>
    <</div>
</dl>
```

### DIFF
--- a/output/forms.eslint.txt
+++ b/output/forms.eslint.txt
@@ -23,6 +23,9 @@ The react/require-extension rule is deprecated. Please use the import/extensions
 /home/travis/build/Talend/ui/packages/forms/src/UIForm/fields/Radios/Radios.component.js
   24:7  error  Form controls using a label to identify them must be programmatically associated with the control using htmlFor  jsx-a11y/label-has-for
 
+/home/travis/build/Talend/ui/packages/forms/src/UIForm/UIForm.component.js
+  21:7  error  'formTemplates' is assigned a value but never used  no-unused-vars
+
 /home/travis/build/Talend/ui/packages/forms/src/UIForm/utils/templates.js
   1:8  error  Using exported name 'DefaultArrayTemplate' as identifier for default export  import/no-named-as-default
 
@@ -30,5 +33,5 @@ The react/require-extension rule is deprecated. Please use the import/extensions
   68:8  error  Expected indentation of 6 tab characters but found 7  react/jsx-indent
   78:2  error  Mixed spaces and tabs                                 no-mixed-spaces-and-tabs
 
-✖ 12 problems (12 errors, 0 warnings)
+✖ 13 problems (13 errors, 0 warnings)
 
--- a/output/forms.eslint.txt
+++ b/output/forms.eslint.txt
@@ -23,9 +23,6 @@ The react/require-extension rule is deprecated. Please use the import/extensions
 /home/travis/build/Talend/ui/packages/forms/src/UIForm/fields/Radios/Radios.component.js
   24:7  error  Form controls using a label to identify them must be programmatically associated with the control using htmlFor  jsx-a11y/label-has-for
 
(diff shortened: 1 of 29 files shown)


output/forms.eslint.txt
```diff
@@ -23,12 +23,15 @@ The react/require-extension rule is deprecated. Please use the import/extensions
 /home/travis/build/Talend/ui/packages/forms/src/UIForm/fields/Radios/Radios.component.js
   24:7  error  Form controls using a label to identify them must be programmatically associated with the control using htmlFor  jsx-a11y/label-has-for
 
+/home/travis/build/Talend/ui/packages/forms/src/UIForm/UIForm.component.js
+  21:7  error  'formTemplates' is assigned a value but never used  no-unused-vars
```
romainseb: some lint here

nicomaligne:
Just don't forget to fix the linter error if it stills here after the CI.

packages/forms/src/UIForm/fields/TextArea/displayMode/TextMode.component.js
```diff
@@ -0,0 +1,29 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import { TextMode as FieldTemplate } from '../../FieldTemplate';
+
+export default function TextMode({ id, schema, value }) {
+	const { rows = 5, title } = schema;
+
+	return (
+		<FieldTemplate id={id} label={title}>
+			<pre style={{ height: `${rows * 2}rem`, fontSize: 'inherit' }}>{value}</pre>
```
romainseb: do we have to make inline style here ?

packages/forms/src/UIForm/Widget/Widget.component.js
```diff
@@ -7,15 +7,31 @@ import { getError } from '../utils/errors';
 import { getValue } from '../utils/properties';
 import shouldRender from './condition';
 
+function getWidget(displayMode, widgetId, customWidgets) {
+	// resolve the widget id depending on the display mode
+	let id = widgetId;
```
sgangnant: You can use a ternary here to avoid using `let`
`const id = displayMode ? \`${widgetId}_${displayMode}\` : widgetId;`

packages/forms/src/UIForm/fields/TextArea/displayMode/TextMode.component.js
jsomsanith: There is a feature in Textarea about the size. We can pass a row attributes. I just wanted to make the div `almost` have the same size as the textarea.

jsomsanith:
@acateland Done. Added custom widgets doc followed by display mode.

https://github.com/Talend/ui/blob/49daf58ec923a412ef3711a1f2f5e9588cae9d5b/packages/forms/src/UIForm/README.md#custom-widgets

PR #1628, opened 2018-08-21
chore(cmf/middleware): passing headers to onActionresponse

jlame:
In case of file upload the file name is sent in the response headers so we need to have access to it in onResponse callback, might also be usefull in other cases.

adding headers as a parameter of onActionResponse callback.

### DIFF
--- a/packages/cmf/src/middlewares/http/middleware.js
+++ b/packages/cmf/src/middlewares/http/middleware.js
@@ -163,8 +163,7 @@ export function handleResponse(response) {
 		return Promise.resolve({});
 	}
 	if (response.json) {
-		return response.json()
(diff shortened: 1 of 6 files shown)


packages/cmf/src/middlewares/http/middleware.js
```diff
@@ -163,7 +163,9 @@ export function handleResponse(response) {
 		return Promise.resolve({});
 	}
 	if (response.json) {
-		return response.json();
+		return response.json().then(json => {
+			return Promise.resolve({ data: json, headers: response.headers });
```
jsomsanith: You don't need to create a promise that resolves directly within a Promise.
```
return response
	.json()
	.then(json =>({ data: json, headers: response.headers }));
```

PR #1642, opened 2018-08-23
feat(frontend): add className prop to enumeration item

lmaillet:
I need for a functionality in TMC to have different color in Enumeration for inactive users.
See [https://jira.talendforge.org/secure/attachment/152884/screenshot-2.png](https://jira.talendforge.org/secure/attachment/152884/screenshot-2.png)

I've added the new prop `className` in the `Item` component.


packages/components/src/Enumeration/Items/Item/Item.component.js
```diff
@@ -21,10 +21,11 @@ function itemClasses(isSelected) {
 	});
 }
 
-function itemLabelClasses() {
+function itemLabelClasses(item) {
```
sgangnant: You don't need to pass `item` to the method, you can directly pass `item.className` inline in the upper scope

packages/components/stories/Enumeration.css
```diff
@@ -0,0 +1,3 @@
+.tc-enumeration button.inactive {
+  color: silver;
```
jdescottes-talend: Please keep the style in sass for consistency :).

lmaillet: I can't make the style be applied if I rename to `Enumeration.scss`.
So I added the style to `preview-head.html` as it's done for other stories.

PR #1650, opened 2018-08-27
feat(Treeview): improve a11y

jsomsanith:
* component structure is incorrect
* no accessibility gesture implemented

Make it accessible following https://www.w3.org/TR/wai-aria-practices-1.1/examples/treeview/treeview-1/treeview-1b.html

* props.onClick has been replaced with props.onToggle (more explicit)
* props.structure[].selected has been replaces with props.selectedId

Before
```
import TreeView from '@talend/react-components/lib/TreeView'

function MyComponent(props) {
	...

	const structure = [{ id: 'first', name: 'awesome folder', selected: true }];
	return <TreeView structure={structure} onClick={toggleHandler} ... />
}
```

After
```
import TreeView from '@talend/react-components/lib/TreeView'

function MyComponent(props) {
	...
	const structure = [{ id: 'first', name: 'awesome folder' }];
	return <TreeView structure={structure} selectedId={'first'} onToggle={toggleHandler} ... />
}
```


packages/components/src/TreeView/TreeViewItem/TreeViewItem.component.js
```diff
@@ -112,57 +125,202 @@ class TreeViewItem extends React.Component {
...
+				currentFound = value === this.containerRef;
+				hasNext = !done;
+			}
+		} while (hasNext);
+
+		return nextElement;
+	}
+
+	getPreviousItem() {
+		let previousElement;
```
jmfrancois: the difference between getPreviousItem and getNextItem has lots of repetitions. Could you refactor it please ?

jmfrancois: `width deep structure` example is broken
screen shot 2018-09-05 at 13 55 22

jsomsanith: `the difference between getPreviousItem and getNextItem has lots of repetitions. Could you refactor it please ?`
--> not really, there are changes that make it not really possible to factor it. The only thing that I can put in common is the selection of all items. I did that

packages/components/src/TreeView/TreeView.scss
```diff
@@ -2,42 +2,26 @@
...
-				fill: $brand-primary;
-				margin-right: 0;
-			}
+		:global(.btn-link .tc-svg-icon) {
+			stroke: $brand-primary;
 		}
 	}
 
-	.tc-treeview-nav {
+	.tc-treeview-ul {
```
frassinier: Please do not use element in css class name (i.e. `&-list` instead)

PR #1654, opened 2018-08-27
feat(UIForms): improve a11y

jsomsanith:
* errors are not assertive
* inputs are not linked to descriptions and errors
* no aria attributes

* make errors div always rendered, with `aria-live="assertive"`, so any change will be immediatly described by SR
* use aria-describedby to link input with the description and errors
* use aria-required on required values
* use aria-invalid on invalid inputs


output/forms.eslint.txt
```diff
@@ -6,10 +6,12 @@ The react/require-extension rule is deprecated. Please use the import/extensions
...
 
 /home/travis/build/Talend/ui/packages/forms/src/UIForm/fields/CheckBox/SimpleCheckBox.component.js
-  9:4  error  Form controls using a label to identify them must be programmatically associated with the control using htmlFor  jsx-a11y/label-has-for
+  10:4  error  Form controls using a label to identify them must be programmatically associated with the control using htmlFor  jsx-a11y/label-has-for
+  11:5  error  The attribute aria-required is not supported by the role checkbox. This role is implicit on the element input    jsx-a11y/role-supports-aria-props
 
 /home/travis/build/Talend/ui/packages/forms/src/UIForm/fields/Radios/Radios.component.js
-  24:7  error  Form controls using a label to identify them must be programmatically associated with the control using htmlFor  jsx-a11y/label-has-for
+  25:7  error  Form controls using a label to identify them must be programmatically associated with the control using htmlFor  jsx-a11y/label-has-for
+  26:8  error  The attribute aria-required is not supported by the role radio. This role is implicit on the element input       jsx-a11y/role-supports-aria-props
```
talend-glorieux: Fix those errors.

packages/forms/src/UIForm/Message/generateId.js
```diff
@@ -0,0 +1,11 @@
...
+	return id && `${id}-description`;
+}
+
+export function generateErrorId(id) {
+	return id && `${id}-error`;
+}
+
+export function generateDescribedBy(id) {
+	return id && `${generateDescriptionId(id)} ${generateErrorId(id)}`;
+}
```
talend-glorieux: Would be better if the method returned something meaningful if the id isn't passed.
In that case it returns `undefined` if no `id` is passed.

jsomsanith: I changed that to generate an id via uuid.v4() if there is no base id.
So each widget now generate it, and pass it to the input and FieldTemplate to set it on description and errors

packages/forms/src/UIForm/Message/generateId.js
```diff
@@ -0,0 +1,16 @@
+import uuid from 'uuid';
```
sgangnant: Since this is used by other form elements than Message, should't it be somewhere else ?

PR #1656, opened 2018-08-28
fix(UIForms): datalist alignement

jsomsanith:
When datalist is side by side with an input, the form group is not aligned.
capture d ecran 2018-08-28 a 16 49 44

This is du to the label style definition that is placed as material design style, if it follows a `form-control`.
But adding `form-control` on the datalist div container is not ok, it introduce more visual regression, as the theme add input style.

* introduce a `form-control-container` classname, and make the label style applied if it follows it
* add this new classname on datalist widget

capture d ecran 2018-08-28 a 16 38 25


packages/components/src/Datalist/Datalist.scss
```diff
@@ -4,61 +4,55 @@ $tc-datalist-items-shadow-color: $shadow !default;
 $tc-typeahead-input-width: 260px !default;
 $tc-datalist-item-height: 40px !default;
 
-.tc-datalist {
-	display: flex;
-	align-items: center;
+
+.tc-datalist-form {
```
jdescottes-talend: Where is it used ? And if used, shouldn't it be declared in the form/widget scope instead ?

PR #1660, opened 2018-08-29
feat(forms/conditions/TCOMP-1029): add jsonLogic condition [BREAKING CHANGE]

rmannibucau:

Some developpers need to be able to make conditional the visibility using a contains logic.

Use http://jsonlogic.com/ to implement condition.

Before:

```javascript
{
	values: [undefined],
	path: 'undefined',
}
```

After:

```javascript
{
	'==': [{ var: 'undefined' }, undefined],
}
```


packages/forms/src/UIForm/Widget/condition.js
```diff
@@ -1,18 +1,81 @@
 import get from 'lodash/get';
-import includes from 'lodash/includes';
+import findIndex from 'lodash/findIndex';
 
-function getValue(value, strategy) {
-	if (!strategy) {
+function toNumber(value) {
```
acateland: if the value is not a string this act as identity function, wich is not expected for a toNumber function.

packages/forms/src/UIForm/Widget/condition.js
```diff
@@ -1,18 +1,81 @@
...
+
+function toString(value) {
+	if (!value) {
 		return value;
 	}
-	switch (strategy.toLowerCase()) {
+	if (typeof value === 'number') {
+		return JSON.stringify(value);
+	}
+	if (value.toString) {
```
acateland: Wouldn’t it be better tonremove this check and return ´toString(value)’ so the function will consistantly return a string ?

packages/forms/src/UIForm/Widget/condition.js
rmannibucau: any advice? throw an exception?

packages/forms/src/UIForm/Widget/condition.js
rmannibucau: not sure I got it, it will loop if I call it again here and it is not a number, i can likely throw an exception if not a string instead of doing this check which is quite poor for objects

packages/forms/src/UIForm/Widget/condition.js
```diff
@@ -1,18 +1,91 @@
...
+				name: strategy.substring(0, start).toLowerCase(),
+				params,
+			};
+		}
+	}
+	return {
+		name: strategy.toLowerCase(),
+		params: {},
+	};
+}
```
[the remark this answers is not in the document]
talend-glorieux: It could also be a good idea to add a unit test on that function.

packages/forms/src/UIForm/Widget/condition.js
```diff
@@ -1,18 +1,91 @@
...
+
+function toEvaluator(value, strategyConfig) {
+	switch (strategyConfig.name) {
 		case 'length':
 			if (value && value.length) {
-				return value.length;
+				const length = value.length;
+				return expected => length === toNumber(expected);
+			}
+			return expected => expected === 0 || expected === '0';
```
talend-glorieux: `expected` should be normalized to a number so that you don't have to compare to `'0'`.

packages/forms/src/UIForm/Widget/condition.js
```diff
@@ -1,18 +1,91 @@
...
+		case 'contains':
+			if (!value) {
+				return () => false;
 			}
-			return 0;
+			if (strategyConfig.params.lowercase === 'true') {
```
talend-glorieux: Don't use ` 'true'` as a boolean.

packages/forms/src/UIForm/Widget/condition.js
```diff
@@ -21,9 +94,15 @@ function evaluateInlineCondition(properties, condition) {
...
 	const value = get(properties, condition.path);
-	const actual = getValue(value, condition.strategy);
-	return (condition.shouldBe !== false) === includes(condition.values, actual);
+	let evaluator;
+	if (strategyConfig) {
+		evaluator = toEvaluator(value, strategyConfig);
+	} else {
+		evaluator = expected => value === expected || (value && toString(value) === expected);
+	}
+	return (condition.shouldBe !== false) === findIndex(condition.values, evaluator) >= 0;
```
talend-glorieux: I don't understand what's happening on those last seven lines.

packages/forms/src/UIForm/Widget/condition.js
```diff
@@ -1,18 +1,91 @@
...
+		}
+	}
+	return {
+		name: strategy.toLowerCase(),
+		params: {},
+	};
+}
+
+function toEvaluator(value, strategyConfig) {
+	switch (strategyConfig.name) {
```
jdescottes-talend: We could change the switch for a configurable array, with the name as the key, and the evaluator fn as value.
This pluggable way will ease the adding of possible new strategies, remove the need to call manually each evaluator with the args and also to have this file grow over time.

Here it could be like :
```javascript
// [...]/conditions.js
import evaluatorsDefinitions from './evaluators'; //

function getEvaluatorFor(strategyConfig, value) {
    return evaluatorsDefinitions[strategyConfig.name](strategyConfig, value);
}
```

The defintions of evaluators could look like : 
```javascript
// [...]/evaluators/index.js
import containsEvaluator from './contains';

export default {
    'contains': containsEvaluator,
    // Bunch of others evaluators
};
```

An evaluator could look like :
```javascript
// [...]/evaluators/contains.js
export default containsEvaluator(strategyConfig, value) {
   // What actually exists 
   // Signature : return expected => value
}
```

packages/forms/src/UIForm/Widget/condition.js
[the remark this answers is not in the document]
rmannibucau: AFAIK it is consistent {name,params} or undefined/falsy. will add a test

packages/forms/src/UIForm/Widget/condition.js
jdescottes-talend: Moreover : 0 is the default value against checking the expectation ? If so, define the value first (length) with a default (0) if not provided, then do the same check for all.

packages/forms/src/UIForm/Widget/condition.js
```diff
@@ -1,18 +1,167 @@
...
+ * parses a strategy based on this pattern "name(parameters)" and returns
+ * an object containing { name, params }.
+ */
+function parseStrategy(strategy) {
+	if (!strategy) {
+		return strategy;
+	}
+
+	const start = strategy.indexOf('(');
+	if (start > 0) {
```
jdescottes-talend: Parsing a string is entirely the job of RegExp, one that could do the job :
`^([a-zA-Z]{1}\w*)(?:\((.+)?\))?$`
Will extract the name as second match and parameters as third match (or undefined, if no parentheses or empty parentheses)

No need to deal with indexes this way and lot's of checks. There will just remain the need to check if there is parameters against undefined (to verify with lots of units tests).

packages/forms/src/UIForm/Widget/condition.js
```diff
@@ -1,18 +1,167 @@
...
+
+/**
+ * "foo=bar;dummy=1" will return [{foo:"bar",dummy:"1"}]
+ */
+function parseParameters(parameterString) {
+	return parameterString
+		.split(';')
+		.map(it => {
+			const sep = it.indexOf('=');
+			if (sep > 0) {
```
jdescottes-talend: It's still a good case for RegExp
Example : `^(?:(\w+)=)?(.+)$`

Define the key `value` as default if group1 matching is not found

This way we have a consistent unique return too and we not deal with indexes

packages/forms/src/UIForm/Widget/condition.js
```diff
@@ -1,18 +1,167 @@
...
+	return onError(value);
+}
+
+/**
+ * "foo=bar;dummy=1" will return [{foo:"bar",dummy:"1"}]
+ */
+function parseParameters(parameterString) {
+	return parameterString
+		.split(';')
+		.map(it => {
```
jdescottes-talend: What's `it` for ?

packages/forms/src/UIForm/Widget/condition.js
rmannibucau: using indices is faster (keep in mind it is called in render) and is less fragile than your regex which doesnt guarantee the parenthesis symmetry I think :(.

jdescottes-talend:
For the parsing part, something to check is specials characters, especially, those used as delimiters. There will be potential bugs with that, we need to escape them or do something else

rmannibucau:
@jdescottes-talend agree (and also why it is not ',') but let's have a need for that before doing a string parsing by character. Today we only options not requiring this careness and this method is only for options :).

packages/forms/src/UIForm/Widget/condition.js
jdescottes-talend: There is unit tests for robustness.
If the values doesn't change, they must be some caching with memoization somehow somewhere to avoid recomputation. And if the values change often, there is other questions to asks, like why are we changing the strategy really often ?.

I never find the fast argument relevant on frontend for strict computation only. It's more often (always?) some i/o latency, some calls to compute values already computed, some computations not needed at the current time (like a whole list not actually rendered) which can be deferred ; which cause the real bad performances, aside with bad architecture.
I have found a lot of cases, where fast is the first thing searched, resulting in cumbersome code in the big picture, which result in others developers not understanding what's actually done, making others changes on top without refactoring because they cannot change what they don't understand, resulting at last in a poor final architecture app and so bad performances :p

I don't understand the thing with the parenthesis, what's need to be guarantee ?

jdescottes-talend: ps: dealing with indexes is as fragile as regex ^^'

rmannibucau: We can memoize it but it is not really needed with current implementation I think (which is always better since memoization is a workaround due to the lack of listeners :().

If you have "foo(" you must not parse parameters, your regex will let it do it if I read it right.
The other advantage i see using this implementation (and I will agree it is not an advantage for you but it is for me and talend ecosystem ;)) is that it is 1-1 the implementation we have in talend studio and since this is a feature we must offer on both sides it is simpler to align it.
Depending how we make this trivial language evolving, regex can kill us very quickly whereas this parsing will always work and be reliable cause it enables any syntax whereas regex are limited from my past experiences.

packages/forms/src/UIForm/Widget/condition.js
```diff
@@ -1,29 +1,173 @@
...
+ */
+function stringOrNumberToString(
+	value,
+	onError = val => {
+		const error = { error: 'the passed value is not a string or a number', value: val };
+		throw error;
+	},
+) {
+	if (!value) {
+		return value;
```
jdescottes-talend: This check can return any falsy value like 0

packages/forms/src/UIForm/Widget/condition.js
```diff
@@ -1,29 +1,173 @@
...
+	}
+	if (typeof value === 'number') {
+		return JSON.stringify(value);
+	}
+	if (typeof value === 'string') {
+		return value.toString();
```
jdescottes-talend: I don't understand what's done here

rmannibucau: dropping the toString, was a remaning part from previous code, good catch, thks!

packages/forms/src/UIForm/Widget/condition.js
rmannibucau: agree, i picked value since it was here ;)

packages/forms/src/UIForm/Widget/condition.test.js
```diff
@@ -49,6 +50,11 @@ const TRUTHY_CONDITIONS = [
 		strategy: 'length',
 		path: 'arrayEmpty',
 	},
+	{
+		values: ['0'],
+		strategy: 'length',
+		path: 'arrayEmpty',
+	},
```
jmfrancois: no difference with the previous case ?

rmannibucau: it is to check the coercing of the value (string/number)

jmfrancois: 👍

packages/forms/src/UIForm/Widget/condition.js
```diff
@@ -1,29 +1,172 @@
...
+		name: name.toLowerCase(),
+		params: hasParameters ? parseParameters(parameters) : {},
+	};
 }
 
+function areEqualsAsNumbers(expected, actualStringOrNumber) {
+	return expected === stringOrNumberToNumber(actualStringOrNumber);
+}
+
+function containsString(expected, actualStringOrNumber, valueProcessor = v => v) {
```
acateland: i think a function prefixed by contains should always return a boolean

packages/forms/src/UIForm/Widget/condition.js
```diff
@@ -1,29 +1,172 @@
...
+ * It supports string and number for now.
+ */
+function stringOrNumberToNumber(value) {
+	if (typeof value === 'number') {
+		return value;
+	}
+	if (typeof value === 'string') {
+		return Number(value);
+	}
+	const error = { error: 'the passed value is not a string or a number', value };
```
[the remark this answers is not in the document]
rmannibucau: Idea was to be able to catch it and have the value to implement a recover, if i ignore the original value and the direct caller doesn't process it then I'm loosing this info, no?

packages/forms/src/UIForm/Widget/condition.js
```diff
@@ -1,29 +1,172 @@
...
-		case 'length':
-			if (value && value.length) {
-				return value.length;
+	if (typeof value === 'number') {
+		return JSON.stringify(value);
+	}
+	if (typeof value === 'string') {
+		return value;
+	}
+	return onError(value);
```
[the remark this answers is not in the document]
rmannibucau: likely the same

packages/forms/src/UIForm/Widget/condition.js
rmannibucau: do you mean you want a !! at the beginning?

packages/forms/src/UIForm/Widget/condition.js
```diff
@@ -1,29 +1,172 @@
...
+ * Tries to make the returned value a string.
+ * It supports number and string as input.
+ * In case of an error it will call onError. By default it throws an exception.
+ */
+function stringOrNumberToString(
+	value,
+	onError = val => {
+		const error = { error: 'the passed value is not a string or a number', value: val };
+		throw error;
+	},
```
talend-glorieux: Don't define the default function inline.
Also useless constant.

packages/forms/src/UIForm/Widget/condition.js
```diff
@@ -1,29 +1,172 @@
...
+			};
+		})
+		.reduce(
+			(a, v) => ({
+				...a,
+				...v,
+			}),
+			{},
+		);
+}
```
talend-glorieux: Needs simplification refactoring.

packages/forms/src/UIForm/Widget/condition.js
```diff
@@ -1,29 +1,172 @@
...
+	const name = isMatching ? matches[1] : strategy;
+	const parameters = isMatching ? matches[2] : undefined;
+
+	const hasParameters = parameters !== undefined;
+
+	return {
+		name: name.toLowerCase(),
+		params: hasParameters ? parseParameters(parameters) : {},
+	};
 }
```
talend-glorieux: Use regular if/else to see all `return` possibilities.

packages/forms/src/UIForm/Widget/condition.js
```diff
@@ -1,29 +1,172 @@
...
+function areEqualsAsString(expected, actualStringOrNumber) {
+	return actualStringOrNumber && stringOrNumberToString(actualStringOrNumber, v => v) === expected;
+}
+
+/**
+ * For an array or a string if compares the expected value to its length.
+ */
+function toLengthEvaluator(value) {
+	if (value && value.length) {
+		const length = value.length;
```
talend-glorieux: Useless constant.

packages/forms/src/UIForm/Widget/condition.js
```diff
@@ -1,29 +1,172 @@
...
+
+function containsString(expected, actualStringOrNumber, valueProcessor = v => v) {
+	return (
+		actualStringOrNumber &&
+		valueProcessor(stringOrNumberToString(actualStringOrNumber)).includes(expected)
+	);
+}
+
+// this will not fail in case actualStringOrNumber is not a string or number
+// but will return false sinc eit is used as a fallback comparison using coercing
```
talend-glorieux: Keep comments formatting consistent.

packages/forms/src/UIForm/Widget/condition.js
```diff
@@ -1,29 +1,172 @@
...
+	if (!actual) {
+		return () => false;
+	}
+	if (Array.isArray(actual)) {
+		if (options.lowercase === 'true') {
+			return expected => actual.map(it => it.toLowerCase()).indexOf(expected) >= 0;
+		}
+		return expected => actual.indexOf(expected) >= 0;
+	}
+	if (options.lowercase === 'true') {
```
talend-glorieux: Please don't use `'true'` as a boolean expression. It should be normalized beforehand.

packages/forms/src/UIForm/Widget/condition.js
```diff
@@ -1,29 +1,172 @@
...
+function containsString(expected, actualStringOrNumber, valueProcessor = v => v) {
+	return (
+		actualStringOrNumber &&
+		valueProcessor(stringOrNumberToString(actualStringOrNumber)).includes(expected)
+	);
+}
+
+// this will not fail in case actualStringOrNumber is not a string or number
+// but will return false sinc eit is used as a fallback comparison using coercing
+function areEqualsAsString(expected, actualStringOrNumber) {
```
talend-glorieux: What does `actual` mean?

packages/forms/src/UIForm/Widget/condition.js
```diff
@@ -90,3 +233,10 @@ export default function shouldRender(condition, properties) {
 	// navigate the nested graph
 	return evaluateChildrenCondition(properties, condition);
 }
+
+export const $internals = {
```
talend-glorieux: Please dont use `$` prefix.

packages/containers/src/ComponentForm/ComponentForm.component.js
```diff
@@ -141,6 +141,7 @@ export class TCompForm extends React.Component {
 			// - race conditions,
 			// - trigger result that is does not fit user entry anymore,
 			// - erase a good value put by the enduser
+			// TODO(TUI-203): move this in UIForm.container
```
acateland: guess you can remove this

packages/forms/src/UIForm/README.md
```diff
@@ -283,17 +283,8 @@ Those validations change the `errors` object accordingly.
 ### Conditional rendering
 
 It is possible to render parts of the forms defined in uiSchema, depending on properties values.
-The uiSchema accepts a `condition` property, which defines recursively all conditions to match to be rendered.
+The uiSchema accepts a `condition` property, which use [jsonLogic](http://jsonlogic.com). So please firt go their and read the doc.
```
acateland: So please first go there and read the doc.

packages/forms/src/UIForm/Widget/condition.js
```diff
@@ -1,92 +1,33 @@
...
+	if (typeof a === 'number') {
+		return a;
+	} else if (typeof a === 'string') {
+		return parseInt(a, 10);
 	}
-
-	const value = get(properties, condition.path);
-	const actual = getValue(value, condition.strategy);
-	return (condition.shouldBe !== false) === includes(condition.values, actual);
+	throw new Error(`${a.toString()} is not a number`);
```
acateland: throw new TypeError ?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypeError

acateland:
image
 :+1:

PR #1667, opened 2018-09-03
fix(UIForms/TUI-197): nested array items key has now the indexes

jsomsanith:
Array inject the item index only on the first level of schema items.
For more complex item widgets, the key is not propagated.

Example: 
ArrayWidget
.    .CustomFieldset
.    .    .TexWidget // this has not the index in the key

On submit, the value is not passed.

Resolve keys deeply on array item schema computation


packages/forms/src/UIForm/fieldsets/Array/Array.component.js
```diff
@@ -6,12 +6,16 @@ import defaultTemplates from '../../utils/templates';
 import defaultWidgets from '../../utils/widgets';
 
 function adaptKeyWithIndex(keys, index) {
-	let indexedKeys = keys;
-	const firstIndexPlaceholder = indexedKeys.indexOf('');
-	if (firstIndexPlaceholder >= 0) {
-		indexedKeys = [...keys];
-		indexedKeys[firstIndexPlaceholder] = index;
+	if (!keys) {
```
sgangnant: This check could be made by callers before calling adaptKeyWithIndex to have a consistent return type from this method

packages/forms/src/UIForm/fieldsets/Array/Array.component.js
```diff
@@ -6,12 +6,16 @@ import defaultTemplates from '../../utils/templates';
...
+		return;
 	}
+
+	let firstIndexPlaceholder = keys.indexOf('');
```
acateland: ok i don't understand why you need to add this bit of code

jsomsanith: The cases : 
- key = ["my", "array", "", "nested"] for nested items fields
- key = ["my", "array"] : this defines array itself. Each item will receive a key ["my", "array", index]
To check that, I spot the first empty string in the key
- find it:  and replace it, it's a nested element
- not found : it's an item key, I add the index after

packages/forms/src/UIForm/fieldsets/Array/Array.component.js
jsomsanith: Thanks, fixed

PR #1669, opened 2018-09-04
feat(containers/DeleteResource): Add `onCancelRedirectUrl` prop

yyanwang:
Currently we pass `redirecetUrl` to DeleteResource container, and it will redirect user to that url after removing resource as well as cancel the delete operation. in TMC, we need to redirect to different urls on cancel/validate.
Also add the logic to remove resource from redux store when DELETE call returned successfully.
Add a prop `onCancelRedirectUrl` to which user will be redirected to when cancel delete. if not set, then will use `redirectUrl`  for backward compatibility.


packages/containers/src/DeleteResource/sagas.js
```diff
@@ -23,8 +23,7 @@ got ${resourcePath}`,
 	return resourceType;
 }
 
-export function* redirect(action) {
-	const url = get(action, 'data.model.redirectUrl');
+export function* redirect(url) {
 	if (!url) {
 		throw new Error('redirect action must have data.model.redirectUrl value');
```
sgangnant: This message should be updated since we're no longer testing the `redirectUrl` object's key

packages/containers/src/DeleteResource/sagas.test.js
```diff
@@ -21,7 +21,7 @@ describe('internals', () => {
 	});
 	describe('redirect', () => {
 		it('should put a redirect action', () => {
-			const gen = internals.redirect({ data: { model: { redirectUrl: '/foo' } } });
+			const gen = internals.redirect('/foo');
```
sgangnant: Those changes are purely internals or could we expect breaking changes ?

yyanwang: `redirect` is used only internal, should be no breaking changes.

packages/containers/src/DeleteResource/sagas.js
```diff
@@ -75,17 +74,26 @@ export function* deleteResourceValidate(
...
 				model: {
+					id: safeId,
 					labelResource: resource.get('label') || resource.get('name', ''),
 				},
 			});
+			yield put(
+				cmf.actions.collections.mutate(resourceLocator, {
+					delete: [safeId],
+				}),
+			);
```
talend-glorieux: That's a good idea but it won't word on CQRS based applications.
Under that architecture we have to wait for an acknowledgment from the back-end in order to actually remove the element. That's why we've kept the DELETE request and the collection removal steps separated.

packages/containers/src/DeleteResource/sagas.js
```diff
@@ -75,17 +74,26 @@ export function* deleteResourceValidate(
...
 		}
-		yield call(redirect, action);
+		const url = get(action, 'data.model.redirectUrl');
```
talend-glorieux: Useless constant. Could be used directly inside `call`.
`yield call(redirect, get(action, 'data.model.redirectUrl'))`.

packages/containers/src/DeleteResource/sagas.js
```diff
@@ -75,17 +74,26 @@ export function* deleteResourceValidate(
...
+		yield call(redirect, url);
 	}
 }
 
 export function* deleteResourceCancel() {
 	const action = yield take(deleteResourceConst.DIALOG_BOX_DELETE_RESOURCE_CANCEL);
-	yield call(redirect, action);
+	const url =
+		get(action, 'data.model.onCancelRedirectUrl') || get(action, 'data.model.redirectUrl');
```
talend-glorieux: Could be simplified by using `get` default value.
`get(action, 'data.model.onCancelRedirectUrl', get(action, 'data.model.redirectUrl'))`

packages/containers/src/DeleteResource/sagas.js
```diff
@@ -23,10 +23,9 @@ got ${resourcePath}`,
 	return resourceType;
 }
 
-export function* redirect(action) {
-	const url = get(action, 'data.model.redirectUrl');
+export function* redirect(url) {
 	if (!url) {
-		throw new Error('redirect action must have data.model.redirectUrl value');
+		throw new Error('redirect url can not be empty');
```
acateland: maybe lodash isString can be a good fit on the condition ?

packages/containers/src/DeleteResource/README.md
```diff
@@ -90,6 +90,7 @@ the delete service will use it to check if the resource exist in your applicatio
 * **female** : Only for i18n, allow to set the i18nkey to tell of the resource type if female or not
 * **collectionId** : specify the collection which stores resource. if not provided, then use `resourceType` as collection name
 * **resourceUri** : is the backend api to delete resource. if not provided, then `${uri}/${resourceType}/${id}` will be used.
+* **onCancelRedirectUrl** : is the url to redirect when cancel delete. if not provided then `redirectUrl` will be used.
```
acateland: :+1:

packages/containers/src/DeleteResource/sagas.js
yyanwang: ok, thank you for this info. then i will remove this part.

yyanwang:
@jmfrancois thank you! from ux, we have something like a resource info page with subheader view and have delete button on it, when cancel delete it will stay at info page. so need this change.

PR #1674, opened 2018-09-05
feat(containers/TUI-189): add trigger to update properties

jmfrancois:

The backend can t create a trigger to change values in the form.

Add a new `update` trigger in the defaultRegistry.
Add demo in the sandbox

Demo: http://recordit.co/tcKasOLKE6

### DIFF
--- a/packages/containers/src/ComponentForm/kit/defaultRegistry.js
+++ b/packages/containers/src/ComponentForm/kit/defaultRegistry.js
@@ -111,10 +111,59 @@ function onError({ errors, error, schema }) {
 	return { errors: getNewErrors(errors, schema, extractErrorMessage(error)) };
 }
 
+/**
+ * Update the content of the properties
+ * @param body is the new value
+ * @param trigger the trigger that call it
+ * @example
+const trigger = {
+	"action":"guessMe",
+	"family":"test",
+	"options":[
+		{
+			"path":"root.updatable_config",
+			"type":"object" // or "string"
+		}
+	],
+	"parameters":[
+		{
+			"key":"arg0.name",
+			"path":"root.updatable_config.name"
+		}
+	],
+	"type":"update"
+};
+ */
+function updateProperties({ body, trigger, properties }) {
+	const newProperties = { ...properties };
+	const targetPath = trigger.options[0].path;
+	const getValue = () => {
+		if (trigger.options[0].type === 'object') {
+			return body;
+		}
+		return body.data;
+	};
+	const targetSplitted = targetPath.split('.');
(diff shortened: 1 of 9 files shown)


packages/containers/src/ComponentForm/kit/defaultRegistry.js
```diff
@@ -111,10 +111,59 @@ function onError({ errors, error, schema }) {
...
+		if (trigger.options[0].type === 'object') {
+			return body;
+		}
+		return body.data;
+	};
+	const targetSplitted = targetPath.split('.');
+	targetSplitted.reduce((acc, current, index) => {
+		if (!acc[current]) {
+			// eslint-disable-next-line no-param-reassign
+			acc[current] = {};
```
smouillour: it's really a good thing to do that? #ByPassESLint

sgangnant: In the context of a `reduce()` call it's not that shocking, and regarding eslint automatically flagging it, it is/has been debated on the eslint repo itself https://github.com/eslint/eslint/issues/8581
A workaround for that would be to return the deconstructed/reconstructed accumulator with values update

packages/containers/src/ComponentForm/kit/defaultRegistry.js
```diff
@@ -111,10 +111,59 @@ function onError({ errors, error, schema }) {
...
+	const newProperties = { ...properties };
+	const targetPath = trigger.options[0].path;
+	const getValue = () => {
+		if (trigger.options[0].type === 'object') {
+			return body;
+		}
+		return body.data;
+	};
+	const targetSplitted = targetPath.split('.');
+	targetSplitted.reduce((acc, current, index) => {
```
jsomsanith: You don't need to rewrite that : 
```
import { mutateValue } from '@talend/react-forms/lib/UIForm/utils/properties';

const schema = { key: targetPath.split('.') };
const value = trigger.options[0].type === 'object' ? body : body.data;
return {
	properties: mutateValue(properties, schema, value),
};
```

packages/containers/src/ComponentForm/kit/defaultRegistry.js
acateland: not shocking but yet easy to avoid

packages/containers/src/ComponentForm/kit/defaultRegistry.js
acateland: please don't mutate external scoped value inside a function that return explicitly said result, it make the code hard to read.

PR #1681, opened 2018-09-06
feat(Text) Overlay support

sgangnant:
Text fields don't support overlay at the moment (useful for RichTooltip)

Read overlay component from props and render field with overlay support


packages/forms/src/UIForm/fields/Text/Text.component.js
```diff
@@ -23,6 +35,37 @@ export default function Text(props) {
...
+	if (overlayComponent) {
+		const overlayProps = {
+			overlay: <Popover>{overlayComponent}</Popover>,
+			placement: overlayPlacement || 'right',
+		};
+
+		field = <OverlayTrigger {...overlayProps}>{input}</OverlayTrigger>;
+	} else {
+		field = input;
+	}
```
talend-glorieux: That entire logic might be better in its own function.

packages/forms/src/UIForm/fields/Text/Text.component.js
```diff
@@ -23,6 +35,37 @@ export default function Text(props) {
...
+			aria-describedby={`${descriptionId} ${errorId}`}
+		/>
+	);
+
+	let field;
+
+	if (overlayComponent) {
+		const overlayProps = {
+			overlay: <Popover>{overlayComponent}</Popover>,
+			placement: overlayPlacement || 'right',
```
talend-glorieux: `'right' should be a default prop.

packages/forms/src/UIForm/fields/Text/Text.component.js
```diff
@@ -1,12 +1,24 @@
...
+		id,
+		isValid,
+		errorMessage,
+		onChange,
+		onFinish,
+		schema,
+		value,
+		overlayComponent,
+		overlayPlacement,
+	} = props;
```
talend-glorieux: No need to spread the entire `props` object if most of the values are only used once.

packages/forms/src/UIForm/fields/Text/Text.component.test.js
```diff
@@ -191,4 +191,18 @@ describe('Text field', () => {
...
+
+	it('should render input with tooltip', () => {
+		// given
+		const props = {
+			schema,
+			overlayComponent: <span>Foo</span>,
+		};
+
+		// when
+		const wrapper = shallow(<Text {...props} />);
```
talend-glorieux: Why do you define the props outside of the `Text` component definition?

PR #1694, opened 2018-09-12
fix(containers/DeleteResource): Add onHide handler

yyanwang:
When click outside delete modal, there will be js errors about onClick not a function.
Pass `onHide` handler from DeleteResource. when click outside modal, it'll hide.


packages/containers/src/DeleteResource/DeleteResource.container.js
```diff
@@ -106,6 +106,9 @@ export class DeleteResource extends React.Component {
 				cancelAction={cancelAction}
 				validateAction={validateAction}
 				getComponent={this.props.getComponent}
+				onHide={() =>
+					this.props.dispatchActionCreator('DeleteResource#cancel', event, { model: resourceInfo })
```
jsomsanith: To avoid passing a new function ref at each render, can you make it a class function ?

PR #1697, opened 2018-09-13
feat(UIForm): import Code widget a11y

jsomsanith:
* no link between editor and error/description div
* keyboard trap within ace-editor (talend/ui > react-ace > brace > ace)

* generate ids for errors and descriptions, pass them to the textarea
* implement a 2 esc keydown gesture to focus out of the editor
* bonus: import missing imports to fix warnings


packages/forms/src/UIForm/fields/Code/Code.component.js
```diff
@@ -32,42 +45,79 @@ try {
 			this.props.onFinish(event, { schema: this.props.schema });
 		}
 
+		onKeyDown(event) {
+			if (event.keyCode === keyCode.codes.esc) {
```
acateland: while i find this feature to be a nice thing, how do we hint users that it is even a possible move ?

packages/forms/src/UIForm/fields/Code/Code.component.js
```diff
@@ -32,42 +45,79 @@ try {
...
+				const now = Date.now();
+				if (this.lastEsc && this.lastEsc - now < 1000) {
```
acateland: why not put it in the component state ?
if not why not initializing the property on component instantiation ?

packages/forms/src/UIForm/fields/Code/Code.component.js
```diff
@@ -32,42 +45,79 @@ try {
...
+		}
+
+		onBlur() {
+			this.editor.textInput.getElement().removeAttribute('tabindex');
+		}
+
 		onLoad(editor) {
-			editor.textInput.getElement().setAttribute('id', this.props.id);
+			this.editor = editor;
+			this.componentDidUpdate();
```
acateland: im not sure this should be considered as a good idea to manually call for React's life-cycle function.

packages/forms/src/UIForm/fields/Code/Code.component.js
jsomsanith: In fact I was thinking that this is something that will naturally be tried. During our a11y training, the blind instructor was faced to this issue, and he tried lot of things : (shift) tabs, (multiple) esc, SR shortcuts, ...

But if you think it is really needed, we can think about a hidden description div with this (translated instruction), then linnk the textarea with an `aria-describedby`.

packages/forms/src/UIForm/fields/Code/Code.component.js
jsomsanith: It would be a little big to trigger a render to perform the action I want. Here I just want to add the attributes to the Ace textarea. And to avoid writing this twice, I called the componentDidUpdate that only have this code.
But I understand the issue, I'll extract that into another method, that will be called in cDU and here.

packages/forms/src/UIForm/fields/Code/Code.component.js
jsomsanith: ok I add you that with a comment above it

packages/forms/src/UIForm/fields/Code/Code.component.js
jsomsanith: Added the instruction div

PR #1698, opened 2018-09-13
fix(list): add title on header icon

jare-talend:
There are one tooltip on the header with a name but there isn't no tooltip on the icon.

add title on the header icon

### DIFF
--- a/packages/components/src/VirtualizedList/HeaderIcon/HeaderIcon.component.js
+++ b/packages/components/src/VirtualizedList/HeaderIcon/HeaderIcon.component.js
@@ -13,8 +13,10 @@ import theme from './HeaderIcon.scss';
 function HeaderIcon({ label, sortBy, dataKey, sortDirection, columnData }) {
 	return (
 		<div className={classnames(theme['tc-header-icon'], 'tc-header-icon')}>
-			{columnData.iconName && <Icon name={columnData.iconName} title={label} />}
-			{sortBy === dataKey && <SortIndicator sortDirection={sortDirection} />}
+			<span title={label} aria-label={label}>
+				{columnData.iconName && <Icon name={columnData.iconName} />}
+				{sortBy === dataKey && <SortIndicator sortDirection={sortDirection} />}
+			</span>
 		</div>
 	);
 }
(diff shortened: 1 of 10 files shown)


packages/components/src/VirtualizedList/HeaderIcon/HeaderIcon.component.js
```diff
@@ -13,8 +13,10 @@ import theme from './HeaderIcon.scss';
 function HeaderIcon({ label, sortBy, dataKey, sortDirection, columnData }) {
 	return (
 		<div className={classnames(theme['tc-header-icon'], 'tc-header-icon')}>
-			{columnData.iconName && <Icon name={columnData.iconName} title={label} />}
-			{sortBy === dataKey && <SortIndicator sortDirection={sortDirection} />}
+			<span title={label} aria-label={label}>
```
jdescottes-talend: Why not change the div for a span and merge the props ?

packages/components/src/VirtualizedList/HeaderIcon/HeaderIcon.scss
```diff
@@ -4,5 +4,9 @@ $tc-list-header-icon-margin-top: 3px !default;
 	display: flex;
 	flex-grow: 1;
 	justify-content: center;
-	margin-top: $tc-list-header-icon-margin-top;
+	padding-top: $tc-list-header-icon-margin-top;
```
jdescottes-talend: Adding some meta-attributes shouldn't change the style.
And btw, the new `padding` declaration use a `margin` variable

packages/components/src/VirtualizedList/HeaderIcon/HeaderIcon.scss
```diff
@@ -4,5 +4,9 @@ $tc-list-header-icon-margin-top: 3px !default;
...
+
+	span {
+		height: 1.6rem;
```
jdescottes-talend: Should not be needed

packages/components/src/VirtualizedList/HeaderIcon/HeaderIcon.component.js
jare-talend: i doesn't change the div by a span. I added a span to get a tooltip on the icon not on the whole div to get the same way that with the header with a string

packages/components/src/VirtualizedList/HeaderIcon/HeaderIcon.scss
jare-talend: the span create a height and a different renderer

packages/components/src/VirtualizedList/HeaderIcon/HeaderIcon.scss
jare-talend: yes i change that

PR #1699, opened 2018-09-14
feat(UIForm): improve tabs widget a11y

jsomsanith:
* Tabs widget use react-bootstrap tabs that doesn't implement all accessibility gestures
* tab with error holds the info only with red color. Need text aternative.

* use components TabBar
* add text alternative to tabs with error

### DIFF
--- a/packages/components/src/TabBar/TabBar.component.js
+++ b/packages/components/src/TabBar/TabBar.component.js
@@ -52,11 +52,12 @@ class TabBar extends React.Component {
 	}
 
 	render() {
-		const { id, items, selectedKey, children, generateChildId } = this.props;
+		const { className, id, items, selectedKey, children, generateChildId } = this.props;
 		return (
 			<Tab.Container
 				id={id}
 				activeKey={selectedKey}
+				className={className}
 				onSelect={this.handleSelect}
 				onKeyDown={this.handleKeyDown}
 				generateChildId={generateChildId}
@@ -78,6 +79,7 @@ class TabBar extends React.Component {
 					<Tab.Content>
 						{items.map(item => (
 							<Tab.Pane eventKey={item.key} key={item.key}>
+								{item.children}
 								{selectedKey === item.key ? children : null}
 							</Tab.Pane>
 						))}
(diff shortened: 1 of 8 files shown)


packages/components/src/TabBar/TabBar.component.js
```diff
@@ -102,7 +105,7 @@ TabBar.propTypes = {
 		}).isRequired,
 	).isRequired,
 	onSelect: PropTypes.func.isRequired,
-	selectedKey: PropTypes.string,
+	selectedKey: PropTypes.any,
```
sgangnant: So the component will work evenly wether with a string/number/object/array/boolean ?

jsomsanith: This is passed to react-bootstrap tabs, which propType for this props is any. I guess they shallow compare them, so any key with the same value/reference will work.

packages/forms/src/UIForm/fieldsets/Tabs/Tabs.component.js
```diff
@@ -1,48 +1,70 @@
...
-		</RBTabs>
-	);
+	onSelect(event, item) {
+		this.setState({ selectedKey: item.key });
+	}
+
+	render() {
+		const { schema, t, ...restProps } = this.props;
+		const tabs = schema.items.map((item, index) => {
+			const tabIsValid = isValid(item, restProps.errors);
```
acateland: im puzzled, how a tab is supposed to be invalid ? What purpose does it serve ?

jsomsanith: It's something I added a while ago.
When you have an error in 1 tab, but don't display it's content, the submit will not be done, because of the error. But there is no indication because it is on a not-displayed input.
The idea is to give a visual indication (red tab) that there is an error in this tab.

And using only color is not good accessibility, so I added a text alternative.

acateland: shouldn't it be handled by https://reactjs.org/docs/react-component.html#componentdidcatch

acateland: not in this PR i mean

jsomsanith: By error I mean form error. If a field is mandatory for example, you press on submit, it will trigger a full validation, then add error messages on the fields.
But if it is not displayed because it's not in the active tab, you won't know that there is an error.
If the tab holds the info too, you will see it.

PR #1702, opened 2018-09-17
feat(List/Large): improve a11y

jsomsanith:
Large display is not accessible. 
- the markup doesn't allow to be understandable by SR
- the navigation is hard, with keyboard we have to go over every element before reaching next element.

Old markup
```
<div role="grid">
	<div role="rowgroup">
		<div class="tc-list-item" role="button">
			<div class="tc-list-large-row inner-box">
				<div class="header" />
				<ul>
					<li/>
				</ul>
			</div>
		</div>
	</div>
</div>
```

- list/listitem roles
- definition list
```diff
-<div role="grid">
+<div role="list">

-	<div role="rowgroup">
-		<div class="tc-list-item" role="button">
+	<div role="listitem" class="tc-list-large-row tc-list-item" tabindex="0">
		<div class=" inner-box">
			<div class="tc-list-large-row">
				<div class="header" />

-				<ul>
-					<li/>
-				</ul>
+				<dl>
+					<dt/>
+					<dd/>
+				</dl>

			</div>
		</div>
	</div>
</div>
```

- gesture to go from item to item (up/down)

(Waiting for #1703)


packages/components/src/VirtualizedList/CellTitle/CellTitleActions.component.js
```diff
@@ -166,6 +166,9 @@ export function CellTitleActionsComponent({
 		<div
 			id={id}
 			className={classNames('main-title-actions-group', theme['main-title-actions-group'])}
+			onKeyDown={e => {
+				e.stopPropagation();
```
jsomsanith: This allows dropdown gesture without interference with the item to item gesture (up/down arrow)

packages/components/src/Gesture/withListGesture.js
```diff
@@ -0,0 +1,84 @@
...
+	}
+}
+
+export default function withListGesture(WrappedComponent) {
+	function ListGesture(props) {
+		return <WrappedComponent {...props} lol={'lol'} onKeyDown={onKeyDown} />;
+	}
+
+	ListGesture.propTypes = {
+		...omit(WrappedComponent.propTypes, 'onKeyDown'),
```
jdescottes-talend: Is it necessary to defined the `propTypes` of the `wrappedComponent` if they are not used in this layer ?

packages/components/src/Gesture/withListGesture.js
```diff
@@ -0,0 +1,84 @@
...
+	if (element) {
+		element.focus();
+	}
+}
+
+function getAllItems(ref) {
+	return ref.closest('[role="list"]').querySelectorAll('[role="listitem"]');
+}
+
+function getNextItem(ref) {
```
jdescottes-talend: What if just use `element.nextElementSibling` and `element.previousElementSibling` ?

packages/components/src/Gesture/withListGesture.js
```diff
@@ -0,0 +1,84 @@
+import React from 'react';
```
jdescottes-talend: I don't understand why this file is in a standalone folder, while it have the `List` semantic

packages/components/src/VirtualizedList/RowLarge/RowLarge.component.js
```diff
@@ -12,55 +20,76 @@ const { LARGE } = listTypes;
...
+				id={id}
+				onClick={onRowClick}
+				onDoubleClick={onRowDoubleClick}
+				onKeyDown={e => onKeyDown(e, this.ref)}
+				style={style}
+				ref={ref => {
+					this.ref = ref;
+				}}
+				role="listitem"
+				tabIndex="0"
```
acateland: this field should be dynamic i think since this component is mapped from a value, anyway since tabIndex have a global incidence on the page, im not sure it will play well with other tabIndex definition outside of this list.

packages/components/src/VirtualizedList/RowLarge/RowLarge.component.js
```diff
@@ -12,55 +20,76 @@ const { LARGE } = listTypes;
...
+				aria-setsize={parent.props.rowCount}
+				aria-label={titleField && getCellData(titleField, parent, index)}
+			>
+				<div className={theme['inner-box']} key="inner-box">
+					<div className={theme.header} key="header">
+						{titleCell}
+						{selectionCell}
+					</div>
+					<dl className={`tc-list-large-content ${theme.content}`} key="content">
+						{otherFields.map((field, fieldIndex) => {
```
acateland: the anonymous function should probably be moved out of the jsx declaration for readability

packages/components/src/Gesture/withListGesture.js
```diff
@@ -0,0 +1,84 @@
...
+			focusOn(event, getPreviousItem(ref));
+			break;
+		default:
+			break;
+	}
+}
+
+export default function withListGesture(WrappedComponent) {
+	function ListGesture(props) {
+		return <WrappedComponent {...props} lol={'lol'} onKeyDown={onKeyDown} />;
```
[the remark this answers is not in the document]
jsomsanith: lol, let's say that yes :P

packages/components/src/Gesture/withListGesture.js
jsomsanith: Good point, I remove it.

packages/components/src/VirtualizedList/ListGrid/ListGrid.component.js
```diff
@@ -41,6 +41,9 @@ function ListGrid(props) {
 			rowCount={collection.length}
 			rowRenderer={enhancedRowRenderer}
 			rowGetter={index => collection[index]}
+			role="group"
+			aria-label="list"
```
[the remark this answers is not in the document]
jsomsanith: It's a default one. It is set, but overridden if user pass an aria-label in props.
It is encouraged to pass it, but not mandatory.

packages/components/src/VirtualizedList/RowLarge/RowLarge.component.js
jsomsanith: I don't think so. `tabindex` 0 just add the elements in the natural html order flow. If used with value > 0, yes it would have a global incidence and this kind of thing if highly discouraged.
Here it's like I used a button element instead.

packages/components/src/VirtualizedList/RowLarge/RowLarge.component.js
jsomsanith: I extract that.

PR #1707, opened 2018-09-19
feat(datagrid): other type render like string

jare-talend:
boolean value are empty in the datagrid

fix that


packages/datagrid/stories/datagrid.component.js
```diff
@@ -4,11 +4,37 @@ import { checkA11y } from '@storybook/addon-a11y';
...
+		case 'DefaultPinHeaderRenderer':
+			return DefaultPinHeaderRenderer;
+		case 'DefaultCellRenderer':
+			return DefaultCellRenderer;
+		case 'DefaultStringCellRenderer':
+			return DefaultRenderer;
+		default:
+			console.error(component);
+	}
+}
```
nicomaligne: This can become a very big switch case in the future if we have exotic need / feature...

jare-talend: it is only for the story.

PR #1709, opened 2018-09-20
feat(generators): add register in index + create settings json

jmfrancois:

When you use the generator in a project and you add a component
You will have to register that component in src/app/component/index.js
and you will maybe add a settings json for it.

add the following questions to the generator of components:

* register into src/app/component/index.js
* add src/settings/MyComponent.json

Bonus:

* Add missing displayName in the existing components
* Remove useless mapStateToProps example
* Remove the useless arrow function component


packages/generator/generators/react-component/index.js
```diff
@@ -106,5 +130,49 @@ module.exports = yeoman.Base.extend({
...
+		if (this.props.parentIndex) {
+			const indexPath = path.join(this.props.path, './index.js');
+			let code;
+			try {
+				code = fs.readFileSync(indexPath);
+			} catch (error) {
+				this.log(error.message);
+			}
+			if (code) {
+				const ast = recast.parse(code);
```
smouillour: Could you change the variable's name to be more explicit?

packages/generator/generators/react-component/index.js
```diff
@@ -55,9 +58,30 @@ module.exports = yeoman.Base.extend({
 			name: 'path',
 			message: 'path',
 			default: 'src/app/components',
+		}, {
+			type: 'input',
+			name: 'settings',
+			message(a) {
```
smouillour: what is `a`?

packages/generator/generators/react-component/index.js
```diff
@@ -106,5 +130,49 @@ module.exports = yeoman.Base.extend({
...
+			try {
+				code = fs.readFileSync(indexPath);
+			} catch (error) {
+				this.log(error.message);
+			}
+			if (code) {
+				const ast = recast.parse(code);
+				const b = recast.types.builders;
+				// first find last import
+				const imports = ast.program.body.filter(item => item.type === 'ImportDeclaration');
```
smouillour: that's don't need any verification on `ast` before use `filter`?

packages/generator/generators/react-component/index.js
```diff
@@ -106,5 +130,49 @@ module.exports = yeoman.Base.extend({
...
+				const lastIndexImport = ast.program.body.indexOf(imports[imports.length - 1]);
+				// inject the new import after the last one
+				ast.program.body.splice(lastIndexImport + 1, 0, b.importDeclaration([
```
smouillour: same for slice?

packages/generator/generators/react-component/index.js
jdescottes-talend: mostly `b` below

packages/generator/generators/react-component/index.js
```diff
@@ -106,5 +130,49 @@ module.exports = yeoman.Base.extend({
...
+				code = fs.readFileSync(indexPath);
+			} catch (error) {
+				this.log(error.message);
+			}
+			if (code) {
+				const ast = recast.parse(code);
+				const b = recast.types.builders;
+				// first find last import
+				const imports = ast.program.body.filter(item => item.type === 'ImportDeclaration');
+				const lastIndexImport = ast.program.body.indexOf(imports[imports.length - 1]);
```
jdescottes-talend: please don't make computation inside function call, extract the `lastImport` in a variable

packages/generator/generators/react-component/index.js
jdescottes-talend: same here, define the new import declaration inside a variable.

packages/generator/generators/react-component/index.js
jmfrancois: parsedIndex ?

PR #1710, opened 2018-09-21
feat(cmf): extract router

jsomsanith-tlnd:
- CMF is too tied to react-router v3.
- There are issues with other routers (ex: react-router v4) that have `<Route>` not directly as `<RouteProvider>` child. Because of react-redux, some components are not re-rendered, which means that `<Route>` under it won't be rerender.

- Extract router part in a dedicated module.
- `cmf.bootstrap` can then configure router with the passed configuration.
- `cmfConnect` manage any `router` settings, making the component render or not depending on the route matching (~ react-router v4)
- we will release 2 major versions of of `@talend/react-cmf-router`, that follows the major versions of `react-router` (v3 / v4)

Router is not managed directly by cmf anymore. Here is a list of previous features that is NOT done anymore:

- cmf middleware that catch and run `config.routerPush`/`config.routerReplace` action
- `UIRouter` component automatically rendered in `App` component. This `UIRouter` was converting route settings into react-router v3 components.

This has been moved to `@talend/react-cmf-router`.
To have the same features again, follow this migration path :

#### 1. Install @talend/react-cmf-router

`yarn add @talend/react-cmf-router@^3.0.0`

#### 2. Change of imports

| Previous imports | New imports | Description |
|---|---|---|
| `import { route } from '@talend/react-cmf'` / `cmf.route` | `import { route } from '@talend/react-cmf-router'` | It's the exact same module except the reoval of deprecated functions. (See next lines).  |
| `cmf.route.getComponentFromRegistry()` | `cmf.component.get()` | Previously deprecated |
| `cmf.route.registerComponent()` | `cmf.component.register()` | Previously deprecated |
| `import { UIRouter } from '@talend/react-cmf'` | `import { UIRouter } from '@talend/react-cmf-router'` | This has been exposed to ensure compatibility, but it was never intended to be used externally. If you do that, please don't, we might remove the export in next releases. |
| `import { history } from '@talend/react-cmf'; history.getStore(store)` | `import router from '@talend/react-cmf-router'; router.getStoreHistory(history, store)` |  |

#### 3. action.cmf.routerPush / action.cmf.routerReplace
Do not use `action.cmf.routerPush` / `action.cmf.routerReplace`. This is not mandatory, but using an abstraction instead of make code tied to the specific router is better for the future.

Before
```
// redux
dispatch({
	type: '...',
	cmf: { routerPush: '/my/new/route' }
});

// redux-saga
yield put({
	type: '...',
	cmf: { routerPush: '/my/new/route' }
});
```

After
```
import { routerActions } from '@talend/react-cmf-router';

// redux
dispatch(routerActions.push('/my/new/route'));

// redux-saga
yield put(routerActions.push('/my/new/route'));
```

#### 4. Inject back the router middleware, reducer, and route provider

Before
```
import React from 'react';
import { render } from 'react-dom';
import { App } from '@talend/react-cmf';
import { AppLoader } from '@talend/react-components';

... // react-cmf configuration

render(
	<App store={store} loading={AppLoader} />,
	document.getElementById('app'),
);
```
(or if you are a good person :) )
```
import React from 'react';
import cmf from '@talend/react-cmf';

cmf.bootstrap({
	..., // bootstrap configuration
});
```

After
```diff
import React from 'react';
import { render } from 'react-dom';
import { App } from '@talend/react-cmf';
+import { bootstrapRouter } from '@talend/react-cmf/lib/bootstrap';
+import routerConfiguration from '@talend/react-cmf-router';
import { AppLoader } from '@talend/react-components';
import { syncHistoryWithStore } from 'react-router-redux';

// this must be before saga and store initialisation
// saga : it needs the history to run sagaRouter
// store : it needs router middleware
+const { history, getStoreHistory, rootComponent } = bootstrapRouter(routerConfiguration);

... // react-cmf configuration considering history

render(
-	<App store={store} loading={AppLoader} />,
+	<App store={store} loading={AppLoader} history={getStoreHistory(history)}>
+		{rootComponent}
+	</App>,
	document.getElementById('app'),
);
```
(or if you are a good person , everything is easier)
```diff
import React from 'react';
import cmf from '@talend/react-cmf';
+import routerConfiguration from '@talend/react-cmf-router';

cmf.bootstrap({
	..., // bootstrap configuration
+	router: routerConfiguration,
});
```


packages/cmf-router/src/defaultRouterConfiguration.js
```diff
@@ -0,0 +1,70 @@
...
+import UIRouter from './UIRouter';
+
+const cmfMiddleware = store => next => action => {
+	const config = action.cmf;
+	if (!config) {
+		return next(action);
+	}
+	if (config.routerPush || config.routerReplace) {
+		let route = config.routerPush || config.routerReplace;
+		if (typeof route === 'function') {
```
acateland: Shouldn't we use cmf mechanism to solve function, instead of allowing to dispatch action containing function ?

packages/cmf-router/src/defaultRouterConfiguration.js
```diff
@@ -0,0 +1,70 @@
...
+}
+
+function getReduxMiddleware(history) {
+	return [cmfMiddleware, routerMiddleware(history)];
+}
+
+function insertReducerFactory() {
+	return rootReducer =>
+		function reducerWithRoute(state, action) {
+			const routerState = state ? state.routing : undefined;
```
acateland: im not sure what this portion of code is supposed to do can you explain it ?

packages/cmf-router/src/route.js
```diff
@@ -33,16 +18,16 @@ function registerFunction(id, func) {
...
 
 /**
  * return a function from the router configuration
  * @param  {string} id
- * @param  {object} contextcmf context
+ * @param  {object} context cmf context
  */
 function getFunction(id, context) {
-	return registry.getFromRegistry(`${CONST.REGISTRY_HOOK_PREFIX}:${id}`, context);
+	return cmf.registry.getFromRegistry(`${REGISTRY_HOOK_PREFIX}:${id}`, context);
```
acateland: I think we should take this opportunity to add documentation as why a cmf-user need to register and resolve function from the route configuration.

nicomaligne: Yep don't see why we have registered func here...
Normally <Route> only make render of a component and give props to it, nop ?

output/cmf.eslint.txt
```diff
@@ -16,5 +19,5 @@ The react/require-extension rule is deprecated. Please use the import/extensions
 /home/travis/build/Talend/ui/packages/cmf/src/sagas/collection.js
   10:1  error  Prefer default export  import/prefer-default-export
 
-✖ 6 problems (3 errors, 3 warnings)
+✖ 7 problems (4 errors, 3 warnings)
```
nicomaligne: Quick fix

packages/cmf-router/src/route.js
jsomsanith: It's to support router hooks (onEnter, onLeave).
That's what is already in place. But you're right, a little bit of documentation can be good.

PR #1711, opened 2018-09-24
fix(cmf): do not parse json if no content

ncomont:

`content-type: json` + `content-length: 0` = KO

Follow the RFC.
https://tools.ietf.org/html/rfc7231#section-3.1.1.5
https://stackoverflow.com/questions/29784398/should-content-type-header-be-present-when-the-message-body-is-empty


packages/cmf/src/sagas/http.js
```diff
@@ -80,7 +80,9 @@ export function handleHttpResponse(response) {
 	if (!testHTTPCode.isSuccess(response.status)) {
 		return Promise.reject(response);
 	}
-	if (response.status === HTTP_STATUS.NO_CONTENT) {
+
+	const length = parseInt(response.headers.get('content-length'), 10);
+	if (response.status === HTTP_STATUS.NO_CONTENT || length === 0) {
```
acateland: shouldn't the promise reject in this case ?
i think it will be more clear for the developer to know that something probably unexpected failed.

sgangnant: `testHTTPCode` earlier will catch any non-2xx responses and return a rejected promise. I'm not sure every 2xx will have a body's response (maybe only `204` is _supposed_ to but in some real use cases it can be different).

frassinier:
We will change backend part in order to return `204 NO-CONTENT`

PR #1714, opened 2018-09-25
fix(TUI-233): ensure conditional rendering with array indices resolution

rmannibucau:

When the back sends a widget condition for an array which is related to the array we were used to use the [] syntax (ex: names[]). The front was making it relative to "current" - from schema.key - indices recursively.

But with the use of jsonlogic, this is not the case anymore, array indices are not resolved, the condition is always false.

Remap the condition "var" values before evaluating the condition, replacing `[]` by the current index.

Side note: adding a custom operation in jsonlogic registry is not that trivial since we don't have the schema in the parameters and we can't have it.


packages/forms/src/UIForm/utils/condition.js
```diff
@@ -16,18 +17,84 @@ function toNumber(a) {
...
+ */
+function replaceArrayNotationByIndexes(path, key) {
+	if (!path || !path.includes('[]')) {
+		return path;
+	}
+	return path
+		.split(/\.|\[/)
+		.map((part, index) => {
+			if (part === ']') {
+				return key[index];
```
jmfrancois: note: https://github.com/jwadhams/json-logic-js/blob/master/logic.js#L134
json logic work that way foo.0.bar

PR #1719, opened 2018-09-28
update(collapsiblePanel/TPSVC-6044): update by frontify guideline

inna-i:
https://jira.talendforge.org/browse/TPSVC-6044
http://guidelines.talend.com/document/92132#/navigation-layout/expandable-collapsible-panels

- onToggle event is bind to the Panel
- new type **TYPE_CUSTOM** is added, to have ability to use custom element
- onHover styles are updated
- props **tooltipLabel** is added (tooltip text can include some tips)


packages/components/src/CollapsiblePanel/CollapsiblePanel.component.js
```diff
@@ -109,13 +120,25 @@ function CollapsiblePanelHeader(props) {
...
 					{elements}
 				</div>
 			);
 		}
 		return (
-			<div key={index} className={classNames(css[headerItem.className], css[headerColumnClass])}>
+			<div
+				key={index}
+				className={classNames(css[headerItem.className], css[headerColumnClass])}
+				onClick={onToggle}
```
sgangnant: This (those) onClick are what trigger linting errors.
I've encountered this issue before, don't know yet how to solve them and it's kind of annoying ... `Talend/UI` Action maybe ?

inna-i: I agree with. Actually _div_ can be replaced with _button_ as an option

packages/components/src/CollapsiblePanel/CollapsiblePanel.component.js
```diff
@@ -109,13 +120,25 @@ function CollapsiblePanelHeader(props) {
 		if (Array.isArray(headerItem)) {
 			const elements = headerItem.map(renderHeaderItem);
 			return (
-				<div key={index} className={classNames(css.group, css[headerColumnClass])}>
+				<div
+					key={index}
+					className={classNames(css.group, css[headerColumnClass])}
+					onClick={onToggle}
```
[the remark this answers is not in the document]
inna-i: @frassinier maybe you know how to hadle lint error *no-noninteractive-element-interactions*?

packages/components/src/CollapsiblePanel/CollapsiblePanel.component.js
```diff
@@ -106,18 +117,22 @@ function CollapsiblePanelHeader(props) {
...
-			<div key={index} className={classNames(css[headerItem.className], css[headerColumnClass])}>
-				{renderHeaderItem(headerItem)}
-			</div>
+			/* eslint-disable jsx-a11y/no-static-element-interactions */
+			<div
+				key={index}
+				className={classNames(css.group, css[headerColumnClass])}
+				onClick={onToggle}
+				onKeyPress={onToggle}
+				role="button"
```
jsomsanith: Why don't you turn the header div into a button (role=button) instead of all elements in the header ?
Plus, the user doesn't know what the buttons do.

My proposal : 
- the header div  manage the toggle, instead of every element in it
- the header should not be focusable, or even have button semantic, so SR doesn't consider it a button. Instead, to toggle the element, there is the caret, which is a button with a clear text alternative.

What do you think @frassinier ?

jsomsanith: BTW I see a mistake in a11y. 
Can you add 
- an `aria-expanded` attribute on the toggle button to indicate that it is expanded or not  
- aria-controls="panel-id" to indicates that it controls the targeted panel

packages/components/src/CollapsiblePanel/CollapsiblePanel.scss
```diff
@@ -72,27 +84,45 @@ $tc-toggle-color: #555964 !default;
 	&.muted,
 	&.warning {
 		padding-left: 1px;
-		border-left: 4px solid;
+		border-left: 5px solid;
+		:global(.tc-status-label) {
+			font-weight: 600;
```
jsomsanith: I don't see anything in the guideline and the result seem bolder. 
Moreover, using 600 (not one of the font-weight that is natively downloaded) cans introduce some anti-aliasing issue in some OS/Browser.
Can you remove that and let the normal (400) weight ?

packages/components/src/CollapsiblePanel/__snapshots__/CollapsiblePanel.snapshot.test.js.snap
```diff
@@ -529,6 +536,62 @@ exports[`CollapsiblePanel should render default without content 1`] = `
...
+            onMouseOver={[Function]}
+          >
+            Version 1 Version 1
+          </span>
+          <span
+            className="theme-undefined"
+            onFocus={[Function]}
+            onMouseOver={[Function]}
+          >
+            Custom
```
jsomsanith: Hmm I don't really understand. In you test, the custom element is `<span>Custom element</span>` but the rendered one is not that

packages/components/src/CollapsiblePanel/CollapsiblePanel.component.js
```diff
@@ -126,14 +133,24 @@ function CollapsiblePanelHeader(props) {
 		onSelect ? (
 			<Button
 				className={classNames(css['panel-title'], 'panel-title')}
+				blah="blah"
```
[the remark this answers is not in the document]
inna-i: oopps :)

packages/components/src/CollapsiblePanel/CollapsiblePanel.scss
inna-i: @jsomsanith do we have a *font*, that can be used instead of adding font-weight?

packages/components/src/CollapsiblePanel/CollapsiblePanel.component.js
```diff
@@ -126,14 +133,24 @@ function CollapsiblePanelHeader(props) {
...
-			<div className={classNames(css['panel-title'], 'panel-title')} key={1}>
+			/* eslint-disable jsx-a11y/no-static-element-interactions */
+			<div
+				aria-controls={id}
+				className={classNames(css['panel-title'], 'panel-title')}
+				key={1}
+				onClick={onToggle}
+				onKeyPress={onToggle}
+				role="button"
+				tabIndex="-1"
```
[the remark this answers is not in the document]
inna-i: @frassinier, we need your help :point_up_2:

packages/components/src/CollapsiblePanel/CollapsiblePanel.component.js
```diff
@@ -70,20 +72,20 @@ function renderHeaderItem({ displayMode, className, ...headerItem }, key) {
...
 			);
 		}
 		default: {
-			const { label, tooltipPlacement } = headerItem;
+			const { element, label, tooltipLabel, tooltipPlacement } = headerItem;
 			return (
-				<TooltipTrigger key={key} label={label} tooltipPlacement={tooltipPlacement}>
-					<span className={css[className]}>{label}</span>
+				<TooltipTrigger key={key} label={tooltipLabel || label} tooltipPlacement={tooltipPlacement}>
+					<React.Fragment>{element || label}</React.Fragment>
```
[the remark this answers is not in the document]
inna-i: @sgangnant what do you think about removing this *element* property and just change PropTypes of *label* - can be element or string?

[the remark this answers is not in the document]
inna-i: or actually that `<span>` can be replaced with `<div>`, to have ability inject some custom block element and don't get new warnings in the console

packages/components/src/CollapsiblePanel/CollapsiblePanel.component.js
```diff
@@ -112,8 +112,12 @@ function CollapsiblePanelHeader(props) {
 			? headerItem.map(renderHeaderItem)
 			: renderHeaderItem(headerItem);
 
+		const selectors = Array.isArray(headerItem)
```
sgangnant: Array.isArray() statement result could be stored in a boolean const to avoid duplicate calls (few lines above in the file)

packages/components/src/CollapsiblePanel/CollapsiblePanel.scss
jsomsanith: In summary, we provide some fonts, and their variations (weight, italic, ...) in some versions only. (Example 400, 700, 900).
If we set a font-weight that is not provided natively, the browser will try to interpolate the font rendering. Depending on the OS and browser, the result can be good (OSX/safari) or bad (Windows/IE).
But either way that's not in the guideline (and does not match the guideline screenshot neither), please remove it.

PR #1721, opened 2018-09-28
fix(Container) HeaderBar API remote response handling

sgangnant:
Backend API has been updated and HeaderBar container no longer handles properly its response.
Also, language parameter is actually not needed.

Update mock with final backend response content and adapt the component.

### DIFF
--- a/packages/containers/sandbox/mockBackend/mock/header-bar/products-list.json
+++ b/packages/containers/sandbox/mockBackend/mock/header-bar/products-list.json
@@ -1,17 +1,26 @@
 [
 	{
-		"icon": "talend-tdp-colored",
-		"uri": "http://www.talend.com/products#tdp",
-		"label": "My"
+		"id": "TMC",
+		"name": "Management Console",
+		"url": "https://www.talend.com/products#tmc",
+		"icon": "tmc"
 	},
 	{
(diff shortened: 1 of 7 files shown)


jsomsanith:
This is a common backend service. It means that all projects are impacted.
This is a breaking change BTW. So
1- you need to fill properly the breaking change part in description and wiki.
2- you need to see with everyone (backend and frontend) how and when to migrate that. Everyone should migrate in the same time, or its corresponding frontend projet will be blocked in Talend/ui version.

sgangnant:
No one has implemented it since backend was not ready, those changes only reflect the new/first version of a functionning backend.

jsomsanith:
Ok, if it's the first version of the backend. But this is still a breaking change. Every project has their app switcher mock that follows the last api.
Please fill the breaking change log

sgangnant:
@jsomsanith I don't think anyone has implemented it yet, but fair enough.

jsomsanith:
Every project has the app switcher, with their way to handle the list of apps. They adapt this list to the format that header bar wants.
And it's still a breaking change

sgangnant:
Sure, but here I change the `container`, not the `component` (which they most likely use)

PR #1724, opened 2018-10-01
feat(TPSVC-6043): collapsible panels virtualized list

andrii-holovko-dev:
Create common component for collapsible panels virtualized list

Use VirtualizedList to render CollapsiblePanels

* [ ] Tests for the changes have been added (for bug fixes / features) And [non reg](./screenshots.md)


packages/components/src/VirtualizedList/RowCollapsiblePanel/RowCollapsiblePanel.component.js
```diff
@@ -0,0 +1,90 @@
...
+		let onRowDoubleClick;
+
+		if (parent.props.onRowDoubleClick) {
+			onRowDoubleClick = event => parent.props.onRowDoubleClick({ event, rowData });
+		}
+
+		const onToggle = (event, measure) => {
+			if (parent.props.onRowClick) {
+				parent.props.onRowClick({ event, rowData: { ...rowData, index } });
+				setTimeout(measure, 0);
```
jsomsanith: I don't understand the defer here.
But that doesn't work if you already have expanded panels as initial props.
You can instead use onLoad attribute on the div to execute measure()

packages/components/src/VirtualizedList/RowCollapsiblePanel/RowCollapsiblePanel.component.js
```diff
@@ -0,0 +1,90 @@
...
+				columnIndex={0}
+				key={index}
+				parent={parent}
+				rowIndex={index}
+			>
+				{({ measure }) => (
+					<div
+						className={classNames(css['tc-collapsible-row'], rowData.className, className)}
+						id={id}
+						onDoubleClick={onRowDoubleClick}
```
jsomsanith: Why do you set a double clic gesture ?

packages/components/src/VirtualizedList/RowCollapsiblePanel/RowCollapsiblePanel.component.js
```diff
@@ -0,0 +1,90 @@
...
+						onKeyDown={e => onKeyDown(e, this.ref)}
+						ref={ref => {
+							this.ref = ref;
+						}}
+						role="listitem"
+						tabIndex="0"
+						aria-posinset={index + 1}
+						aria-setsize={parent.props.rowCount}
+						aria-label={titleField && getCellData(titleField, parent, index)}
```
jsomsanith: This is always undefined. You don't pass any column definition so no title column, so the `extractSpecialFields` doesn't work here.
You need to use the panel info to get the title.

packages/components/src/VirtualizedList/RowCollapsiblePanel/RowCollapsiblePanel.component.js
```diff
@@ -0,0 +1,90 @@
...
+						style={{ ...style, maxHeight: rowData.expanded ? 'none' : `${DEFAULT_HEIGHT}px` }}
```
jsomsanith: Why `maxHeight ` is redefined here ? Shouldn't you let CellMeasurer do the workd to set the height ?

packages/components/src/VirtualizedList/RowCollapsiblePanel/__snapshots__/RowCollapsiblePanel.test.js.snap
```diff
@@ -0,0 +1,91 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`RowCollapsiblePanel should render collapsible panel row 1`] = `
+<VirtualizedList(RowCollapsiblePanel)
```
jsomsanith: Your snapshot isn't meaningful. It prints your component with its props, but that doesn't test your component.

packages/components/src/VirtualizedList/RendererSelector.component.js
```diff
@@ -94,6 +99,14 @@ class RendererSelector extends React.Component {
 		} else {
 			ListRenderer = ListGrid;
 			customProps = { rowRenderer: getRowRenderer(type, this.props.rowRenderers) };
+
+			if (type === COLLAPSIBLE_PANEL) {
```
jsomsanith: Please avoid specific code in the RendererSeletor. If you need to pass options, perhaps you can :
- consider that the rowRenderer can hold an attribute `options`
- so 
```
ListRenderer = ListGrid;
const rowRenderer = getRowRenderer(type, this.props.rowRenderers);
customProps = { rowRenderer, ...rowRenderer.options };
```
- in RowCollapsiblePanel/index.js
```
const cache = new CellMeasurerCache({ fixedWidth: true });
RowCollapsiblePanel.options = {
	deferredMeasurementCache: cache,
	rowHeight: cache.rowHeight,
};
export default RowCollapsiblePanel;
```

packages/components/src/index.js
```diff
@@ -155,7 +155,6 @@ export {
 	HttpError,
 	Icon,
 	IconsProvider,
-	Image,
```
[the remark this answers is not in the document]
andrii-holovko-dev: Image is not imported here

packages/components/src/CollapsiblePanel/CollapsiblePanel.scss
```diff
@@ -271,3 +271,7 @@ $tc-toggle-color: #555964 !default;
 		}
 	}
 }
+
+.panel-list {
```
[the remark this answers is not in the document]
andrii-holovko-dev: I pass this class name as "theme" property of CollapsiblePanel. It is used to make CollapsiblePanel inside list fit the guideline

packages/components/stories/VirtualizedList.js
```diff
@@ -710,4 +774,41 @@ storiesOf('Virtualized List', module)
 				</VirtualizedList>
 			</section>
 		</div>
-	));
+	))
+	.add('List > Debug Virtualized Collapsible list', () => {
```
jmfrancois: please remove this example, seems you have added to debug.

packages/components/src/CollapsiblePanel/CollapsiblePanel.scss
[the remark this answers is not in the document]
jmfrancois: This is not the way to fit the guideline.

the RowCollapsiblePanel component already customize the CollapsiblePanel.

So please move the margin-bottom there.

packages/components/src/VirtualizedList/RowCollapsiblePanel/RowCollapsiblePanel.component.js
```diff
@@ -0,0 +1,90 @@
...
+				cache={options.deferredMeasurementCache}
+				columnIndex={0}
+				key={this.props.index}
+				parent={this.props.parent}
+				rowIndex={index}
+			>
+				{({ measure }) => (
+					// eslint-disable-next-line jsx-a11y/no-static-element-interactions
+					<div
+						className={classNames(css['tc-collapsible-row'], rowData.className, className)}
```
jmfrancois: So I would add it here (the global class) for customization: `tc-collapsible-row`

packages/components/src/VirtualizedList/RowCollapsiblePanel/RowCollapsiblePanel.scss
```diff
@@ -0,0 +1,9 @@
+.tc-collapsible-row {
+	:global(.panel-collapse) {
+		transition-duration: 0s;
+	}
```
jmfrancois: You can add your margin 10 px here.:
```
:global(.panel) {
  margin-bottom: $padding-small;
}
```

PR #1731, opened 2018-10-02
feat(Components/Slider): Implement Action Icon

nicomaligne:
We want the icon on sliderbar to be clickable, to be an action.
Example [here](http://talend.surge.sh/components/?selectedKind=Slider&selectedStory=default&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel)
Add a new props actions that will render specific action if needed. You can still render simple icons.


packages/components/src/Slider/Slider.component.js
```diff
@@ -33,6 +34,31 @@ export function getSelectedIconPosition(icons, value, min, max) {
...
 }
 
+/**
+ * This function allow to get the actions components
+ * @param {array} actions
+ * @param {number} value
+ * @param {number} min
+ * @param {number} max
+ */
+export function getActions(actions, value, min, max) {
```
acateland: Wouldn’t renderActions be a better name ?

jmfrancois:
Why icons are changed in the PR ?

packages/components/src/Slider/Slider.component.js
jare-talend: The component has already a event onChange. We have here an implementation very coupled with CMF. 

When we click on an icon, he should just trigger an event onChange with the value.

nicomaligne: Erf didn't know the icon can render a button... 
Love the if that changes the rendering of a component... is so handy.
So yeah you're right for the implem, back to work

nicomaligne: Hmmm another problem is that we have less posibility with the icon than the action. No tooltip for example.

PR #1732, opened 2018-10-02
fix(ComponentForm): extract complex items in trigger parameters

jsomsanith:
Component trigger extract parameters to pass to the trigger call, based on param keys.
This extraction includes only native parameters, but objects and arrays are not included. Keys that point to them are associated to undefined.

Include complex items in parameters extraction

### DIFF
--- a/packages/containers/src/ComponentForm/kit/flatten.js
+++ b/packages/containers/src/ComponentForm/kit/flatten.js
@@ -28,13 +28,13 @@ function step(anything, key, payload) {
 		if (Array.isArray(anything)) {
 			anything.forEach((item, index) => {
 				const itemKey = `[${index}]`;
-				step(item, itemKey, subPayload, false);
+				step(item, itemKey, subPayload);
 			});
 		} else {
 			Object.keys(anything).forEach(nextKey => {
-				const itemKey = key ? `.${nextKey}` : nextKey;
+				const itemKey = `.${nextKey}`;
 				const item = anything[nextKey];
-				step(item, itemKey, subPayload, false);
+				step(item, itemKey, subPayload);
 			});
 		}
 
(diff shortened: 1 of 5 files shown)


packages/containers/src/ComponentForm/kit/flatten.js
```diff
@@ -13,42 +14,50 @@
...
+	} else {
+		// for objects (including arrays), we flatten it. The we store it in 2 forms.
+		// Let's say that we have a key 'my-object'.
+		// - Form 1: { 'my-object' : {<flattenedObject>} }
+		// - Form 2: { 'my-object.key1': value1, my-object.key2: value2, ... }
+		const subPayload = {};
+		if (Array.isArray(anything)) {
+			anything.forEach((item, index) => {
+				const itemKey = `[${index}]`;
+				step(item, itemKey, subPayload, false);
```
jmfrancois: the last argument doesn t exists in the function definition, please remove

packages/containers/src/ComponentForm/kit/flatten.js
```diff
@@ -13,42 +14,50 @@
...
-		} else if (typeof anything === 'object') {
-			Object.keys(anything).forEach(nextKey => {
-				const itemKey = key ? `${key}.${nextKey}` : nextKey;
-				const item = anything[nextKey];
-				step(item, itemKey);
-			});
-		}
-	}
-	step(obj, '');
+	step(obj, '', payload);
```
jmfrancois: this really looks like a reducer, you take obj to create payload.
It would be more comprehensive to do Sth like Object.keys(obj).reduce here.
And it would remove that key='' in the first call that I don t like and don t see well how it s handled

packages/containers/src/ComponentForm/kit/flatten.js
jsomsanith: Oh right, I forgot to remove it after the Nth solution :D

PR #1722, opened 2018-09-28
fix(components): tabBar with no children

frassinier:
**What is the problem this PR is trying to solve?**
TabBar with no children produces extra div even if you pass ids of existing tab contents

**What is the chosen solution to this problem?**
Hide this extra div if tabBar content already exists

**Please check if the PR fulfills these requirements**

* [x] The PR commit message follows our [guidelines](https://github.com/talend/tools/blob/master/tools-root-github/CONTRIBUTING.md)
* [x] Tests for the changes have been added (for bug fixes / features) And [non reg](./screenshots.md) done before need review
* [ ] Docs have been added / updated (for bug fixes / features)
* [ ] Related design / discussions / pages (not in jira), if any, are all linked or available in the PR

**[ ] This PR introduces a breaking change**

PR #1738, opened 2018-10-04
fix(TabBar): children from item definition are displayed again

jsomsanith:
#1722 introduced a regression on TabBar.
There are 3 ways to manage children in TabBar : 
- user pass children props, it is displayed everytime
- user pass children in items definition, they are set in tab panels. react-bootstrap manages the display
- no children at all, user manage the display under the tabs

The second is broken, because we test if there is a children prop.

Check if we have children attribute in items to see if we display the tab panel

### DIFF
--- a/packages/components/src/TabBar/TabBar.component.js
+++ b/packages/components/src/TabBar/TabBar.component.js
@@ -77,7 +77,7 @@ class TabBar extends React.Component {
 							</NavItem>
 						))}
 					</Nav>
-					{hasChildren ? (
+					{hasChildren && (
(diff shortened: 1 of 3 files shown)


packages/components/src/TabBar/TabBar.component.js
```diff
@@ -76,7 +77,7 @@ class TabBar extends React.Component {
 							</NavItem>
 						))}
 					</Nav>
-					{children && (
+					{hasChildren ? (
```
jmfrancois: please use hasChildren &&

PR #1749, opened 2018-10-09
feat(cmf/TUI-256): add more control on injected props

jmfrancois:

We have tons of props warning from react 16.
The problems:

1. When you cmfConnect a component the original component may need some tools from CMF not all.
For example you can take the layout, you will just need `getComponent`.

2. When you create a container you want to be able to remove some props from `this.props` before spread them to the pure component

3. When react-router inject a component it adds many props to it the component may don t know.

All that create a noise in the console.error which becomes aweful.

Add API to cmfConnect to let you choose what tools you want to be added by props.

`withState` will add you the props.state props.setState ...

This solution (kind of whitelist) is friendly with the fact that we want to split cmfConnect implementation in multiple piece of code to compose them later.

This will be a breaking change in 2.0.0 where the with params will be required.


packages/cmf/src/cmfConnect.js
```diff
@@ -190,8 +190,31 @@ export default function cmfConnect({
 	mapStateToProps,
 	mapDispatchToProps,
 	mergeProps,
+	omitCMFProps = false, // will be removed and considered to true in 2.0.0
```
jsomsanith: Not sure if we should define this property.
Your wrote a codemod to adapt the code. Let's consider that we ask for no cmf props, and only use the detailed one. The codemod will ensure that the existing code works like today

PR #1750, opened 2018-10-09
chore(components/Inject.getReactElement): remove key warning

jmfrancois:

When you use the Inject API with an array React complains to not have keys on it

Add key when use array. (build it using component/componentId)

### DIFF
--- a/packages/components/src/Inject/Inject.component.js
+++ b/packages/components/src/Inject/Inject.component.js
@@ -124,7 +124,7 @@ Inject.getReactElement = function getReactElement(
 	} else if (data === null) {
 		return data;
 	} else if (typeof data === 'string') {
-		const props = {getComponent, component: data };
+		const props = { getComponent, component: data };
 		if (withKey) {
 			props.key = `${data}#default`;
 		}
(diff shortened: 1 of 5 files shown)


packages/forms/src/UIForm/fields/CheckBox/SimpleCheckBox.component.js
```diff
@@ -43,7 +43,7 @@ if (process.env.NODE_ENV !== 'production') {
 	SimpleCheckBox.propTypes = {
 		describedby: PropTypes.string.isRequired,
 		id: PropTypes.string,
-		isValid: PropTypes.string,
+		isValid: PropTypes.bool,
```
jmfrancois: bonus: isValid is bool everywhere in UIForm except here

packages/components/src/Inject/Inject.component.js
```diff
@@ -113,17 +113,34 @@ Inject.getAll = function injectGetAll(getComponent, config) {
...
+		return data.map(info => getReactElement(getComponent, info, CustomInject, true));
 	} else if (data === null) {
 		return data;
 	} else if (typeof data === 'string') {
-		return <CustomInject getComponent={getComponent} component={data} />;
+		return (
+			<CustomInject
+				getComponent={getComponent}
+				component={data}
+				key={withKey && `${data}#default`}
```
sgendre: if `withKey` is false, `key` value will be `undefined` which may happen on multiple components and React will complain about having sibling elements with the same `undefined` `key`, it's maybe better to not define this attritute at all when `withKey` is false

PR #1766, opened 2018-10-17
feat(TUI-231/TUI-90/SidePanel): make the container work with link

jmfrancois:

Use the SidePanel container encourage to use deprecated actionIds and do not handle well the selected props and the current route matching.

use `actions` props from the SidePanel.component and parse it to find href
if `href` is found, use it to compare it with the currentRoute.

Cache both mapStateToProps to avoid re-render on state mutation.

So now your props can and should look like this:

```javascript
{
  "props": {
    "SidePanel#default": {
       actions: [
	{
		componentId: 'first',
		href: '/storybook',
	},
	{
		componentId: 'second',
		href: '/foo',
	},
	{
		componentId: 'configuration',
		href: '/configuration',
	},
      ]
    }
  }
}
```


packages/containers/package.json
```diff
@@ -78,6 +78,7 @@
     "invariant": "^2.2.2",
     "jest": "20.0.3",
     "jest-cli": "20.0.3",
+    "jest-in-case": "^1.0.2",
```
acateland: You don’t need jest in case anymore jest support the same features

packages/containers/src/SidePanel/SidePanel.test.js
```diff
@@ -19,42 +22,214 @@ describe('SidePanel', () => {
...
 
+		state.routing.locationBeforeTransitions.pathname = '/href';
 		const href = mapStateToProps(state, {
-			location: { pathname: '/href' },
 			actionIds: ['menu:href'],
 		});
 		expect(href.actions[0].active).toBe(true);
 	});
 
 	xit('should handle actionCreator with href', () => {
```
ncomont: why is it disabled ?

packages/containers/src/SidePanel/SidePanel.connect.js
```diff
@@ -1,39 +1,93 @@
...
+				if (!route) {
+					route = get(info, 'payload.cmf.routerPush');
 				}
-			}
-			return action;
-		});
+				if (!route) {
+					route = get(info, 'href');
+				}
+				if (route) {
```
ncomont: ?

packages/containers/src/SidePanel/SidePanel.connect.js
```diff
@@ -1,39 +1,93 @@
...
+					if (currentRoute.indexOf(route) !== -1) {
+						action.active = true;
+					}
```
ncomont: You could use `includes()` :

`if (currentRoute.includes(route)` or `action.active = currentRoute.includes(route);`

packages/containers/src/SidePanel/SidePanel.test.js
jmfrancois: It was already, the case was not covered in the code.

packages/containers/src/SidePanel/SidePanel.connect.js
```diff
@@ -1,39 +1,93 @@
...
+	const fullPathSplit = splitPath(fullPath);
+
+	return zip(pathToCheckSplit, fullPathSplit)
+		.filter(([pathToCheckPart]) => pathToCheckPart !== undefined)
+		.every(([pathToCheckPart, fullPathPart]) => pathToCheckPart === fullPathPart);
+}
+
+const getActionsWrapped = actions =>
+	actions.map(action => {
+		if (!action.href) {
```
ncomont: could you remove this if please ?

output/containers.eslint.txt
```diff
@@ -15,11 +15,14 @@ The react/require-extension rule is deprecated. Please use the import/extensions
 /home/travis/build/Talend/ui/packages/containers/src/register.js
   6:1  error  Prefer default export  import/prefer-default-export
 
-/home/travis/build/Talend/ui/packages/containers/src/SidePanel/SidePanel.connect.js
-  3:1  error  'react-router' should be listed in the project's dependencies, not devDependencies  import/no-extraneous-dependencies
+/home/travis/build/Talend/ui/packages/containers/src/SidePanel/constants.js
```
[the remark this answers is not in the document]
jmfrancois: https://github.com/Talend/tools/pull/74

packages/containers/src/SidePanel/SidePanel.connect.js
jmfrancois: I have finally moved to the new check which work better

packages/containers/src/SidePanel/SidePanel.connect.js
```diff
@@ -1,39 +1,93 @@
...
+}
+
+function isBasePathOf(pathToCheck, fullPath) {
+	const pathToCheckSplit = splitPath(pathToCheck);
+	const fullPathSplit = splitPath(fullPath);
+
+	return zip(pathToCheckSplit, fullPathSplit)
+		.filter(([pathToCheckPart]) => pathToCheckPart !== undefined)
+		.every(([pathToCheckPart, fullPathPart]) => pathToCheckPart === fullPathPart);
+}
```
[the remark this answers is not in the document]
jmfrancois: Yes for sure you can take a look in the tests the main idea is to compare slug by slug.
So `/foo/hello` is different from `/foo/hell` which is not the case with a simple includes or indexOf.
So here we split on `/` then we compare each slug.

packages/cmf/src/onEvent.js
```diff
@@ -100,7 +100,7 @@ function addOnEventSupport(handlerType, instance, props, key) {
 		}
 		props.toOmit.push(key);
 		const handlerKey = key.replace(CONSTANT[`IS_HANDLER_${handlerType}`], '');
-		const originalEventHandler = props[handlerKey];
+		const originalEventHandler = props[handlerKey] || instance.props[handlerKey] || undefined;
```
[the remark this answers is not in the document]
jmfrancois: oups

jsomsanith:
There is an issue with `<a>` and onclick handler. Clicking on a with href will triggers the browser native behavior. Even if there is an onClick, I think it's react that doesn't prevent the default.

packages/containers/src/SidePanel/SidePanel.test.js
```diff
@@ -19,42 +22,226 @@ describe('SidePanel', () => {
...
+				{
+					name:
+						'Subset path starting at the beginning but without the exact last segment (currentRoute shorter)',
+					currentRoute: '/a/usual/pat',
+					itemRoute: '/a/usual/path',
+					isMatching: false,
+				},
+				{
+					name:
+						'Subset path starting at the beginning but without the last item path segment matching (currentRoute with more segments)',
```
jsomsanith: I don't understand the difference with the one testing `'/a/usual/path'` with `'/a/usual/pat'`

packages/containers/src/SidePanel/SidePanel.test.js
```diff
@@ -19,42 +22,226 @@ describe('SidePanel', () => {
...
+					isMatching: false,
+				},
+				{
+					name: 'Subset path but not starting at the beginning',
+					currentRoute: '/a/usual/path',
+					itemRoute: '/usual/path',
+					isMatching: false,
+				},
+				{
+					name: 'Just a total other path',
```
jsomsanith: I don't see the difference with the previous one

packages/containers/src/SidePanel/SidePanel.test.js
```diff
@@ -19,42 +22,226 @@ describe('SidePanel', () => {
...
+					currentRoute: '/a/usual/path',
+					itemRoute: '/this/path/has/nothing/in/common',
+					isMatching: false,
+				},
+				{
+					name: 'More specific path than the current route',
```
jsomsanith: Same here. In a generic way, the feature is that it should match url that starts with the one we set. we should test 3 cases : 
* it starts with
* it doesn't starts with
* edge case : it starts with but the last word doesn't match

packages/containers/src/SidePanel/SidePanel.connect.js
```diff
@@ -1,39 +1,112 @@
...
+		}
+		return action;
+	});
+}
+
+function getSelectedAction(currentRoute, actions) {
+	return actions.find(action => action.href && isBasePathOf(action.href, currentRoute));
+}
+
+function getAction(id, currentRoute, state) {
```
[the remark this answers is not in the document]
jmfrancois: This would be a breaking change -> no

packages/containers/src/SidePanel/SidePanel.connect.js
```diff
@@ -1,39 +1,112 @@
...
+		},
+		id,
+	);
+	action.label = info.label;
+	action.id = info.id;
+
+	const route = get(
+		info,
+		'payload.cmf.routerReplace',
+		get(info, 'payload.cmf.routerPush', get(info, 'href')),
```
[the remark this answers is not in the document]
jmfrancois: Yes, but at the moment this task means extract the router first.
Then we will be able to work on containers for that.
This will not change the href implementation, we will just be able to get ride of that later

packages/containers/src/SidePanel/SidePanel.connect.js
```diff
@@ -1,39 +1,112 @@
...
+			currentRoute,
+		);
+	} else if (ownProps.actions) {
+		props.actions = memoizeOne(
+			() => getActionsWrapped(ownProps.actions),
+			componentId,
+			ownProps.actions,
+			currentRoute,
+		);
+		props.selected = getSelectedAction(currentRoute, props.actions);
```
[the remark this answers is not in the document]
jmfrancois: no it would be a breaking change.
The plan here is to make app migrate to the href to remove this code which is in the app at the moment. CF the jira ticket

packages/containers/src/SidePanel/SidePanel.connect.js
```diff
@@ -1,39 +1,112 @@
...
+import { ACTION_TYPE_LINK } from './constants';
+
+const cache = {};
+
+/**
+ * Memoize per component on a key the result of the fn call
+ * @param {function} fn
+ * @param {string} componentId,
+ */
+function memoizeOne(fn, componentId = 'default', actions, currentRoute) {
```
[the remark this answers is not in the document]
jmfrancois: impossible, we want one per component id.
But I can rename the function which do not do any memoize for sure

packages/containers/src/SidePanel/SidePanel.test.js
```diff
@@ -19,42 +22,201 @@ describe('SidePanel', () => {
...
+			},
+			[
+				{
+					name: 'Usual simple path',
+					currentRoute: '/a/usual/path',
+					itemRoute: '/a/usual/path',
+					isMatching: true,
+				},
+				{
+					name: 'Usual simple path with trainling /',
```
jsomsanith: `trailing`

packages/containers/src/SidePanel/SidePanel.test.js
```diff
@@ -19,42 +22,201 @@ describe('SidePanel', () => {
...
+					currentRoute: '/a/usual/path/',
+					itemRoute: '/a/usual/path',
+					isMatching: true,
+				},
+				{
+					name: 'Usual simple path with trainling / on match',
```
jsomsanith: `trailing`

PR #1769, opened 2018-10-17
fix(datagrid): null value in cell

jare-talend:
when the value is null, null is display in the cell

detect null and replace by empty value


packages/datagrid/src/components/DefaultCellRenderer/DefaultValueRenderer.component.js
```diff
@@ -39,7 +39,14 @@ export default class DefaultValueRenderer extends React.Component {
 	}
 
 	render() {
-		const stringValue = String(this.props.value);
+		let stringValue;
+
+		if (this.props.value === null) {
```
romainseb: How about
```suggestion
		if (!this.props.value) {
```
To handle weird cases like undefined ?

PR #1771, opened 2018-10-17
feat(icon): update sharing-owner icon

AnLefevre:
Last version of the icon wasn't correct. Please update it by this one. The only one to change is sharing-owner.
Last PR #1735 


packages/components/src/IconsProvider/__snapshots__/IconsProvider.test.js.snap
```diff
@@ -6154,7 +6155,8 @@ exports[`IconsProvider should render default talend-icons and custom icons defin
...
+          fill="#010202"
```
acateland: all other icons use black ass default fill are we sure we want to use this ?

AnLefevre: you're right i don't know why the color turns to #010202 instead of #202020. But it's not very impacting our future design because it will use $Dove gray and $Brand primary in the design

PR #1774, opened 2018-10-18
fix(components/DatePicker) Widget error handling with UIForm - poc2

jdescottes-talend:
Avoid loss of synchronicity between form property value and its `errorMessage`.
Having a not too bad way to handle errors. To just have at least a way to do it

Let spread all values not handled in the conversion, and have the form validation handle it.
Let the custom uiSchema `validationMessage` override all the error messages to avoid displaying an understandable message to the user.

### DIFF
--- a/packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.test.js
+++ b/packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.test.js
@@ -186,6 +186,111 @@ describe('InputDateTimePicker', () => {
 			expect(stateDatetime).toBeInstanceOf(Date);
 			expect(isNaN(stateDatetime.getTime())).toBe(true);
 		});
+
+		cases(
+			'should not update state when the "selectedDateTime" prop is exactly the "datetime" state',
+			options => {
+				const invariableInputValue = "whatever input value which musn't change";
+				const defaultDate = new Date(2014, 1, 9, 12, 21, 3, 452);
+				const wrapper = shallow(
+					<InputDateTimePicker id={DEFAULT_ID} selectedDateTime={defaultDate} />,
+					{
+						disableLifecycleMethods: true,
+					},
+				);
+
+				wrapper.setState({
+					textInput: invariableInputValue,
+					datetime: options.selectedDateTime,
+				});
+
+				wrapper.update();
+
+				wrapper.setProps({
+					selectedDateTime: options.selectedDateTime,
+				});
+
+				wrapper.update();
+
+				expect(wrapper.state('textInput')).toBe(invariableInputValue);
+			},
+			[
+				{ name: 'InvalidDate value', selectedDateTime: new Date('whatever wrong selected date') },
+				{ name: 'undefined value', selectedDateTime: undefined },
+				{ name: 'Any valid Date value', selectedDateTime: new Date(2018, 4, 5, 12, 56) },
+			],
+		);
+
+		cases(
+			'should update state when the "selectedDateTime" prop diverge from the "datetime" state',
+			options => {
+				const defaultInputValue = 'whatever input value which must change';
+				const defaultDate = new Date(2014, 1, 9, 12, 21, 3, 452);
+				const wrapper = shallow(
+					<InputDateTimePicker id={DEFAULT_ID} selectedDateTime={defaultDate} />,
+					{
+						disableLifecycleMethods: true,
+					},
+				);
+
+				wrapper.setState({
+					textInput: defaultInputValue,
+					datetime: options.defaultSelectedDateTime,
+				});
+
+				wrapper.update();
+
+				wrapper.setProps({
+					selectedDateTime: options.newSelectedDateTime,
+				});
+
+				wrapper.update();
+
+				expect(wrapper.state('textInput')).not.toBe(defaultInputValue);
+			},
+			[
+				{
+					name: 'InvalidDate before and InvalidDate after',
+					defaultSelectedDateTime: new Date('whatever wrong selected date'),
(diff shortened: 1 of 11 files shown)


packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.test.js
```diff
@@ -1224,5 +1348,59 @@ describe('InputDateTimePicker', () => {
 				},
 			],
 		);
+
+		cases(
+			'should NOT apply an "invalid placeholder" on input AND override the regular one when InvalidDate value is internal, ie: wrong input value to keep visible',
```
jsomsanith: This seems to be 2 different tests as you execute different branches using `if`. Please split it into 2 simple tests

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.test.js
```diff
@@ -1224,5 +1348,59 @@ describe('InputDateTimePicker', () => {
...
+				}
+
+				wrapper.setState({
+					inputFocused: false,
+				});
+
+				wrapper.update();
+
+				const inputWrapper = wrapper.find('DebounceInput');
+				const placeholder = inputWrapper.prop('placeholder');
```
jsomsanith: `const placeholder = wrapper.find('DebounceInput').prop('placeholder');` + remove inputWrapper

packages/forms/src/UIForm/fields/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -4,19 +4,12 @@ import memoize from 'lodash/memoize';
...
-const INVALID_DATE = new Date('');
 const HANDLE_CONVERTION_TYPE = ['string', 'number'];
 
-function isDateValid(date) {
-	if (date === undefined) {
-		return true;
-	}
-
-	return date instanceof Date && !isNaN(date.getTime());
+function generateInvalidDate() {
```
jsomsanith: Why did you replaced the const INVALID_DATE by a function that create a new one each time ? You even have a test on that, guess you really want that for a reason.

packages/forms/src/UIForm/fields/InputDateTimePicker/README.md
```diff
@@ -75,23 +76,8 @@ Example : "2018-01-01T10:35:48.951Z"
 		"autoFocus": false,
 		"disabled": false,
 		"placeholder": "Type the date here...",
-		"readOnly": false
+		"readOnly": false,
+		"validationMessage": "The date is not valid"
```
jsomsanith: As this is an example, and people will copy/paste it, can you improve the message to push good practices ?
Something like `The date format is not valid. Expected format: YYYY/MM/DD.`

packages/forms/src/UIForm/fields/InputDateTimePicker/InputDateTimePicker.component.js
jdescottes-talend: Effectively :)
There is multiple possibilities to spread an InvalidDate to the component. When we switch for one error possibility to another still giving an InvalidDate, we couldn't figure out a change if it's the same instance.
And for the component point of view, the prop don't change from the last one, so it can't update itself.
One case :
- spread an InvalidDate due to 'error possibility X'
- define a valid date in the picker/component
- new error in the widget so spread a new InvalidDate due to 'error possibility Y'
(I lost the specific real case :/)

It's only for specific cases when errors could be raised. But I think just by principle, it's a good idea to have a different instance for different errors/data. And if it's the same data flowing into the widget, this will be memoized, so the same error and so, same instance.

packages/forms/src/UIForm/fields/InputDateTimePicker/README.md
jdescottes-talend: yep good point

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.test.js
jdescottes-talend: ✔

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.test.js
jdescottes-talend: ✔

packages/forms/src/UIForm/fields/InputDateTimePicker/README.md
jdescottes-talend: Update the behavior to have the message defined directly in the widget, as there could be only one.
In the goal to:
- enforce all applications to have the same message
- avoid having to add a new uiSchema property to use it properly
- reduce futur changes in uiSchemas when this error handling will evolve to a more friendly one and the unique message will not be needed

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.test.js
```diff
@@ -1224,5 +1347,56 @@ describe('InputDateTimePicker', () => {
...
+				});
+
+				wrapper.setState({
+					inputFocused: false,
+				});
+
+				wrapper.update();
+
+				const inputWrapperAfter = wrapper.find('DebounceInput');
+				const placeholder = inputWrapperAfter.prop('placeholder');
```
jsomsanith: Same remark as next test 
```javascript
const placeholder = wrapper.find('DebounceInput').prop('placeholder');
```

PR #1775, opened 2018-10-19
chore(TUI-149): update jest to 23

jmfrancois:

we are in jest 20, so we can t use new api like it.each

update to jest 23 in 'versions' script
use version scripts on our dependencies
update some tests to work with jest 23
update all snapshots (this is what makes the number of files so big)
fix some of the failing tests

it speeds up the CI which now take only 10 minutes !!
But it's not only the update its because we use jest in series instead of parallel using `--runInBand`
Sometimes people believe // execution is faster, this is a proof that it's not always the case.


npm-audit/package.json
```diff
@@ -82,8 +82,8 @@
     "eslint-plugin-import": "^1.16.0",
     "eslint-plugin-jsx-a11y": "^2.2.2",
     "eslint-plugin-react": "^6.3.0",
-    "jest": "20.0.3",
-    "jest-cli": "20.0.3",
+    "jest": "^23.0.0",
+    "jest-cli": "^23.0.0",
```
acateland: is the `^` intended ?

jmfrancois: Yes I trust the semver on jest.

jmfrancois:
Now I have time to work on it

npm-audit/package.json
jsomsanith: Yeah I think we can trust facebook to follow semver

packages/html-webpack-plugin/package.json
```diff
@@ -17,7 +17,7 @@
     "eslint": "3.4.0",
     "eslint-config-airbnb": "10.0.1",
     "eslint-plugin-import": "1.14.0",
-    "jest-cli": "20.0.3"
+    "jest": "20.0.3"
```
jsomsanith: The version is ko

packages/logging/package.json
```diff
@@ -56,7 +56,6 @@
     "eslint-plugin-jsx-a11y": "^2.2.2",
     "eslint-plugin-react": "^6.3.0",
     "jest": "20.0.3",
```
jsomsanith: Version is ko

packages/storybook-cmf/package.json
```diff
@@ -40,7 +40,7 @@
     "eslint-plugin-react": "^6.3.0",
     "immutable": "3.8.1",
     "invariant": "2.2.2",
-    "jest": "20.0.3",
+    "jest": "^23.0.0",
```
jsomsanith: Why do we have sometimes `"^23.0.0"` and othee times `"^23.6.0"` ?
We should align them

package.json
```diff
@@ -20,7 +20,7 @@
     "build-forms": "lerna exec --scope=@talend/react-forms -- yarn run prepublish",
     "build-icons": "lerna exec --scope=@talend/icons -- yarn run prepublish && lerna exec --scope=@talend/react-components -- yarn test -- -u",
     "build-theme": "lerna exec --scope=@talend/bootstrap-theme -- yarn run prepublish",
-    "test": "lerna exec --concurrency 1 -- TZ=Europe/Paris yarn test",
+    "test": "lerna exec --concurrency 1 -- TZ=Europe/Paris yarn test --detectOpenHandles --forceExit",
```
jsomsanith: I don't really understand this option. Can you explain why you added it ?

jmfrancois: it was to debug the CI which hang with jest. I have finally used --runInBand which fix the issue.
I have used --silent also which speed up the global execution which is now under 10 minutes :D

packages/containers/src/ShortcutManager/ShortcutManager.test.js
```diff
@@ -9,8 +9,9 @@ import Connected from './ShortcutManager.connect';
 
 describe('Container ShortcutManager', () => {
 	it('should render', () => {
-		const wrapper = shallow(<Container />);
+		const wrapper = shallow(<Container redirectMap={{}} />);
 		expect(wrapper.getElement()).toBeNull();
+		wrapper.unmount();
```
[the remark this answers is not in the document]
jmfrancois: to remove the addevent listener which is bind by the didMount. 
I don t know why we don t have the issue in the previous version

packages/containers/package.json
```diff
@@ -72,8 +72,8 @@
     "i18next-parser": "^0.13.0",
     "immutable": "^3.8.1",
     "invariant": "^2.2.2",
-    "jest": "20.0.3",
-    "jest-cli": "20.0.3",
+    "jest": "^23.6.0",
+    "jest-environment-jsdom": "^23.0.0",
```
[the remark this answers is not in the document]
jmfrancois: It s not anymore by default and we have some event listener. So we need it in order to make the test run

packages/logging/package.json
jmfrancois: I have added an entry in the travis matrix to test logging and html plugin.

PR #1777, opened 2018-10-22
feat: skip links

jsomsanith:
To provide skip links to all react projects using Talend/ui, we can add it into layout component.
So next upgrade will automatically bring them in the apps.

The choice of Layout is to control the ids, z-index and placement in 1 place.

Implement that, so it's free for the apps.

### DIFF
--- a/packages/components/src/Layout/SkipLinks/SkipLinks.component.js
+++ b/packages/components/src/Layout/SkipLinks/SkipLinks.component.js
@@ -23,6 +23,11 @@ SkipTo.propTypes = {
 	label: PropTypes.string.isRequired,
 };
 
+/**
+ * Skip links are a technique to ease keyboard navigation.
+ * It consists in giving some links as the firsts focusable elements of the page
+ * to go directly to some meaningful content.
+ */
 function SkipLinks({ mainId, navigationId, t }) {
 	return (
 		<nav
--- a/packages/components/src/Layout/SkipLinks/SkipLinks.component.js
+++ b/packages/components/src/Layout/SkipLinks/SkipLinks.component.js
@@ -1,0 +1,38 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import theme from './SkipLinks.scss';
+
+function SkipLinks(props) {
+	const children = [];
+
+	if (props.navigation) {
+		children.push(
+			<li>
+				<a href="#tc-layout-side-menu">{'Skip header to navigation'}</a>
+			</li>,
+		);
+	}
+
+	if (props.main) {
+		children.push(
+			<li>
+				<a href="#tc-layout-main">{'Skip navigation to main content'}</a>
+			</li>,
+		);
+	}
+
+	return <ul className={theme['tc-skip-links']}>{children}</ul>;
+}
+
+SkipLinks.defaultProps = {
+	navigation: true,
+	main: true,
+};
+
+SkipLinks.propTypes = {
+	navigation: PropTypes.bool,
+	main: PropTypes.bool,
+};
+
+export default SkipLinks;
--- a/packages/components/src/Layout/SkipLinks/SkipLinks.component.js
+++ b/packages/components/src/Layout/SkipLinks/SkipLinks.component.js
@@ -1,38 +1,66 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { translate } from 'react-i18next';
+
+import Icon from '../../Icon';
+import I18N_DOMAIN_COMPONENTS from '../../constants';
+import getDefaultT from '../../translate';
 
 import theme from './SkipLinks.scss';
 
-function SkipLinks(props) {
-	const children = [];
+function SkipTo({ href, label }) {
+	return (
+		<a href={href}>
+			<span className={theme.icon}>
+				<Icon transform="rotate-270" name="talend-arrow-left" />
+			</span>
+			<span className={theme.text}>{label}</span>
+		</a>
+	);
+}
+SkipTo.propTypes = {
+	href: PropTypes.string.isRequired,
+	label: PropTypes.string.isRequired,
+};
 
-	if (props.navigation) {
-		children.push(
-			<li>
-				<a href="#tc-layout-side-menu">{'Skip header to navigation'}</a>
-			</li>,
-		);
-	}
-
-	if (props.main) {
-		children.push(
-			<li>
-				<a href="#tc-layout-main">{'Skip navigation to main content'}</a>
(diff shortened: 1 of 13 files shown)


packages/components/src/Layout/SkipLinks/SkipLinks.component.js
```diff
@@ -0,0 +1,66 @@
...
+			<span className={theme.text}>{label}</span>
+		</a>
+	);
+}
+SkipTo.propTypes = {
+	href: PropTypes.string.isRequired,
+	label: PropTypes.string.isRequired,
+};
+
+function SkipLinks({ mainId, navigationId, t }) {
```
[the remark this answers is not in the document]
jsomsanith: The idea is not to display every part of the page. To do that you have SR, with the right attributes (role search/search-input).

packages/components/src/Layout/SkipLinks/SkipLinks.component.js
```diff
@@ -0,0 +1,66 @@
...
+import I18N_DOMAIN_COMPONENTS from '../../constants';
+import getDefaultT from '../../translate';
+
+import theme from './SkipLinks.scss';
+
+function SkipTo({ href, label }) {
+	return (
+		<a href={href}>
+			<span className={theme.icon}>
+				<Icon transform="rotate-270" name="talend-arrow-left" />
```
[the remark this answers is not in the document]
jsomsanith: It doesn't provide any content as the icon is aria-hidden.
Does it bring anything to hide the span as well ?

packages/components/src/Layout/SkipLinks/SkipLinks.component.js
```diff
@@ -0,0 +1,66 @@
...
+		<nav
+			className={theme['tc-skip-links']}
+			aria-label={t('SKIP_LINKS', { defaultValue: 'Skip links' })}
+		>
+			<ul>
+				{navigationId && (
+					<li key="navigation">
+						<SkipTo
+							href={navigationId}
+							label={t('SKIP_TO_NAV', { defaultValue: 'Skip header to navigation' })}
```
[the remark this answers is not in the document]
jsomsanith: Fixed for the 2 links

packages/components/src/Layout/SkipLinks/SkipLinks.component.js
```diff
@@ -0,0 +1,71 @@
...
+ */
+function SkipLinks({ mainId, navigationId, t }) {
+	return (
+		<nav
+			className={theme['tc-skip-links']}
+			aria-label={t('SKIP_LINKS', { defaultValue: 'Skip links' })}
+		>
+			<ul>
+				{navigationId && (
+					<li key="navigation">
```
annilsson: Why a key here?

packages/components/src/Layout/SkipLinks/SkipLinks.component.js
```diff
@@ -0,0 +1,71 @@
...
+						<SkipTo
+							href={navigationId}
+							label={t('SKIP_TO_NAV', { defaultValue: 'Skip to navigation' })}
+						/>
+					</li>
+				)}
+				{mainId && (
+					<li key="main">
```
annilsson: Why a key here?

packages/components/src/Layout/SkipLinks/SkipLinks.component.js
```diff
@@ -0,0 +1,71 @@
...
+ * to go directly to some meaningful content.
+ */
+function SkipLinks({ mainId, navigationId, t }) {
+	return (
+		<nav
+			className={theme['tc-skip-links']}
+			aria-label={t('SKIP_LINKS', { defaultValue: 'Skip links' })}
+		>
+			<ul>
+				{navigationId && (
```
annilsson: There's a defaultProp here so we don't need this conditional. And I don't think we need to make it optional unless we specifically need it disabled in an app.

packages/components/src/Layout/SkipLinks/SkipLinks.component.js
```diff
@@ -0,0 +1,71 @@
...
+					<li key="navigation">
+						<SkipTo
+							href={navigationId}
+							label={t('SKIP_TO_NAV', { defaultValue: 'Skip to navigation' })}
+						/>
+					</li>
+				)}
+				{mainId && (
```
annilsson: There's a defaultProp here so we don't need this conditional. And I don't think we need to make it optional unless we specifically need it disabled in an app.

packages/components/src/Layout/SkipLinks/SkipLinks.component.js
jsomsanith: because there are conditions to insert or not the `<li>`. And for react reconciliation, we need to provide unique keys so react knows what are the corresponding nodes.

packages/components/src/Layout/SkipLinks/SkipLinks.component.js
jsomsanith: See answer above

packages/components/src/Layout/SkipLinks/SkipLinks.component.js
jsomsanith: as we can disable them we need them. For example, a page where we have a layout with only a main, and no navigation side panel

packages/components/src/Layout/SkipLinks/SkipLinks.component.js
jsomsanith: Yes, for main, there is no use case where we want to remove that, I'll remove the condition

packages/components/src/Layout/SkipLinks/SkipLinks.component.js
jsomsanith: In fact, since we use it explicitly in Layout component, I'll remove the default props, so the api is better.
By default we don't want it, and if the id is explicitly passed, we add the link

packages/components/src/Layout/SkipLinks/SkipLinks.component.js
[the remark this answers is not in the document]
frassinier: It looks good, OK

packages/components/src/Layout/SkipLinks/SkipLinks.component.js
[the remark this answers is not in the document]
frassinier: It will be nice to access nav, main content **and** search. 
But TDP is the only one with a search bar, and I will not use this component for now. 
Let's see that in the second run.

PR #1779, opened 2018-10-22
fix(generator/cmf): make it work

jmfrancois:

The current cmf template doesn't work. It require a cmf.json file and the talend-script configuration to run.
More it uses talend/script and adds many dot files for nothing.

we have many security issues on this package.

generators/dotfile
* improve generator to accept more options

generators/cmf
* call dotfile to ignore babelrc, eslint, travis, sasslint
* add the two required files cmf.json and talend-scripts.json
* add forgotten test-setup.js file

security:
* remove generator-git-init (package not updated in 3 years)
* update mocha, yeoman-assert,yeoman-test, and yeoman generator (3 major versions)

remove coveralls which is not used.


packages/generator/.babelrc
```diff
@@ -1,4 +1,6 @@
 {
-  "extends": "../../.babelrc",
+  "presets": [
+    ["env", { "exclude": ["transform-es2015-classes"] }]
+  ],
```
jmfrancois: this has been needed because the new yeoman need real classes handled by node.

packages/generator/generators/dotfiles/index.js
```diff
@@ -32,11 +47,11 @@ module.exports = yeoman.Base.extend({
 		}
 		this.destinationRoot('.');
 		this.props = { name: this.options.name() };
-	},
+	}
 
 	configuring() {
 		const jsFormatterPath = 'tools-javascript';
-		const jsDotFiles = ['.editorconfig', '.eslintrc', '.travis.yml', '.npmignore'];
+		const jsDotFiles = ['.editorconfig', '.npmignore'];
```
ncomont: is `.prettierrc` missing ?

packages/generator/generators/react-cmf/templates/src/app/reducers/reducers.md
```diff
@@ -0,0 +1,3 @@
+You should not need any reducer using CMF, because it propose already some low level reducers.
```
jsomsanith: Disagree because I'm not fond of the props.setState
:) so please don't

acateland: while it looked like a good idea, in the long run i am not fond of props.setState to, because states update did loose conveying their intent and are less constrained.
Making thing hard to understand and maintain.

jmfrancois: `I'm not fond` is very subjective. please guys try to be more objective on this kind of review (the PR is not about that at all).

And this is not an idea but this was a need. So I will try to write down why do we have this API.

We want to embrace redux. Why ?
May be read the following article may help you to understand.
https://medium.com/@dan_abramov/you-might-not-need-redux-be46360cf367

So we want all the state of all components to be there so our bug reports will be really nice to work on.

Now how to achieve that ? Why do we have added `props.state` and `props.setState`  API ?

The second goal was to not have to write at the app level a `reducer` an `action` and a `selector` just to put a component state in it. It would make the app level just crasy with a mix of real app level stuff with some internal related to components stuff.

To make it easy to use and easy to understand the idea was to just have the same API as react state && setState. but under the hood it s just one action, one reducer and one selector written in the framework.

The only things I regret about this API is it force you to use immutable js. And about this we have the following ticket: 

[TUI-266](https://jira.talendforge.org/browse/TUI-266) make immutablejs an option in CMF

> states update did loose conveying their intent and are less constrained.

less constrained by what ? which intent ? I don't understand that point.

> Making thing hard to understand and maintain.

What is hard to understand and maintain ? Same point here , very subjective and I dont get the point ,could be more precise ?

* What is hard to understand ?
* What is hard to maintain ?

jsomsanith: Saw with @jmfrancois, that's not the goal of this PR. That's a way bigger topic, let's merge this one to fix the generator and iterate again

packages/generator/generators/dotfiles/index.js
jmfrancois: nice catch 👍

PR #1780, opened 2018-10-22
fix(log): update error logger to work on react 16

jmfrancois:

current release of @talend/log doesn't work on React 16.

Use error boundary to create a component which will post every issues on backend server URL.


acateland:
i see a few issues.

First it address only rendering related errors.
Second if an error rise up to the top level component, propably the app will become unstable and unusable, we should provide feedback to the user (IE show an alternate top level component)
Third, if a developer use error boundaries somewhere down in the tree, the error will not be catched and send to the server by the top component.

So i think this should provide an errorBoundaries factory of some sort. So you can create an error boundary component one in you app and apply it in many place so you can provide sensible data handling to the end user and error collecting.
We can keep having a top level errorBoundary so we don't miss any of those errors.

PR #1781, opened 2018-10-22
fix(components/ConfirmDialog): onHide called only if provided

sgangnant:
There is a warning poping when a `ConfirmDialog` component is hidden but no `onHide` handler has been provided.

Only call the handler if it is provided (since prop is not marked as required).

### DIFF
--- a/packages/components/src/ConfirmDialog/ConfirmDialog.component.js
+++ b/packages/components/src/ConfirmDialog/ConfirmDialog.component.js
@@ -63,7 +63,7 @@ function ConfirmDialog({
 		if (cancelAction && cancelAction.onClick) {
 			cancelAction.onClick();
 		}
-		return onHide(event);
+		return onHide ? onHide(event) : null;
 	}
 	return (
 		<Dialog


packages/components/src/ConfirmDialog/ConfirmDialog.component.js
```diff
@@ -63,7 +63,7 @@ function ConfirmDialog({
 		if (cancelAction && cancelAction.onClick) {
 			cancelAction.onClick();
 		}
-		return onHide(event);
+		return onHide ? onHide(event) : null;
```
acateland: you don't need to return anything in this function

sgangnant: Linter is unhappy when there is inconsistency with the return statement of a method :/

jmfrancois: in that case, remove the return?

sgangnant: It was there before, didn't want to potentially introduce a breaking change

PR #1792, opened 2018-10-26
fix(UIForms): array with fieldseption validation

jsomsanith:
Everything works fine when we define a UISchema with an array that defined each field directly as items.
```json
{
  "key": "comments",
  "itemWidget": "myItemWidget",
  "items": [{ "key": "comments[].name" }, { "key": "comments[].content" }]
}
```

But if we introduce extra levels of fieldsets (fieldseption), the global validation at submit fails to resolve the indexes in keys.

```json
{
  "key": "comments",
  "itemWidget": "myItemWidget",
  "items": [{
      "widget": "fieldset",
      "items": [{ "key": "comments[].name" }, { "key": "comments[].content" }]
  }]
}
```

We need to resolve the indexed deeply when we encounter an array, during validation.

* resolve the indexed deeply when we encounter an array, during validation.
* mutualize index resolution between global validation and during array rendering.


packages/forms/src/UIForm/utils/array.js
```diff
@@ -0,0 +1,52 @@
+function arrayStartsWith(prefix, arrayToCheck) {
+	return prefix.every((next, index) => arrayToCheck[index] === next);
+}
+
+function adaptArrayItemKey(arraySchema, item, itemIndex) {
```
acateland: please describe what does it do

packages/forms/src/UIForm/utils/array.js
```diff
@@ -0,0 +1,52 @@
...
+	const arrayKey = arraySchema.key;
+	const itemKey = item.key;
+	const itemChildren = item.items;
```
acateland:
```suggestion
	const itemChildrens = item.items;
```

packages/forms/src/UIForm/utils/array.js
```diff
@@ -0,0 +1,52 @@
...
+		...item,
+	};
+
+	if (itemKey) {
+		const indexedKey = [...itemKey];
+		indexedKey[arrayKey.length] = itemIndex;
+		schema.key = indexedKey;
+	}
+
+	if (itemChildren) {
```
acateland:
```suggestion
	if (itemChildrens) {
```

packages/forms/src/UIForm/utils/array.js
```diff
@@ -0,0 +1,52 @@
...
+		schema.items = itemChildren.map(child => adaptArrayItemKey(arraySchema, child, itemIndex));
```
acateland:
```suggestion
		schema.items = itemChildrens.map(child => adaptArrayItemKey(arraySchema, child, itemIndex));
```

PR #1801, opened 2018-10-31
feat(UIForms): date time

jsomsanith:
No widget for date/time picking

Component change : 
* InputDateTimePicker accepts timestamp, Date, or string following dateFormat as value

Form : 
* form widget
* iso-datetime and timestamp format with validation
* if string is required, without any format, the uiSchema widget configuration defines the format. It is used for display, and value is passed as is to the form.
* InputDateTimePicker error message is displayed as is for now. Another PR to have a proper translated error system.


packages/forms/src/UIForm/fields/Date/DateTime/DateTime.component.js
```diff
@@ -0,0 +1,150 @@
...
+import { isoDateTimeRegExp } from '../../../customFormats';
+import { WidgetUnhandleTypeError, WidgetUnexpectedTypeError } from '../WrongTypeError';
+import { generateDescriptionId, generateErrorId } from '../../../Message/generateId';
+
+const HANDLE_CONVERTION_TYPE = ['string', 'number'];
+const UNIQUE_ERROR_MESSAGE = 'The date is invalid. Expected format: YYYY-MM-DD HH:mm';
+const INVALID_DATE = new Date('');
+
+function convertToDate({ schema }, value) {
+	if (value === undefined) {
```
sgendre: what about null ?

packages/forms/src/UIForm/fields/Date/DateTime/DateTime.component.js
```diff
@@ -0,0 +1,150 @@
...
+import memoizeOne from 'memoize-one';
+import InputDateTimePicker from '@talend/react-components/lib/DateTimePickers';
+import FieldTemplate from '../../FieldTemplate';
+import { isoDateTimeRegExp } from '../../../customFormats';
+import { WidgetUnhandleTypeError, WidgetUnexpectedTypeError } from '../WrongTypeError';
+import { generateDescriptionId, generateErrorId } from '../../../Message/generateId';
+
+const HANDLE_CONVERTION_TYPE = ['string', 'number'];
+const UNIQUE_ERROR_MESSAGE = 'The date is invalid. Expected format: YYYY-MM-DD HH:mm';
+const INVALID_DATE = new Date('');
```
sgendre: not a big fan to have invalid objects, it is harder to grab the edge case from the caller, why not throw or null ?

packages/forms/src/UIForm/fields/Date/DateTime/DateTime.component.js
```diff
@@ -0,0 +1,150 @@
...
+			return isoDateTimeRegExp.test(value) ? new Date(value) : INVALID_DATE;
+		default:
+			// eslint-disable-next-line no-console
+			console.error(new WidgetUnhandleTypeError(HANDLE_CONVERTION_TYPE, type));
+			return INVALID_DATE;
+	}
+}
+
+function convertFromDate({ schema }, date) {
+	if (date === undefined) {
```
sgendre: please add null case

packages/forms/src/UIForm/fields/Date/DateTime/DateTime.component.js
```diff
@@ -0,0 +1,150 @@
...
+	constructor(props) {
+		super(props);
+
+		this.onChange = this.onChange.bind(this);
+		this.onBlur = this.onBlur.bind(this);
+		this.convertToDate = memoizeOne(convertToDate);
+	}
+
+	onChange(event, { datetime, errorMessage, origin }) {
+		const hasError = errorMessage !== undefined;
```
sgendre: if errorMessage is null, it considers that there is an error?

packages/forms/src/UIForm/customFormats.js
```diff
@@ -16,9 +16,21 @@ const urlGit = new RegExp(
...
+// EcmaScript ISO 8601 Extended Format: https://www.ecma-international.org/ecma-262/5.1/#sec-15.9.1.15
+// Only handle full string (date and time with full second fraction) and only UTC timezone
+export const isoDateTimeRegExp = /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}.\d{3}Z$/;
+
 const leadingTralingSpaceRegExp = /^\s|\s$/;
 const stringWithoutSpaceRegExp = /^\S+$/;
 
+const MAX_TIMESTAMP = 8640000000000000;
+
+function isValidDate(rawString) {
```
jlame: why not directly use the date-fns function ?

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -33,7 +33,11 @@ const PROPS_TO_OMIT_FOR_INPUT = [
 class InputDateTimePicker extends React.Component {
 	static propTypes = {
 		id: PropTypes.string.isRequired,
-		selectedDateTime: PropTypes.instanceOf(Date),
+		selectedDateTime: PropTypes.oneOfType([
```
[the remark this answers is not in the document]
jsomsanith: Done

packages/components/src/DateTimePickers/InputDateTimePicker/date-extraction.js
```diff
@@ -380,8 +380,40 @@ function extractPartsFromTextInput(textInput, options) {
 	};
 }
 
+/**
+ * Extract parts (date, time, date/time, textInput) from a value with
+ * different possible types
+ * @param value {string | Date \ number}
```
sgendre: typo

packages/components/src/DateTimePickers/InputDateTimePicker/date-extraction.js
```diff
@@ -380,8 +380,40 @@ function extractPartsFromTextInput(textInput, options) {
...
+ * @param options {Object}
+ * @returns
+ *	{{
+ *		date: Date,
+ *		time: { hours: string, minutes: string, seconds: string },
+ *		datetime: Date,
+ *		textInput: string
+ * 	}}
+ */
+function extractParts(value, options) {
```
sgendre: imo, this file would be better named by 'normalize.js'

packages/components/src/DateTimePickers/InputDateTimePicker/date-extraction.js
```diff
@@ -380,8 +380,40 @@ function extractPartsFromTextInput(textInput, options) {
...
+	const typeOfValue = typeof value;
+	if (typeOfValue === 'number') {
+		return extractPartsFromDateTime(new Date(value), options);
+	} else if (typeOfValue === 'string') {
+		return extractPartsFromTextInput(value, options);
+	} else if (value instanceof Date) {
+		return extractPartsFromDateTime(value, options);
+	}
+
+	return {
```
sgendre: what about  `return null` instead of an invalid object? 
I got very bad experience with those kind of empty/invalid instances

packages/components/src/DateTimePickers/InputDateTimePicker/date-extraction.test.js
```diff
@@ -43,44 +44,98 @@ describe('Date extraction', () => {
...
+			const options = { dateFormat: 'YYYY-MM-DD' };
+
+			// when
+			const parts = extractParts(validDate, options);
+
+			// then
+			expect(parts).toEqual({
+				date: new Date(2015, 8, 15),
+				datetime: validDate,
+				textInput: '2015-09-15',
```
sgendre: so is it 8 or 9?

packages/components/src/DateTimePickers/InputDateTimePicker/date-extraction.test.js
```diff
@@ -43,44 +44,98 @@ describe('Date extraction', () => {
...
+			const options = { dateFormat: 'YYYY-MM-DD' };
+
+			// when
+			const parts = extractParts(timestamp, options);
+
+			// then
+			expect(parts).toEqual({
+				date: new Date(2015, 8, 15),
+				datetime: date,
+				textInput: '2015-09-15',
```
sgendre: 8 or 9? there might be something that I don't get

packages/forms/src/UIForm/fields/Date/README.md
```diff
@@ -0,0 +1,76 @@
+# InputDateTimePicker
+
+This widget allows you to render a date/time picker.
+
+## Data conversion
+
+You can request 3 format :
+* timestamp (number)
+* ISO-8601 ewtended format (string)
```
sgendre: typo

packages/components/src/DateTimePickers/InputDateTimePicker/date-extraction.js
jsomsanith: No, because we expose extraction functions. Normalize is quite generic and we usually call normalization something that takes a value, transform it and return this same value, but with a defined format.
Here we extract things : date, hours, and convert it : date and string

packages/components/src/DateTimePickers/InputDateTimePicker/date-extraction.js
jsomsanith: No, the values in the returned object are used in different locations.
For example, the date and time are used in the date picker, the textInput, in the input.
All those info will live in the InputDateTimePicker state. So we need the define each one of them for setState to replace the old values.

packages/components/src/DateTimePickers/InputDateTimePicker/date-extraction.test.js
jsomsanith: In date api 8 is the month index, that is used for september.
It is transformed to reflext `09` in the input to be human understandable.

packages/components/src/DateTimePickers/InputDateTimePicker/date-extraction.test.js
jsomsanith: Same explanation

packages/components/src/DateTimePickers/InputDateTimePicker/date-extraction.js
sgendre: inflate?

packages/components/src/DateTimePickers/InputDateTimePicker/date-extraction.test.js
sgendre: right !
it's been a while since I worked with datetime

packages/components/src/DateTimePickers/InputDateTimePicker/date-extraction.js
jsomsanith: This word is not really clear neither. But I'll draw the card "Not in this PR scope" as it's not something I just created and don't want to discuss about this naming because the ideas we have for now is not better than the one in place.
But if we find something better later, an easy PR can be done.

PR #1803, opened 2018-11-06
feat(form): add form reset handlign on UIForm

acateland:
There is no handling of reset button on the lib.

Split the state in two parts, an initialState representing the last stable formSchema (jsonSchema, UISchema, data), and a liveState that is used to update formData on form change.

This allow on reset to overwritte liveState with initialState.

the way both of those state are handled
image


packages/forms/src/UIForm/UIForm.container.js
```diff
@@ -5,28 +5,87 @@ import React from 'react';
...
+});
+
+/**
+ * update live state with new properties derived from user interacting
+ * with the form
+ * @param {Object} properties
+ * @param {State} prevState
+ * @return {State}
+ */
+const change = properties => prevState => ({
```
jsomsanith: As you externalized those pieces of code, I think you should give them more explicit names, explaining what they do instead of when they are called.

For example:
reinitLiveState 👍 
change --> setLiveStateProperties
setErrors --> setLiveStateErrors
submit --> setLiveAsInitialState

packages/forms/src/UIForm/UIForm.container.js
```diff
@@ -5,28 +5,87 @@ import React from 'react';
...
+});
+
+/**
+ * update initialState with liveState after a user submission of the form
+ * @param {Object} newProperties
+ * @param {State} prevState
+ * @return {State}
+ */
+const submit = newProperties => prevState => {
+	const newFormSchema = { ...prevState.liveState, properties: newProperties };
```
jsomsanith: You don't need the replace the properties / create a new object here.
On submit, the submitted properties are the current live form properties.

packages/forms/src/UIForm/UIForm.container.js
```diff
@@ -40,15 +99,35 @@ export default class UIForm extends React.Component {
...
+		if (typeof this.props.onSubmit === 'function') {
+			this.props.onSubmit(event, properties);
+		}
+	}
+
+	/**
+	 * On user reset change local state and call this.props.onReset
+	 */
+	onReset() {
+		this.setState(prevState => ({ ...prevState, liveState: prevState.initialState }));
```
jsomsanith: Totally ok, just noticed that you inline this one but externalized the others

packages/forms/src/UIForm/UIForm.container.js
```diff
@@ -5,28 +5,94 @@ import React from 'react';
...
 export default class UIForm extends React.Component {
 	static displayName = 'Container(UIForm)';
 	constructor(props) {
 		super(props);
-		this.state = { ...this.props.data };
-		if (!this.state.errors) {
-			this.state.errors = {};
-		}
+		this.state = {
+			initialState: addErrorObject(this.props.data),
```
jare-talend: interesting solution to separate the initial data with the living data. So, addErrorObject make a spread on the first level. We lose the reference on the first level but if the formdata has nested data, we have a reference with initialState, Would  it is interesting to make a deep shallow to avoid that ?

acateland: using https://lodash.com/docs/4.17.11#cloneDeep for example ?

jare-talend: yes

jsomsanith: But everything is treated as immutable. If the user get it and mutate data, it's a bad practice and he has to fix that.

acateland: im not sure that we need a deep clone here and on each state changes:
- because it wasn't done before and it worked fine
- because PureComponent and other méthod usualy do a shallow check on props

PR #1806, opened 2018-11-06
fix(components/Slider): align icons with slider handler and add disabled props

nicomaligne:
We need to align the icons with the button value of the slider.
rating25
You can see the guideline here : http://guidelines.talend.com/document/92132#/specific-features/rating-sliders
Tweak some css and html
We also add a disabled props that allow to disabled the buttons and the slider.


packages/components/stories/Slider.js
```diff
@@ -48,52 +48,56 @@ const actions = [
...
+		icon: 'talend-smiley-satisfied',
+		'data-feature': 'action',
+		link: true,
+		hideLabel: true,
+	},
 ];
 
 const functionToFormat = value => `${value}-test`;
-const functionFormatFloor = value => Math.floor(value);
+const functionFormatFloor = value => `${Math.floor(value)}k`;
```
jare-talend: what k ?

packages/components/src/Slider/Slider.component.js
```diff
@@ -219,17 +224,19 @@ class Slider extends React.Component {
...
-					id={id}
-					value={value}
-					min={min}
-					max={max}
-					handle={noValue ? undefined : this.state.handle}
-					className={classnames(theme['tc-slider-rc-slider'], 'tc-slider-rc-slider')}
-					onChange={onChange}
-					{...rest}
-				/>
+			<div>
```
jare-talend: why double div nesting ? one only

packages/components/stories/Slider.js
nicomaligne: It was just to look like the visual on frontify. I can remove it.

packages/components/src/Slider/Slider.component.js
nicomaligne: Need the slider padding applied only to the slider not the icons.

PR #1811, opened 2018-11-08
fix(form): remove warning messages

lmaillet:
Lot's of warning messages in the console while using uiforms

```
Warning: Failed prop type: The prop `descriptionId` is marked as required in `FieldTemplate`, but its value is `undefined`.
    in FieldTemplate (created by Button)
    in Button (created by Widget)
```
	
```
Warning: Failed prop type: The prop `errorId` is marked as required in `FieldTemplate`, but its value is `undefined`.
    in FieldTemplate (created by Button)
    in Button (created by Widget)
```
	
```
Warning: Each child in an array or iterator should have a unique "key" prop.
Check the render method of `TalendUIForm`. See https://fb.me/react-warning-keys for more information.
    in div (created by TalendUIForm)
    in TalendUIForm (created by I18n)
```
	
```
Warning: Failed prop type: You provided a `value` prop to a form field without an `onChange` handler. This will render a read-only field. If the field should be mutable use `defaultValue`. Otherwise, set either `onChange` or `readOnly`.
    in input (created by FileWidget)
    in div (created by FileWidget)
    in div (created by FieldTemplate)
    in FieldTemplate (created by FileWidget)
    in FileWidget (created by I18n)
```

In order to remove warning messages in forms and widgets
* Add mandatory props `descriptionId` and `errorId` in Button component
* Add forgotten `key` prop in UIForm component
* Add `onChange` in File widget


packages/forms/src/UIForm/fields/Button/Button.component.js
```diff
@@ -2,13 +2,18 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import FieldTemplate from '../FieldTemplate';
 import SingleButton from './SingleButton.component';
+import { generateDescriptionId, generateErrorId } from '../../Message/generateId';
```
jmfrancois: Not the scope of the PR but wanna talk about that:
First it doens t generate, it generate only if id is undefined. So I would like to rename that module.
Then I would like it to be shared with all the package.

so name could be `enforceId(id, suffix)` or sth like this

packages/forms/src/UIForm/UIForm.component.js
```diff
@@ -262,7 +262,7 @@ export class UIFormComponent extends React.Component {
 				/>
 			));
 		const buttonsRenderer = () => (
-			<div className={classNames(theme['form-actions'], 'tf-actions-wrapper')}>
+			<div className={classNames(theme['form-actions'], 'tf-actions-wrapper')} key="form-buttons">
```
acateland: since buttonsRenderer is called multiple time it should have a unique key associated with each execution in sequence.

https://reactjs.org/docs/lists-and-keys.html#keys

jmfrancois: for me it s not the case it s called at each render but it identify the form submit item. So it s identity is form-buttons

It s not the widget.

screen shot 2018-11-13 at 11 58 40

acateland: ok it is because buttonsRenderer is called inside an array, i think a comment on the code should be put, so we do not forget to migrate to React.Fragment later, because a static key is pretty counter intuitive here

PR #1813, opened 2018-11-08
fix(CMF/TUI-291): use spawn to not break all sagas

jmfrancois:

If a saga throw an error all sagas are cancelled.

Replace all fork calls inside CMF to make all sagas completly independant.
So if a saga fails just this saga will be cancelled.


jmfrancois:
yes, they will.
All sagas today starts from the main which is created by cmf.bootstrap.

Then you start three sagas:
1. the component saga. The fact that we fork it from the main is ok if its a zombie. You are sure your component saga will do its work.
2. then the saga router. same as above
3. the main saga of the app. which can be a merge.

Then the component saga starts component saga using spawn but controls its life. so everything is OK.
The sagaRouter is aligned on that too.

So the main question is more for the saga you write in the App.
You are on your own so I return the question to you: how should you handle an error in app sagas?

This patch seems perfect to me as we have only the three sagas which are handled correctly and the scope is maintained as just start other sagas.

PR #1822, opened 2018-11-13
fix(forms/TUI-93): required checkbox put * at the end

jmfrancois:

screen shot 2018-03-26 at 11 07 34

screen shot 2018-11-13 at 11 51 24


packages/forms/src/UIForm/fields/FieldTemplate/FieldTemplate.scss
```diff
@@ -2,4 +2,8 @@
 	&:global(.required) :global(.control-label):after {
 		content: '*';
 	}
+	&:global(.required) :global(.checkbox) :global(.control-label):after {
```
frassinier:
```suggestion
.template {
	:global {
		.required .checkbox .control-label:after {
			position: initial;
			content: '*';
			outline: none;
		}
	}
}
```

packages/forms/.storybook/addons.js
```diff
@@ -1,3 +1,2 @@
-import '@storybook/addon-knobs/register';
```
frassinier: It should be split in another PR I think

packages/forms/src/UIForm/fields/FieldTemplate/FieldTemplate.scss
frassinier: `position: initial` is not compatible with IE 11
https://caniuse.com/#feat=css-initial-value

PR #1831, opened 2018-11-15
feat(ActionFile): pass accept props

VolodymyrKovalM:
Sometimes need to make input file accept only specific extensions
(of course it can be changed in system window)
We have such case in TMC when uploading resource of directory type (need to accept only archives with zip extension)


packages/components/src/Actions/ActionFile/ActionFile.component.js
```diff
@@ -117,6 +117,7 @@ class ActionFile extends React.Component {
 			tooltipLabel,
 			label,
 			tooltipPlacement,
+			accept,
```
frassinier: we can accept all others props instead
```suggestion
			...rest,
		}
```

packages/components/src/Actions/ActionFile/ActionFile.test.js
```diff
@@ -30,6 +30,11 @@ describe('ActionFile', () => {
 		expect(wrapper.getElement()).toMatchSnapshot();
 	});
 
+	it('should render with accept attribute passed to it', () => {
+		const wrapper = shallow(<ActionFile {...myAction} accept=".zip" />);
+		expect(wrapper.getElement()).toMatchSnapshot();
```
jmfrancois: please go to find the accept on the input instead of a simple snapshot.

packages/components/src/Actions/ActionFile/ActionFile.component.js
jmfrancois: no please don t do that. It generate lots of proptype warning.
```
React does not recognize the `iconTransform` prop on a DOM element
```

PR #1832, opened 2018-11-16
fix(ComponentForm): reset 

acateland:
While reset on UIForm do work now, reset on ComponentForm does not handle some case.

Lets start by having a ComponentForm setup with a definitionUrl props `/form/1`
If we type some values in the displayed form, then reset it.
The form display the initial value as seen on the initial form display.

So far so fine.

But let say now that without the component still mounted, we change the definitionUrl props to `form/2` this displaying a different form with differents values.

if we press reset now, the value shown are those initialy seen on the first load of the form attached to `/form/1`

this mean, UIForm can't be reset programatically.

First thing add an optionnal `initialData` props to the UIForm container.
if this `initialData` is provided, it is going to be used to set the initialState of the UIForm container.
if this `initialData` is updated (important to use immutable datastructure he), this new version will update the initialState of the UIForm container.

This give use the ability to set the UIForm intialState without unMounting -> mounting the component.

And for the ComponentForm we will also create in the redux store an initialState, that will be provided to the UIForm container.
This initialState will be used on reset tu update the live FormComponent redux state.
This initialState will updated when submiting the form (success or not) and when fetching form definitions.

This is linked to https://github.com/Talend/ui/pull/1803

### DIFF
--- a/packages/containers/src/ComponentForm/ComponentForm.component.js
+++ b/packages/containers/src/ComponentForm/ComponentForm.component.js
@@ -4,7 +4,7 @@ import cmf, { cmfConnect } from '@talend/react-cmf';
 import Form from '@talend/react-forms';
 import { getValue } from '@talend/react-forms/lib/UIForm/utils/properties';
 import omit from 'lodash/omit';
-import { Map, fromJS } from 'immutable';
+import { Map } from 'immutable';
 import memoizeOne from 'memoize-one';
 import kit from './kit';
 import tcompFieldsWidgets from './fields';
--- a/packages/containers/src/ComponentForm/ComponentForm.component.js
+++ b/packages/containers/src/ComponentForm/ComponentForm.component.js
@@ -4,7 +4,7 @@ import cmf, { cmfConnect } from '@talend/react-cmf';
 import Form from '@talend/react-forms';
 import { getValue } from '@talend/react-forms/lib/UIForm/utils/properties';
 import omit from 'lodash/omit';
-import { Map } from 'immutable';
+import { Map, fromJS } from 'immutable';
 import memoizeOne from 'memoize-one';
 import kit from './kit';
 import tcompFieldsWidgets from './fields';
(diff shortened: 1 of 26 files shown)


packages/containers/src/ComponentForm/ComponentForm.component.js
```diff
@@ -4,7 +4,7 @@ import cmf, { cmfConnect } from '@talend/react-cmf';
 import Form from '@talend/react-forms';
 import { getValue } from '@talend/react-forms/lib/UIForm/utils/properties';
 import omit from 'lodash/omit';
-import { Map } from 'immutable';
+import { Map, fromJS } from 'immutable';
```
talend-glorieux:
```suggestion
import { Map } from 'immutable';
```

packages/containers/src/ComponentForm/ComponentForm.component.js
```diff
@@ -22,6 +22,7 @@ const TO_OMIT = [
 
 export const DEFAULT_STATE = new Map({
 	dirty: false,
+	initialState: {},
```
sgangnant: Wouldn't it be less confusing to call this `initialData` ?

packages/forms/src/UIForm/README.md
```diff
@@ -548,3 +548,18 @@ function MyComponent(props) {
...
+When the definition url is used to load the first state of the form a copy is kept so in case the user use this reset action the live form state will be `reset` to this initial state.
+
+Said copy is updated when the user submit the form or when the definition url is changed and that a new form definition is loaded.
+
+#### Programatic reset
+`initialData` this additional props serve one use, create a new saved initial state programaticaly.
+This is specially usefull when this component is used by `containers/ComponentForm`.
+
+When this value is provided, it is going to be used a the new initialState, if this value change over time, it will update the initialState.
+(So it is very important to no create new reference or useless mutation here)
```
jsomsanith: `not to create`

packages/containers/src/ComponentForm/ComponentForm.component.js
acateland: in UIForm it is already called initialState didn't wan't to create discrepancies now.

PR #1833, opened 2018-11-16
fix(components/subheaderbar): design adjustments

ncomont:

Subheader does not follow the actual guidelines.

Align it.


packages/components/src/Actions/Action/index.js
```diff
@@ -1,4 +1,4 @@
-import Action, { wrapOnClick } from './Action.component';
+import Action, { wrapOnClick, ACTION_TYPES } from './Action.component';
 
-export { wrapOnClick };
+export { wrapOnClick, ACTION_TYPES };
```
jmfrancois: why do you export ACTION_TYPES ?

packages/components/src/EditableText/EditableText.scss
```diff
@@ -46,6 +46,8 @@ $tc-circle-editable-text-size: 16px !default;
 		padding-left: $padding-small;
 		padding-right: $padding-small;
 		flex: 0 0 auto;
+		line-height: inherit;
+		min-height: inherit;
```
frassinier: Could you try with IE 11, please?
https://caniuse.com/#feat=css-unset-value
Just to be sure

packages/components/src/Actions/Action/index.js
jmfrancois: I prefer you to attach it to the Action component so you can write
```
Action.DISPLAY_MODE_FILE
```
etc.

packages/components/src/SubHeaderBar/SubHeaderBar.scss
```diff
@@ -14,15 +14,17 @@ $tc-svg-icon-size: $padding-large !default;
 	width: 100%;
 
 	button {
```
jmfrancois: nothing seems attached to button, Is this possible to remove it ?

PR #1841, opened 2018-11-21
chore(DatePicker): seconds management

jlame:
The widget doesn't manage second for the moment 

Manage seconds in background and allowing it to be change in the input. 


packages/components/src/DateTimePickers/generator.js
```diff
@@ -6,6 +6,15 @@ import endOfMonth from 'date-fns/end_of_month';
 import format from 'date-fns/format';
 import setDay from 'date-fns/set_day';
 import startOfWeek from 'date-fns/start_of_week';
+import { getCurrentLanguage } from '../translate';
+
+export const locales = {
+	fr: require('date-fns/locale/fr'),
+	zh: require('date-fns/locale/zh_cn'),
```
jlame: the issue with that part is that most app use zh as local for now we don't manage regions. if anyone has a proposal for that part feel free.

jsomsanith: This is not how we should do the things pour date-fns locales. Here it is quite strict, if we want to support another language, we have to change the code.
We decided to write our own date-fns locale, that use our i18next system. You can take a look at components/src/DateFnsLocale/locale.js.
It doesn't contain all translation, so if something is missing (like what you use here I think), please add it in the locale.
You can take example on any date-fns translation to see the api (https://github.com/date-fns/date-fns/tree/master/src/locale/fr)

packages/components/src/DateTimePickers/InputDateTimePicker/date-extraction.js
```diff
@@ -48,23 +51,28 @@ function checkTime({ hours, minutes }) {
...
+ * Convert hour minutes and seconds into seconds
  * @param hours {string}
  * @param minutes {string}
+ * @param seconds {string}
  * @returns {number}
  */
-function hoursAndMinutesToTime(hours, minutes) {
-	checkTime({ hours, minutes });
-	return Number(hours) * 60 + Number(minutes);
+function hoursAndMinutesAndSecondsToTime(hours, minutes, seconds) {
```
jsomsanith: Perhaps you can change the naming :D `timeToSeconds`

packages/components/src/DateTimePickers/InputDateTimePicker/date-extraction.js
```diff
@@ -76,14 +84,17 @@ function dateTimeToStr(date, time) {
...
 
-	const { hours, minutes } = time;
+	const { hours, minutes, seconds } = time;
 	try {
-		const timeInMinutes = hoursAndMinutesToTime(hours, minutes);
-		const fullDate = setMinutes(date, timeInMinutes);
-		return format(fullDate, INPUT_FULL_FORMAT);
+		const timeInSeconds = hoursAndMinutesAndSecondsToTime(hours, minutes, seconds);
+		const fullDate = setSeconds(date, timeInSeconds);
+		return format(fullDate, seconds === '00' ? INPUT_WITHOUT_SECOND_FORMAT : INPUT_FULL_FORMAT);
```
jsomsanith: I don't think we should manage the 2 formats. It will be very hard to manage.
Just use the format with seconds, we should tackle the format change (by user or language).

packages/components/src/DateTimePickers/InputDateTimePicker/date-extraction.js
```diff
@@ -76,14 +84,17 @@ function dateTimeToStr(date, time) {
...
 	} catch (e) {
 		const dateStr = format(date, INPUT_DATE_ONLY_FORMAT);
 		if (hours !== '' && minutes !== '') {
+			if (seconds !== undefined) {
```
jsomsanith: So don't need the `if` here

packages/components/src/DateTimePickers/InputDateTimePicker/date-extraction.js
```diff
@@ -102,9 +113,9 @@ function dateAndTimeToDateTime(date, time) {
 	}
 
 	try {
-		const { hours, minutes } = time;
-		const timeInMinutes = hoursAndMinutesToTime(hours, minutes);
-		return setMinutes(date, timeInMinutes);
+		const { hours, minutes, seconds = '00' } = time;
```
jsomsanith: I don't think we should manage seconds differently than the others

packages/components/src/DateTimePickers/InputDateTimePicker/date-extraction.js
```diff
@@ -148,15 +159,18 @@ function strToDate(strToParse) {
...
 	if (!timeMatches) {
 		throw new Error('TIME - INCORRECT FORMAT');
 	}
 
 	const hours = timeMatches[1];
 	const minutes = timeMatches[2];
 
-	return { hours, minutes };
+	const seconds = timeMatches[3] === undefined ? '00' : timeMatches[3];
+	return { hours, minutes, seconds };
```
jsomsanith: This function would be impacted too

packages/components/src/DateTimePickers/InputDateTimePicker/date-extraction.js
jlame: right haha

packages/components/src/DateTimePickers/InputDateTimePicker/date-extraction.js
jlame: useSeconds to make the switch between format

dateTimeToStr(date, time, option) {useTime, useSecond} // use time in another PR

PR #1842, opened 2018-11-21
chore(CMF/TUI-271): extract router as cmfModule

jmfrancois:

same as #1710. I just find it too much complex. I would like it to be a simple cmfModule

done as a cmfModule and remove allmost all references of react-router from cmf. The only one which is not possible without a big breaking change is the context.

project which use context.router need the router so we can let the contextType router in cmfConnect and withExpression it doesn t break if you don t have router context.router will be undefined if you setup @talend/react-cmf-router it will be as before.


packages/router/src/sagaRouter/matchPath.js
```diff
@@ -0,0 +1,58 @@
+/**
```
jsomsanith: This file is duplicated in cmf and in router

packages/router/package.json
```diff
@@ -0,0 +1,81 @@
...
+    "node-fetch": "^1.7.3",
+    "react": "^16.0.0",
+    "react-dom": "^16.0.0",
+    "react-test-renderer": "^16.0.0",
+    "redux-mock-store": "^1.2.3",
+    "redux-saga-tester": "^1.0.345",
+    "rimraf": "^2.6.1"
+  },
+  "dependencies": {
+    "@talend/react-cmf": "1.7.0",
```
jsomsanith: It should be peer dep. Otherwise, we will have to update this dep when we release cmf, or we can end up with 2 react-cmf loaded.
Perhaps it's the same for redux/...

lerna.json
```diff
@@ -12,6 +12,7 @@
     "packages/forms",
     "packages/generator",
     "packages/icons",
+    "packages/router",
```
jsomsanith: Let's not have the same versioning, instead, let's follow react-router versions

packages/containers/sandbox/src/app/index.js
```diff
@@ -5,18 +5,22 @@
...
+const routerModule = getRouter({});
+
 const app = {
 	components: { ComponentForm },
 	settingsURL: '/settings.json',
 	actionCreators: actions,
 	middlewares: [createLogger({})],
-	modules: [containersModule],
+	modules: [routerModule, containersModule],
+	RootComponent: routerModule.components.Router,
```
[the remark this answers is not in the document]
jmfrancois: ok to improve it. In your proposal it means getRouter return an object so let say router with two attributes which starts with `router`.

An other proposal in the same direction:

```
const router = getRouter();

const app = {
	components: { ComponentForm },
	settingsURL: '/settings.json',
	actionCreators: actions,
	middlewares: [createLogger({})],
	modules: [router.cmfModule, containersModule],
	RootComponent: router.RootComponent,
};
```

packages/router/src/sagaRouter/matchPath.js
jmfrancois: yes I have removed it from the router.
This one is a pure utility

packages/router/package.json
jmfrancois: ok I have moved it to peer, make sens for cmf because you need cmf to use this one.
I have also pin CMF to >= 1.7

packages/cmf/BREAKING-CHANGES.md
```diff
@@ -0,0 +1,84 @@
+# 1.0 -> 2.0
```
jsomsanith: This doesn't seem to cover all the breaking changes. Some changes I spotted :
* `bootstrap` that doesn't accept any router conf anymore (history, etc). The README link is not sufficient, it's not a breaking change log explaining the diff.
* `WithExpression HOC` don’t inject context.router anymore

packages/cmf/BREAKING-CHANGES.md
```diff
@@ -0,0 +1,84 @@
...
+```javascript
+// before
+import cmf from '@talend/react-cmf';
+
+function* mainSaga() {
+    yield fork(cmf.sagas.changeDocumentTitle);
+    //...
+}
+// after
+// just remove that the setup of the addon is doing it.
```
jsomsanith: Please replace `addon` by `@talend/react-cmf-router`

packages/containers/src/DeleteResource/actions.test.js
```diff
@@ -6,7 +6,11 @@ describe('deleteResource actions', () => {
 		it('should return an action DIALOG_BOX_DELETE_RESOURCE object', () => {
 			// Given
 			const context = {
-				router: { getCurrentLocation: jest.fn(() => ({ pathname: 'currentUrl' })) },
+				store: {
+					getState: jest.fn(() => ({
+						routing: { locationBeforeTransitions: { pathname: 'currentUrl' } },
```
jsomsanith: To avoid having test depending on the implem of the router from an external module, we could mock it

packages/cmf/BREAKING-CHANGES.md
```diff
@@ -0,0 +1,84 @@
...
+    }
+    return {
+        type: 'MY_ACTION_WITHOUT_ROUTER',
+    };
+}
+```
+
+Why do I need to write down a path here ?
+Because before you rely on the router configuration, so you may without know it
+adding a route that will match this and provide a `myParamId`. This is not enough explicit, your code should not rely on configuration.
```
acateland: can we agree to remove the whole cmfRouter thing then ? this abstraction was created so we could leverage settigns to create route and stuff.

packages/containers/src/ShortcutManager/ShortcutManager.container.js
```diff
@@ -17,7 +17,7 @@ class ShortcutManager extends React.Component {
 	};
 
 	static contextTypes = {
-		router: PropTypes.object.isRequired,
+		store: PropTypes.object.isRequired,
```
jsomsanith: Context doesn't seem to be used. (Don't forget to remove them from tests too)

packages/containers/src/ShortcutManager/ShortcutManager.test.js
```diff
@@ -18,6 +18,19 @@ describe('Shortcut container', () => {
...
+});
+
+describe('handles routes', () => {
+	const state = mock.state();
+	state.cmf.settings.props.shortcuts = {
+		redirectMap: {},
+	};
+	state.cmf.components = new Map();
+	state.routing = {
+		locationBeforeTransitions: {
```
jsomsanith: Same here, shouldn't we mock routerAPI instead ?

packages/containers/src/register.test.js
```diff
@@ -9,6 +9,16 @@ jest.mock('@talend/react-cmf', () => ({
...
+	routerAPI: {
+		selectors: {
+			matchPath: jest.fn(),
+			getPath: jest.fn(),
+		},
+		matchPath: jest.fn(),
+	},
+}));
+
 describe('#register contianers', () => {
```
jsomsanith: Can you fix the typo please ?

packages/containers/package.json
```diff
@@ -146,6 +148,7 @@
   "dependencies": {
     "memoize-one": "^4.0.0",
     "react-immutable-proptypes": "^2.1.0",
-    "react-router": "^3.2.0"
+    "react-router": "^3.2.0",
+    "@talend/react-cmf-router": "3.0.0-2"
```
jsomsanith: Is it dependency or a peer dep ? You added them in both section

packages/cmf/src/settings.md
```diff
@@ -96,96 +95,3 @@ It should follow some rules:
...
-
-```json
-{
-  "path": "*",
-  "component": "Redirect",
-  "to": "/404"
-}
-```
-
-You can use onEnter/onLeave lifecycle hooks. To do that, you need to:
```
jsomsanith: There is no doc that explains that anymore

packages/containers/package.json
```diff
@@ -146,6 +148,7 @@
   "dependencies": {
     "memoize-one": "^4.0.0",
     "react-immutable-proptypes": "^2.1.0",
-    "react-router": "^3.2.0"
+    "react-router": "^3.2.0",
```
jsomsanith: This is a dep of @talend/react-cmf-router, you can drop it

packages/router/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+const router = getRouter({ history, sagaRouterConfig });
+
+cmf.bootstrap({
+    modules: [router.cmfModule],
+    RootComponent: router.RootComponent,
+});
+```
+
+Then every cmfConnect component do not have anymore the context.router. That means you have to search for context.router in your code and update it.
```
jsomsanith: This has nothing to do in the module readme. This is a breaking change migration guideline

packages/router/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+before
+```
+function myaction(event, data context) {
```
jsomsanith: (included in previous comment, but fix the typo please)

packages/router/README.md
```diff
@@ -0,0 +1,37 @@
...
+```
+
+## routerAPI
+
+routerAPI is an object which expose the following api:
+
+
+`routerAPI.selectors.getLocation(state)` return the current [location object](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/location.md) which is persisted from `history` into redux store
+
+`routerAPI.selectors.getPath(state)` return the current path (a string) which is the fragment so you can apply matchPath to it.
```
jsomsanith: Can you turn this 2 texts into a table please for readability.

packages/router/test-setup.js
```diff
@@ -0,0 +1,24 @@
+/* eslint-disable global-require,no-plusplus */
+import 'babel-polyfill';
```
jsomsanith: This needs to be in package.json
This should be spotted by linter, can you add the module in travis linter part ?

packages/cmf/BREAKING-CHANGES.md
```diff
@@ -0,0 +1,84 @@
...
+
+Worse, action creators should not rely on the current router params. That means your action is not an event, it s an entier effect. Your sagas should do that kind of check.
```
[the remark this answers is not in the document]
acateland: action are not event it is missleading.
why should redux action payload and action creator should not rely on router params ? it s an `entier effect`
i do not understand this part of the sentence

packages/router/README.md
acateland: `cmfConnect` is an higher order component, i think you should rephrase it like this:
Now components wrapped inside a `cmfConnect` higher order component don't have router's context injected as props.
This mean that you have to use the React context API yoursef in the component wrapped by `cmfConnect`

packages/cmf/src/middlewares/cmf/index.md
```diff
@@ -33,6 +33,9 @@ Further information about managing collections is available [here](how-to-manage
 
 ## Router
 
+
+Note: this will work only if you have setup `@talend/react-cmf-router`
```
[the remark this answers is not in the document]
jmfrancois: I have extracted the part of middleware in the module too

packages/cmf/BREAKING-CHANGES.md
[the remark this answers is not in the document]
jmfrancois: I have revert that breaking change

packages/cmf/BREAKING-CHANGES.md
jmfrancois: I have revert/re-add context.router on WithExpression

packages/containers/package.json
jmfrancois: no it s used by the SidePanel explicitly

packages/router/README.md
jmfrancois: context BR has been revert

output/router.eslint.txt
```diff
@@ -0,0 +1,10 @@
+The react/require-extension rule is deprecated. Please use the import/extensions rule from eslint-plugin-import instead.
+
+/home/travis/build/Talend/ui/packages/router/src/sagaRouter.js
+  163:54  error  There should be no spaces inside this paren  space-in-parens
+  177:51  error  There should be no spaces inside this paren  space-in-parens
```
jsomsanith: Please fix those errors

output/containers.eslint.txt
```diff
@@ -15,6 +15,9 @@ The react/require-extension rule is deprecated. Please use the import/extensions
 /home/travis/build/Talend/ui/packages/containers/src/Notification/Notification.test.js
   11:80  error  'notifications' is missing in props validation  react/prop-types
 
-✖ 5 problems (5 errors, 0 warnings)
+/home/travis/build/Talend/ui/packages/containers/src/Redirect/index.js
+  1:8  error  Using exported name 'Redirect' as identifier for default export  import/no-named-as-default
```
jsomsanith: Please fix this error

output/router.eslint.txt
jmfrancois: I can t it comes form prettier

PR #1844, opened 2018-11-22
feat(cmf): support multiple cmf settings

annilsson:
We need to be able to produce a build with different settings files than default. 

Support multiple settings objects in cmf.json

### DIFF
--- a/packages/cmf/scripts/cmf-settings.merge.js
+++ b/packages/cmf/scripts/cmf-settings.merge.js
@@ -17,6 +17,14 @@ const {
 
 const DEFAULT_CONFIG_FILENAME = 'cmf.json';
 
+function getCmfconfig(cmfconfigPath, onError) {
+	const cmfconfig = importAndValidate(cmfconfigPath, onError);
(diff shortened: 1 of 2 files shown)


packages/cmf-webpack-plugin/src/index.js
```diff
@@ -27,17 +27,26 @@ function ReactCMFWebpackPlugin(options = {}) {
 	};
 }
 
+function getCmfconfig(cmfconfigPath) {
+	const cmfconfig = require(cmfconfigPath);
+	if (process.env.CMF_ENV) {
+		return cmfconfig[process.env.CMF_ENV];
+	}
+	return cmfconfig;
+}
```
jmfrancois: it really looks like a duplicate.

annilsson: cmf-webpack-plugin has cmf as dependency, but cmf also invalidates the require cache in this function - cmf-webpack-plugin does not. It would be a change to cmf-webpack-plugin that I don't know is a good idea or not.

PR #1846, opened 2018-11-23
chore(TUI-307/CMF): DEPRECATED redux-storage

jmfrancois:

redux-storage API has never been exposed but has been used.
we want to remove it from CMF

First lets deprecate it.
Expose a new API way more simple to use to get the initialState.


packages/cmf/src/localStorage.js
```diff
@@ -0,0 +1,68 @@
...
+		source.cmf.components = Immutable.fromJS(source.cmf.components);
+	}
+	if (source.cmf.collections) {
+		source.cmf.collections = Immutable.fromJS(source.cmf.collections);
+	}
+	return source;
+}
+
+/**
+ * getStoreCallback read localStorage and create a initilState for redux
```
sgangnant:
```suggestion
 * getStoreCallback read localStorage and create a initialState for redux
```

test-setup.js
```diff
@@ -37,6 +37,10 @@ const fetch = jest.fn(
 );
 global.fetch = fetch;
 global.Headers = Headers;
+global.localStorage = {};
+global.localStorage.setItem = (key, value) => {
+	global.localStorage[key] = value;
+};
```
jsomsanith:
```javascript
global.localStorage = {
	setItem(key, value) {
		global.localStorage[key] = value;
	}
}
```

But, is it really necessary to mock that ? Perhaps the use of localStorage + `afterEach` to clear it will do the trick

packages/cmf/__tests__/localStorage.test.js
```diff
@@ -0,0 +1,69 @@
...
+		localStorage.setItem(KEY, undefined);
+	});
+	it('should getStoreCallback return a function', () => {
+		const callback = localStorageAPI.getStoreCallback(KEY, PATHS);
+		expect(typeof callback).toBe('function');
+		const realEventListener = window.addEventListener;
+		const mock = jest.fn();
+		window.addEventListener = mock;
+		callback({ getState: () => state });
+		window.addEventListener = realEventListener;
```
jsomsanith: Be careful, something breaks before, you don't set back the real event listener

packages/cmf/src/localStorage.md
```diff
@@ -0,0 +1,22 @@
...
+]);
+
+cmf.bootstrap({
+    preloadedState,
+    storeCallback,
+});
+```
+
+From now the redux-storage api provided in CMF is DEPRECATED.
+It will be removed in the next major release.
```
frassinier: How do you deal with this removal? Another TUI?

jmfrancois: good point, I have added a subtask for that and update description

PR #1848, opened 2018-11-26
chore(DatePicker): date only

jlame:
The time picker was always displayed

Adding a prop to show/hide it


packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -44,11 +47,13 @@ class InputDateTimePicker extends React.Component {
 		onBlur: PropTypes.func,
 		dateFormat: PropTypes.string,
 		useSeconds: PropTypes.bool,
+		useTime: PropTypes.bool,
 	};
 
 	static defaultProps = {
 		dateFormat: 'YYYY-MM-DD',
 		useSeconds: false,
+		useTime: true,
```
jsomsanith: We should not have time by default.
So the classical use is only date, and when we want time, we just write `<InputDateTimePicker useTime>`.
(And I think we should rename the component to `InputDatePicker`)

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -63,11 +68,16 @@ class InputDateTimePicker extends React.Component {
 		}
 
 		checkSupportedDateFormat(props.dateFormat);
+		const selectedDate =
+			props.useTime === false
```
jsomsanith: if default is `false` no need to test if it's explicitly false, just use the truthyness `props.useTime ? ... : ...`

packages/components/src/DateTimePickers/views/DateTimeView/DateTimeView.component.js
```diff
@@ -89,14 +102,7 @@ class DateTimeView extends React.PureComponent {
...
 				</div>
-				<div className={theme.time}>
-					<TimePicker
-						allowFocus={this.props.allowFocus}
-						value={this.props.selectedTime}
-						onChange={this.props.onSelectTime}
-						useSeconds={this.props.useSeconds}
-					/>
-				</div>
+				{this.getTimePicker()}
```
jsomsanith: Can you replace that by `this.props.useTime && this.getTimePicker()`. Then in getTimePicker, you don't need the ternary

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
jlame: i guess if that component exist its already used somewhere so renaming it my lead to breaking changes ?

jsomsanith: No it's not used, as it's still in progress. It's a breaking changes but as it `should` not be used (warning in constructor) and it is not used, it's ok to do this breaking change.
But no pb to do that in another PR

PR #1857, opened 2018-11-28
fix(components/SubHeaderBar): data-feature on title edition button

frassinier:
SubHeaderBar title edition button does not have `data-feature` attribute

Place it with a validated-by-PM-value


packages/components/src/EditableText/EditableText.component.js
```diff
@@ -54,14 +61,22 @@ export function EditableTextComponent({ editMode, loading, inProgress, ...rest }
 		return <Skeleton type={Skeleton.TYPES.text} size={Skeleton.SIZES.large} />;
 	}
 
-	const Component = editMode ? InlineForm : PlainTextTitle;
+	let Component;
+	let restWithoutFeature = { ...rest };
+	if (editMode) {
+		Component = InlineForm;
+		restWithoutFeature = omit(rest, 'feature');
```
jsomsanith: I would prefer the reverse. Instead of integrate and omit, what about destructure, and explicitly add the feature.
```
function EditableTextComponent({ editMode, loading, inProgress, feature, ...rest }) {
	...
	const  propsToSpread = { ...rest };
	if (editMode) {
		...
	} else {
		propsToSpread.feature = feature;
	}
```

annilsson: Or maybe even a function so it can be tested easily? the number of `let`'s are also increasing :) 
```
function getComponentProps(editMode, feature, rest) {
	if (editMode) {
		return { feature, ...rest };
	}
	return rest;
}
function EditableTextComponent({ editMode, loading, inProgress, feature, ...rest }) {
	...
	const Component = editMode ? InlineForm : PlainTextTitle;
	const componentProps = getComponentProps(editMode, feature, rest);
	...
	<Component inProgress={inProgress} {...componentProps} />
```

packages/components/src/EditableText/EditableText.component.js
```diff
@@ -70,7 +85,7 @@ export function EditableTextComponent({ editMode, loading, inProgress, ...rest }
 			})}
 			{...allyProps}
 		>
-			<Component inProgress={inProgress} {...rest} />
+			<Component inProgress={inProgress} {...restWithoutFeature} />
```
annilsson: I think `restWithoutFeature` could contain `feature` if editMode is false? If that's the case I think it's better to rename to something else `componentProps` or anything?

PR #1865, opened 2018-11-30
fix(components/VirtualizedList): hide key/value pairs when there is no value

frassinier:
I want to hide key/value pair when the value does not exist (only for expanded list view).
But, for now, expanded list view presents non-natural order which can produce issue when you want to hide one or many cells.

image 10

Fix flow order and be able to hide cell when the value does not exist.

image


annilsson:
This is different from the guidelines images so we should also update the guidelines if we do this? http://guidelines.talend.com/document/92132#/navigation-layout/list

frassinier:
@annilsson order is not defined in the guidelines? 
Summon @tlnd-clalande, @AnLefevre and @mhuet 🙏

packages/components/src/VirtualizedList/RowLarge/RowLarge.component.js
```diff
@@ -28,19 +29,25 @@ class RowLarge extends React.Component {
...
-				</dt>
-				<dd className={theme['field-value']} title={tooltip}>
-					{cellContent}
-				</dd>
-			</div>
-		);
+		const rawContent = getRowData(parent, index);
+		const dataKey = getDataKey(field);
+		const value = rawContent[dataKey];
+		if (value != null && `${value}`.trim().length) {
```
jdescottes-talend: Why checking if need to display based on `rawContent` when its `cellContent` which is used ?

frassinier: In my case, `cellContent` contains extra div with empty content.
I can't lead on.

jdescottes-talend: yep nevermind, I didn't chek the apis ;)

tlnd-clalande:
@frassinier (checked with @mhuet)
The order is Left to the right and back to line (like your 2nd screen).

Question ? is ne number of "column" is responsive ? or is it only 3 ?

frassinier:
@tlnd-clalande 
>The order is Left to the right and back to line (like your 2nd screen).
Thanks! 🎉 

>Question ? is ne number of "column" is responsive ? or is it only 3 ?
Only 3 at the moment, nothing was changed

packages/components/src/VirtualizedList/RowLarge/RowLarge.component.js
```diff
@@ -28,19 +29,25 @@ class RowLarge extends React.Component {
...
+					<dt key={fieldIndex} className={theme['field-label']}>
+						{label}
+					</dt>
+					<dd className={theme['field-value']} title={tooltip}>
+						{cellContent}
+					</dd>
+				</div>
+			);
+		}
+		return null;
```
jmfrancois: please escape as soon as possible and finish by returning the JSX

frassinier:
👋 @jdescottes-talend and @jmfrancois 

I took your change requests into account

PR #1867, opened 2018-12-02
feat(componentForm): dispatch failed event when submit is failed

romainseb:
Currently, we have a success event when the tcomp form is ok but we have no knowledge when there is a failure
Add an event to know when the submit has failed


packages/containers/src/ComponentForm/ComponentForm.sagas.js
```diff
@@ -75,13 +75,18 @@ function* onFormSubmit(componentId, submitURL, action) {
 	}
 	const { response, data } = yield call(cmf.sagas.http.post, submitURL, action.properties);
 	if (!response.ok) {
-		return;
+		yield put({
```
sgangnant: I don't see unit tests for `onFormSubmit` generator, any particular reason ?

frassinier: same `put`, DRY, just switch type value regarding the response status

packages/containers/src/ComponentForm/ComponentForm.sagas.js
```diff
@@ -75,13 +75,18 @@ function* onFormSubmit(componentId, submitURL, action) {
...
+			type: Component.ON_SUBMIT_FAILED,
+			data,
+			componentId,
```
jmfrancois: please add element from the response in it so we can build any effects on it

romainseb: I add the response :) good call

PR #1868, opened 2018-12-03
fix(UIForm): array template error and description linked with list

jsomsanith:
Arrays can have errors and descriptions.
For example, if an array can only have 3 elements max, the error will be displayed. But those are not linked to the array in html.
This is an accessibilty issue and it gives required proptypes warning.

Link the error/description to the array.


packages/forms/src/UIForm/fieldsets/Array/DefaultArrayTemplate.component.test.js
```diff
@@ -107,4 +107,25 @@ describe('Default Array Template component', () => {
...
+				renderItem={index => <div>Render item {index}</div>}
+				schema={schema}
+				value={value}
+				errorMessage={'This is an error'}
+				isValid={false}
+			/>,
+		);
+
+		// then
+		expect(wrapper.getElement()).toMatchSnapshot();
```
jdescottes-talend: Prefer expecting specific things (like having an errorId and descriptionId in the Message rendered) than a snapshot.
To avoid having to update n unrelated snapshots for just a minor change. And also not knowing what to look at in the snapshot to check if the expectation is fulfilled.

jsomsanith:
@jdescottes-talend yes that's intentional because it is not really easy to spot it when there are a lot of items. And basically the errors will only be on the number of elements. So it happens on add button click. In the default template, the button is next to the title, on top, I put the error message here as well.
There is no guideline, so that's my decision. If it doesn't make sense to you, we can discuss the reasons.

PR #1869, opened 2018-12-03
chore: fix warnings

jsomsanith:
There are warnings in talend forms because
* schema.key is an array that not only contains string (array index for example)
* TextArea uses schema.key to set the name attribute, but in array of text, name = index, with doesn't make sense, and it expects a string
* Array expects an array of object as value, but it can be an array of anything (object, number, string, boolean, ...)
* ArrayItem have missing keys in the children
* Collapsible fieldset pass empty string as icon transform when there is no transformation, but an enum is defined. Plus it uses classnames lib to avoid writting an if :/

Fix all that


packages/forms/src/UIForm/fieldsets/Array/Array.component.js
```diff
@@ -149,7 +149,9 @@ if (process.env.NODE_ENV !== 'production') {
 		onFinish: PropTypes.func.isRequired,
 		schema: PropTypes.object.isRequired,
 		templates: PropTypes.object.isRequired,
-		value: PropTypes.arrayOf(PropTypes.object).isRequired,
+		value: PropTypes.arrayOf(
```
acateland: maybe boolean to ?

packages/forms/src/UIForm/fieldsets/CollapsibleFieldset/CollapsibleFieldset.component.js
```diff
@@ -33,7 +33,7 @@ export default function createCollapsibleFieldset(title = defaultTitle) {
 		render() {
 			const { id, schema, value, ...restProps } = this.props;
 			const { items } = schema;
-			const iconTransform = classNames({ 'flip-vertical': !this.props.value.isClosed });
+			const iconTransform = !this.props.value.isClosed ? 'flip-vertical' : undefined;
```
acateland: i think `null` would better convey the fact that iconTransform was intentionally set to this value instead of maybe being an error.

packages/forms/src/UIForm/fieldsets/Array/Array.component.js
jsomsanith: Oh thanks

jdescottes-talend: I was wondering if it's not one case to use `any` as it could really be an array of anything, even an array of array potentially...

jsomsanith: Deal, changed, pushed

PR #1870, opened 2018-12-03
fix(cmf/http): don't stringify if not json type

ncomont:

`JSON.stringify` is applied on the payload, even if the content type does not contain `json`.

Check if the content-type contains `json` before calling `JSON.stringify` on it.


packages/cmf/src/sagas/http.js
```diff
@@ -117,26 +117,29 @@ export function httpFetch(url, config, method, payload) {
...
+			method,
+		},
+		{
+			...HTTP.defaultConfig,
+			...config,
+		},
+	);
+
+	const type = params.headers['Content-Type'];
+	if (type && type.includes('json')) {
```
jsomsanith: AFAIK, the only content type is `application/json`

ncomont: I don't understand your point. There are tens of HTTP valid content-type, so even if Talend recommends to use `application/json`, I don't see any reason to forbids the use of the other types.
`application/json` is already the default content-type, I guess this is the maximum we could do for that.

packages/cmf/src/sagas/http.js
```diff
@@ -117,26 +117,29 @@ export function httpFetch(url, config, method, payload) {
...
+			headers: defaultHeaders,
+			method,
+		},
+		{
+			...HTTP.defaultConfig,
+			...config,
+		},
+	);
+
+	const type = params.headers['Content-Type'];
```
acateland: i would see this as a function to pipelined with the others

PR #1897, opened 2018-12-11
chore: move current talend/scripts into a webapp folder

jsomsanith:
Talend/scripts will contains other scripts, not related to webapp.
But the folder hierarchy is not really friendly for that.

Move current scripts into a webapp/ folder


packages/scripts/README.md
```diff
@@ -1,140 +1,3 @@
...
-```
-
-The issue is well known, still not fixed in the lib or any of its dependencies.
-This [github issue](https://github.com/facebook/jest/issues/1767) brings you a lot of info.
-
-As a workaround, consider installing [`watchman`](https://facebook.github.io/watchman/).
-```
-brew install watchman
-```
+[Webapp scripts](./webapp/README.md)
```
jmfrancois: this is way too much empty. 
I would like here a list of all sub commands.

packages/scripts/index.js
```diff
@@ -1,44 +1,57 @@
...
 		break;
-	case 'test':
-		result = require('./scripts/test')(env, presetApi, options);
+	}
+	case 'test': {
+		const { env, presetApi } = getWebappEnv();
+		result = require('./webapp/scripts/test')(env, presetApi, options);
 		break;
+	}
 	default:
```
jmfrancois: Here we should do the required based on the case, so you don t have to modify that file to add a new command.

jsomsanith: *Draw card* --> Out of scope :)

PR #1905, opened 2018-12-13
feat(cmf-cqrs): be able to configure absolute url

frassinier:
Be able to create CQRS WS middleware only with relative path
 
Be able to configure absolute URL


packages/cmf-cqrs/src/middleware/socketMiddleware.test.js
```diff
@@ -27,6 +27,20 @@ const state = {
 	},
 };
 
+describe('hasWebSocketProtocol', () => {
+	it('should return true for Web Socket url', () => {
+		expect(isWebSocketAbsoluteUrl('ws://provider/api/v1/ws/relay')).toBeTruthy();
```
acateland: i think you should test against `toBe(true)`, truthy can be many thing else than a true boolean

PR #1909, opened 2018-12-13
feat(Scripts): i18n to github

jsomsanith:
Next step in i18n target: upload to github

Implement that + doc about how the locales repo is managed

### DIFF
--- a/packages/scripts/internationalization/README.md
+++ b/packages/scripts/internationalization/README.md
@@ -186,7 +186,7 @@ This scripts will extract the version (major.minor) from files at project root.
 * package.json
 * pom.xml
 
-The i18n files will be pushed to a branch `release/{XTM_project}/{version}`, and a tag `{XTM_project}/{version}` is set at each new push.
+The i18n files will be pushed to a branch `{XTM_project}/major.minor.x`, and a tag `{XTM_project}/major.minor` is set at each new push.
 
 With this branch and tag structure, the locales for your project can be downloaded from a defined url.
 ```
--- a/packages/scripts/internationalization/README.md
+++ b/packages/scripts/internationalization/README.md
@@ -13,7 +13,7 @@ The i18n process has multiple steps
 4- Download translated files to the bundle
 
 
-The talend-i18n module targets steps 1 - 2 - 3.
+The talend-scripts module targets steps 1 - 2 - 3.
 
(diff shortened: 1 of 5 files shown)


packages/scripts/internationalization/README.md
```diff
@@ -149,3 +150,50 @@ This step will download, unzip and transform files from XTM.
...
+| Configuration | Description |
+|---|---|
+| url | The https git url. |
+
+This scripts will extract the version (major.minor) from files at project root.
+* lerna.json
+* package.json
+* pom.xml
+
+The i18n files will be pushed to a branch `release/{XTM_project}/{version}`, and a tag `{XTM_project}/{version}` is set at each new push.
```
jmfrancois: avoid `/` in tag name please prefer `-` and may be prefix with v to make it very different from branch name.
May be the branch can be different from release but just be `{XTM_project}/{version}`

packages/scripts/internationalization/scripts/to-github.js
```diff
@@ -0,0 +1,204 @@
...
+}
+
+function checkI18nFolder(i18nFolder) {
+	printSection('Check i18n files');
+	if (
+		!fs.existsSync(i18nFolder) || // folder doesn't exist
+		!fs.lstatSync(i18nFolder).isDirectory() || // is not a folder
+		!fs.readdirSync(i18nFolder).length // folder is empty
+	) {
+		error(`
```
jmfrancois: Here I would prefere dedicated error.
Something like
```
`no folder found at ${i18nFolder}. Are you sure you have run "talend-scripts download" before ?`
```

packages/scripts/internationalization/README.md
jsomsanith: avoid `/` --> why ? We do that in Talend since a long time.
I added the `release` because it collide with tag name.

jmfrancois: This create folders.
Branches and tags are both reference. We should in that particular case keep all the references under the XTM_project.

`{XTM_project}/2.6.1` for tag
`{XTM_project}/2.x.x` for branch

jsomsanith: Done for branch name. Tag name stays `{XTM_project}/2.6`

PR #1915, opened 2018-12-17
feat(Components): Give possibility to inject directly a component reference

jdescottes-talend:
Slots injection in `components` package with `Inject.all` require to retrieve the component usage definition with `getComponent` so with some kind of registry.
We need a way to directly give the component to inject without fetching it somewhere else if we have a direct reference to it.

Directly use the data given in the `components` prop injection config if it's a React valid component.


packages/components/src/Inject/Inject.test.js
```diff
@@ -165,6 +165,15 @@ describe('Inject.all', () => {
 		};
 		expect(Inject.all(null, components)()).toEqual(null);
 	});
+	it('should return a function which return a react component', () => {
+		const getComponent = jest.fn();
```
jmfrancois: please add expect it has not been called

jsomsanith: I don't think this is necessary. You just created it here, without passing it to something that is executed.
Checking that it has not been called here would be equivalent to 
```javascript
const obj = { value: 'lol' };
expect(obj.value).toBe('lol'); // not really useful

execute(obj);
expect(obj.value).toBe('other');
```

PR #1917, opened 2018-12-17
feat(Scripts): extract by expression

jsomsanith-tlnd:
talend-scripts i18n-extract has no way to get all files that match an expression.

Add this extract mode, that will get all files with
* a provided expression supporting `*` wildcard
* a provided folder to start the search

### DIFF
--- a/packages/datagrid/src/components/DataGrid/DataGrid.scss
+++ b/packages/datagrid/src/components/DataGrid/DataGrid.scss
@@ -28,7 +28,7 @@ $td-datagrid-skeleton-height: 9.6rem !default;
 	}
 
 	/* ==== HEADER ==== */
-	:global(.ag-header-container) {
+	:global(.ag-header-viewport) {
 		background-color: $white;
 		color: $dove-gray;
 	}


packages/scripts/internationalization/scripts/extract.js
```diff
@@ -61,6 +61,22 @@ function extractFiles({ files, target }) {
 	});
 }
 
+function extractByExpression({ expression, rootPath, target }) {
+	const child = spawn.sync('find', [rootPath, '-iname', `${expression}`]);
```
jmfrancois: does find exists on windows ?

packages/scripts/internationalization/README.md
```diff
@@ -92,6 +93,25 @@ There are 3 methods of extraction
...
+    "target": "./i18n"
+  }
+}
+```
+
+| Configuration | Description |
+|---|---|
+| method | `expressionn` |
+| rootPath | The folder to start the search. |
+| expression | Expression describing the files name. You can use `*` as wildcard. `message*.json` means every files that start with `message` and end with `.json` (ex: message-errors.json, message.json, ...), whereas `message.json` matches only the `message.json` files. |
```
jmfrancois: please provide a link to what is possible and not possible to write as expression here. regexp ? May be a link to find expression option.

packages/scripts/internationalization/scripts/extract.js
jsomsanith: Nope, but it's not the only shell command that is done. I used the simpler way (shell or node) depending on the need.
The scripts will be executed in CI on linux distributions, I chose the easiest way to write it, no windows compatibility

packages/scripts/internationalization/README.md
jsomsanith: This is not a regex. If we need it, let's consider it, but for now it's only what is written here, the *, managed by find command

PR #1918, opened 2018-12-18
fix(UIForm/Datalist): do not add empty value in title map

jsomsanith:
In datalist widget, when value is empty (`''`), an empty value is added in the title map.

Do not add the entry when it is empty.

If you were
* passing an renderItemContainer props to `Typeahead` or `Datalist`
* rely on children props passed to this renderer
It won't render anymore if the value is empty (`''`). In this case, children is empty, so you have to replace your implementation with a simpler one : render prop passed as children.

```diff
-function renderItemsContainer({ children }) {
-    return (
-        <div>
-            {children}
-            {children && <button>Click</button>}
-        </div>
-    );
-}

<Typeahead
    {...props}
-    renderItemsContainer={renderItemsContainer}
>
+    (content, {searching, loading, noResult, isShown}) =>(
+        <div>
+            {content}
+            {isShown && <button>Click</button>}
+        </div>
+    )}
</Typeahead>
```


packages/forms/src/UIForm/fields/Datalist/Datalist.component.js
```diff
@@ -89,7 +89,7 @@ class Datalist extends Component {
 
 		let titleMapFind = titleMap;
 
-		if (!restricted) {
+		if (!restricted && propsValue) {
```
jdescottes-talend: Can you check with an explicit condition rather than with truthy/falsy check ?
And also swap the condition to an early return.

jsomsanith:
@jmfrancois Looking at react-autowhatever code, I don't know how this is possible.
In the render, they call `renderItemsContainer()` with or without result, with props.children as items to render.

If you override this function to add the button, it should be rendered depending on how to implement the function.

packages/forms/src/UIForm/fields/Datalist/Datalist.component.js
jsomsanith: `propsValue` is a string, and we don't want to add it if it's empty or undefined.
--> basically it's a falsy check
I'll let that check, but I change the condition to have an early return

jdescottes-talend: It's just implicit vs explicit, it tells many things :
- we expect a string (no need to go check the props def) => yes a typed language would be more convenient ...
- it tells you have added this condition strictly for checking empty string and nothing else (null, undefined, whatever). And if the prop type change, it will need some revision to check if condition is handle properly. It's not hidden and will seem to be checked where it's not.

It's just about communicating intention explicitly and avoid useless thinking.
But whatever, yes it's working

jsomsanith:
Hmm ok so we need to provide the empty info to the renderItemContainer through a `renderItemContainerFactory` to be able to differentiate the noMatch values and the not opened suggestions.

PR #1921, opened 2018-12-20
feat(components): Add rounding label on PieChart

romainseb:
Currently, we have some issues when we show some pie chart aside -> 
image
To fix this behaviour, the chosen solution is to round the percentage & if the value is < to 1% show "< 1%" & if it's > to 99%, show "> 99%"
This implementation doesn't change nothing if we already use rounded values


packages/components/src/PieChart/PieChartIcon.component.js
```diff
@@ -191,18 +191,35 @@ export function getPercentageToIndex(values, index) {
...
+ * @param {number} percentage percentage to show as label
+ */
+function getPercentageAndPrefix(percentage) {
+	if (percentage > 0 && percentage < 1) {
+		return [1, '< '];
+	} else if (percentage > 99 && percentage < 100) {
+		return [99, '> '];
+	}
+	return [Math.round(percentage), ''];
+}
```
jmfrancois: just a comment, do as you wish but I would return prefix as first argument.

PR #1923, opened 2018-12-20
feat(scripts): i18n version

jsomsanith:
To avoid the creation of N projects for N versions, an interesting idea came out about wrapping the i18n files in a version folder.

*Extract*
The transform is done here instead of in the download step.
Wrap extracted files in a version folder.

*Upload*
XTM will receive a different path for a different version, considering it like a new file, or overriding the existing files of the same version.

*Download*
If there is a version detected, it will download only the files from the wanted version, based on the file name (containing the path, so the version).
Otherwise it downloads all the files.
A transformation is done after download to have a final folder hierarchy
```
i18n-from-xtm
    1.10
            en_US
                    file1.json
                    file2.json
            fr_FR
                    file1.json
                    file2.json
    1.11
            en_US
                    file1.json
                    file2.json
            fr_FR
                    file1.json
                    file2.json
```

To perform the download on selected files, we have to pass the file ids as ... query params. The url is limited in character length, so we perform the download in multiple parts, handling up to 100 files per chunk.

*To-github*
It handles each version from download step, creating each branch, and pushing the files.


packages/scripts/internationalization/common/xtm.js
```diff
@@ -110,40 +116,89 @@ function uploadFile(data) {
...
-		.then(() => data);
+		.then(() => printSuccess(`Translations downloaded to ${filePath}`));
+}
+
+function downloadFiles(data) {
+	const { targetPath, project } = data;
+
+	printRunning('Download files from XTM project...');
+	rimraf.sync(data.targetPath);
+	shell.mkdir('-p', targetPath);
```
jmfrancois: you have mkdirp as deps, any reason not to use it here ?

PR #1927, opened 2018-12-21
fix(datagrid): update serialize to handle quality or not

jare-talend:
Currently, the serializer handle data with quality. We show the data with or without quality. Rather add a custom serializer to do that, add it on the current serializer.

Also, the type can be an array of string or an array of object. It is weird but that is the current response from the backend, they don't change that so the client should be able to manage this two way.

fix that


packages/datagrid/src/components/DatasetSerializer/datasetSerializer.js
```diff
@@ -67,10 +85,7 @@ export function getTypeValue(type, optional) {
  */
 export function getType(type) {
 	if (Array.isArray(type)) {
-		return getTypeValue(
-			type.find(subType => subType.type !== 'null'),
-			type.find(subType => subType.type === 'null'),
-		);
+		return getTypeValue(type.find(negate(isNull)), type.find(isNull));
```
nicomaligne: This seems odd...
Can you explain ?

packages/datagrid/src/components/DatasetSerializer/datasetSerializer.js
```diff
@@ -55,9 +56,26 @@ export function sanitizeAvro(avro) {
...
  */
 export function getTypeValue(type, optional) {
+	if (typeof type === 'string') {
+		return `${type}${optional ? '' : '*'}`;
+	}
 	return `${type.dqType || type.type}${optional ? '' : '*'}`;
 }
 
+/**
+ * isNull - check if the type is null
```
jsomsanith: No need to repeat the function name
```suggestion
  * Check if the type is null
```

packages/datagrid/src/components/DatasetSerializer/datasetSerializer.js
```diff
@@ -55,9 +56,26 @@ export function sanitizeAvro(avro) {
...
+ *
+ * @param  {type} string|object type to check
```
jsomsanith:
```suggestion
  * @param  {string|object} type to check
```

packages/datagrid/src/components/DatasetSerializer/datasetSerializer.js
jsomsanith: Yep I had the same feeling.
From what I understand : 
- take the type label of the first non null type
- it is mandatory if there is no type that is "null"

Perhaps you can name functions better
```javascript
return getTypeValue(getFirstNonNull(type), canBeOptional(type));
```

But is it ok to get the label of only the first type ?

packages/datagrid/src/components/DatasetSerializer/datasetSerializer.js
```diff
@@ -88,20 +103,28 @@ export function getQuality(qualityTotal, rowsTotal) {
  */
 export function getQualityValue(type) {
 	if (isArray(type)) {
-		return type.find(value => value[QUALITY_KEY] !== undefined)[QUALITY_KEY];
+		return get(type.find(value => value[QUALITY_KEY] !== undefined), QUALITY_KEY, null);
```
jsomsanith: I think you don't need to set `null` as default value.
(same for next line)

jsomsanith: And why do you use lodash `get()` ? The only reason I can see is that type can be null/undefined ?

packages/datagrid/src/components/DatasetSerializer/datasetSerializer.js
```diff
@@ -55,9 +56,26 @@ export function sanitizeAvro(avro) {
...
 	return `${type.dqType || type.type}${optional ? '' : '*'}`;
 }
 
+/**
+ * isNull - check if the type is null
+ *
+ * @param  {type} string|object type to check
+ * @return {boolean} return if the type is null
+ */
+export function isNull(type) {
```
jsomsanith: Why do you export it ?
It's supposed to be internal. If it's for the tests, this is useless, don't export internal things for tests, they are tested with the exposed api implicitly.
Doing that makes it less maintainable, as you can change the implementation by refactoring internal code, it will break the tests even if there is no impact on exposed api.

packages/datagrid/src/components/DatasetSerializer/datasetSerializer.test.js
```diff
@@ -246,6 +248,13 @@ describe('#getTypeValue', () => {
 	it('should return the dqType', () => {
 		expect(getTypeValue({ type: 'hello', dqType: 'world' }, true)).toEqual('world');
 	});
+
+	it('should return the type with a star (type is string)', () => {
+		expect(getTypeValue('string')).toEqual('string*');
+	});
+	it('should return the optionnal type (type is string)', () => {
```
jsomsanith: Typo
```suggestion
 	it('should return the optional type (type is string)', () => {
```

packages/datagrid/src/components/DatasetSerializer/datasetSerializer.test.js
```diff
@@ -368,8 +377,18 @@ describe('getQuality', () => {
 });
 
 describe('getFieldQuality', () => {
-	it('should enrich the quality', () => {
-		expect(getFieldQuality(sample.schema.fields[0][QUALITY_KEY])).toMatchSnapshot();
+	it('should enrich the quality with the quality if exists', () => {
```
jsomsanith:
```suggestion
 	it('should compute quality metrics if there are quality values', () => {
```

packages/datagrid/src/components/DatasetSerializer/datasetSerializer.test.js
```diff
@@ -368,8 +377,18 @@ describe('getQuality', () => {
...
+		expect(getFieldQuality(sample.schema.fields[0].type)).toEqual({
+			'@talend-quality@': {
+				'-1': { percentage: 62, total: 62 },
+				0: { percentage: 0, total: 0 },
+				1: { percentage: 38, total: 38 },
+			},
+		});
+	});
+
+	it('should do nothing if the quality do not exists', () => {
```
jsomsanith:
```suggestion
 	it('should return if there are no quality metrics', () => {
```

packages/datagrid/src/components/DatasetSerializer/datasetSerializer.test.js
```diff
@@ -431,3 +450,21 @@ describe('convertSample', () => {
 		});
 	});
 });
+
+describe('isNull', () => {
```
jsomsanith: Remove that, and replace with other tests on the exposed api if necessary

packages/datagrid/src/components/DatasetSerializer/datasetSerializer.js
jare-talend: done

packages/datagrid/src/components/DatasetSerializer/datasetSerializer.js
jare-talend: exact i have undefined i no get, it is suffisant. get avoid to do redundant check if object exist, take this property

packages/datagrid/src/components/DatasetSerializer/datasetSerializer.js
```diff
@@ -55,9 +56,26 @@ export function sanitizeAvro(avro) {
  * @param {boolean} optional;
  */
 export function getTypeValue(type, optional) {
+	if (typeof type === 'string') {
+		return `${type}${optional ? '' : '*'}`;
+	}
```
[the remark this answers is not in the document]
jare-talend: done

packages/datagrid/src/components/DatasetSerializer/datasetSerializer.js
jare-talend: done remove it

PR #1929, opened 2018-12-21
fix(ObjectViewer): perf

acateland:
a json document containing a list of 50 object, those object 70 key take 1.2 Gig or ram and 1minute of processing on Table mode

Move the translate HOC where it is indeed used.
120 mo ram and 3 sec processing


packages/components/src/ObjectViewer/JSONLike/JSONLike.component.js
```diff
@@ -382,6 +385,8 @@ ComplexItem.propTypes = {
 	t: PropTypes.func,
 };
 
+export const ComplexItem = translate(I18N_DOMAIN_COMPONENTS)(ComplexItemUntranslated);
```
nicomaligne: Not found of ComplexItemUntranslated name.
The complex item is not exported and it's display name is complex item.
Every component will have the suffix untranslated with this logic.
Can we inverse the naming calling something translated than untranslated ?

PR #1933, opened 2018-12-28
fix: SubHeader and Dialog headings

jsomsanith:
Dialogs and Subheader titles should be h1, without custom font style.

* fix the markup
* remove style
* editable text accepts now a componentClass props

The text mode style in EditableText has changed. We don't apply any default font style because you can insert any component you'd like.
To have the same rendering as before you must style yourself the text

```css
.tc-editable-text-wording {
	color: $black;
	font-size: 1.4rem;
	font-weight: 900;
}
```

### DIFF
--- a/packages/components/src/EditableText/EditableText.component.js
+++ b/packages/components/src/EditableText/EditableText.component.js
@@ -17,12 +17,12 @@ export function PlainTextTitle({ componentClass, onEdit, disabled, text, inProgr
 	return (
 		<div className={theme['tc-editable-text-title']}>
 			<TooltipTrigger label={text} tooltipPlacement="bottom">
-				<componentClass
+				<ComponentClass
 					className={classNames(theme['tc-editable-text-wording'], 'tc-editable-text-wording')}
 					onDoubleClick={isDisabled ? undefined : onEdit}
 				>
 					{text}
-				</componentClass>
+				</ComponentClass>
 			</TooltipTrigger>
 			<Action
 				name="action-edit"
--- a/packages/components/src/EditableText/EditableText.component.js
+++ b/packages/components/src/EditableText/EditableText.component.js
@@ -11,17 +11,18 @@ import getDefaultT from '../translate';
 
 import I18N_DOMAIN_COMPONENTS from '../constants';
 
-export function PlainTextTitle({ onEdit, disabled, text, inProgress, feature, t }) {
+export function PlainTextTitle({ componentClass, onEdit, disabled, text, inProgress, feature, t }) {
 	const isDisabled = disabled || inProgress;
+	const ComponentClass = componentClass;
 	return (
 		<div className={theme['tc-editable-text-title']}>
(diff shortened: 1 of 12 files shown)


packages/components/src/EditableText/EditableText.component.js
```diff
@@ -37,17 +38,18 @@ export function PlainTextTitle({ onEdit, disabled, text, inProgress, feature, t
 		</div>
 	);
 }
-
 PlainTextTitle.propTypes = {
-	text: PropTypes.string.isRequired,
-	onEdit: PropTypes.func.isRequired,
+	componentClass: PropTypes.string,
```
jmfrancois: that means I can t use a componetn but only native. PLease add oneOf(string, func)

packages/components/src/EditableText/EditableText.component.js
```diff
@@ -11,17 +11,18 @@ import getDefaultT from '../translate';
...
 
-export function PlainTextTitle({ onEdit, disabled, text, inProgress, feature, t }) {
+export function PlainTextTitle({ componentClass, onEdit, disabled, text, inProgress, feature, t }) {
 	const isDisabled = disabled || inProgress;
+	const ComponentClass = componentClass;
 	return (
 		<div className={theme['tc-editable-text-title']}>
 			<TooltipTrigger label={text} tooltipPlacement="bottom">
-				<span
+				<ComponentClass
```
jmfrancois: you can t write ```<componentClass``` ?

packages/components/src/EditableText/EditableText.component.js
```diff
@@ -80,19 +82,20 @@ export function EditableTextComponent({ editMode, loading, inProgress, ...rest }
 EditableTextComponent.displayName = 'EditableText';
 
 EditableTextComponent.propTypes = {
-	text: PropTypes.string.isRequired,
+	componentClass: PropTypes.string,
```
annilsson: (non-blocking comment) Personally I find componentClass misleading and would rather see ElementType or something (with leading capital too). I know bootstrap uses it, but this component is our own and that way we don't confuse it with a bootstrap component anyway. What do you think about it?

packages/components/src/EditableText/EditableText.component.js
jsomsanith: No, react consider it as a native markup

packages/components/src/EditableText/EditableText.component.js
jsomsanith: I find it more confusing to have 2 different props. So user have to know if it's something from ui or from bootstrap to know which one to use.
Having the same one remove this possible confusion IMO

annilsson: I meant the name only, not necessarily splitting it into 2 props. `"h1"` is not a class, so the name made no sense to me

annilsson: So you still have to know if it's from bootstrap so you can go to bootstrap to see what it actually means :)

PR #1935, opened 2018-12-28
fix(datagrid): header with type string

jare-talend:
when the type is ['null', 'string'], the type is serialized like {type: 'string'} instead one object { type: { type: string }}

So, the cell can use the good renderer to render the data

fix that


output/datagrid.eslint.txt
```diff
@@ -1,3 +1,9 @@
 > eslint --config ../../.eslintrc src
 
 The react/require-extension rule is deprecated. Please use the import/extensions rule from eslint-plugin-import instead.
+
+/home/travis/build/Talend/ui/packages/datagrid/src/components/DatasetSerializer/datasetSerializer.js
+  79:8  error  'avroType' is assigned a value but never used  no-unused-vars
```
jsomsanith: Please remove the unused var

PR #1936, opened 2018-12-31
fix(UIFrom) : hidden error not clean

smouillour:
In case of an error exist for an hidden field, it's not possible to submit the form.

Validate the form with only visible field.

In validation.js:
- the functions validateArray and validateSimple return an object with, for field without error, a value equal to null (only visible field are validated)

in UIForm.component.js: 
- keep triggered errors, clean errors with value not defined, and clean error for hidden fields.


packages/forms/src/UIForm/UIForm.component.js
```diff
@@ -215,7 +215,20 @@ export class UIFormComponent extends React.Component {
 		const { mergedSchema } = this.state;
 		const { properties, customValidation } = this.props;
 		const newErrors = validateAll(mergedSchema, properties, customValidation);
-		const errors = { ...this.props.errors, ...newErrors };
+		Object.keys(this.props.errors).forEach(errorKey => {
```
jare-talend: If i understand, you want keet the reference, can you do that ?
```suggestion

		const errors = Object.assign(this.props.errors, newErrors);
```

smouillour: no because errors is define with only keys present on newErrors to skip potential new hidden field with error

packages/forms/src/UIForm/UIForm.component.js
```diff
@@ -215,7 +215,20 @@ export class UIFormComponent extends React.Component {
...
+			if (newErrors[errorKey] !== undefined) {
+				newErrors[errorKey] = this.props.errors[errorKey];
+			}
+		});
+		const errors = {};
+		Object.keys(newErrors).forEach(errorKey => {
+			if (newErrors[errorKey]) {
+				errors[errorKey] = newErrors[errorKey];
+			}
+		});
```
jsomsanith:
```javascript
const errors = Object.entries(newErrors)
    .filter(entry => entry[1])
    .reduce((accu, [key, value]) => { accu[key] = value; }, {});
```

smouillour: with that i have 2 lint errors:

Expected to return a value in arrow function
Assignment to property of function parameter 'accu'
it's ok for you to ignore them?

PR #1941, opened 2019-01-07
fix(tui-97/filterbar): set padding in tc-list-toolbar

jmfrancois:

The filterbar in the the list toolbar is not well aligned.
It create in some case an ugly visual effect.

update the margin when in the list toolbar

### DIFF
--- a/packages/components/src/FilterBar/FilterBar.scss
+++ b/packages/components/src/FilterBar/FilterBar.scss
@@ -83,7 +83,7 @@ $tc-filter-bar-width: 250px !default;
 	background-color: $tc-filter-bar-highlight-color;
 }
 
-:global(.tc-list-toolbar) .filter{
+:global(.tc-list-toolbar) .filter {
 	margin-top: 9px;
 	margin-bottom: 9px;
 }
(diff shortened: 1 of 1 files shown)


packages/components/src/FilterBar/FilterBar.scss
```diff
@@ -83,6 +83,11 @@ $tc-filter-bar-width: 250px !default;
 	background-color: $tc-filter-bar-highlight-color;
 }
 
+:global(.tc-list-toolbar) .filter {
```
jsomsanith: Shouldn't this be defined in list toolbar as it's a component placement within toolbar ?

jmfrancois: AFAIK @frassinier request on that point to use this approach to list all context where a component is used in the component.
So when you update a component you see all the context where is used and modified

frassinier: @jmfrancois I don't remember but if it's what I said, I would like to correct by the opposite: I prefer not using `:global`statement at all, and move it to ListToolbar.
With that, if you remove List (and its Toolbar), you can forget this piece of code and embed dead code.

packages/components/src/FilterBar/FilterBar.scss
```diff
@@ -83,6 +83,11 @@ $tc-filter-bar-width: 250px !default;
...
+	margin-top: 9px;
+	margin-bottom: 9px;
```
frassinier: `9px` looks like a bit magic! Are guidelines referring to? I can't reach it.
At least, could you use a variable with `rm` unit, please?

PR #1948, opened 2019-01-08
fix(containers): SidePanel and router

jmfrancois:

The SidePanel doesn t work as expected in 2.0
The ActionList component do not receive the getComponent props and
the router do not initialize the redux middleware to handle the `action.cmf.router[Replace|Push]`.

So even when the SidePanel dispatch the action no middleware treat it.

This is why I have put the two changes in the same PR.

fix the connect of ActionList to add getComponent props.
Add router middleware init.

### DIFF
--- a/packages/containers/src/index.js
+++ b/packages/containers/src/index.js
@@ -8,7 +8,11 @@ const components = Object.keys(allComponents).reduce((acc, key) => {
 		let options = {};
 		if (key === 'Layout') {
 			options = {
-				omitCMFProps: true,
+				withComponentRegistry: true,
+			};
+		}
+		if (key === 'ActionList') {
+			options = {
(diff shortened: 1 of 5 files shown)


packages/containers/src/index.js
```diff
@@ -8,7 +8,11 @@ const components = Object.keys(allComponents).reduce((acc, key) => {
...
 			options = {
-				omitCMFProps: true,
+				withComponentRegistry: true,
+			};
+		}
+		if (key === 'ActionList') {
+			options = {
 				withComponentRegistry: true,
 			};
 		}
```
ncomont: what about `if (key === 'Layout' || key === 'ActionList' )`, or `if (['ActionList', 'Layout'].includes(key)` ?

PR #1949, opened 2019-01-08
chore(DatePicker): form mode

jlame:
The picker is not very user friendly.

Adding auto close when date and validate button when time is being used.


packages/components/src/DateTimePickers/views/ViewLayout/ViewLayout.scss
```diff
@@ -31,10 +32,18 @@
 
 	.body {
 		flex: 1 1 auto;
+		padding: 0 15px;
+		padding-bottom: $padding-larger;
```
jsomsanith: Can you see with the UX if this is right ? I see it in guideline but the guideline spelcifies a space between the text and the bottom instead of between the day button and the bottom. And this space seems very big in practice.

packages/components/src/DateTimePickers/views/ViewLayout/ViewLayout.scss
```diff
@@ -4,6 +4,7 @@
 	flex-direction: column;
 
 	.header {
+		padding: 0 15px;
```
jsomsanith: Please use `$padding-normal`

packages/components/src/DateTimePickers/views/ViewLayout/ViewLayout.scss
```diff
@@ -31,10 +32,18 @@
 
 	.body {
 		flex: 1 1 auto;
+		padding: 0 15px;
```
jsomsanith: Please use `$padding-normal`

packages/components/src/DateTimePickers/views/ViewLayout/ViewLayout.scss
```diff
@@ -31,10 +32,18 @@
...
 	}
 
 	.element-container > * {
 		width: 100%;
 		height: 100%;
 	}
+
+	.footer {
+		align-items: center;
+		padding: 0 15px;
```
jsomsanith: Please use `$padding-normal`

packages/components/src/DateTimePickers/views/DateTimeView/DateTimeView.component.js
```diff
@@ -81,6 +83,22 @@ class DateTimeView extends React.PureComponent {
 		);
 	}
 
+	renderValidationButton(t) {
```
jsomsanith: Please avoid that. Extract it in another functional component (that can be in the same file), and wrap it into a translate HOC to get the t.
You can then remove the t from this component

packages/components/src/DateTimePickers/views/DateTimeView/DateTimeView.component.js
```diff
@@ -81,6 +83,22 @@ class DateTimeView extends React.PureComponent {
...
+		const validateLabel = t('DATEPICKER_VALIDATE_BUTTON', { defaultValue: 'Done' });
+		return (
+			<div className={theme.footer}>
+				<Action
+					name="action-datepicker-validate"
+					bsStyle="primary"
+					role="button"
+					label={validateLabel}
+					onClick={this.props.onFinish}
+					disabled={!this.props.isValid}
```
jsomsanith: You don't have any indication where the error is. Hard for user to fix. So IMO you should not disable it.
Let the error system do its job. (validate that with ux, exposing this point of vue)
If so, you don't need the isValid props anymore

packages/components/src/DateTimePickers/views/DateTimeView/DateTimeView.component.js
```diff
@@ -81,6 +83,22 @@ class DateTimeView extends React.PureComponent {
 		);
 	}
 
+	renderValidationButton(t) {
+		const validateLabel = t('DATEPICKER_VALIDATE_BUTTON', { defaultValue: 'Done' });
```
jsomsanith: Inline this instead of a new var

packages/components/src/DateTimePickers/views/DateTimeView/DateTimeView.scss
```diff
@@ -14,3 +14,9 @@
 		padding-left: $padding-normal;
 	}
 }
+
+.footer {
+	flex-grow: 1;
+	text-align: right;
+	padding: 10px;
```
jsomsanith: This introduce extra padding on left and right, making the button not aligned to the rest.
And use padding variables please.
`padding: $padding-small 0;`

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -39,6 +39,7 @@ class InputDateTimePicker extends React.Component {
 			PropTypes.number,
 			PropTypes.string,
 		]),
+		onFinish: PropTypes.func,
```
jsomsanith: If you introduce the button and the automatic finish if no time, you should change the way changes are passed to the outside world. You don't need a `props.onFinish()`

This means that 
* `this.onPickerChange` updates only the internal state + call `this.onFinish()` if no time
* `this.onFinish()` call `props.onChange()` and close the picker

This has an impact on form widget
* onChange : trigger form onFinish() if it comes from picker
* on input blur : trigger form onFinish()

There is another behavior that this button implies. It seems to be a validation button, so it means that if we don't click on it, but close the picker (esc for example), it should not apply the `temporary value` ? Let's check with UX on this one too.

jlame: for the last sentence i guess yes.

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -138,14 +141,23 @@ class InputDateTimePicker extends React.Component {
 					this.setPickerVisibility(true);
 				}
 				break;
+			case keycode.codes.enter:
+				// commiting existing change on enter
+				this.onFinish(event, 'INPUT');
```
jsomsanith: This is problematic because it adds additional behavior on buttons too. Each date/month/year is a button. And enter intends to press the button. Not submit the datetime-form

Furthermore, in practice, the enter close the picker and the selection is after, so it's equivalent to a cancel.

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -68,13 +70,15 @@ class InputDateTimePicker extends React.Component {
...
+		this.initialState = extractParts(props.selectedDateTime, this.getDateOptions());
 		this.state = {
-			...extractParts(props.selectedDateTime, this.getDateOptions()),
+			...this.initialState,
 			showPicker: false,
 		};
 
 		this.onBlur = this.onBlur.bind(this);
 		this.onFocus = this.onFocus.bind(this);
+		this.onFinish = this.onFinish.bind(this);
```
jsomsanith: Can you rename that to onSubmit ? As we talk about forms now

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -100,31 +104,30 @@ class InputDateTimePicker extends React.Component {
...
+		if (this.props.onChange && (this.state.datetimeUpdated || errorUpdated)) {
+			this.props.onChange(event, { errorMessage, datetime, textInput, origin });
+		}
 	}
 
 	onInputChange(event) {
 		const textInput = event.target.value;
 		const nextState = extractPartsFromTextInput(textInput, this.getDateOptions());
-		return this.onChange(event, nextState, 'INPUT');
+		const datetimeUpdated = this.dateHasChanged(nextState);
```
jsomsanith: Can you remove all those computations everytime there is a change ?
You have everything you need in this.state and this.initialState, you can compute it in `this.onChange()` to see if you trigger the onChange callback.
It's less error prone to have it in 1 place

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -234,6 +264,9 @@ class InputDateTimePicker extends React.Component {
 								useTime={this.props.useTime}
 								useSeconds={this.props.useSeconds}
 								useUTC={this.props.useUTC}
+								onFinish={this.onFinish}
+								isValid={!this.state.errorMessage}
```
jsomsanith: It's more an error message than a valid flag.
And discussed together, it will be a list of errors definitions

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -234,6 +264,9 @@ class InputDateTimePicker extends React.Component {
...
+								formMode={this.props.formMode}
```
jsomsanith: I think you can avoid passing `formMode` and `onSubmit` (onFinish).
It can be the responsibility of `InputDatePicker` to display the form mode.
- wrap the date picker with a form,
- insert the submit button in footer
- insert the error message in footer
- manage a proper form onSubmit attribute so it handles the time inputs enter naturally

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
jlame: i'll remove this for now as we don't plan to disable the button and it was just for that for the moment.

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
jlame: > insert the error message in footer
was thinking of doing that in another PR

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -203,41 +261,45 @@ class InputDateTimePicker extends React.Component {
...
-								}}
-								onSubmit={this.onPickerChange}
-								useTime={this.props.useTime}
-								useSeconds={this.props.useSeconds}
-								useUTC={this.props.useUTC}
-							/>
-						</Popover>
-					</Overlay>
-				</div>
+				<form onSubmit={this.onSubmit}>
```
jsomsanith: This should be here only if we are in formMode.
Otherwise we can have a form in a form if DateTimePicker is used in a form

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -203,41 +261,45 @@ class InputDateTimePicker extends React.Component {
...
+									manageFocus
+									selection={{
+										date: this.state.date,
+										time: this.state.time,
+									}}
+									onSubmit={this.onPickerChange}
+									useTime={this.props.useTime}
+									useSeconds={this.props.useSeconds}
+									useUTC={this.props.useUTC}
+									formMode={this.props.formMode}
```
jsomsanith: Why do you pass formMode to the DateTimePicker ?

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -168,6 +208,13 @@ class InputDateTimePicker extends React.Component {
 		this.openPicker();
 	}
 
+	onSubmit(event, origin) {
+		this.onChange(event, origin);
+		if (this.props.formMode) {
+			this.onBlur(event);
```
jsomsanith: Why do you call onBlur ?

packages/components/src/DateTimePickers/InputDateTimePicker/README.md
```diff
@@ -12,5 +12,5 @@ This component display an input with a datetime picker.
 | dateFormat       | Default `YYYY-MM-DD`. A combination of `YYYY` (year), `MM` (month), `DD` (day). Other format are consider invalid. |
 | useSeconds       | Display and convert the seconds |
 | useUTC           | Default `false` convert the date to UTC for the output and convert the input selectedDate (UTC) to retrieve the utc values. |
-
+| formMode         | Display the validation button and use the component in form mode (date selection/validation).
```
jsomsanith:
```suggestion
| formMode         | Date picker is a form. Only the click on a submit button will validate the change. |
```

packages/components/src/DateTimePickers/views/ViewLayout/ViewLayout.component.js
```diff
@@ -14,6 +14,7 @@ function ViewLayout(props) {
 				<div className={classNames(theme['element-container'], theme.right)}>{rightElement}</div>
 			</div>
 			<div className={classNames(theme['element-container'], theme.body)}>{props.bodyElement}</div>
+			<div className={theme.footer}>{props.footerElement}</div>
```
jsomsanith: You can revert all ViewLayout change (component, snapshot, css)

packages/components/stories/DatePicker.js
```diff
@@ -83,10 +83,55 @@ storiesOf('DatePicker', module)
...
+				Default date picker doesn't include time<br />
+				It comes with classical keyboard gesture.
+			</p>
+			<TestPickerWrapper
+				id="my-date-picker"
+				selectedDateTime={new Date(2018, 4, 13, 12, 30)}
+				onChange={action('onChange')}
+				onFinish={action('onFinish')}
+				name="Datetime"
+				formMode={true}
```
jsomsanith:
```suggestion
				formMode
```

packages/components/stories/DatePicker.js
```diff
@@ -83,10 +83,55 @@ storiesOf('DatePicker', module)
...
+			<form style={{ width: 320 }}>
+				<InputDateTimePicker
+					id="my-date-picker"
+					name="Datetime"
+					onBlur={action('onBlur')}
+					onChange={action('onChange')}
+					onFinish={action('onFinish')}
+					selectedDateTime={new Date(2018, 4, 13, 12, 30, 44)}
+					useTime
+					formMode={true}
```
jsomsanith:
```suggestion
					formMode
```

packages/components/stories/DatePicker.js
```diff
@@ -83,10 +83,55 @@ storiesOf('DatePicker', module)
...
 			/>
 		</div>
 	))
+	.add('Date picker form mode', () => (
+		<div>
+			<IconsProvider />
+			<h1>DatePicker</h1>
+			<p>
+				Default date picker doesn't include time<br />
+				It comes with classical keyboard gesture.
```
jsomsanith: Please update the description. Here we don't want to highlight that, but the form mode

packages/components/stories/DatePicker.js
```diff
@@ -83,10 +83,55 @@ storiesOf('DatePicker', module)
...
+				name="Datetime"
+				formMode={true}
+			/>
+		</div>
+	))
+	.add('Date picker - time form mode', () => (
+		<div>
+			<IconsProvider />
+			<h1>DatePicker with time</h1>
+			<p>You can require time with a simple "useTime" props.</p>
```
jsomsanith: Same here, please update the description, the form mode is important here
Perhaps you can merge the 2 stories about form mode to display 2 pickers (with/without time)

packages/components/stories/DatePicker.js
```diff
@@ -83,10 +83,55 @@ storiesOf('DatePicker', module)
 				id="my-date-picker"
 				selectedDateTime={new Date(2018, 4, 13, 12, 30)}
 				onChange={action('onChange')}
+				onFinish={action('onFinish')}
```
jsomsanith: This doesn't exist anymore
Remove that from all the instances here please

output/components.eslint.txt
```diff
@@ -2,6 +2,9 @@
 
 The react/require-extension rule is deprecated. Please use the import/extensions rule from eslint-plugin-import instead.
 
+/home/travis/build/Talend/ui/packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.test.js
```
jsomsanith: Please fix the linter error

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
jlame: forgot to remove it

PR #1951, opened 2019-01-09
fix(containers/ComponentForm): security config from saga http

frassinier:
CSRF info comes from props even if http saga is already configured.

Take http configuration into account if props don't contain security info.

### DIFF
--- a/packages/containers/src/ComponentForm/ComponentForm.component.js
+++ b/packages/containers/src/ComponentForm/ComponentForm.component.js
@@ -175,14 +175,15 @@ export class TCompForm extends React.Component {
 
 	setupTrigger(props) {
 		const config = cmf.sagas.http.getDefaultConfig() || {};
+		const defaultSecurity = config.security || {};
 		this.trigger = kit.createTriggers({
 			url: props.triggerURL,
 			customRegistry: props.customTriggers,
 			headers: config.headers,
 			lang: props.lang,
 			security: {
-				CSRFTokenCookieKey: props.CSRFTokenCookieKey,
-				CSRFTokenHeaderKey: props.CSRFTokenHeaderKey,
+				CSRFTokenCookieKey: props.CSRFTokenCookieKey || defaultSecurity.CSRFTokenCookieKey,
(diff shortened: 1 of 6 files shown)


packages/containers/src/ComponentForm/ComponentForm.component.js
```diff
@@ -175,14 +175,15 @@ export class TCompForm extends React.Component {
...
 		this.trigger = kit.createTriggers({
 			url: props.triggerURL,
 			customRegistry: props.customTriggers,
 			headers: config.headers,
 			lang: props.lang,
 			security: {
-				CSRFTokenCookieKey: props.CSRFTokenCookieKey,
-				CSRFTokenHeaderKey: props.CSRFTokenHeaderKey,
+				CSRFTokenCookieKey: props.CSRFTokenCookieKey || defaultSecurity.CSRFTokenCookieKey,
+				CSRFTokenHeaderKey: props.CSRFTokenHeaderKey || defaultSecurity.CSRFTokenHeaderKey,
```
jmfrancois: why not a default Props ?

jmfrancois: I would prefer the default to be handled in the createTriggers function.

frassinier: Yes, good point

packages/containers/src/ComponentForm/kit/createTriggers.test.js
```diff
@@ -87,6 +89,21 @@ describe('createTriggers', () => {
 			done();
 		});
 	});
+	it('should handle security specified by default', done => {
+		cmf.sagas.http.setDefaultConfig({
+			security: {
+				CSRFTokenCookieKey: 'otherToken',
+				CSRFTokenHeaderKey: 'X-CSRF-OTHER',
+			},
+		});
```
jmfrancois: you set a default value, you should unset it at the end of the test

frassinier: I've changed it in favor of using a mock. 👍

PR #1953, opened 2019-01-10
fix(frontend): TMC-15611 allow spaces (%20) on GIT url validation

yonmey:
https://jira.talendforge.org/browse/TMC-15611

We have had a case where the Azure repository urls can have a space between its name, this was not handled by the RegEx to validate the GIT urls in the UIForm custom formats.

I added the % character to the validation RegEx so that we allow it for this case.


jdescottes-talend:
This could have been checked specifically for percent followed by 2 hexadecimals numbers, like so : 
`([a-zA-Z0-9/.\-_]|%[0-9A-F]{2}])+`

PR #1955, opened 2019-01-11
fix: all target blank and window.open

jmfrancois:

https://dev.to/ben/the-targetblank-vulnerability-by-example

`target="_blank"` must be followed by `rel="noopener noreferrer"`

The ActionButton do not enforce that.

window.open must also be followed by

```
		opened.opener = null;
		opened.location = action.payload.url;
```

The HeaderBar is impacted

Fix HeaderBar
Enforce this pattern in ActionButton

### DIFF
--- a/packages/containers/src/HeaderBar/HeaderBar.sagas.test.js
+++ b/packages/containers/src/HeaderBar/HeaderBar.sagas.test.js
@@ -70,8 +70,7 @@ describe('HeaderBar sagas', () => {
 	});
 
 	describe('handleOpenProduct', () => {
-		const result = { opened: {} };
-		global.open = jest.fn(() => result.opened);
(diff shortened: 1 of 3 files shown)


packages/containers/src/HeaderBar/HeaderBar.sagas.test.js
```diff
@@ -70,17 +70,23 @@ describe('HeaderBar sagas', () => {
 	});
 
 	describe('handleOpenProduct', () => {
-		global.open = jest.fn();
+		const result = { opened: {} };
+		global.open = jest.fn(() => result.opened);
```
jsomsanith: You don't need to init those values here, it's done in beforeEach. Just declare `result`

PR #1960, opened 2019-01-14
chore(component/ActionBar): Prepare component for Count extraction

jdescottes-talend:
The `Count` component is strongly coupled to the `ActionBar` while it's used only for the `List`.
This make difficult to tune the `Count` component without forwarding unrelated props inside the `ActionBar`component.
Besides, there is some unrelated `Count` stuff and checks occurring each time the `ActionBar` is used anywhere else where there is clearly no need to it.

Start to extract the `Count` component to the top level layer of `ActionBar` with the help of `slot injection`. Make it ready to be moved in a next PR to the `List`.

N.B. I have intentionally not made additional tests for the slot injection for now. I prefer to do it in a next changes when all the `Count` references will be gone.
 

packages/components/stories/ActionBar.js
```diff
@@ -249,11 +249,12 @@ storiesOf('ActionBar', module)
 					selected={1}
 					actions={{ left: [primary] }}
 					multiSelectActions={{ right: [multi3] }}
+					getComponent="whatever"
 				/>
 			</div>
 			<p>Toolbar with btn-group and only icons/ Layout: left, center, right</p>
 			<div id="btn-group">
-				<ActionBar actions={btnGroupActions} />
+				<ActionBar actions={btnGroupActions} getComponent="whatever" />
```
sgangnant: `ActionBar.propTypes` mentions `func` for this prop, shouldn't it be updated or the story changed then ?

jdescottes-talend: Yep I was wondering about this prop too, it's not really used as we inject just a known React component. But it's actually necessary in the `injectAll` mechanic for now (no injection if no `getComponent`).
I will change here the story with an empty function to match the proptypes. And later I will make the correction in the injection mechanic in an other PR.

sgangnant: If it's a component, maybe `PropTypes.node` then :+1:

jdescottes-talend:
@sgangnant, I updated to remove the need for the `getComponent` injector. We now just need the `components` to inject if they already are React components.
No more prop warning in the stories :)
@jmfrancois, Can you check if the changes in the `Inject.all` seems right ?

PR #1966, opened 2019-01-14
feat(i18n): generate npm modules before pushing to github

jsomsanith:
To get back the translation files in each project, we are going to publish the translation files in a module (npm, mvn, etc).

- `i18n-to-github` is renamed to `i18n-deploy`
- it generates the required files before pushing to github
- not implemented, but the missing step is to publish the module

In this PR, only npm is done. Mvn is coming.


packages/scripts/internationalization/generators/npm/readme.generator.js
```diff
@@ -0,0 +1,60 @@
...
+i18n
+	.init({
+		ns: [
+			...<%= minimalName %>Namespaces,
+			...myProjectNamespaces
+		],
+		resources: merge(<%= minimalName %>Locales, myProjectLocales),
+	});
+
+export default i18n;
```
jmfrancois: Please add in the readme where the JSON are located, so if I want to merge assets to use fetch instead of require all langues I can do it.

PR #1987, opened 2019-01-25
feat(form): provide a skeleton while a trigger is loading

sgendre:
on a field with a trigger, the user doesn't have any feedback about the fact that smtg is going on (backend call), and then suddenly the rest of teh form is displayed

provide a loading feedback as soon as a trigger is needed


packages/forms/src/UIForm/utils/widgets.js
```diff
@@ -71,6 +73,7 @@ const widgets = {
 	radioOrSelect: RadioOrSelect,
 	resourcePicker: ResourcePicker,
 	toggle: Toggle,
+	loadingPlaceholder: TriggerSkeleton,
```
jmfrancois: skeleton would be better I guess

PR #1990, opened 2019-01-28
feat(TMC-15843): disabled item in datalist

VolodymyrKovalM:
https://jira.talendforge.org/browse/TMC-15843
In TMC we need a possibility to make disabled item in Datalist


packages/components/src/Datalist/Datalist.component.js
```diff
@@ -170,6 +170,10 @@ class Datalist extends Component {
 		if (this.props.multiSection) {
 			newValue = this.state.suggestions[sectionIndex].suggestions[itemIndex];
 		}
+		if (newValue.disabled) {
```
frassinier: We need to follow that PR I guess https://github.com/moroshko/react-autowhatever/pull/35

VolodymyrKovalM: Yes, also saw that PR, but it was done around 4 months ago (
But we can follow of course

VolodymyrKovalM: Yes, also saw that PR, but it was done around 4 months ago (
But we can follow of course

frassinier:
I validated even if I consider that disabled items should not be reachable by keyboard like a simple `<select>` does actually.

PR #1995, opened 2019-01-28
chore(cmf): remove componentWillReceiveProps

jsomsanith-tlnd:
We need to remove deprecated react 15 lifecycle methods.
This PR is only on cmf package.

- Remove the call from `componentDidMount()`
- Remove the call from `componentWillReceiveProps()`
- Replace them with a call in `render()` instead. There were no check on the 2 methods above, so the behavior is exactly the same :/


packages/cmf/src/Dispatcher.js
```diff
@@ -89,10 +60,19 @@ export class Dispatcher extends React.Component {
...
+				actionCreator.get(this.context, this.props[name]);
+			}
+		});
+	}
+
 	/**
 	 * @return {object} ReactElement
 	 */
 	render() {
+		this.checkIfActionInfoExist();
```
jmfrancois: AFAIK if you don t bind it it should not have access to this.props and this.context

jsomsanith: This is a method call, not a function call. So this is ok, it has the context

PR #2004, opened 2019-01-31
fix(TUI-332): make ComponentForm support error onTrigger

jmfrancois:

When a trigger issue a 500 the componentform fails on can t access attribute properties of undefined.
The onTrigger promise is resolve with a undefined data object.

Add usecase in sandbox to work on it
Add missing return in onError

Bonus:
If you use the widget "button" with a key it will receive the type define by the jsonschema. So It can get a type="text" which is bad.
Button should not be sth else than "submit", "button", "reset"


packages/containers/sandbox/src/assets/settings.json
```diff
@@ -100,13 +100,13 @@
...
       "triggerURL": "/api/v1/application/action",
-      "className": "well"
+      "className": "full-form"
     },
     "ComponentForm#validation": {
       "definitionURL": "/api/v1/forms/validation",
       "uiSpecPath": "ui",
       "triggerURL": "/api/v1/application/action",
-      "className": "well"
+      "className": "full-form"
```
jmfrancois: I have removed well class because it doesn't work with scrolling

PR #2008, opened 2019-02-01
feat(forms): add toggle widget in text mode

romainseb:
The toggle widget is not in text mode while the form is
Add a display mode to make it not editable
image


packages/forms/src/UIForm/fields/Toggle/displayMode/TextMode.component.js
```diff
@@ -0,0 +1,49 @@
...
+			</dd>
+		</div>
+	);
+}
+
+TextModeToggle.defaultProps = {
+	value: false,
+};
+
+if (process.env.NODE_ENV !== 'production') {
```
nicomaligne: Why make a condition on env on proptypes ?

nicomaligne: And none of this props are required ?

PR #2011, opened 2019-02-01
fix(components/Dialog): support icon in header

frassinier:
`Icon` cannot be displayed in `Dialog` header block

Fix that 👊 

https://company-57688.frontify.com/document/135457#/pages/sharing


packages/components/src/Dialog/Dialog.test.js
```diff
@@ -78,6 +84,10 @@ describe('Dialog', () => {
 		const wrapper = shallow(<Dialog {...subtitleProps}>{children}</Dialog>);
 		expect(wrapper.getElement()).toMatchSnapshot();
 	});
+	it('should render icon and subtitle', () => {
+		const wrapper = shallow(<Dialog {...iconAndSubtitleProps}>{children}</Dialog>);
+		expect(wrapper.getElement()).toMatchSnapshot();
```
annilsson: Instead of snapshot is it possible to simply check if `<Icon />` is present? 

Personally I feel like that has a higher chance to prevent regression than a snapshot, because nobody likes looking at snapshots 😄

jmfrancois:
If I remove subtitle 
screenshot 2019-02-04 at 09 12 47

PR #2013, opened 2019-02-03
fix: remove warn from console

romainseb:
There is some warns
Wrong prop passed to button in ActionIconToggle
CSSTransition.childContextTypes has to be declared, even empty since React 16 ( https://github.com/wikiwi/react-css-transition/issues/5 )
Skeleton props are not matched


packages/components/src/Actions/ActionIconToggle/ActionIconToggle.component.js
```diff
@@ -25,7 +25,7 @@ function ActionIconToggle(props) {
 				className={cn}
 				aria-label={label}
 				aria-pressed={active}
-				bsStyle=""
+				bsStyle="default"
```
jdescottes-talend: You sure it didn't break the style ? I tested that too but, the colors change a bit (different grey or blue) :s

romainseb: i have use link bsStyle to have the same behaviour now :)

PR #2016, opened 2019-02-04
fix(TUI-335/Tabs): hide tab panes legends

jsomsanith:
In forms tabs, the tab panes are fieldsets.
A fieldset must be with a legend (a11y). But in dtabs case, the legend is a duplication of the tab text.

Hide the fieldset legend in case of tabs using bootstrap `sr-only` classname.
- Fieldsets have a `hideTitle` option (we speak about titles, and not legends because the field in uiSchema is `title`
- Tabs force this option while building the tab pane fieldsets schema


packages/forms/src/UIForm/fieldsets/Fieldset/__snapshots__/Fieldset.component.test.js.snap
```diff
@@ -4,7 +4,9 @@ exports[`Fieldset widget should render fieldset 1`] = `
 <fieldset
   className="form-group"
 >
-  <legend>
```
jmfrancois: as title is required please update the tests

PR #2019, opened 2019-02-04
fix(components): add FocusManager to handle focus in datalist component

romainseb:
The datalist automaticly close in some case :
screen recording 2019-02-04 at 17 51 39

Create a FocusManager component that wrap some behaviour to not close when the focus is inside him.
This component could be used in the date picker

### DIFF
--- a/packages/components/src/Datalist/Datalist.component.js
+++ b/packages/components/src/Datalist/Datalist.component.js
@@ -6,6 +6,7 @@ import keycode from 'keycode';
 import get from 'lodash/get';
 import Typeahead from '../Typeahead';
 import theme from './Datalist.scss';
+import FocusManager from '../FocusManager';
 
 export function escapeRegexCharacters(str) {
 	return str.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
@@ -59,7 +60,6 @@ class Datalist extends Component {
 	 * @param event The blur event
 	 */
 	onBlur(event) {
-		this.resetSuggestions();
 		const { value, previousValue } = this.state;
 
 		if (value !== previousValue) {
@@ -174,6 +174,8 @@ class Datalist extends Component {
 			event.preventDefault();
 			return;
 		}
+
+		this.resetSuggestions();
 		this.updateValue(event, newValue, true);
 	}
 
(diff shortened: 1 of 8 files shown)


packages/components/src/Datalist/Datalist.component.js
```diff
@@ -351,21 +353,27 @@ class Datalist extends Component {
...
-				onFocus={this.onFocus}
-				onKeyDown={this.onKeyDown}
-				onSelect={this.onSelect}
-				theme={this.theme}
-				value={label}
-				caret
-			/>
+			<FocusManager
+				onFocusOut={() => {
+					this.resetSuggestions();
```
jsomsanith: Can you bind it in contructor please ?
```javascript
<FocusManager onFocusOut={this.resetSuggestions} />
```

packages/components/src/Datalist/Datalist.component.test.js
```diff
@@ -163,7 +163,6 @@ describe('Datalist component', () => {
 		// then
 		const payload = { value: 'foobar' };
 		expect(onChange).toBeCalledWith(expect.anything(), payload);
-		expect(wrapper.find(Typeahead).props().items).toBe(null);
```
jsomsanith: Can you update the test description accordingly please ? (no reset suggestion anymore)
Something like `should trigger change callback on blur ....`

packages/components/src/Datalist/Datalist.component.test.js
```diff
@@ -190,7 +189,6 @@ describe('Datalist component', () => {
 		// then
 		const payload = { value: 'foooo' };
 		expect(onChange).toBeCalledWith(expect.anything(), payload);
-		expect(wrapper.find(Typeahead).props().items).toBe(null);
```
jsomsanith: Same here

packages/components/src/FocusManager/FocusManager.component.js
```diff
@@ -0,0 +1,30 @@
...
+ * We loose the focus on the datalist field but we don't want to close the suggestions
+ * until we click outside of the component
+ */
+export default class FocusManager extends Component {
+	static propTypes = {
+		children: PropTypes.element.isRequired,
+		onFocusOut: PropTypes.func.isRequired,
+	};
+
+	onJSOFocus() {
```
jsomsanith: You let the name I set during the PoC :D
(same for blur)

packages/components/src/FocusManager/FocusManager.component.js
```diff
@@ -0,0 +1,30 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+
+/**
+ * This component handle some specifics case like when we click in the datalist
```
jsomsanith: Typo 
-`handle` -->`handles`
- `specifics case` --> `specific cases`

Your description is good as example. Can you add a little generic description before, explaining that this allows to  capture the blur that occurs outside of complex components ?

packages/components/src/FocusManager/FocusManager.test.js
```diff
@@ -0,0 +1,30 @@
+import FocusManager from './FocusManager.component';
+
+describe('FocusManager', () => {
+	it('should call onFocusOut if there is no onJSOFocus called', done => {
```
jsomsanith: `onJSOFocus` :)
(same for next test)

packages/components/src/FocusManager/FocusManager.test.js
```diff
@@ -0,0 +1,30 @@
+import FocusManager from './FocusManager.component';
+
+describe('FocusManager', () => {
```
jsomsanith: Instead of calling directly the class methods, can you set more real cases ?
- create test component 
- wrap it with FocusManager
- simulate a blur in the test component : check that the onFocusOut is called
- simulate a blur and focus in the test component : check that the onFocusOut is not called

packages/components/stories/Datalist.js
```diff
@@ -23,6 +23,19 @@ const propsMultiSection = {
 			suggestions: [{ name: 'My foobar', value: 'foobar', description: 'foobar description' }],
 		},
 		{ title: 'cat 4', suggestions: [{ name: 'My lol', value: 'lol' }] },
+		{
+			title: 'cat 1',
```
jsomsanith: I think changing this storybook is irrelevant here.
And you add the same existing values, which is not a supported case :D .
I think you can revert it

romainseb: Forgot i pushed it 🤦‍♂️

PR #2021, opened 2019-02-05
feat(forms/TUI-336): add updating feedback

jmfrancois:

When a trigger update fields (using `options`) we need to give a feedback to the enduser

add heartbeat on corresponding fields
add circularprogress to the clicked button

demo http://2021.talend.surge.sh/forms/?selectedKind=Core%20concepts&selectedStory=Core%20Concept%20updating


babel.config.js
```diff
@@ -1,4 +1,4 @@
-module.exports = function(api) {
+module.exports = function babel(api) {
```
jmfrancois: this to fix lint on the whole function

packages/containers/sandbox/mockBackend/kit.js
```diff
@@ -87,6 +88,7 @@ function suggestionBig() {
 		function onEnd() {
 			console.log('onEnd', body);
 			cache.photos = {
+				cacheable: true,
```
jmfrancois: fix the trigger return so we do not have loading again and again on the same field

packages/theme/src/theme/_forms.scss
```diff
@@ -57,7 +57,7 @@ form {
 
 		label {
 			&.control-label {
-				color: $silver-chalice;
+				color: $dove-gray;
```
jmfrancois: to respect the guideline

packages/forms/src/UIForm/Widget/Widget.component.js
```diff
@@ -20,6 +20,15 @@ function getWidget(displayMode, widgetId, customWidgets) {
 	return widget;
 }
 
+function isUpdating(updatingKeys = [], key) {
+	if (!updatingKeys.length || !key) {
```
sgendre:
```suggestion
	if (updatingKeys.length === 0 || !key) {
```

packages/forms/src/UIForm/fields/Select/Select.component.js
```diff
@@ -30,13 +39,14 @@ export default function Select({ id, isValid, errorMessage, onChange, onFinish,
 			label={title}
 			labelAfter
 			required={schema.required}
+			valueIsUpdating={valueIsUpdating}
```
sgendre: replacing by 'updatingValue' would have the same meaning

jsomsanith: `would have the same meaning` --> So let's not change it :D

PR #2027, opened 2019-02-07
doc: UIForm

jsomsanith:
Doc is not up to date and contains too much info

Split into many parts.
Add triggers that becomes quite complicated without strong documentation.


packages/forms/src/UIForm/doc/triggers.md
```diff
@@ -0,0 +1,181 @@
+# Triggers
+
+Triggers are advanced ways to controle properties and errors.
```
ncomont: to control

packages/forms/src/UIForm/doc/triggers.md
```diff
@@ -0,0 +1,181 @@
...
+
+*Warning*
+Use only triggers if you can't perform what you want with another feature listed in the [documentation](../README.md).
```
ncomont: Use triggers only if you can't

packages/forms/src/UIForm/doc/how-to-use.md
```diff
@@ -0,0 +1,116 @@
...
+}
+```
+
+Changing UIForm's `props.data` will replace the existing pieces in the form.
+
+So you need to synchronize the UIForm's state with your own state system.
+
+## Lifecycle
+
+![lifecycle schema](http://www.plantuml.com/plantuml/png/ZP11IyGm48NlyolcBEp1rLCFibV1Wx17yH2RqGuaQKYcBYB-TsEJ3gq8zLHU-jx7bvooc2IJTp2m9IGVvGCjvJvn51u3F5qzu-5mX_U1XOqxYn7ga6l2rp5vFFS3UGFh1cgbnao2ClvdvlYTTZTG6vlffwhE-RghYtumuOsWdgtnKxJbO_72ClnHdKc529eNycqZ9og6i1mzhBrdw4QlzN21RboFZPYd5yyDULm_uGQhLrYAQleBUuDTMJY7YNeg8unfZ1dglqxrb1xeylZJQ3BrgMjRxxTVLPqlE56matVV)
```
jmfrancois: I would be more confident to use local text file transformed. You can use vscode plantuml plugin to achieve this.

packages/forms/src/UIForm/doc/triggers.md
```diff
@@ -0,0 +1,181 @@
...
+
+## What are triggers ?
+
+Triggers are asynchronous calls that you can perform in your custom widgets.
```
jmfrancois: You do not need custom widgets to use or understand triggers (do not mix widget dev and form user)

What do you think of
```suggestion
Triggers are custom functions called during event handling. They support asynchronous calls using Promises.
````

File: packages/forms/src/UIForm/doc/triggers.md [hunk 5]
```diff
@@ -0,0 +1,181 @@
...
+You can do whatever you want outside of the forms, such as
+- getting info from outside, REST calls, etc
+- perform advanced validation
+- change value in another widget
+- ...
+
+## TD;LR
+
+To create a trigger you need
+- (widget) to define events that will run the trigger
```
jmfrancois: trigger can be called by any widget. I don t understand the point here

packages/forms/src/UIForm/doc/how-to-use.md
jsomsanith: I just copied/pasted the existing schema

packages/forms/src/UIForm/doc/triggers.md
jsomsanith: Depending on the widget implementation, it defines a set of triggers that will be called.
For example, a widget will trigger on blur, but another won't.

Here I try to explain that there are
- a set of trigger events a widget supports
- a set of actions you implement in you app
- a schema definition that links the 2 sets

packages/forms/src/UIForm/doc/triggers.md
jsomsanith: Changed to avoid custom widgets in the explanation 👍

PR #2028, opened 2019-02-11
chore(DatePicker): refactor to make its layout flexible

jsomsanith:
Date picker Layout is hard coded.
There is no way to use it with another layout, and there will be the need soon.

Use Manager/children pattern with context communication.
- Manager: context provider, synchronize data between all the chidren
- Input: get data and callback from context consumer, display the main ibput
- Picker: get data and callback from context consumer, display the picker
- Validation: get data and callback from context consumer, display the footer (errors and submit)

```javascript
<DateTime.Manager>
    <DateTime.Input />
    <DateTime.Picker />
    <DateTime.Validation />
</DateTime.Manager>
```

Existing InputDateTimePicker use those components to compose the Layout with input + dropdown.


packages/components/stories/index.js
```diff
@@ -1,51 +1,51 @@
-import './AboutDialog';
```
jlame: don't forget to remove the comment before merging the PR

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -178,115 +86,35 @@ class InputDateTimePicker extends React.Component {
...
-			useTime: this.props.useTime,
-			useSeconds: this.props.useSeconds,
-			useUTC: this.props.useUTC,
-			required: this.props.required,
-		};
+	onChange(event, payload) {
+		this.props.onChange(event, payload);
+		if (
+			this.props.formMode ||
+			(!this.props.formMode && !this.props.useTime && payload.origin !== 'INPUT')
```
jlame: Closing it while selecting a date in non formMode is added here, but if we use the input we have to either tab, esc, click outside maybe adding enter keydown to this makes sense ?

jlame:
formMode should be by default when using time, as we now automatically close the picker when selecting a date in non formMode.

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -89,86 +58,25 @@ class InputDateTimePicker extends React.Component {
...
-				this.onChange(event, 'INPUT');
-			}
-		});
-	}
-
-	onKeyDown(event) {
+	onKeyDown(event, { onReset }) {
 		switch (event.keyCode) {
 			case keycode.codes.esc:
+				onReset();
```
jlame: is it possible to have blur called here ? 
In the grid case we need this event to be called, click outside and escape should work the same

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
jlame: behaviour out of scope

PR #2031, opened 2019-02-13
fix(frontend): translate date picker error messages

yonmey:
The Date component takes its local state error messages in priority instead of the ones passed in the props.

I changed the priority of the error messages so it takes before the ones from the props.


yonmey:
As I have seen in some of the JsonSchemas of our forms, a `validationMessage` attribute is set so a custom message can be passed to the form validation: I don’t know if this was a workaround to show the messages, cause what I see are the same as the ones coming from the date picker. I will cancel the PR if this is the correct priority, however, I have trace the errors in our forms for the date picker and I saw that there’s an object sent and therefore an error saying that we’re expecting a string. I believe this comes from the DatePickerException

yonmey:
I have changed some things in the code, still in WIP. The way I've done it, the translations are working, however the translation's extractor won't work, I think my modifications should change in a way. Any advice? @jsomsanith

yonmey:
I changed the translation logic inside the Manager component, the messages are getting translated, however I would like to make some unit tests for this method. How can I pass the `t` (translation) function to test the error's translation method?
@jsomsanith

I see some of the enzyme tests are broken (the ones that use the InputDateTimePicker).

jsomsanith:
I don't think translating all the messages at component instantiation is a good idea.
Moreover, there are then 2 lists to maintain (the one you introduced, and the one in Error.component.js.

What I propose : 
- create an `error-messages.js` that expose a getErrorMessage(key, options)
- this js file won't have the `t()` function, you can use i18next directly (import then `i18next.t()`).
- use it in `Error.component.js` instead of the defined list
- use it in `date-extraction.js` to replace the error message in errors

packages/components/src/DateTimePickers/DateTime/Validation/Error.component.test.js
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import { shallow } from 'enzyme';
+import { mount } from 'enzyme';
```
jsomsanith: Please restore this file, changing just the `Error.WrappedComponent` to `Error`.

The mount is not relevant here and it turn the snapshot to something that is not what we want (mount wrappers should be snapshot with enzyme-to-json, but here just a shallow is sufficient).

packages/components/src/DateTimePickers/DateTime/Validation/Error.component.js
```diff
@@ -2,67 +2,21 @@ import React from 'react';
...
-}
-
-function Error({ hidden, errors, id, t }) {
+export default function Error({ hidden, errors, id }) {
 	const classNames = classnames({ 'sr-only': hidden });
 	return (
 		<div id={id} className={classNames}>
 			{errors.map((error, index) => (
-				<span key={index}>{getDateErrorLabel(t, error.message)}</span>
+				<span key={index}>{getErrorMessage(error.message)}</span>
```
jlame: this is not needed anymore, you just need to use error.message because it is already translated when you created the exception.

PR #2033, opened 2019-02-14
feat: loadable hoc

ncomont:
According to UX team (cc @marc, @Anne), we should use skeletons at every single place where we could find _unsynchronized_ elements.

Actually, we check for _loading_, _inProgress_, _isLoading_, _whatever_ props, and then we perform things like :

```
export function MyComponent({ inProgress, ...rest }) {
	if (loading) {
		return <Skeleton type={Skeleton.TYPES.text} size={Skeleton.SIZES.large} />;
	}
	// blah
	// blah blah
	// blah

	return <h1>yolo</h1>;
}
```

Which is, I think, not so clean and hard to maintain. I would like to propose a kind of standard for _loadable_ components, to never write again `if (isLoading)` and stuff like that.

The idea is to use this dead simple HOC on every component that should be able to render as `Skeleton` : https://github.com/Talend/ui/blob/de89f9374b903e275566189de284836de0efbbb7/packages/components/src/Loadable/Loadable.component.js

The benefit is not so huge in this case, but this could work out of the box for more complex components trees :

`SubHeader` > `Content` > `RightActions` > `Action A` | `Action B`

- If `loading` prop is passed to `Action A` we’ll have :

`SubHeader` > `Content` > `RightActions` > `SKELETON` | `Action B`

- If `loading` prop is passed to `RightActions` we’ll have :

`SubHeader` > `Content` > `SELETON`

And the only think I had to do was to export each of my _loadable_ components like that :

```
export default loadable(TitleSubHeader, {
	type: 'text',
	size: 'large',
});

```


romainseb:
Just some toughs about it, i think would like better some implementation this way :
Having a component like Suspense's idea ( https://www.youtube.com/watch?v=z-6JC0_cOns ) was first presented. This could be used this way :

```javascript
<Loadable 
        status={loadingStatus}
        fallback={<Skeleton type="text" size="large" />}
>
	<Action label="i'm Loaded" />
</Loadable>
```

We can also go further with some placeholder definitions in the components maybe like

```javascript
const actionProps = { label: 'hello' }
<Loadable 
        status={loadingStatus}
        fallback={<Action.LoadingPlaceholder {...actionProps} />}
>
	<Action {...actionProps}/>
</Loadable>
```

As we load data, we also could imagine some behaviour to show if there is some error : 

```javascript
<Loadable 
        status={loadingStatus}
        fallback={<Skeleton type="text" size="large" />}
        errorFallback={<div>Something wrong happened, sry 😢</div>}
>
	<Action label="i'm Loaded" />
</Loadable>
```

At least, i think in the guidelines, we show skeleton for the first loading, but then, when we refresh the data, we only want some heartbeat effect. As i see it, you also can use function as children to pass some props but it's more for a container implementation with some context or store connection, idk if it fit you case ( maybe it's a bit too far here )

```javascript
<Loadable
        fallback={<Skeleton type="text" size="large" />}
        errorFallback={<div>Something wrong happened, sry 😢</div>}
>
        {({ inProgress }) => {
        	<Action inProgress label="i'm Loaded" />;
        }}
</Loadable>
```

We have a component in this way of thinking in a @talend private github project ( i can send you the link in DM if you want it )
Your question here is a really good thing, we could have some loading component to help us handle loading feedback 👍 

Meanwhile, i think the best would be to not go too far from what i will going on in the future react updates as we could evolve with the library

annilsson:
I like the idea of at least trying to find a default way of loading. Personally I prefer explicitly using it inside render though, but perhaps something like @romainseb mentioned.  

For the HOC I see the **pros**:
- Clear and default way to handle the loading as long as it's documented
- the render would be very clean in all components, compared to now. 

**Cons**:
- We need to forbid the prop `loading` in all UI components (especially considering we often pass `{...props}`), because the prop would be swallowed as soon as the HOC is added. If so I think this rule should be enforced somehow. 
- Currently you can test components with the `translate` HOC by using `MyComponent.WrappedComponent` but if we add another HOC it will be more tricky as that would not give you the original component anymore.

sgendre:
about the HoC, I do not like the user experience that it will provide on forms, having the whole form being replaced by a skeleton whereas the user as already set some input is really weird.

I prefer having just one skeleton at the end of the form and keep existing form displayed (though maybe disabled) during the trigger

jmfrancois:
Hey the need is clearly here. We need to expose all options and decide what is the best options.

The FormSwitcher component do that.
My two cents: each component should expose it s Skeleton.

The only point is how do we trigger the display of it:

- props: internal HOC hidden (like FormSwitcher)
- per business case: HOC on project side

And for sure @sam the guideline explains the Skeleton is only on the initial loading.
We have the same to do with refreshing data (hearbeat)

PR #2034, opened 2019-02-14
feat(components/subheader): subtitle in loading mode

ncomont:

Subtitle cannot be rendered in loading mode

Add a new prop to manage it


jsomsanith:
The placement is not the same in loading more or not. The skeleton is 14px height and the text is 19px height.
Can you try to fix them to avoid a title jump when it switches status ?

PR #2035, opened 2019-02-14
feat(cmf): provide more precise options about saga onHttpError

sgendre:
Atm, there is only a  `silent` boolean that is all or nothing about dispatching HTTP_ERRORS and that does not allow to differenciate notification and redirection behavior
Keep the generic boolean working but also add the possibility to provide an object for more precise behavior definition
Documentation is updated


packages/cmf/src/sagas/http.js
```diff
@@ -170,14 +170,15 @@ export function httpFetch(url, config, method, payload) {
...
-	if (!get(options, 'silent') && answer instanceof Error) {
+	const silent = get(options, 'silent');
+	if (silent !== true && answer instanceof Error) {
 		yield put({
 			error: { message: answer.data.message, stack: { status: answer.response.status } },
 			url,
 			config,
 			method,
 			payload,
+			options,
```
annilsson: Just a note, here people could send anything, like `IwantANotification` and `redirectMe`. If we pass options, should we not validate it then? For example
```
options: {
  notify,
  redirect,
}
```

But on the other hand if the option doesn't actually do anything inside CMF, maybe we shouldn't even document it?

sgendre: those 2 options will be fully integrated within cmf once notification and redirect will be cmf-modules

packages/cmf/src/sagas/index.md
```diff
@@ -77,6 +77,20 @@ const { data, response } = yield call(http.get, `${API['dataset-sample']}/${data
...
+		redirect: false,
+	},
+};
+
+const { data, response } = yield call(http.get, `${API['dataset-sample']}/${datasetId}`, config, options);
+```
+Passing an object instead of a boolean will make cmf dispatch an action of type ```@@HTTP/ERRORS``` and allow you to deal with it in your redirect and notification sagas.<br/>
+If ```notify``` is ```false```, your onHttpErrorNotification saga should send the notification as it is within a ```silent```, if ```true``` it should not.<br/>
+If ```redirect``` is ```false```, your onHttpErrorRedirect saga should redirect to dedicated error page base on the status code as it is within a ```silent```, if ```true``` it should do nothing.<br/>
+
```
jmfrancois: look like this doc could be outside of setDefault config.
Could you please create a title  for it ?

packages/cmf/src/sagas/http.js
jsomsanith: Agree with @annilsson, if the options are not used by cmf code, it should not be specified in its doc.
If you do a cmf external module that consume that, this should be in this module documentation

PR #2037, opened 2019-02-18
fix(cmf): http saga options documentation

jsomsanith:
The cmf http saga documentation describes options that are not cmf related.

Change this part to explain that you can pass any options that is given back to the app on http error dispatch.

### DIFF
--- a/packages/cmf/src/sagas/index.md
+++ b/packages/cmf/src/sagas/index.md
@@ -88,8 +88,8 @@ Example
 
 ```javascript
 const options = {
-    notify: false,
-    redirect: false,
+    toto: false,
+    tata: false,
 };
 
 const { data, response } = yield call(http.get, `${API['dataset-sample']}/${datasetId}`, config, options);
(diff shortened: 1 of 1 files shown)


packages/cmf/src/sagas/index.md
```diff
@@ -68,32 +68,33 @@ const options = {
...
+
+Example
 
 ```javascript
 const options = {
-	silent: {
-		notify: false,
-		redirect: false,
-	},
+    notify: false,
```
sgendre: please add documentation about `silent : false` as it belongs to cmf

Plus providing aside properties will lead to the possibility to pass inconsistent options: eg: 
`{
silent: true,
notify: true,
redirect: true
}` which was avoided by the previous implementation and aligned documention

jsomsanith: This is an example. not a suggestion of what to write in your app.
It'll change for `toto`, `tata` to avoid confusion

PR #2040, opened 2019-02-20
feat(UIForm): trigger properties management from ComponentForm to UIForm

jsomsanith:
This reverts commit af155554a16912e0e93d684cf25f6ad6ef3cfb12.

UIForms triggers can update properties. But there are 2 modes :
* trigger returns a `properties` object that will override existing properties
* trigger returns a `properties()` function to modify existing properties

Both are not in the same place, one is in component form, so available only for tacokit, the other one in UIForm component

Move them in a common place : UIForm.

The case where triggers return an object introduce a breaking change. Before this PR, returning a properties object from non-ComponentForm trigger was doing nothing.  
Now it will override the form data.
More details in [migration path](https://github.com/Talend/ui/wiki/Migration-path-:-2.x-to-3.x#uiform)


packages/forms/src/UIForm/UIForm.container.js
```diff
@@ -151,9 +151,11 @@ export default class UIForm extends React.Component {
...
 				this.setErrors(event, errors);
 			}
-			if (typeof data.properties === 'function') {
+			if (data.properties) {
 				let properties = data.properties;
-				properties = data.properties(liveState.properties);
+				if (typeof data.properties === 'function') {
+					properties = data.properties(liveState.properties);
+				}
 				this.onChange(event, { properties });
```
jmfrancois: this cause TUI-350. Please update this onChange call

jsomsanith: As discussed together, this will come in another PR that doesn't introduce a breaking change, so we don't wait for 3.0 to have that

jsomsanith: #2041

packages/forms/src/UIForm/UIForm.container.js
```diff
@@ -151,9 +151,11 @@ export default class UIForm extends React.Component {
 				}
 				this.setErrors(event, errors);
 			}
-			if (typeof data.properties === 'function') {
+			if (data.properties) {
 				let properties = data.properties;
-				properties = data.properties(liveState.properties);
+				if (typeof data.properties === 'function') {
```
sgendre:
```suggestion
				if (typeof properties === 'function') {
```

packages/forms/src/UIForm/UIForm.container.js
```diff
@@ -151,9 +151,11 @@ export default class UIForm extends React.Component {
...
+					properties = data.properties(liveState.properties);
```
sgendre:
```suggestion
					properties = properties(liveState.properties);
```

PR #2046, opened 2019-02-20
feat(Scripts): option to disable css modules

jsomsanith:
Some app don't use css modules, but talend/scripts default preset use css modules and there is no way to disable it.

Add an option to disable it in talend preset

### DIFF
--- a/packages/scripts/webapp/preset/config/webpack.config.js
+++ b/packages/scripts/webapp/preset/config/webpack.config.js
@@ -108,7 +108,13 @@ module.exports = ({ getUserConfig, mode }) => {
 				{
 					test: /\.scss$/,
 					use: getSassLoaders(true, sassData, mode),
+					include: /@talend/,
 					exclude: /bootstrap-theme/,
+				},
+				{
+					test: /\.scss$/,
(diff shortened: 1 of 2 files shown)


packages/scripts/webapp/preset/config/webpack.config.js
```diff
@@ -108,8 +108,14 @@ module.exports = ({ getUserConfig, mode }) => {
 				{
 					test: /\.scss$/,
 					use: getSassLoaders(true, sassData, mode),
+					include: /@talend/,
 					exclude: /bootstrap-theme/,
 				},
+				{
+					test: /\.scss$/,
+					use: getSassLoaders(getUserConfig(['sass', 'modules'], true), sassData, mode),
+					exclude: /@talend/,
```
annilsson: Wondering if this should also exclude `bootstrap-theme` since it's already covered on line 106, and it's excluded when modules is set to true above?

annilsson: Yeah never mind, as you said "bootstrap-theme" is inside a "@talend" module so it seems good

PR #2048, opened 2019-02-20
feat(Scripts): disable cmf webpack plugin by default

jsomsanith:
There is an incompatibility between cmf-webpack-plugin and copy-webpack-plugin.  
This is not new, and well documented in talend/scripts doc.

But most of the time, projects use settings in the way that it's incompatible. They will have to set `cmf: false` option, that doesn't make sense.

Change the default to not include the cmf plugin. To enable it, the cmf option need to be set to true.

But it's talend/scripts that is not in 1.0.

### DIFF
--- a/packages/scripts/webapp/preset/config/webpack.config.dev.js
+++ b/packages/scripts/webapp/preset/config/webpack.config.dev.js
@@ -2,7 +2,7 @@ const path = require('path');
 
 module.exports = ({ getUserConfig }) => {
 	const plugins = [];
-	if (getUserConfig('cmf') === true) {
+	if (getUserConfig('cmf')) {
 		// eslint-disable-next-line global-require,import/newline-after-import
 		const ReactCMFWebpackPlugin = require('@talend/react-cmf-webpack-plugin');
 		plugins.push(new ReactCMFWebpackPlugin({ watch: true }));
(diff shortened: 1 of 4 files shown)


packages/scripts/webapp/preset/config/webpack.config.dev.js
```diff
@@ -2,7 +2,7 @@ const path = require('path');
 
 module.exports = ({ getUserConfig }) => {
 	const plugins = [];
-	if (getUserConfig('cmf') !== false) {
+	if (getUserConfig('cmf') === true) {
```
romainseb:
```suggestion
	if (getUserConfig('cmf')) {
```

packages/scripts/webapp/preset/config/webpack.config.prod.js
```diff
@@ -3,7 +3,7 @@ const UglifyJsPlugin = require('uglifyjs-webpack-plugin');
 
 module.exports = ({ getUserConfig }) => {
 	const plugins = [];
-	if (getUserConfig('cmf') !== false) {
+	if (getUserConfig('cmf') === true) {
```
romainseb:
```suggestion
	if (getUserConfig('cmf')) {
```

PR #2049, opened 2019-02-21
feat(Scripts): analyze option and spread user cli options

jsomsanith:
- no webpack bundle analyzer
- no way to pass extra options to webpack cli

- add a `--env.analyze` to add webpack-bundle-analyzer
- spread user cli options


packages/scripts/webapp/scripts/start.js
```diff
@@ -3,16 +3,18 @@ const { hereRelative, resolveBin } = require('../utils/path-resolver');
...
 const webpackDevServer = resolveBin('webpack-dev-server');
 
-module.exports = function start(env) {
+module.exports = function start(env, _, options) {
 	return spawn.sync(
 		webpackDevServer,
 		[
 			'--inline',
-			'--config', hereRelative(__dirname, '../config/webpack.config.js'),
-			'--content-base', 'build/',
```
annilsson: Just double checking --content-base was removed on purpose?

packages/scripts/webapp/scripts/build.js
```diff
@@ -1,23 +1,12 @@
 const spawn = require('cross-spawn');
 const { hereRelative, resolveBin } = require('../utils/path-resolver');
 
-const rimraf = resolveBin('rimraf');
 const webpack = resolveBin('webpack');
 
-module.exports = function build(env) {
-	// remove old builds folder
```
annilsson: We don't need to do this anymore?

jsomsanith: Sorry I mixed 2 PRs, while you were writing that, I was reverting it :)
This will come after, the target forlder removal will come from a webpack plugin instead, more maintainable because it takes the real target folder instead of duplicating it here

packages/scripts/webapp/scripts/start.js
jsomsanith: Yes, in fact the path is wrong. Let's manage that in the devServer config

PR #2059, opened 2019-02-25
chore(Scripts): upgrade eslint

jsomsanith:
We have an issue with linter. Talend/scripts depends on enzyme, it setup enzyme adapter.  
But in project, the import from enzyme will produce a lint error because it is pulled from talend/scripts, not explicitly (`import/no-extraneous-dependencies`).

There is an option to set the paths to package.json files for `import/no-extraneous-dependencies` but it is introduced in import plugin 2 and we are using version 1.

The eslint versions and plugins are tied to airbnb plugin. Upgrading the eslint versions implies to upgrade the plugins, and airbnb plugin have a lot of changes since version 11 (current version is 17).

* Upgrade eslint version and all the plugins
* disable some rules that have changed
* keep some new rules that make totally sense
* (bonus) fixed dependency warnings

(v0.13.0-0 is published with this feature)


packages/scripts/package.json
```diff
@@ -42,15 +42,15 @@
...
+    "eslint": "^5.3.0",
+    "eslint-config-airbnb": "^17.1.0",
+    "eslint-plugin-import": "^2.16.0",
+    "eslint-plugin-jsx-a11y": "^6.2.1",
+    "eslint-plugin-react": "^7.12.4",
     "file-loader": "^2.0.0",
     "html-webpack-plugin": "3.2.0",
     "jest": "^24.0.0",
-    "jest-environment-jsdom": "^24.0.0",
+    "jest-environment-jsdom": "^23.0.0",
```
romainseb: Why is there a downgrade here ?

jsomsanith: Oh yeah I fixed the dependencies warnings too.
`jest-environment-jsdom-global` that is used as testing env, is requesting `jest-environment-jsdom` 23

packages/scripts/package.json
```diff
@@ -73,9 +73,15 @@
...
   },
+  "devDependencies": {
+    "react": "^16.8.3",
+    "react-dom": "^16.8.3"
+  },
   "peerDependencies": {
     "@talend/bootstrap-theme": "1.10.0",
-    "@talend/react-components": "1.10.0"
+    "@talend/react-components": "1.10.0",
+    "react": "^16.0.0",
```
romainseb: Maybe this peer dep is more >= 16.8.0 instead of 16.0.0 ?

packages/scripts/webapp/preset/config/.eslintrc
```diff
@@ -3,16 +3,33 @@
...
+    "react/forbid-prop-types": 0,
+    "react/jsx-filename-extension": [1, { "extensions": [".js"] }],
     "react/jsx-indent": ["error", "tab"],
     "react/jsx-indent-props": ["error", "tab"],
-    "react/jsx-filename-extension": [1, { "extensions": [".js"] }],
+    "react/jsx-one-expression-per-line": 0,
+    "react/no-array-index-key": 0,
+    "react/no-this-in-sfc": 0, // because of ref affectation
     "react/no-unused-prop-types": [2, { "skipShapeProps": true }],
-    "react/forbid-prop-types": [0]
```
ncomont: Should the `tools-javascript` project be updated (or deprecated) ?
https://github.com/Talend/tools/blob/master/tools-javascript/.eslintrc

jsomsanith: In fact I'm not that sure we should merge this PR.
At least we have to determine a path, because this will impact a lot the current code. So it's in stand by now.
I think we should deprecate the one in tools, and talend/scripts should be the only source of truth

jmfrancois: Yes lets drop Talend/tools.

jsomsanith: PR to remove it : https://github.com/Talend/tools/pull/79

PR #2060, opened 2019-02-25
fix(container): Drawer.Title behavior

sgendre:
If there is no title, children are not displayed which breaks original layout

Display children even if title is empty.
Display nothing only if title is empty and there is no children


packages/components/src/Drawer/Drawer.component.js
```diff
@@ -107,7 +107,7 @@ export function subtitleComponent(subtitle) {
 }
 
 function DrawerTitle({ title, subtitle, children, onCancelAction, getComponent }) {
-	if (!title) {
+	if (!title && !children) {
```
jmfrancois: this change do the exact inverse of the PR description.

sgendre: no it's not, is return null only on a more specific condition which comply with description

jmfrancois: yep sorry, my bad

PR #2061, opened 2019-02-25
feat(Scripts): get versions and sha info, et as global var

jsomsanith:
Feature needed in some app : having the talend libs, versions and sha1.

Add a TALEND_APP_INFO (name is historical) global var
- app version from package.json
- app revision from git command
- talend libs : name, version, sha (if provided)


packages/scripts/webapp/preset/README.md
```diff
@@ -188,6 +188,18 @@ It has an incompatibility with `copy-webpack-plugin`. To use it correctly
...
 3. In your cmf app index file, you can fetch the settings from `/settings.json`.
 
+## Versions
+
+Some variables are injected as global.
+
+| Variable | Description |
+|---|---|
+| BUILD_TIMESTAMP | The built timestamp. |
+| TALEND_APP_INFO | Versioning of current app and @talend libs. To get the libs git SHA1, you need to provide `sha1.json` containing the sha of the packages. |
```
annilsson: Great doc, could we also have an example `sha1.json` code block here somewhere? Even if it's only key value I'd feel more confident as a user seeing an example confirming that. 

Also I'm not sure about this but it looks like `@talend/ui` has only one sha1 with the key `@talend/ui` right? That might be worth clarifying in that example, so you don't have one for each subpackage (components, containers, forms etc)

jsomsanith: Done : https://github.com/Talend/ui/blob/31b12298aa281c10bcc79ce788fa159f66cd2103/packages/scripts/webapp/preset/README.md#versions

- Add example with package.json / sha1.json / result
- include `@talend/ui` merge explanation + include in example
- include example case
** version with sha,
** version without sha,
** sha that is not used (not a dependency)

jsomsanith: (Feel free to tell me if something is still unclear, even if the PR is merged)

annilsson: That looks perfect with the example, thanks a lot! 👍

PR #2063, opened 2019-02-26
fix(TPSVC-7920/CollapsiblePanel): use onEntered/onExited

lmaillet:
Collapsible panel in VirtualizedList does not work in Firefox and Edge
See https://jira.talendforge.org/browse/TPSVC-7920
Because it use setTimeout with a cross fingers renderer has happened.

Force the VirtualizedList to measure again the panel after each expand or collapse using `onEntered` or  `onExited` callback (https://react-bootstrap.github.io/utilities/transitions/#collapse).

There is still a problem in storybook if you load directly the page, because storybook loads the css after javascript and so css is loaded after the first render. But it works in _normal_ app.

@jmfrancois I've removed the `appear` prop, as it breaks the UT and looks like it's not needed anymore.
image


packages/components/src/VirtualizedList/RowCollapsiblePanel/RowCollapsiblePanel.component.js
```diff
@@ -62,7 +58,12 @@ class RowCollapsiblePanel extends React.Component {
 						aria-label={get(rowData, 'header[0].label')}
 						style={style}
 					>
-						<PureCollapsiblePanel {...{ rowData, measure, onToggle: this.onToggle }} />
+						<CollapsiblePanel
+							onEntered={() => onMeasure(measure)}
+							onExited={() => onMeasure(measure)}
```
jmfrancois: could you please try onEntererd={measure} / same for onExited and remove onMeasure ?

lmaillet: Done 👍

packages/components/src/VirtualizedList/RowCollapsiblePanel/RowCollapsiblePanel.component.js
```diff
@@ -62,7 +55,12 @@ class RowCollapsiblePanel extends React.Component {
 						aria-label={get(rowData, 'header[0].label')}
 						style={style}
 					>
-						<PureCollapsiblePanel {...{ rowData, measure, onToggle: this.onToggle }} />
+						<CollapsiblePanel
+							onEntered={measure}
```
sgendre: I rather name it onEnter

packages/components/src/VirtualizedList/RowCollapsiblePanel/RowCollapsiblePanel.component.js
```diff
@@ -62,7 +55,12 @@ class RowCollapsiblePanel extends React.Component {
...
+							onExited={measure}
```
sgendre: I rather name it onExit

lmaillet: I used `onEntered` and `onExited` to keep consistency with bootstrap collapse.
`onEnter` callback is called **before** component expands and `onEntered` is called **after** component has expanded. Same for `onExit` and `onExited`.
And I would like `react-virtualized` to measure the row after expand/collapse.
See https://react-bootstrap.github.io/utilities/transitions/#collapse

packages/components/src/VirtualizedList/RowCollapsiblePanel/RowCollapsiblePanel.component.js
jsomsanith: It seems that `react-bootstrap` collapsible panel that our component use, has a prop `onEntered`. Let's not change the name to do a mapping in our component.

PR #2067, opened 2019-02-26
feat: scripts build lib

jsomsanith:
Talend/scripts has a goal to build webapp using webpack. But the libs use only babel. There is no way with current version, to build a lib.

Add a `build:lib` goal, running 
- remove target folder
- run babel
- remove tests files
- copy assets (css, json)


packages/scripts/package.json
```diff
@@ -18,15 +18,15 @@
...
   },
   "dependencies": {
-    "@babel/cli": "^7.2.0",
-    "@babel/core": "^7.2.0",
-    "@babel/plugin-proposal-class-properties": "^7.2.0",
-    "@babel/plugin-proposal-export-default-from": "^7.2.0",
-    "@babel/plugin-proposal-export-namespace-from": "^7.2.0",
-    "@babel/plugin-proposal-object-rest-spread": "^7.2.0",
-    "@babel/plugin-transform-object-assign": "^7.2.0",
+    "@babel/cli": "^7.1.0",
```
jsomsanith: We need to downgrade babel because Storybook 4 rely on `babel-preset-react-app` that rely on babel 7.1.0 (fixed).
So if we don't have compatible versions, @talend/scripts babel version will be in @talend/scripts node_modules. Jest (babel-jest) fails to find the babel plugins :(

PR #2069, opened 2019-02-27
fix(components/Datalist): select suggestion value from typed title

jare-talend:
On the datalist, after one search, the input is erased.
ui_datalist_before

On the datalist, after one search, we kept the value only if a value matching with one suggestion
ui_datalist_after

This behavior is checked with marc.


frassinier:
It's funny since it's exactly what a datalist does not do?
> https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist

jare-talend:
> It's funny since it's exactly what a datalist does not do?
> 
> > https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist

yes, it doesn't. also it made less thing. It accepts free type, no validation control, ...

frassinier:
@jare-talend so it's more a `<select />` behavior that you want to apply to a `<datalist />`
I personally not convinced at all by this feature which it's pretty disturbing

jare-talend:
> @jare-talend so it's more a `<select />` behavior that you want to apply to a `<datalist />`
> I'm personally not convinced at all by this feature which it's pretty disturbing

search and select, that is wanted. @jsomsanith set a no go on that PR. So, I wait reviews with UX team to see the wished behavior.

packages/components/src/Datalist/Datalist.component.js
```diff
@@ -64,7 +64,18 @@ class Datalist extends Component {
 		const { value, previousValue } = this.state;
 
 		if (value !== previousValue) {
-			this.updateValue(event, value, true);
+			let newValue = value;
+			if (this.props.restricted && !this.props.multiSection && this.state.suggestions && value) {
+				const hasMatchingSuggestion = this.state.suggestions.find(
```
jsomsanith: You should base your search on titleMap props, not suggestions. 
Here it means that it doesn't take the suggestion if you close suggestions.

packages/components/src/Datalist/Datalist.component.test.js
```diff
@@ -214,10 +214,39 @@ describe('Datalist component', () => {
...
+				multiSection={false}
+				errorMessage={'This should be correct'}
+				onChange={onChange}
+				{...props}
+				value={'foo'}
+				restricted
+			/>,
+		);
+		const input = wrapper.find('input').at(0);
+		input.simulate('change', { target: { value: 'foobar' } });
```
jsomsanith: Can you change you use case to have title != value please ?
So we know that in you case, with `{title: 'lol', value: 'mdr'}`, typing `lol` propagates `mdr`

PR #2081, opened 2019-03-05
fix(cmf): remove automatic event from onEvent [BC]

sgendre:
The SyntheticEvent that is passed can be quite fat, leading devTools to crash by oom or **browser to lag**

a static object  to define the action to provide to a button
```javascript

const submitButtonProps = {
	type: 'submit',
	label: 'Select that',
	onClickDispatch: { type: SUBMIT_ACTION },
};
```

If the event is a react SyntheticEvent, provide an empty object instead of nested object with potential circular dependencies

if you need data from the event initial, implement your own handler and extract your data to build your own minimum meaningful payload
```javascript
const submitButtonProps = dispatch => {
	const dispatchSubmit = e =>
		dispatch({ type: SUBMIT_ACTION, payload: { target: e.target } });
	return {
		type: 'submit',
		label: 'Select that',
		onClick: dispatchSubmit,
	};
};
```

Migration path:

```snippet
// before
const submitButtonProps = {
	type: 'submit',
	label: 'Select that',
	onClickDispatch: { type: SUBMIT_ACTION },
};

// after
const submitButtonProps = dispatch => {
	const dispatchSubmit = e =>
		dispatch({ type: SUBMIT_ACTION, payload: { target: e.target } });
	return {
		type: 'submit',
		label: 'Select that',
		onClick: dispatchSubmit,
	};
};
```

alternative solution can be found [here](https://github.com/zalmoxisus/redux-devtools-extension/blob/master/docs/Troubleshooting.md#it-fails-to-serialize-data-when-passing-synthetic-events-or-calling-an-action-directly-with-redux-actions)

If you look for the original native event, you will only get an empty object.
The best is to catch that native event yourself and dispatch your own event with the minimal payload.


packages/cmf/src/onEvent.js
```diff
@@ -5,6 +5,7 @@ import CONSTANT from './constant';
 function serializeEvent(event) {
 	if (event.persist) {
 		event.persist();
+		return {};
```
jsomsanith: If the event is not used/propagated, why do you still persist it ?

sgendre: I bet that some middleware or js-logger could have some reference on it?

jmfrancois: No you can t have any reference to it if you don t dispatch it.

sgendre: k thx, updated

PR #2082, opened 2019-03-06
chore(TUI-364): deprecate react-jsonschema-form

jmfrancois:

Lets deprecate react-jsonschema-form and remove it by default.
Most of apps has migrate to UIForm.

If you still use the old form library please configure webpack to define `process.env.FORM_MOZ = true`; should work as before

So as deprecated we have:

- remove existing documentation about it
- remove tests
- move the code into a deprecated folder.

So if you want to use the old form change the import that way

```diff
- import Form from '@talend/react-forms/lib/Form'
+ import Form from '@talend/react-forms/lib/deprecated/Form'
-import { customWidgets } from '@talend/react-forms';
+import { customWidgets } from '@talend/react-forms/lib/deprecated';
-import { renderActionIcon } from '@talend/react-forms';
+import { renderActionIcon } from '@talend/react-forms/lib/deprecated';
-import { renderActions } from '@talend/react-forms';
+import { renderActions } from '@talend/react-forms/lib/deprecated';
-import { DataPropTypes } from '@talend/react-forms';
+import { DataPropTypes } from '@talend/react-forms/lib/deprecated';
-import { ActionsPropTypes } from '@talend/react-forms';
+import { ActionsPropTypes } from '@talend/react-forms/lib/deprecated';
```

Note: it will be removed in 5.0

* [x] **This PR introduces a breaking change**

# `uiform` props

Before this PR you can use previous uiSpec (uiSchema as object {} notation to extends jsonschema) and ask Form component to migrate the old uiSpec to the new one using a boolean props `uiform`.

this feature has been removed because it has never been usable. (lots of bug founds and way to much hard to maintain)

```javascript
<Form data={{jsonSchema: {...}, uiSchema: {}} uiform />
```

If you do that you will now have a error message displayed in alert-danger style.
So you can just remove the uiform and the form will be displayed using the deprecated form library or just write a uiSchema.


packages/forms/src/UIForm/merge.js
```diff
@@ -1,252 +1,10 @@
-import React from 'react';
```
annilsson: just to make sure; we don't need this anymore even if you use the deprecated version?

jmfrancois: Yes, this one has never been used and is full of bugs (the main idea was automigrate from old schema to new schema);

annilsson: On our side we don't use it. But now I found `wrapCustomWidget` being used here: https://github.com/Talend/ui/pull/2082/files#diff-0ff6e5979591c52a9139e5f1cff133cdR9 (packages/forms/src/Form.js → packages/forms/src/deprecated/Form.js)

output/containers.eslint.txt
```diff
@@ -8,9 +8,12 @@ The react/require-extension rule is deprecated. Please use the import/extensions
 /home/travis/build/Talend/ui/packages/containers/src/EditableText/EditableText.selectors.js
   8:1  error  Prefer default export  import/prefer-default-export
 
+/home/travis/build/Talend/ui/packages/containers/src/Form/Form.test.js
+  2:10  error  'mount' is defined but never used  no-unused-vars
```
jsomsanith-tlnd: Please remove it

packages/forms/src/deprecated/Form.js
```diff
@@ -122,16 +120,9 @@ class Form extends React.Component {
...
-			if (props.widgets) {
-				Object.keys(props.widgets)
-					.filter(key => props.widgets[key].displayName !== 'TFMigratedWidget')
-					.forEach(key => {
-						props.widgets[key] = wrapCustomWidget(props.widgets[key]);
-					});
-			}
-			return <UIForm {...props} />;
+			return (
+				<div className="alert alert-danger">The combination of old UISpec with props.uiform is not supported anymore.</div>
```
jsomsanith-tlnd: You removed, the compat mode, this need to be documented in a migration guide.
Perhaps you can add it in the `breaking changes` part of this PR description, so we can validate it and we just have to copy/paste in the real migration guide

packages/forms/src/FormSwitcher.js
```diff
@@ -1,17 +1,29 @@
...
 
+let Form = () => (
+	<div className="alert alert-danger">
+		Error: You should compile using FORM_MOZ to use deprecated Form
+	</div>
+);
+
+if (process.env.FORM_MOZ) {
+	// eslint-disable-next-line global-require
+	Form = require('./deprecated/Form');
```
jsomsanith-tlnd: Nice 👍

packages/forms/src/deprecated/Form.js
jmfrancois: good point, done

packages/forms/src/UIForm/merge.js
annilsson: Looks like this is fixed now

PR #2089, opened 2019-03-11
fix(DatePicker): call onChange on Done button

lmaillet:
In the `DatePicker`, if the user don't change the date and click on the **Done** button, nothing happens.

Always trigger `onChange` if user click on the **Done** button, even if the date has not been changed, in order to close the `DatePicker`.


packages/components/src/DateTimePickers/DateTime/Manager/Manager.component.js
```diff
@@ -103,12 +94,8 @@ class ContextualManager extends React.Component {
 	}
 
 	onChange(event, origin) {
-		const { errorMessage, datetime, textInput, errors, previousErrors } = this.state;
-
-		if (
-			this.props.onChange &&
-			(this.dateHasChanged() || haveErrorsChanged(previousErrors, errors))
-		) {
+		const { errorMessage, datetime, textInput, errors } = this.state;
```
jsomsanith: You can move that in the if.
Or better, make a fast return instead `if (!this.props.onChange) { return; } // then add the onChange code`

PR #2090, opened 2019-03-12
fix(UiForm): remove value from state if empty

romainseb:
There is some cases where we have : 
Input not required but with regex.
If you type something in the field, the regex validation would be triggered, if you clear the field, there is still remaining an empty string, so the validation pattern take the empty string to proceed the validation.
If your regex doesn't allow an empty string, you will have no feedback, but the form sould not be submit ( as it should be if the concerned input is pristine )
Remove the property in the state if the input is empty


packages/forms/src/UIForm/utils/properties.test.js
```diff
@@ -59,13 +59,21 @@ describe('Properties utils', () => {
 			expect(convertedValue).toBe(3.5);
 		});
 
-		it('should convert empty string to undefined value', () => {
+		it('should convert empty number to undefined value', () => {
 			const value = '';
```
jsomsanith: You forgot to change that to number

packages/forms/src/UIForm/utils/properties.test.js
```diff
@@ -105,5 +113,26 @@ describe('Properties utils', () => {
 				},
 			});
 		});
+
+		it('should delete the value, creating nested objects', () => {
```
jsomsanith: Can you change the description please ? `creating nested objects` is not good here

PR #2095, opened 2019-03-12
chore: remove unused jest-environment-jsdom dependency

jsomsanith:
`jest-environment-jsdom` 23 was set on multiple packages, as test environment.
But it turns out that we don't need it, no feature of this package is used in our tests.

Remove it


packages/forms/src/UIForm/fields/ResourcePicker/__snapshots__/ResourcePicker.component.test.js.snap
```diff
@@ -76,7 +76,7 @@ exports[`ResourcePicker field should render simple select 1`] = `
         ],
         "results": Array [
           Object {
-            "isThrow": false,
+            "type": "return",
```
jsomsanith: Snapshots have changed because removing the lib with version 23, removed jest duplications (23).
Seems that things have changed from jest 23 to 24 in mocks, but either way, this is a very bad practice to start with: snapshoting a fn mock...

PR #2097, opened 2019-03-13
chore(components/Typeahead): Add ref container to items container renderer

nicomaligne:
In dataset we have a use case that mix the typeahead component (react-autowhatever) and a button.
We need to pass the ref from the container generated by react-autowhatever to the children as function renderer


packages/components/src/Typeahead/Typeahead.component.renderers.js
```diff
@@ -95,17 +92,26 @@ export function renderItemsContainerFactory(
...
+				key={containerProps.key}
+				ref={containerProps.ref}
+				role={containerProps.role}
+			>
+				{render(content, {
+					searching,
+					loading,
+					noResult,
+					isShown,
+					refContainer: containerProps.ref,
```
jsomsanith: You pass the same ref for the div above and the render prop. Isn't that problematic ?

PR #2099, opened 2019-03-13
fix(forms/MultiSelectTag): TUI-373 - Scrollbar can't be used with click

sgangnant:
https://jira.talendforge.org/browse/TUI-373
Scrollbar can't be used on MultiSelectTag fields when they are shown in a drawer
multiselect

Use `<FocusManager />` do the job of handling blur/focusOut.


packages/forms/src/UIForm/fields/MultiSelectTag/MultiSelectTag.component.js
```diff
@@ -139,6 +140,9 @@ export default class MultiSelectTag extends React.Component {
 		this.props.onChange(event, payload);
 		this.props.onFinish(event, payload);
 		this.updateSuggestions('');
+		setTimeout(() => {
+			this.resetSuggestions();
```
jsomsanith: Can you explain why you have this timeout ?

sgangnant: Without, component seem to not re-render and suggestions remain visible.

annilsson: I have a feeling that doing both `this.updateSuggestions` and  `this.resetSuggestions` might cause bugs because they both setState on the same keys (focusedItemIndex and suggestions), plus one `setState` uses a function with oldState while the other does not. 

Any chance it's possible to simply skip `this.updateSuggestions` here? It feels like it's trying to reset them anyway right? That might allow you to remove the `setTimeout`.

annilsson: Actually I see that `onChange` *also* call `updateSuggestions` making the setState run 3 times. 😱

sgangnant: Wanted to make as less changes as possible, but I can take a look ;)

jsomsanith:
Capture d’écran 2019-03-15 à 10 09 53

You have an outline problem.
I think you can remove the outline here. For accessibility it's ok because you have a visual indicator on input focus and currently highlighted element.
If user use the mouse to scroll and select the item, no outline is ok.

PR #2100, opened 2019-03-18
feat(List): composition - display mode

jsomsanith:
See #2026 for PoC.

Split the PoC into multiple smaller PRs. This one introduces the compound component structure with a controlled/uncontrolled display mode.


packages/components/src/List/ListComposition/Manager.component.js
```diff
@@ -0,0 +1,56 @@
...
+	static displayName = 'List.Manager';
+	static propTypes = {
+		children: PropTypes.node,
+
+		displayMode: PropTypes.string, // eslint-disable-line react/no-unused-prop-types
+		onDisplayModeChange: PropTypes.func,
+	};
+
+	static getDerivedStateFromProps(props) {
+		return { collection: props.collection };
```
jsomsanith: It is bad to just copy props into state, but this is to prepare the structure for next iterations where we will have built-in sorting and paging for uncontrolled features.

packages/components/src/List/ListComposition/README.md
```diff
@@ -0,0 +1,362 @@
...
+
+After
+```javascript
+const toolBar = (
+    <List.Toolbar>
+        <ActionBar left={[
+            { id: 'add', label: 'Add Folder' },
+            { id: 'remove', label: 'Remove Folder' },
+        ]} />
+        <List.DisplayMode id="my-list-displayMode" />
```
nicomaligne: Don't we want to enforce the relation of the compound components ?
`
<List>
<List.Toolbar>
<Toolbar.DisplayMode />
</List.Toolbar>
</List>
`

nicomaligne: IMHO it's better, it enforces visually how the composition works.
And can emphasize the css relation wihout opening code.

packages/components/src/List/ListComposition/DisplayMode/DisplayMode.component.js
```diff
@@ -0,0 +1,109 @@
...
+			case 'large':
+				return t('LIST_SELECT_DISPLAY_MODE_LARGE', { defaultValue: 'Expanded' });
+			default:
+				return t('LIST_SELECT_DISPLAY_MODE_TABLE', { defaultValue: 'Table' });
+		}
+	}
+
+	getMenuItems() {
+		return this.props.displayModes.map(option => {
+			const displayModeLabel = this.getLabel(option);
```
nicomaligne: Can't you keep a consistency between your variable name. Option is a display mode, you call it option and in getLabel is named mode.
One of my convention is Entities.map => entity

packages/components/src/List/ListComposition/DisplayMode/DisplayMode.component.test.js
```diff
@@ -0,0 +1,50 @@
...
+
+		// when
+		const wrapper = mount(
+			<ListContext.Provider value={contextValue}>
+				<DisplayMode.WrappedComponent id="myDisplayMode" />
+			</ListContext.Provider>,
+		);
+
+		// then
+		expect(toJson(wrapper)).toMatchSnapshot();
```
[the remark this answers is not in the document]
jsomsanith: I use mount here, so yes
For shallow, getElement is sufficient

packages/components/src/List/ListComposition/DisplayMode/DisplayMode.component.js
```diff
@@ -0,0 +1,109 @@
...
+		case 'table':
+			return 'talend-table';
+		case 'large':
+			return 'talend-expanded';
+		default:
+			return 'talend-table';
+	}
+}
+
+class DisplayMode extends React.Component {
```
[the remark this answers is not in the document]
jsomsanith: I merged ContextualDisplayMode and DisplayMode.
It's now a functional component. No memo because it relies on context too, not only props

packages/components/src/List/ListComposition/DisplayMode/DisplayMode.component.js
```diff
@@ -0,0 +1,109 @@
...
+				return t('LIST_SELECT_DISPLAY_MODE_TABLE', { defaultValue: 'Table' });
+			case 'large':
+				return t('LIST_SELECT_DISPLAY_MODE_LARGE', { defaultValue: 'Expanded' });
+			default:
+				return t('LIST_SELECT_DISPLAY_MODE_TABLE', { defaultValue: 'Table' });
+		}
+	}
+
+	getMenuItems() {
+		return this.props.displayModes.map(option => {
```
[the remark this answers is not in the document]
jsomsanith: I changed that to set it directly in render. I don't create another component that wrap MenuItem, it doesn't seem to work with react-boostrap if there is a component in between :/

packages/components/src/List/ListComposition/Manager.component.js
```diff
@@ -0,0 +1,56 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { ListContext } from './context';
+
+export default class Manager extends React.Component {
```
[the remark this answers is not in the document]
jsomsanith: The display mode management is conditional. In controlled mode, we don't use it, in uncontrolled mode, it's on.
Conditional hook is forbidden

packages/components/src/List/ListComposition/Manager.component.js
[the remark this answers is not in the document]
jsomsanith: I added a comment between your review start and submission.
It's to prepare the structure for furthur features, where we will filter/sort/page/.. the collections in uncontrolled mode. But I'll remove it and add it in another PR

packages/components/src/List/ListComposition/Manager.component.js
```diff
@@ -0,0 +1,56 @@
...
+		onDisplayModeChange: PropTypes.func,
+	};
+
+	static getDerivedStateFromProps(props) {
+		return { collection: props.collection };
+	}
+
+	constructor(props) {
+		super(props);
+		this.onDisplayModeChange = this.onDisplayModeChange.bind(this);
```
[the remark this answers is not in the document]
jsomsanith: Yes I like this way to bind this, in a class :D

packages/components/src/List/ListComposition/Manager.component.js
```diff
@@ -0,0 +1,56 @@
...
+		if (this.props.onDisplayModeChange) {
+			// controlled display mode
+			this.props.onDisplayModeChange(event, displayMode);
+		} else {
+			// uncontrolled
+			this.setState({ displayMode });
+		}
+	}
+
+	getCurrentValue(keys) {
```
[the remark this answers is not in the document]
jsomsanith: I hate passing `props` and `state` to external function...
In my head they are kind of internal, personal, and component must which prop/state it passes.

We don't need it anywhere for now, let's leave it here, we'll reconsider when it's needed

packages/components/src/List/ListComposition/Manager.component.js
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	constructor(props) {
+		super(props);
+		this.onDisplayModeChange = this.onDisplayModeChange.bind(this);
+
+		this.state = {};
+	}
+
+	onDisplayModeChange(event, displayMode) {
```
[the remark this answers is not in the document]
jsomsanith: As explained in another of your comment, it's conditional

packages/components/src/List/ListComposition/Manager.component.js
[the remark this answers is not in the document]
jsomsanith: This is removed with hooks

packages/components/src/List/ListComposition/Manager.component.js
```diff
@@ -0,0 +1,56 @@
...
+import PropTypes from 'prop-types';
+
+import { ListContext } from './context';
+
+export default class Manager extends React.Component {
+	static displayName = 'List.Manager';
+	static propTypes = {
+		children: PropTypes.node,
+
+		displayMode: PropTypes.string, // eslint-disable-line react/no-unused-prop-types
```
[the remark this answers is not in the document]
jsomsanith: Add an initialDisplayMode prop

packages/components/src/List/ListComposition/Manager.component.js
```diff
@@ -0,0 +1,56 @@
...
+	constructor(props) {
+		super(props);
+		this.onDisplayModeChange = this.onDisplayModeChange.bind(this);
+
+		this.state = {};
+	}
+
+	onDisplayModeChange(event, displayMode) {
+		if (this.props.onDisplayModeChange) {
+			// controlled display mode
```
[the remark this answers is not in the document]
jsomsanith: The controlled/uncontrolled mode is now managed by hook and context value.
Let's add a doc afterward that will be used as guideline

packages/components/src/List/ListComposition/README.md
jsomsanith: I prefer to stay with 1 unique prefix. Just reading your example, I wonder where Toolbar group comes from ? Does it work properly with the List ? Does it use the same context for synchro ? etc

packages/components/src/List/ListComposition/DisplayMode/DisplayMode.component.js
jsomsanith: `displayMode` is already the name used in context.

packages/components/src/List/ListComposition/Toolbar/Toolbar.component.js
```diff
@@ -0,0 +1,20 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import Navbar from 'react-bootstrap/lib/Navbar';
+
+import theme from './Toolbar.scss';
+
+function Toolbar(props) {
+	return (
+		<Navbar componentClass="div" className={theme['tc-list-toolbar']} role="toolbar" fluid>
+			{props.children}
```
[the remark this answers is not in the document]
jsomsanith: No the toolbar is an empty toolbar with default style. You add anything you want inside.

packages/components/src/List/ListComposition/DisplayMode/DisplayMode.component.js
```diff
@@ -0,0 +1,86 @@
...
+					id={id}
+					title={<Icon name={getIcon(mode)} />}
+					onSelect={(value, event) => onChange(event, value)}
+					aria-label={t('LIST_CHANGE_DISPLAY_MODE', {
+						defaultValue: 'Change display mode. Current display mode: {{displayMode}}.',
+						displayMode: mode,
+					})}
+				>
+					{displayModes.map(option => (
+						<MenuItem
```
nicomaligne:
```suggestion
function DisplayModeItem({ id, option, t }) {
	return (
		<MenuItem
			id={`${id}-${option}`}
			key={option}
			eventKey={option}
			aria-label={t('LIST_SELECT_DISPLAY_MODE', {
				defaultValue: 'Set {{displayMode}} as current display mode.',
				displayMode: getLabel(option, t),
			})}
		>
			<Icon name={getIcon(option)} />
			{getLabel(option, t)}
		</MenuItem>
	);
}

function DisplayMode(props) {
	const { displayMode, onDisplayModeChange } = useListContext();
	const { id, displayModes, mode = displayMode, onChange = onDisplayModeChange, t } = props;
	return (
		<React.Fragment>
			<Navbar.Text>
				<label htmlFor={id}>{t('LIST_TOOLBAR_DISPLAY', { defaultValue: 'Display:' })}</label>
			</Navbar.Text>
			<Nav>
				<NavDropdown
					id={id}
					title={<Icon name={getIcon(mode)} />}
					onSelect={(value, event) => onChange(event, value)}
					aria-label={t('LIST_CHANGE_DISPLAY_MODE', {
						defaultValue: 'Change display mode. Current display mode: {{displayMode}}.',
						displayMode: mode,
					})}
				>
					{displayModes.map(item => (
						<DisplayModeItem id={id} option={item} t={t} />
					))}
				</NavDropdown>
			</Nav>
		</React.Fragment>
	);
}
```

nicomaligne: ☝️  This was what I had in mind. I find it more readable. More component oriented.
Also what about put the translate t in the list context to avoid props passing ?
And lt might be crazy but what about a translate context ? It's something we want to share everywhere.

packages/components/src/List/ListComposition/Manager/Manager.component.js
```diff
@@ -0,0 +1,26 @@
...
+
+import { ListContext } from '../context';
+import useDisplayMode from './useDisplayMode';
+
+export default function Manager(props) {
+	const { initialDisplayMode, onDisplayModeChange } = props;
+	const [displayMode, setDisplayMode] = useDisplayMode(initialDisplayMode, onDisplayModeChange);
+	const contextValues = {
+		collection: props.collection,
+		displayMode: props.displayMode || displayMode,
```
nicomaligne: Do we share the display mode between multiple nested component in the list ?
It seems more natural to me to override the displaymode hook value directly in the displaymode component:

<List.DisplayMode displayMode='myCustomDisplayMode' />

nicomaligne: This leads to another question, do we want to use useDisplayMode here or in the displaymode component...
The way its design here can lead to an overwhelming props settings for the manager :
<List.Manager onChangeDisplayMode={() => myStuff} onSortBy={() =mySortBy etc... />

packages/components/src/List/ListComposition/Manager/useDisplayMode.js
```diff
@@ -0,0 +1,14 @@
...
+
+export default function useDisplayMode(initialDisplayMode, onChange) {
+	const [displayMode, setDisplayMode] = useState(initialDisplayMode);
+
+	function onDisplayModeChange(event, value) {
+		if (onChange) {
+			onChange(event, value);
+		} else {
+			setDisplayMode(value);
+		}
```
nicomaligne: What about this :
```suggestion
export default function useDisplayMode(initialDisplayMode, onChange, overrideDisplayMode){
	const [displayMode, setDisplayMode] = useState(initialDisplayMode);
	if (overrideDisplayMode) {
	     setDisplayMode(overrideDisplayMode)
	}

	function onDisplayModeChange(event, value) {
		setDisplayMode(value);
		if (onChange) {
			onChange(event, value);
		}
	}
	return [displayMode, onDisplayModeChange];
}

```
Just an idea I didn't test it, have to check what happens in render cycle. But with this solution we are making our hook the unique source of truth and with ovveriding possibility.

packages/components/src/List/ListComposition/Manager/Manager.component.js
nicomaligne: I just saw that the display mode value is used in the vlist (make sense :P)
I think we have to talk here, discuss about design and maintenability.

packages/components/src/List/ListComposition/DisplayMode/DisplayMode.component.js
jsomsanith: Yes but that doesn't work.
That's what I said about react-bootstrap. It seems that if `MenuItem` is not directly under `NavDropdown` it doesn't work. (I haven't looked at react-bootstrap code)

packages/components/src/List/ListComposition/Manager/Manager.component.js
jsomsanith: Yes, the display mode is shared between VList and DisplayMode selector.
But I think we can manage to do something about it. Let me try to move that to display mode

packages/components/src/List/ListComposition/Manager/useDisplayMode.js
jsomsanith: Not sure it's a good idea.
But that will collide with the previous comment subject. Let's try something for the previous one first

packages/components/src/List/ListComposition/README.md
```diff
@@ -0,0 +1,380 @@
+# Compound components
```
romainseb: nice markdown here 👏

packages/components/src/List/ListComposition/DisplayMode/DisplayMode.component.js
```diff
@@ -0,0 +1,103 @@
...
+	const { displayMode, setDisplayMode, t } = useListContext();
+	const {
+		id,
+		displayModes,
+		initialDisplayMode,
+		onChange,
+		selectedDisplayMode = displayMode,
+	} = props;
+
+	useEffect(() => {
```
romainseb: Is this useEffect handle some kind of componentDidMount stuff ?
My main concern is why do we trigger it on onChange modification ?

packages/components/src/List/ListComposition/Toolbar/Toolbar.component.js
```diff
@@ -0,0 +1,20 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import Navbar from 'react-bootstrap/lib/Navbar';
+
+import theme from './Toolbar.scss';
+
+function Toolbar(props) {
+	return (
+		<Navbar componentClass="div" className={theme['tc-list-toolbar']} role="toolbar" fluid>
```
romainseb: Could we have the "tc-list-toolbar" without cssModules too ? 😇

packages/components/stories/ListComposition/ListComposition.js
```diff
@@ -0,0 +1,166 @@
...
+import { action } from '@storybook/addon-actions';
+import talendIcons from '@talend/icons/dist/react';
+
+import { simpleCollection } from './collection';
+import { IconsProvider } from '../../src/index';
+import List from '../../src/List/ListComposition';
+import CellTitle from '../../src/VirtualizedList/CellTitle';
+import CellBadge from '../../src/VirtualizedList/CellBadge';
+
+const icons = {
```
romainseb: Would be great to not have this icons to be defined for this story ( but i don't care if it's not modified for now )

packages/components/src/List/ListComposition/DisplayMode/DisplayMode.component.js
jsomsanith: You're right, it's not supposed to support controlled/uncontrolled switch.
And if this changes, it will reset the state. In all cases it's not right. I remove the onChange from dependency to just execute the effect once.

packages/components/src/List/ListComposition/README.md
jsomsanith: Thanks :)

packages/components/stories/ListComposition/ListComposition.js
jsomsanith: I removed it

PR #2104, opened 2019-03-20
fix(forms/TDC-1885): focus the first error after submit

jare-talend:
when a wide form is invalid and the user submits the form we can't see why we can't valid a form.

We focus the first element in error on the form

* [X] Tests for the changes have been added (for bug fixes / features) And [non reg](./screenshots.md) done before need review


output/forms.eslint.txt
```diff
@@ -5,5 +5,8 @@ The react/require-extension rule is deprecated. Please use the import/extensions
 /home/travis/build/Talend/ui/packages/forms/src/fields/CollapsibleFieldset.js
   82:8  error  Visible, non-interactive elements should not have mouse or keyboard event listeners  jsx-a11y/no-static-element-interactions
 
-✖ 1 problem (1 error, 0 warnings)
+/home/travis/build/Talend/ui/packages/forms/src/UIForm/UIForm.component.test.js
+  6:18  error  Absolute imports should come before relative imports  import/imports-first
```
romainseb: Could you fix this lint ?

jare-talend: done

packages/forms/src/UIForm/UIForm.component.js
```diff
@@ -231,19 +236,32 @@ export class UIFormComponent extends React.Component {
 				return accu;
 			}, {});
 
-		this.props.setErrors(event, errors);
-
 		const isValid = !Object.keys(errors).length;
+		this.props.setErrors(event, errors, isValid ? noop : this.focusFirstError);
```
jsomsanith: Please remove the noop. Pass undefined

packages/forms/src/UIForm/UIForm.component.test.js
```diff
@@ -252,6 +253,11 @@ describe('UIForm component', () => {
 	});
 
 	describe('#onSubmit', () => {
+		const validProperties = {
```
jsomsanith: This seems to be used in 1 test, please move that back in the test

packages/forms/src/UIForm/UIForm.component.test.js
```diff
@@ -273,9 +279,13 @@ describe('UIForm component', () => {
...
 			// then
-			expect(props.setErrors).toBeCalledWith(submitEvent, {
-				firstname: 'Missing required field',
-			});
+			expect(props.setErrors).toBeCalledWith(
+				submitEvent,
+				{
+					firstname: 'Missing required field',
+				},
+				wrapper.instance().focusFirstError,
```
jsomsanith: Replace that to `expect.anything()`.
This is not the goal of this test, and this 3rd arg is tested in a specific test

packages/forms/src/UIForm/UIForm.component.test.js
```diff
@@ -295,11 +305,15 @@ describe('UIForm component', () => {
...
-				check: 'error added via a trigger',
-			});
+			expect(props.setErrors).toBeCalledWith(
+				submitEvent,
+				{
+					firstname: 'Missing required field',
+					lastname: 'String is too short (6 chars), minimum 10',
+					check: 'error added via a trigger',
+				},
+				wrapper.instance().focusFirstError,
```
jsomsanith: Same here

packages/forms/src/UIForm/UIForm.component.test.js
```diff
@@ -314,9 +328,13 @@ describe('UIForm component', () => {
...
 			// then
-			expect(props.setErrors).toBeCalledWith(submitEvent, {
-				firstname: 'is required',
-			});
+			expect(props.setErrors).toBeCalledWith(
+				submitEvent,
+				{
+					firstname: 'is required',
+				},
+				wrapper.instance().focusFirstError,
```
jsomsanith: same here

packages/forms/src/UIForm/UIForm.component.test.js
```diff
@@ -332,18 +350,39 @@ describe('UIForm component', () => {
...
+
+	describe('#focusFirstError', () => {
+		it('should focus the first element', () => {
+			const focus = jest.fn();
+			const querySelector = jest.fn(() => ({
+				focus,
+			}));
+			const wrapper = shallow(<UIFormComponent {...data} {...props} />);
+			wrapper.instance().formRef = { querySelector };
+			wrapper.instance().focusFirstError();
```
jsomsanith: Instead of getting an internal function, can you
1- pass a mock as setErrors
2- mount it
3- submit the form with errors
4- get the 3rd argument of setErrors
5- call it
6- check the activeElement

packages/forms/src/UIForm/UIForm.component.test.js
```diff
@@ -332,18 +350,39 @@ describe('UIForm component', () => {
...
+
+			expect(querySelector).toHaveBeenCalledWith(INVALID_FIELD_SELECTOR);
+			expect(focus).toHaveBeenCalled();
+		});
+	});
+
+	describe('#setFormRef', () => {
```
jsomsanith: Remove this test, it's about internal implem, it's indirectly tested with the focus

packages/forms/src/UIForm/UIForm.component.js
jare-talend: done

packages/forms/src/UIForm/UIForm.component.test.js
jare-talend: done

packages/forms/src/UIForm/UIForm.component.test.js
jare-talend: done

packages/forms/src/UIForm/UIForm.component.test.js
jare-talend: done

packages/forms/src/UIForm/UIForm.component.test.js
jare-talend: done

packages/forms/src/UIForm/UIForm.component.test.js
jare-talend: done

packages/forms/src/UIForm/UIForm.component.test.js
jare-talend: done

packages/forms/src/UIForm/UIForm.component.js
```diff
@@ -18,6 +18,10 @@ import { I18N_DOMAIN_FORMS } from '../constants';
 import '../translate';
 import theme from './UIForm.scss';
 
+export const INVALID_ATTRIBUTE = 'aria-invalid';
+
+const INVALID_SELECTOR = `[${INVALID_ATTRIBUTE}="true"]`;
```
jsomsanith: This feels like `const selector = 'div'` and use `selector` everywhere instead of `div`.
You can remove those 2 vars and inline the string in the only function that use it, and in the test. It's not something that will change.

packages/forms/src/UIForm/UIForm.component.test.js
```diff
@@ -331,19 +343,37 @@ describe('UIForm component', () => {
 		});
 
 		it('should call submit callback when form is valid', () => {
-			// given
 			const validProperties = {
 				...data.properties,
 				lastname: 'This has at least 10 characters',
 				firstname: 'This is required',
 			};
+			// given
```
jsomsanith: wat
The const just above is part of the `given`, why did you move this comment ?

PR #2109, opened 2019-03-22
poc(components): add css module util

romainseb:
Who here has not written this code ?
```
className={classNames(theme['myclass'], 'myclass')
```
Writing the class without the css module encapsulation is a good idea, we should keep it.
I'm just tired to write things in double.

Join with me a bright future where we can only write the class once & if this class is present in a given cssModules files it would set them both 😄
so in our component, we could only write ( with some code before oubviously :) )
```
className={theme('myclass')
```


packages/components/src/theme.js
```diff
@@ -0,0 +1,55 @@
...
+ * const theme = getTheme(myCSS);
+ * const rendererThing = <div className={theme('error', 'test')}></div>
+ * // This will output -> <div class="error test error_X341DZ"/>
+ * }
+ */
+export function getTheme(...cssThemes) {
+	return (...params) => {
+		const classnamesParams = params.reduce((acc, param) => {
+			if (typeof param === 'object') {
+				const newObj = Object.keys(param).reduce((objAcc, key) => {
```
jsomsanith: Not really important, but you can use Object.entries() to avoid calling `param[key]` :)

packages/components/src/theme.js
```diff
@@ -0,0 +1,55 @@
...
+						if (cssTheme[key]) {
+							// eslint-disable-next-line no-param-reassign
+							objAcc[cssTheme[key]] = param[key];
+						}
+					});
+					return objAcc;
+				}, {});
+				acc.push(newObj);
+			}
+			if (typeof param === 'string') {
```
sgendre: else if

packages/components/src/theme.test.js
```diff
@@ -0,0 +1,43 @@
...
+		// when
+		const result = cssTheme(className);
+		// then
+		expect(result).toEqual('prop prop_mod1 prop_mod2');
+	});
+
+	it('should add the theme classes with conditions', () => {
+		// given
+		// when
+		const result = cssTheme('toto', { prop: false, prop2: true });
```
sgendre: can you add a UT with the explicit case of an Array for your second parameter as its typeof is also Object ?

packages/components/src/theme.test.js
```diff
@@ -0,0 +1,55 @@
...
+			'prop prop_mod1 prop_mod2 test-multiple-classes test-multiple-classes-mod1 test-multiple-classes-mod2',
+		);
+	});
+
+	it('should add with arrays classes', () => {
+		// given
+		const className = 'prop';
+		const className2 = 'test-multiple-classes';
+		// when
+		const result = cssTheme([className, className2]);
```
sgendre: what about 
`cssTheme(classname, [className1, className2])` ?

romainseb: it should work, i can modify the unit test but it should work :)

PR #2114, opened 2019-03-26
feat(forms): Move delete button from arrayItem to closeable widgets

annilsson:

form components: `ArrayItem` and `CollapsibleFieldset`

talend surge examples: http://talend.surge.sh/forms/?selectedKind=Core%20fieldsets&selectedStory=Core-arrays&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

after example: http://2114.talend.surge.sh/forms/?selectedKind=Core%20fieldsets&selectedStory=Core-arrays&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

Change some icons to clarify the difference between "reorder" (left) and "expand/collapse" (right). 

**BEFORE**:
Screen Shot 2019-03-27 at 13 54 30
Screen Shot 2019-03-27 at 13 54 20

**AFTER**:
Screen Shot 2019-03-27 at 13 49 30
Screen Shot 2019-03-27 at 13 49 02

**THE PROBLEM**:
We need to move the delete and collapse button to the same component so we can reposition them. 
collapsiblefieldset

- Update icons
- Pass `transform` to the `Icon` component in `ActionIconToggle` (to rotate the arrow icons)
- check if the widget is `isCloseable` (it has the "collapse" icon). If yes, render the delete and collapse buttons from inside the widget. If no, render only the delete button from the `ArrayItem`. 


packages/forms/src/UIForm/fieldsets/CollapsibleFieldset/CollapsibleFieldset.component.test.js
```diff
@@ -93,4 +93,34 @@ describe('CollapsibleFieldset', () => {
 			'on icon click': { selector: '#my-fieldset__collapse', actionType: 'click' },
 		},
 	);
+
+	it('should not render Actions component if no actions are provided', () => {
```
jsomsanith: Descriptions have been switched

packages/forms/src/UIForm/fieldsets/Array/Array.component.test.js
```diff
@@ -382,4 +382,36 @@ describe('Array component', () => {
 			]);
 		});
 	});
+
+	describe('#isCloseable', () => {
```
jsomsanith: You should not test explicitly internal functions.
Instead, test the exposed, api. Here you should use a closeable component and check that the props.isCloseable that is passed is correct.

packages/forms/src/UIForm/fieldsets/CollapsibleFieldset/CollapsibleFieldset.scss
```diff
@@ -8,10 +8,18 @@
 	transition: box-shadow 0.3s cubic-bezier(0.25, 0.8, 0.25, 1);
 	margin-bottom: $padding-normal;
 
+	.actions button {
+		opacity: 0;
+	}
+
 	&:hover {
```
jsomsanith: It's not sufficient to display the buttons on hover for a11y.
In components, we use ally-focus-within that set a classname on all node that contains an element that has the focus. I think we should do the same here.

jsomsanith: out of scope, we already had the issue before

PR #2119, opened 2019-03-27
fix(TUI-378/multiselect): handle controlled mode every where

jmfrancois:

multiselect field in component form doesn t play well with create new item feature.

if you select all it push value `create-new` to the parent.
resolveNameForTitleMap doesn t handle new created item

implement in all handler the controlled / uncontrolled version

Add both in storybook: 
http://2119.talend.surge.sh/components/?selectedKind=MultiSelect&selectedStory=default&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

Note: the value must be unique (id)


packages/components/src/MultiSelect/MultiSelect.container.test.js
```diff
@@ -3,9 +3,41 @@ import { shallow } from 'enzyme';
 
 import Container from './MultiSelect.container';
 
+function t(id, options) {
```
jsomsanith: Please replace that `function t()` with a default props `getDefaultT()`

packages/components/src/MultiSelect/MultiSelect.container.test.js
```diff
@@ -3,9 +3,41 @@ import { shallow } from 'enzyme';
...
+
 describe('Container MultiSelect', () => {
-	it('should render MultiSelect with props', () => {
-		const wrapper = shallow(<Container />);
-		expect(wrapper.props()).toMatchSnapshot();
+	it('should render input with props', () => {
+		const wrapper = shallow(<Container.WrappedComponent id="multiselect" />);
+		expect(wrapper.find('input').props()).toMatchSnapshot();
+	});
+	it('getOptions should eleminate duplicate from parent', () => {
```
jsomsanith: Avoid testing internals, that makes the tests not robust, as it fails in case of refacto, even if the component behavior hasn't changed.
You should: 
1- mount the component
2- interact with it
3- assert what is rendered

packages/components/src/MultiSelect/MultiSelect.container.test.js
```diff
@@ -3,9 +3,41 @@ import { shallow } from 'enzyme';
...
+		// in case the parent resolveNames of existing value in selected we must remove duplicate
+		const added = { added: [{ name: 'bar', value: 'bar' }] };
+		const options = [{ name: 'Foo', value: 'foo' }, { name: 'bar', value: 'bar' }];
+		let wrapper = shallow(<Container.WrappedComponent id="multiselect" options={options} t={t} />);
+		wrapper = wrapper.setState(added);
+		expect(wrapper.instance().getOptions().length).toBe(2);
+	});
+	it('selectAll doesn t select create new option', () => {
```
jsomsanith: Same here

packages/components/stories/MultiSelect.js
```diff
@@ -1,15 +1,37 @@
...
+/**
+ * ControlledMultiSelect reproduce the ComponentForm behaviors
+ */
+function ControlledMultiSelect(props) {
+	const [selected, setSelected] = React.useState([]);
+	const onChange = (event, values) => {
+		action('changed', event, values);
+		setSelected(values);
 	};
+	const missing = selected.filter(i => !props.options.find(o => o.value === i));
```
jsomsanith:
```suggestion
	const missing = selected
		.filter(selectedItem => !props.options.find(({ value }) => value === selectedItem))
		.map(selectedItem => ({ value: selectedItem, name: selectedItem });
```

packages/components/stories/MultiSelect.js
```diff
@@ -1,15 +1,37 @@
...
+	let options = props.options;
+	if (missing.length > 0) {
+		options = props.options.concat(missing.map(v => ({value: v, name: v})));
```
jsomsanith:
```suggestion
		options = props.options.concat(missing);
```

packages/components/stories/MultiSelect.js
```diff
@@ -1,15 +1,37 @@
...
 import MultiSelect from '../src/MultiSelect';
 import IconsProvider from '../src/IconsProvider';
 
-function onSelect(value, title) {
-	return {
-		onSelect: action(`selec ${value} ${title}`),
+/**
+ * ControlledMultiSelect reproduce the ComponentForm behaviors
+ */
+function ControlledMultiSelect(props) {
```
jsomsanith: I don't get this component. It has the same behavior than the uncontrolled one.
To demonstrate the controlled component, the state should not move, and the actions should be triggered as we interact.
This behavior should be demonstrated in the component form story

packages/components/src/MultiSelect/MultiSelect.container.test.js
jmfrancois: I didn t succeed to mount the virtualized list

PR #2124, opened 2019-04-01
fix(icon): remove icon border in enumeration

lmaillet:
Hovered icons has borders in enumeration list.
See https://jira.talendforge.org/browse/TPSVC-8831

Remove the border on hovered icons and apply the guideline about icons color.
See http://guidelines.talend.com/document/92132#/ui-controls/buttons/icon-buttons

This involves at least :
* Enumerations : http://2124.talend.surge.sh/components/?selectedKind=Enumeration&selectedStory=edit%20mode&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel
* Lists : http://2124.talend.surge.sh/components/?selectedKind=VirtualizedList&selectedStory=List%20%3E%20Table&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel
* Buttons : http://2124.talend.surge.sh/theme/#buttons

:warning: **Finally, default color for icon buttons will be updated to $dove-gray in the guidelines.**


packages/components/src/Enumeration/Enumeration.scss
```diff
@@ -23,14 +23,19 @@ $tc-enumeration-font-size: 1.4rem !default;
...
 				svg {
-					fill: $tc-enumeration-svg-color;
 					height: $tc-enumeration-font-size;
 					width: $tc-enumeration-font-size;
 				}
 			}
 		}
 	}
 
+	btn-link {
```
jmfrancois: this is not a valid selector except if you have writte <btn-link>

jsomsanith: And I don't understand why this has been moved. It seems that the icon buttons are all .btn.btn-link so basically you have the same result as before the change.
Or am I missing something ?

packages/theme/src/theme/_buttons.scss
```diff
@@ -78,13 +78,16 @@
...
+				svg {
+					fill: $scooter;
+				}
 			}
 
 			&:active {
-				color: darken($scooter, 12%);
+				svg {
+					fill: darken($scooter, 12%);
+				}
```
jsomsanith: What about this ?
```css
			/* Existing code */
			&:focus,
			&:hover {
				color: $scooter;
			}

			&:active {
				color: darken($scooter, 12%);
			}
			/* End of existing code */

			svg {
				fill: currentColor;
			}
```

packages/components/src/Enumeration/Enumeration.scss
lmaillet: You're right, it was wrong and useless.

packages/theme/src/theme/_buttons.scss
lmaillet: Following the guideline, icons should not have `currentColor`, but `$silver-chalice`.
See http://guidelines.talend.com/document/92132#/ui-controls/buttons/icon-buttons

jsomsanith: We have (in a line just above my copy/paste) `.btn-icon-only { color: $gray; }`. Is this correct ?
We should either
* remove all color property and style the svg in `.btn-icon-only`
or
* manage the color property and set svg fill to currentColor

lmaillet: Finally removed because globally defined here
https://github.com/Talend/ui/blob/e73302d4644d077aff4d3833adca36730f89e82a/packages/theme/src/theme/_buttons.scss#L32

packages/theme/src/theme/_buttons.scss
```diff
@@ -74,27 +88,12 @@
...
+		@include svg-button-behavior();
 	}
 
 	&-icon-only {
 		display: inline-block;
-
-		svg {
-			vertical-align: middle;
-		}
+		@include svg-button-behavior();
```
lmaillet: Finally, only add mouse behaviors on `btn-icon-only` and `btn-link`.
Can't add the behavior globally on svg because I don't need any behavior on _only_ icons, see http://2124.talend.surge.sh/components/?selectedKind=Enumeration&selectedStory=with%20icon&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

packages/theme/src/theme/_buttons.scss
jsomsanith: Capture d’écran 2019-04-02 à 16 56 10
That made a regression.

Let's step back a little.
* you have a svg fill set to currentColor (once you revert the line). It makes sense because the icon should fit the text color
* **on btn-link and icon-only combined**, you have a specific case, where it must be `$silver-chalice`. You have 2 choices: set color to `$silver-chalice`, so the svg is filled with (currentColor), or set the svg fill. You have 2 other cases (hover and active) in this case. Stay with the same way to do it (color of svg fill).

jsomsanith: Another point, you removed the style on focus. Is it what is wanted ?
For a11y, you still have the outline that shows that you have focus on it, so it's acceptable, but need to be sure that it's what UX wants.

lmaillet: Too much dependencies :dizzy_face:
Let's fix it !

It looks strange to me to have blue icon if active, but I will ask Ulcay about that 👍

packages/theme/src/theme/_buttons.scss
```diff
@@ -26,7 +40,7 @@
 	}
 
 	svg {
-		fill: currentColor;
+		fill: $silver-chalice;
```
jsomsanith: This is still ko, all icons in all buttons are `$silver-chalice` now

packages/theme/src/theme/_buttons.scss
```diff
@@ -72,29 +86,14 @@
...
-			}
-		}
 	}
 
 	&-icon-only {
+		@include svg-button-behavior;
 		display: inline-block;
-
-		svg {
-			vertical-align: middle;
```
jsomsanith: This makes the icon appears on top

packages/theme/src/theme/_buttons.scss
```diff
@@ -72,29 +86,14 @@
...
-			}
-
-			&:active {
-				color: darken($scooter, 12%);
-			}
-		}
 	}
 
 	&-icon-only {
+		@include svg-button-behavior;
```
jsomsanith: Even in non link buttons, the hover makes it becomes `$scooter`

lmaillet:
Wait a minute, still _work in progress_ :grinning:.
I'm working on it.

lmaillet:
@jsomsanith should the carret also have _scooter_ as hover color ?

image

lmaillet:
About focus style, here is the answer from Ulcay:

> Laurent Maillet I consulted Mickael Girault about this accessibility issue:
> for focus state we keep the blue outline with the keyboard navigation.
> This is apply just with the keyboard navigation.

So I remove the `:focus` style.

jsomsanith:
For the caret I don't know --> UX

lmaillet:
> For the caret I don't know --> UX

Ulcay and Monia just said **YES, the carret also need hover color**.

packages/theme/src/theme/_buttons.scss
```diff
@@ -72,26 +89,18 @@
...
-				color: darken($scooter, 12%);
-			}
-		}
 	}
 
 	&-icon-only {
 		display: inline-block;
 
+		&.btn-link {
+			@include svg-button-behavior;
```
jsomsanith: It seems that this block is already covered by l.92

lmaillet:
> Buttons are ok, but the colors have changed in enumeration making everything appear disabled

Color is picked form the guideline (http://guidelines.talend.com/document/92132#/ui-controls/buttons/icon-buttons)
$silver-chalice (#ababab)

packages/theme/src/theme/_buttons.scss
lmaillet: It allows to override focus color from bootstrap.
But there is maybe a better way to do ?

image

jsomsanith:
Ok for buttons colors, but this is wrong in enumeration, I think you have to see Enumeration's css

packages/theme/src/theme/_buttons.scss
jsomsanith: Not sure what you mean, but this is from our theme showroom.
I picked a button-icon-only element

* default state, the color is duplicated because of those lines
Capture d’écran 2019-04-05 à 16 38 17

* hover state, the color is also duplicated
Capture d’écran 2019-04-05 à 16 38 54

* same for the svg inside

What I wanted to tell is that L.101 override the same values L.92 introduces. So if you remove it, it produces the exact same values.
Am I missing something ?

lmaillet:
Sorry, I did not understand that the color of the icon buttons were different in the lists and enumerations :smirk:

PR #2134, opened 2019-04-03
feat(components/List): List Toolbar - select all checkbox

yyanwang:
Currently "Select All" checkbox shows on List toolbar.
Should move it on column header for better UX.
image

Related jira: https://jira.talendforge.org/browse/TMC-17011
Following changes are made:
1. Move "Select All" checkbox from List toolbar to header.
2. Remove slot "before-selectall" and "after-selectall" from toolbar.
3. Doc & tests updated.

Slot "before-selectall" and "after-selectall" have been removed from toolbar


packages/components/src/VirtualizedList/HeaderCheckbox/HeaderCheckbox.component.js
```diff
@@ -0,0 +1,56 @@
...
+import getDefaultT from '../../translate';
+
+/**
+ * Header renderer that displays a "Select All" checkbox on header
+ */
+function HeaderCheckbox(props) {
+	const { columnData, t } = props;
+	const { id, onToggleAll, collection, isSelected } = columnData;
+	const isAllSelected = () =>
+		collection.length > 0 && collection.findIndex(item => !isSelected(item)) < 0;
```
jsomsanith:
```suggestion
		collection.length > 0 && collection.some(item => !isSelected(item));
```

jsomsanith: Please memoize it to avoid to compute it on every render

packages/components/src/VirtualizedList/HeaderCheckbox/HeaderCheckbox.component.js
```diff
@@ -0,0 +1,56 @@
...
+	return onToggleAll ? (
```
jsomsanith: Use fast return 
```javascript
function HeaderCheckbox(props) {
	if (!onToggleAll) {
		return null;
	}
	// do the render
}
```

packages/components/src/VirtualizedList/HeaderCheckbox/HeaderCheckbox.component.js
```diff
@@ -0,0 +1,56 @@
...
+		<form className={classnames('tc-list-checkbox', theme['tc-list-checkbox'])}>
+			<div className="checkbox" title={t('LIST_SELECT_ALL', { defaultValue: 'Select All' })}>
+				<label htmlFor={id && `${id}-header-check`}>
+					<input
+						id={id && `${id}-header-check`}
+						type="checkbox"
+						onChange={e => {
+							onToggleAll(e, collection);
```
jsomsanith: The collection comes from outside of the component, I think you can just call onToggleAll without passing the collection.
```
onChange={onToggleAll}
```

packages/components/src/VirtualizedList/HeaderCheckbox/HeaderCheckbox.component.js
```diff
@@ -0,0 +1,56 @@
...
+						}}
+						checked={isAllSelected()}
```
jsomsanith: Why is it a function ? Just calculate it, with memoization and pass it as props.

packages/components/src/VirtualizedList/HeaderCheckbox/HeaderCheckbox.component.js
```diff
@@ -0,0 +1,56 @@
...
+						disabled={!collection.length}
+					/>
+					<span className={'tc-header-checkbox'}>
+						<span className="sr-only">{'Select All'}</span>
```
jsomsanith: Let's avoid this double span. Just add a single `<label>` with `sr-only` classname.
`Select All` must be translated

packages/components/src/VirtualizedList/HeaderCheckbox/HeaderCheckbox.test.js
```diff
@@ -0,0 +1,43 @@
...
+		// then
+		expect(columnData.onToggleAll).toBeCalledWith(event, items);
+	});
+
+	it('should render unchecked & disabled checkbox on header when there is no items', () => {
+		// when
+		const wrapper = shallow(<HeaderCheckbox columnData={{ ...columnData, collection: [] }} />);
+
+		// then
+		expect(wrapper.getElement()).toMatchSnapshot();
```
jsomsanith: Here you just want to verify the value and disabled attributes, let's do that in js intead of a complete snapshot

yyanwang:
VirtualizedList story has been updated too :)

jsomsanith:
For the select all only on table problem, I think we should do it like sorting.
• revert Toolbar.component.js
• In List.component.js we check if we have a table to pass of not the select all to toolbar

Your tests fail because of the useMemo hook and the version of enzyme we use. You can use mount instead of shallow.

packages/components/src/VirtualizedList/HeaderCheckbox/HeaderCheckbox.test.js
```diff
@@ -0,0 +1,43 @@
...
+	collection: items,
+};
+
+describe('Header "Select All" checkbox', () => {
+	it('should render a "Select All" checkbox on header when onToggleAll callback provided', () => {
+		// when
+		const wrapper = mount(<HeaderCheckbox columnData={columnData} />);
+
+		// then
+		expect(wrapper.getElement()).toMatchSnapshot();
```
jsomsanith: With mount, the snapshot is not correct anymore.
Use `enzyme-to-json`.

```javascript
import toJson from 'enzyme-to-json';

expect(toJson(wrapper)).toMatchSnapshot();
```

packages/components/src/VirtualizedList/PropTypes.js
```diff
@@ -28,7 +28,9 @@ export default {
 	onRowDoubleClick: PropTypes.func,
 	// The list scroll
 	onScroll: PropTypes.func,
-	// The row height in ListGrid rendering
+	// handler for selecting all items
+	onToggleAll: PropTypes.func,
+	// The row height in ListGrid renderingf
```
sgangnant:
```suggestion
	// The row height in ListGrid rendering
```

packages/components/src/VirtualizedList/HeaderCheckbox/HeaderCheckbox.component.js
```diff
@@ -0,0 +1,59 @@
...
+				</label>
+			</div>
+		</form>
+	);
+}
+
+HeaderCheckbox.displayName = 'VirtualizedList(HeaderCheckbox)';
+HeaderCheckbox.propTypes = {
+	// The custom props passed to <VirtualizedList.Content columnData={}>.
+	columnData: PropTypes.shape({
```
sgangnant: Do you really need to nest props this way instead of passing `id`, `collection` ... directly as props ?

packages/components/src/VirtualizedList/utils/tablerow.js
```diff
@@ -24,8 +26,15 @@ export function insertSelectionConfiguration({ children, isSelected, selectionTo
 				flexShrink={0}
 				flexGrow={0}
 				cellDataGetter={({ rowData }) => isSelected(rowData)}
-				columnData={{ label: 'Select this element', onChange: selectionToggle }}
+				columnData={{
+					label: 'Select this element',
```
sgangnant: No translation needed ?

packages/components/src/VirtualizedList/HeaderCheckbox/HeaderCheckbox.component.js
yyanwang: yes, I can't pass arbitrary props to Column component, it need changes in react-visualized.

packages/components/src/VirtualizedList/utils/tablerow.js
yyanwang: I didn't add this, just reformat. Let's add translation in another PR.

PR #2139, opened 2019-04-04
fix(TUI-380): Tooltip not shown for actions with overlay

ybaskaran:
Tooltip not displayed for actions with overlay on mouse over
https://jira.talendforge.org/browse/TUI-380

**Cause:**
The OverlayTriggerComponent is created to show PopOver on click **{ trigger: 'click' }** so the events 
onMouseOver and onFocus provided from TooltipTrigger is overridden by the Bootstrap OverlayTrigger.
**Solution:**
Attach those events manually to the OverlayTrigger
**Demo:**
Action with popover ->
http://2139.talend.surge.sh/components/?selectedKind=Action&selectedStory=default&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

### DIFF
--- a/packages/components/src/OverlayTrigger/OverlayTrigger.component.js
+++ b/packages/components/src/OverlayTrigger/OverlayTrigger.component.js
@@ -93,6 +93,11 @@ export default class OverlayTrigger extends React.Component {
 			rootClose: true,
 			trigger: 'click',
 		};
+		const tooltipProps = {
+			onMouseOver: this.props.onMouseOver,
+			onMouseOut: this.props.onMouseOut,
+			onFocus: this.props.onFocus,
+		};
 
 		if (this.props.preventScrolling) {
 			props.container = this;
@@ -105,6 +110,7 @@ export default class OverlayTrigger extends React.Component {
 					theme['tc-action-button-positionned'],
 					'tc-action-button-positionned',
 				)}
+				{...tooltipProps}
 			>
 				<BaseOverlayTrigger {...props}>{this.props.children}</BaseOverlayTrigger>
 			</span>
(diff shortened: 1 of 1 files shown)


packages/components/src/OverlayTrigger/OverlayTrigger.component.js
```diff
@@ -93,6 +93,11 @@ export default class OverlayTrigger extends React.Component {
 			rootClose: true,
 			trigger: 'click',
 		};
+		const tooltipProps = {
+			onMouseOver: this.props.onMouseOver,
+			onMouseOut: this.props.onMouseOut,
+			onFocus: this.props.onFocus,
```
jsomsanith: The span is not the element that is focused. This is never called, we should not pass this one to the span.
BTW there is an issue with tooltip display with keyboard navigation. On focus it appears, but never disappear

packages/components/src/OverlayTrigger/OverlayTrigger.component.js
```diff
@@ -105,6 +110,7 @@ export default class OverlayTrigger extends React.Component {
 					theme['tc-action-button-positionned'],
 					'tc-action-button-positionned',
 				)}
+				{...tooltipProps}
 			>
 				<BaseOverlayTrigger {...props}>{this.props.children}</BaseOverlayTrigger>
```
jsomsanith: You pass the mouseOver/Out to the span and the button

packages/components/src/OverlayTrigger/OverlayTrigger.component.js
ybaskaran: fixed, events are applied to button

packages/components/src/OverlayTrigger/OverlayTrigger.component.js
ybaskaran: fixed

PR #2141, opened 2019-04-05
fix(scripts): Update path for eslint in IDE

annilsson:
When trying to enable eslint in the IDE (with talend-scripts eslint config) the `extends` path will try to fetch an npm module prefixed with `eslint-config-` resulting in `@talend/eslint-config-scripts/preset/config/.eslintrc` (which doesn't exist). 

Update documentation to use relative path into node_modules to get the correct path `./node_modules/@talend/scripts/webapp/preset/config/.eslintrc`

### DIFF
--- a/packages/scripts/webapp/preset/README.md
+++ b/packages/scripts/webapp/preset/README.md
@@ -277,7 +277,7 @@ To use the eslint configuration in your IDE
 
 ```json
 {
-  "extends": "@talend/scritps/preset/config/.eslintrc"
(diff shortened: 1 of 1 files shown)


packages/scripts/webapp/preset/README.md
```diff
@@ -277,7 +277,7 @@ To use the eslint configuration in your IDE
 
 ```json
 {
-  "extends": "@talend/scritps/preset/config/.eslintrc"
```
annilsson: Just to note, I was aware of the typo and path diff also, and it doesn't solve the issue only changing that

PR #2145, opened 2019-04-09
fix(datagrid): implement DefaultDateCellRenderer and use int renderer for all number type

jare-talend:
we have to show a date in ISO 8601 format when the schema has a "logicalType" property to 'timestamp-millis'
For all number or float type (int, long, double, float), we have to show the int to left to right

implement a DefaultDateCellRenderer to manage this type.
implement a mapping for the number or float type to choose the int renderer


packages/datagrid/src/components/DefaultCellRenderer/DefaultValueRenderer.component.js
```diff
@@ -51,7 +51,7 @@ export default class DefaultValueRenderer extends React.Component {
 	render() {
 		let stringValue;
 
-		if (this.props.value === null) {
+		if (this.props.value === null || this.props.value === undefined) {
```
romainseb: Is this should be enought ?
```suggestion
		if (this.props.value == null) {
```

packages/datagrid/src/components/DefaultDateCellRenderer/DefaultDateCellRenderer.component.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+import DefaultValueRenderer from '../DefaultCellRenderer/DefaultValueRenderer.component';
+
+export default function DefaultDateCellRenderer({ data }) {
```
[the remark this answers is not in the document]
jare-talend: The cell value can be invalid or null. It is a case expected. So, it isn't relevant to trigger a Error in the case. So, new Date can throw an error if the type is invalid, I catch it to set the original value.

packages/datagrid/src/components/DefaultCellRenderer/mappingTypeRenderer.js
```diff
@@ -0,0 +1,10 @@
+import { PRIMITIVES_MAPPING, LOGICAL_MAPPING } from '../../constants/avro-type.constant';
+
+// eslint-disable-next-line import/prefer-default-export
```
romainseb: ( i dislike this rule 😏  )

packages/datagrid/src/constants/avro-type.constant.js
```diff
@@ -1 +1,12 @@
-export default ['boolean', 'int', 'long', 'float', 'double', 'bytes', 'string'];
+export const AVRO_TYPES = ['boolean', 'int', 'long', 'float', 'double', 'bytes', 'string'];
+export const LOGICAL_TYPES = ['timestamp-millis'];
```
romainseb: Is the only logical type for the dates ?

packages/datagrid/src/components/DefaultCellRenderer/DefaultValueRenderer.component.js
jare-talend: exact, with abstract equality undefined == null && null == null. But this rule is often bad understand. So, i checked this to be sure https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness#Loose_equality_using ;)

packages/datagrid/src/constants/avro-type.constant.js
jare-talend: yes

packages/datagrid/src/components/DefaultDateCellRenderer/DefaultDateCellRenderer.component.test.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+
+import DefaultDateCellRenderer from './DefaultDateCellRenderer.component';
+import DefaultValueRenderer from '../DefaultCellRenderer/DefaultValueRenderer.component';
+
+describe('#DefaultDateCellRenderer', () => {
+	it('should render DefaultDateCellRenderer with DefaultValueRenderer', () => {
+		const wrapper = shallow(<DefaultDateCellRenderer data={{ value: 1511873062123 }} />);
```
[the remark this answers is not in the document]
jare-talend: There is already a test to test a string value and check that this value isn't modify.

packages/datagrid/src/components/DefaultCellRenderer/DefaultValueRenderer.component.js
romainseb: Alright, i don't care that much on this point ;)

packages/datagrid/src/components/DefaultDateCellRenderer/DefaultDateCellRenderer.component.js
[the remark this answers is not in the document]
romainseb: As we have discussed it by voice, we should not keep the invariant as it's a business case that the data we have could throw this exception.
We don't want to flood the console with some wrong data to display in the datagrid ( that could be a classic use case of the datagrid :P )

packages/datagrid/src/components/DefaultDateCellRenderer/DefaultDateCellRenderer.component.test.js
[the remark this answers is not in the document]
nicomaligne: I meant that certain string format are interpreted : https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Objets_globaux/Date/toISOString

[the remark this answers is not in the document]
nicomaligne: Ok avro spec need a int or long sorry.

PR #2147, opened 2019-04-10
chore: clean dependencies

jsomsanith:
The packages has a lot of libraries as peer dependencies. This cause a lot of pain when we want to use the lib, having to add tons of dependencies that the app won't use, because the lib need them to work.
A lot of them are implementation details: keycode, invariant, immutable, ...

* Set them as dependencies. Dependencies that are more app level (react, i18next, ...) are kept as peer dependencies.
* Change some tools dependencies to accept a more larger range of talend/ui modules. Example: `@talend/html-webpack-plugin` now relies on `@talend/react-components@>2.3.0` instead of fixed version of components.

After the merge : 
* release `@talend/html-webpack-plugin@1.1.0` that relies on `@talend/react-components@>2.3.0` instead of fixed version.
* release `@talend/scripts@0.18.1` (that just upgrade `@talend/html-webpack-plugin` and accept any version of components and bootstrap theme as peer)
* release `@talend/react-cmf-router@3.2.2` that relies on `@talend/react-cmf@>2.1.0` instead of a fixed version.


packages/cmf/package.json
```diff
@@ -32,6 +32,32 @@
   "bin": {
     "cmf-settings": "./scripts/cmf-settings.js"
   },
+  "dependencies": {
+    "@babel/polyfill": "^7.0.0",
```
romainseb: This is not used in CMF source

packages/cmf/package.json
```diff
@@ -32,6 +32,32 @@
...
+    "ajv": "^6.2.1",
```
romainseb: Could we remove this one, & also the unused ajv in https://github.com/Talend/ui/blob/master/packages/cmf/scripts/cmf-settings.utils.js#L75

packages/cmf/package.json
```diff
@@ -32,6 +32,32 @@
...
+    "bson-objectid": "^1.1.5",
+    "classnames": "^2.2.5",
```
romainseb: Don't seems to be used by CMF

packages/components/package.json
```diff
@@ -36,21 +36,30 @@
...
     "immutable": "^3.8.1",
+    "invariant": "^2.2.2",
+    "keycode": "^2.2.0",
     "lodash": "^4.17.4",
     "memoize-one": "^5.0.0",
     "rc-slider": "8.6.3",
     "rc-tooltip": "3.7.3",
     "react-autowhatever": "10.2.0",
+    "react-bootstrap": "0.32.4",
+    "react-css-transition": "^0.7.4",
```
romainseb: Would be great to drop this one in favor of react-transition-group ( but not to do in this PR )

packages/containers/package.json
```diff
@@ -31,6 +31,18 @@
     "type": "git",
     "url": "https://github.com/Talend/ui.git"
   },
+  "dependencies": {
+    "bson-objectid": "^1.1.5",
```
romainseb: Should we use uuid ?

packages/containers/package.json
```diff
@@ -31,6 +31,18 @@
...
+    "classnames": "^2.2.5",
+    "immutable": "^3.8.1",
+    "invariant": "^2.2.2",
+    "keycode": "^2.2.0",
+    "lodash": "^4.17.4",
+    "memoize-one": "^4.0.0",
```
romainseb: Little tought, maybe we should use memo or reselect for this need ( not to change in this pr )

packages/cmf/package.json
jsomsanith: Removed

packages/components/package.json
jsomsanith: Yes it would be great

packages/containers/package.json
jsomsanith: In this case, it's more reselect. It's in a class, so useMemo is not an option, and in the class the usage is to convert form schemas from immutable to js.
But as you meant, it's out of the scope

PR #2154, opened 2019-04-20
chore(ResourcePicker): making sort/filter customizable in form and hi…

jlame:
It wasn't possible to customize filters and sort in form Resource picker and making the author part optional
using attribute in schema to choose which sort we want and hiding author part when the data doesn't exist


packages/components/src/ResourcePicker/Resource/Resource.snap.test.js
```diff
@@ -47,6 +47,29 @@ describe('Resource component snaps', () => {
...
+						collection,
+						rowGetter: index => collection[index],
+					},
+				},
+				index: 0,
+			};
+
+			const wrapper = shallow(<Resource.WrappedComponent {...props} />);
+
+			expect(wrapper.getElement()).toMatchSnapshot();
```
jmfrancois: could you please avoid snapshots here. Just try to find the element and check it exists is enough.
no body check snapshots when they changes

packages/forms/src/UIForm/fields/ResourcePicker/ResourcePicker.component.test.js
```diff
@@ -99,6 +99,28 @@ describe('ResourcePicker field', () => {
...
+						new Promise(resolve => {
+							resolve({ collection });
+							done();
+						}),
+				)}
+				schema={{...schema, filters: ['certified'], sort: ["name"]}}
+			/>,
+		);
+
+		expect(wrapper.getElement()).toMatchSnapshot();
```
jmfrancois: same here. just check sort and filter are here.

packages/forms/stories-core/json/fields/core-resource-picker.json
```diff
@@ -27,7 +27,9 @@
         { "action": "resourcePickerSelected", "onEvent": "change" },
         { "action": "resourcePickerFiltered", "onEvent": "filter" }
       ],
-      "multi": false
+      "multi": false,
+      "filters": ["certified", "selection"],
+      "sort": ["name"]
```
jmfrancois: AFAIK you should use `options` for that

PR #2156, opened 2019-04-23
feat(TPSVC-8903): add action btn to HeaderBar

mkondzera:
We want to have a call to action button on header, according to this item https://jira.talendforge.org/browse/TPSVC-8903
Add call to action button and callToAction prop to show this button.


packages/components/src/HeaderBar/HeaderBar.scss
```diff
@@ -78,6 +78,18 @@ $background-color-on-hover: shade($brand-primary, 30);
 			font-size: $font-size-base;
 			text-transform: uppercase;
 		}
+
+		.tc-header-bar-call-to-action {
+			:global(.btn):global(.btn-info):global(.btn-inverse) {
```
ncomont: I would prefer something like that, WDYT ?

```css
:global {
    .btn.btn-info.btn-inverse {
        // ...
    }
}
```

packages/components/src/HeaderBar/HeaderBar.scss
```diff
@@ -78,6 +78,18 @@ $background-color-on-hover: shade($brand-primary, 30);
...
+				background-color: $white;
+				border: none;
+				margin-right: 15px;
```
ncomont: Please use SASS variable instead

packages/components/src/HeaderBar/HeaderBar.scss
mkondzera: I think it's less verbal and more readable, without redundant :global

packages/components/src/HeaderBar/HeaderBar.component.js
```diff
@@ -87,6 +87,30 @@ function Environment({ getComponent, ...props }) {
 	);
 }
 
+function CallToAction({ getComponent, t, ...props }) {
+	const actionProps = {
+		bsStyle: 'info',
+		className: 'btn-inverse',
+		label: t('HEADERBAR_ACTION', { defaultValue: 'Subscribe now' }),
```
lmaillet: `label` for the Action comes from the props, so no need for default value.
And looks like `t` is not necessary.

PR #2162, opened 2019-04-26
feat(List Toolbar): Merge 2 toolbars into one line

yyanwang:
Apply new design of List Toolbar, merge 2 toolbars into one line.
[jira](https://jira.talendforge.org/browse/TMC-16846)
[guidelines](https://company-57688.frontify.com/document/92132#/navigation-layout/list-toolbar)
toolbar

Changes included:
1. move DisplayMode, Sort, Search to first line & to the right.
2. Pagination: remove Pagination from List Toolbar
3. Sort: 
    - when List in table view, toggle column header to sort; 
    - when in large view, show Sort on toolbar;
    - story "Sort - large" added.
4. Selection:
    - when List in table view, toggle column header to select all;
    - when in large view, show additional line under toolbar with checkbox to select all. 
    - story "Selection - large" added.
5. Display Mode: use icons toggle for switching List view.

1. Pagination has been removed from toolbar.
2. Following slots in List Toolbar has been removed(List.md updated):
    - before-navbar
    - after-navbar
    - before-pagination
    - after-pagination


packages/components/src/List/Toolbar/Toolbar.component.js
```diff
@@ -81,60 +79,47 @@ function Toolbar({
 		};
 	}
 	const displayModeId = id && `${id}-display-mode`;
-	const hasToolbarItem = selectAllCheckbox || display || sort || pagination || filter;
+	const showSelectAll = selectAllCheckbox && display && display.mode === 'large';
+	const showSort = sort && display && display.mode === 'large';
```
jsomsanith: Those 2 are already managed in List.component.js. No need to redefine the conditions here.
The toolbar just displays things based on what is passed as props, and the List just pass things or not with more logic.

packages/components/src/List/Toolbar/Toolbar.component.js
```diff
@@ -81,60 +79,47 @@ function Toolbar({
...
+								</li>
+							)}
+							{injected('after-displaymode')}
+						</ul>
+					</ActionBar.Content>
+				</Renderer.ActionBar>
 			)}
-			{injected('after-navbar')}
+			{injected('after-actionbar')}
+			{showSelectAll && <SelectAll {...selectAllCheckbox} t={t} />}
```
jsomsanith: This is not in the toolbar, it should not be rendered here.
I think List.component.js > ListToolbar is  a better place.

packages/components/src/List/Toolbar/Toolbar.scss
```diff
@@ -1,12 +1,14 @@
 // Toolbar styles
 $tc-list-toolbar-display-color: $silver !default;
-$tc-list-toolbar-background-color: $concrete !default;
+$tc-list-toolbar-background-color: $wild-sand !default;
+$tc-list-toolbar-height: 70px !default;
```
jsomsanith:
```suggestion
$tc-list-toolbar-height: 7rem !default;
```

packages/components/src/List/Toolbar/SelectSortBy/SelectSortBy.component.js
```diff
@@ -75,15 +86,17 @@ function SelectSortBy({ field, id, isDescending, onChange, options, t }) {
...
-					onClick={onChangeOrder}
-					aria-label={t('LIST_CHANGE_SORT_BY_ORDER', {
-						defaultValue: 'Change sort order. Current order: {{sortOrder}}.',
-						sortOrder: currentSortOrderLabel,
-					})}
-				>
-					{currentSortOrderLabel}
+				<NavItem>
+					<OrderChooser
+						id={id ? `${id}-order` : 'tc-list-toolbar-sort-order'}
```
jsomsanith: Don't set a default id if no `props.id`. Or if you need one, generate it. Having the same id for 2 lists in the page is ko.

packages/components/src/List/Toolbar/SelectSortBy/SelectSortBy.component.js
```diff
@@ -75,15 +86,17 @@ function SelectSortBy({ field, id, isDescending, onChange, options, t }) {
...
+						icon={ICONS[isDescending ? ORDERS.DESC : ORDERS.ASC]}
```
jsomsanith: Instead of `ICONS` I think you can inline the values.

packages/components/src/List/Toolbar/SelectSortBy/SelectSortBy.component.js
```diff
@@ -75,15 +86,17 @@ function SelectSortBy({ field, id, isDescending, onChange, options, t }) {
...
-					id={id && `${id}-order`}
-					onClick={onChangeOrder}
-					aria-label={t('LIST_CHANGE_SORT_BY_ORDER', {
-						defaultValue: 'Change sort order. Current order: {{sortOrder}}.',
-						sortOrder: currentSortOrderLabel,
-					})}
-				>
-					{currentSortOrderLabel}
+				<NavItem>
+					<OrderChooser
```
jsomsanith: Using a sub component of another unrelated component feels a little wrong. It's a little hidden, so changes in the ResourcePicker will break the List.
Either externalize the order chooser, either just stay with a simple button here. (BTW the caret seems weird, indicating that it's a dropdown but on click it switch :/ )

packages/components/src/List/Toolbar/DisplayModeToggle/DisplayModeToggle.component.js
```diff
@@ -0,0 +1,66 @@
...
+		case 'table':
+			return t('LIST_SELECT_DISPLAY_MODE_TABLE', { defaultValue: 'Table' });
+		case 'large':
+			return t('LIST_SELECT_DISPLAY_MODE_LARGE', { defaultValue: 'Expanded' });
+		default:
+			return t('LIST_SELECT_DISPLAY_MODE_TABLE', { defaultValue: 'Table' });
+	}
+}
+
+function DisplayModeToggle({ id, displayModes, onChange, mode, t }) {
```
jsomsanith: You can remove `SelectDisplayMode`, this component replaces it

packages/components/src/List/Toolbar/DisplayModeToggle/DisplayModeToggle.component.js
```diff
@@ -0,0 +1,66 @@
...
+	const [active, setActive] = useState(mode);
```
jsomsanith: Your use of hook here duplicate the state. In the List component, everything is pure. We should stick with this behavior.
If a project controls the state externally, your state is desynchronized.

In the List composition mode there will be a controlled (pure) and uncontrolled (with internal state) mode. But for this List legacy, let's not change the feature.

packages/components/src/List/Toolbar/Toolbar.scss
yyanwang: Hi Jimmy,
I understand `rem` should be used here, and on <html> element, font-size is 10px.
but when I set height: 7rem it actually computes to 84px. don't know when font-size changes to 12px.

packages/components/src/List/Toolbar/Toolbar.scss
```diff
@@ -1,12 +1,14 @@
 // Toolbar styles
 $tc-list-toolbar-display-color: $silver !default;
-$tc-list-toolbar-background-color: $concrete !default;
+$tc-list-toolbar-background-color: $wild-sand !default;
+$tc-list-toolbar-height: 5.83rem !default;
```
jsomsanith: This value seems weird. Guideline says 70px --> 7rem

packages/components/src/List/Toolbar/Toolbar.scss
```diff
@@ -1,12 +1,14 @@
...
 
 .tc-list-toolbar {
 	background: $tc-list-toolbar-background-color;
 	border-top: 1px solid $alto;
 	border-bottom: 1px solid $alto;
 	margin-bottom: 0;
+	padding-left: $padding-normal;
```
jsomsanith: The padding is composed by the action bar padding and this padding.
Let's not do that, remove the padding for the actionbar, and set a proper padding here ($padding-larger)

packages/components/src/List/Toolbar/Toolbar.scss
```diff
@@ -27,11 +29,63 @@ $tc-list-toolbar-background-color: $concrete !default;
...
+
+	:global(.tc-actionbar-container) {
+		background-color: $tc-list-toolbar-background-color;
+		padding-right: 0;
+		height: $tc-list-toolbar-height;
+
+		:global(.navbar-left),
+		:global(.navbar-right) {
+			> * {
+				margin-right: $padding-normal;
```
jsomsanith: The last margin interfere with defining proper space round the bar.
What we need is to set space between elements, and not at the edge. Can you remove the margin-right of the last item please ?

packages/components/src/List/Toolbar/Toolbar.scss
```diff
@@ -27,11 +29,63 @@ $tc-list-toolbar-background-color: $concrete !default;
...
+			}
+		}
+	}
+
+	:global(.list-action-bar) {
+		:global(.navbar-right) {
+			& > ul {
```
jsomsanith:
```suggestion
			> ul {
```

packages/components/src/List/Toolbar/Toolbar.scss
```diff
@@ -27,11 +29,63 @@ $tc-list-toolbar-background-color: $concrete !default;
...
+				list-style: none;
+				display: flex;
+				align-items: center;
+				margin: 0 $padding-small 0 0;
```
jsomsanith: Setting a part of the space in different places doesn't seem ok.
Can you remove this margin-right please, to set it to 0.

To manage the space on the right part of the bar : 
* for each element, don't define a space on the right. Make sure it is set to 0 (so it is exactly on the right, the $padding-larger that you have to set on the whole bar will define the the right space.
* for each element, make sure that the left space is ok. Then add a margin of the same space on the left, so it defines the right space of the previous element.
* you will have issues with separator, perhaps you can reverse the the logic. For each element, add a separator (:before instead of :after), and remove it for the first element.

packages/components/src/List/Toolbar/Toolbar.scss
```diff
@@ -27,11 +29,63 @@ $tc-list-toolbar-background-color: $concrete !default;
...
+						}
+
+						:global(.navbar-text) {
+							margin-bottom: $padding-normal;
+							font-weight: normal;
+							color: $gray;
+						}
+
+						a:hover {
+							background-color: $wild-sand;
```
jsomsanith: I don't understand where that applies

packages/components/src/List/Toolbar/Toolbar.component.js
jsomsanith: Still ko. Remove this showSort, `props.sort` already passed depending on this condition in List.component

packages/components/src/List/List.component.js
```diff
@@ -114,6 +115,20 @@ function List({
 }) {
 	const classnames = classNames('tc-list', theme.list);
 	const injected = Inject.all(getComponent, omit(components, ['toolbar', 'list']));
+	let selectAllCheckbox;
+	if (
+		displayMode === 'large' &&
```
jsomsanith:
```suggestion
		displayMode !== 'table' &&
```

jsomsanith: This should be hidden only on table display

packages/components/src/List/Toolbar/Toolbar.scss
jsomsanith: That's weird because in the storybook, 7rem renders 70px which is correct.
If it's something in your app, it should be fixed in your app. The global font size shouldn't be changed from app to app, let the theme handle it.
Or perhaps it's an option in your browser that you change ?

packages/components/src/List/Toolbar/DisplayModeToggle/DisplayModeToggle.component.js
```diff
@@ -0,0 +1,63 @@
...
+	const modes = displayModes || options;
+
+	function getActionIcon(option) {
+		return (
+			<ActionIconToggle
+				key={option}
+				id={`${id}-${option}`}
+				icon={option === 'table' ? 'talend-table' : 'talend-expanded'}
+				label={getLabel(option, t)}
+				aria-label={t('LIST_SELECT_DISPLAY_MODE', {
```
jsomsanith: This doesn't seem to be passed to the button. The label is taken into account as tooltip and read by SR.
Can you remove that and set the message as label prop instead.

```
const label = getLabel(option, t); (BTW this should return option as default, not Table)
<ActionIconToggle
	label={t('LIST_SELECT_DISPLAY_MODE', {
		defaultValue: 'Set {{displayMode}} as current display mode.',
		displayMode: label,
	})}
```

packages/components/src/List/Toolbar/SelectSortBy/SelectSortBy.component.js
```diff
@@ -75,15 +77,22 @@ function SelectSortBy({ field, id, isDescending, onChange, options, t }) {
...
+						})}
+						onClick={onChangeOrder}
+						bsStyle=""
+						className={classNames(
+							theme['tc-list-toolbar-order-chooser'],
+							'tc-list-toolbar-order-chooser',
+						)}
+					>
+						<Icon name={isDescending ? 'talend-sort-desc' : 'talend-sort-asc'} />
+						<div className={classNames('tc-order-indicator', { asc: !isDescending })} />
```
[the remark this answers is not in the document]
yyanwang: Hi @AnLefevre , what do you think? can we remove the arrow right to the sorting icon?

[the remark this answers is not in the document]
AnLefevre: Hello @yyanwang ,for now the icon there’s an issue on the icon (arrow on the left doesn’t move -> needs to be fix soon). For now keep the caret on the icon (as a V1) and in a second we’ll remove it when new icons would be ready. Thanks

packages/components/src/List/Toolbar/SelectSortBy/SelectSortBy.component.js
```diff
@@ -75,16 +77,31 @@ function SelectSortBy({ field, id, isDescending, onChange, options, t }) {
...
-					aria-label={t('LIST_CHANGE_SORT_BY_ORDER', {
-						defaultValue: 'Change sort order. Current order: {{sortOrder}}.',
-						sortOrder: currentSortOrderLabel,
-					})}
-				>
-					{currentSortOrderLabel}
-				</NavItem>
+				<li>
+					<Button
+						id={id ? `${id}-order-chooser` : uuid.v4()}
```
jsomsanith: Don't generate one if no props.id is not provided.
```suggestion
						id={id && `${id}-order-chooser`}
```

packages/components/src/List/Toolbar/DisplayModeToggle/DisplayModeToggle.component.js
jsomsanith: Still KO

PR #2165, opened 2019-04-26
feat(cmf): add http interceptors

jmfrancois:

adding dynamic headers for all http request is not possible today.
Only static headers are supported today accross default http headers.

implement http interceptors API from AngularJS and integrate it into http middleware and http saga API.
https://docs.angularjs.org/api/ng/service/$http#interceptors

add 'interceptors' array support in cmfModule so module can declare interceptors


packages/cmf/src/interceptors.js
```diff
@@ -0,0 +1,152 @@
...
+ * @return {boolean} true if interceptor is compliant with requirements
+ */
+function isInterceptor(interceptor) {
+	if (!interceptor) {
+		return false;
+	}
+	if (typeof interceptor !== 'object') {
+		return false;
+	}
+	if (!interceptor.request && !interceptor.response) {
```
jsomsanith: I'm not sure this condition is ok.
For example we can have an interceptor that check responseError, and if 401 we redirect. This doesn't imply request or response.

packages/cmf/src/interceptors.js
```diff
@@ -0,0 +1,152 @@
...
+/**
+ * @private
+ * onData is the common caller to interceptors
+ * @param {string} event one of ['request', 'response']
+ * @param {Object}
+ */
+function onData(event, data) {
+	const copy = interceptors.slice(0);
+	const result = consume(copy, event, data);
+	if (result.then) {
```
jsomsanith: You can replace all the if/else return by
```
return Promise.resolve(result);
```

packages/cmf/src/interceptors.js
```diff
@@ -0,0 +1,152 @@
...
+		return result
+			.then(newConfig => consume(arr, attr, newConfig))
+			.catch(error => {
+				if (onError) {
+					return onError(error, argument);
+				}
+				throw error;
+			});
+	}
+	return consume(arr, attr, result);
```
jsomsanith: Just a suggestion:
- I don't see any code improvement with the recursion compared to a forEach in readability here.
- This is only called during `onData` that transform it to promise anyway

So what about something like that ? (I haven't tested it, perhaps I'm missing something)
```javascript
function consume(interceptorsArray, attr, argument) {
	const interceptorHandlers = interceptorsArray.map(interceptor => res => {
		const executor = interceptor[attr];
		const onErrorExecutor = interceptor[`${attr}Error`];

		try {
			return executor[attr](res);
		} catch (error) {
			if (!onErrorExecutor) {
				throw error;
			}
			return onErrorExecutor(error, res);
		}
	});

	return interceptorHandlers.reduce(
		(accu, handler) => accu.then(handler),
		Promise.resolve(argument),
	);
}

function onData(event, data) {
	return consume(interceptors.slice(0), event, data);
}
```

packages/cmf/src/interceptors.js
```diff
@@ -0,0 +1,152 @@
...
+	if (!pop[attr]) {
+		return consume(arr, attr, argument);
+	}
+	let result;
+	const onError = pop[`${attr}Error`];
+	try {
+		result = pop[attr](argument);
+	} catch (error) {
+		if (onError) {
+			result = onError(error, argument);
```
jsomsanith: I don't really get the onError. I first thought that it was like in angularJS, but here requestError and responseError are executed as request/response interceptor errors, not http call error

packages/cmf/src/interceptors.js
```diff
@@ -0,0 +1,152 @@
...
+	const pop = arr.pop();
+	if (!pop[attr]) {
+		return consume(arr, attr, argument);
+	}
+	let result;
+	const onError = pop[`${attr}Error`];
+	try {
+		result = pop[attr](argument);
+	} catch (error) {
+		if (onError) {
```
jsomsanith: This never happens because you insert default error handlers in push()

packages/cmf/src/interceptors.js
```diff
@@ -0,0 +1,152 @@
+/* eslint-disable no-console */
+/* eslint-disable no-underscore-dangle */
+const interceptors = [];
```
jsomsanith: Can we call the files and vars httpInterceptors ? Without any http context, it's not obvious, like in bootstrap() for example

packages/cmf/src/interceptors.md
```diff
@@ -0,0 +1,55 @@
...
+So you can do the following:
+
+```javascript
+import cmf from '@talend/react-cmf';
+
+function mySuperUseOfFetch(url) {
+	return cmf.interceptors
+		.onRequest({ url })
+		.then(config => fetch(config.url, config))
+		.then(interceptors.onResponse)
```
jsomsanith:
```suggestion
		.then(cmf.interceptors.onResponse)
```

packages/cmf/src/interceptors.md
```diff
@@ -0,0 +1,55 @@
...
+		.catch(interceptors.onError);
```
jsomsanith: This doesn't exist

packages/cmf/src/interceptors.js
jmfrancois: http call error can be two things which are treated by transformResponse in angular:
https://github.com/angular/angular.js/blob/master/src/ng/http.js#L1166
So either code >= 300 will result in promise reject so it will go in onError

This is where I m don t feel good. The onError can have a JS Error or a fetch response with code > 300.
reading the code let me think this is the same with angular except I don t see any try catch ...

packages/cmf/src/cmfModule.merge.js
```diff
@@ -98,6 +98,7 @@ const MERGE_FNS = {
 	components: mergeObjects,
 	expressions: mergeObjects,
 	actionCreators: mergeObjects,
+	interceptors: mergeArrays,
```
jmfrancois: #2195

PR #2176, opened 2019-05-03
feat(containers):change dirty state component form

romainseb:
Currently, when we have a component form, if the state has changed to dirty, there is no way to revert it to pristine without reset it all
Add an action to make it available
A new entry in the component form sandbox has been added in order to interact with the state


packages/containers/src/ComponentForm/ComponentForm.selectors.js
```diff
@@ -0,0 +1,16 @@
+import get from 'lodash/get';
+import Immutable from 'immutable';
+import { TCompForm } from './ComponentForm.component';
+
+export function getComponentFormState(state, componentName) {
+	return get(state, ['cmf', 'components'], Immutable.Map()).getIn(
+		[TCompForm.displayName, componentName],
+		Immutable.Map(),
+	);
+}
```
jmfrancois: This one exists and is generated by the component: `TCompForm.getState`

packages/containers/src/ComponentForm/ComponentForm.selectors.js
```diff
@@ -0,0 +1,16 @@
...
+
+export function isComponentFormDirty(state, componentName) {
+	return getComponentFormState(state, componentName).get('dirty', false);
```
jmfrancois: so here you can do
```
return TCompForm.getState(state, componentName).get('dirty', false)
```

And by default is use the default state

romainseb: Thx for this feedback, i make the changes ;)

PR #2132, opened 2019-04-03
feat(components/ListComposition): InfiniteScrollList 

sgangnant:
**What is the problem this PR is trying to solve?**
We want to have a list that supports infinite scrolling.

**What is the chosen solution to this problem?**
A new container `<InfiniteScroll />`.
This is a first version/proposal that supports basic features and that should evolve soon (see readme here https://github.com/Talend/ui/blob/ebf5d29270ae6fbecab9f7c2a81f1affc40ba3d5/packages/containers/src/InfiniteScrollList/InfiniteScrollList.md).

**Please check if the PR fulfills these requirements**

* [x] The PR commit message follows our [guidelines](https://github.com/talend/tools/blob/master/tools-root-github/CONTRIBUTING.md)
* [ ] Tests for the changes have been added (for bug fixes / features) And [non reg](./screenshots.md) done before need review
* [ ] Docs have been added / updated (for bug fixes / features)
* [ ] Related design / discussions / pages (not in jira), if any, are all linked or available in the PR

**[ ] This PR introduces a breaking change**

PR #2186, opened 2019-05-09
feat(components/ListComposition): Sort feature

sgangnant:
Implement the "Sort by" feature for the `ListComposition` component

Do the implementation, this is a follow up https://github.com/Talend/ui/pull/2132 which became to cumbersome as a PR.


packages/components/src/List/ListComposition/Manager/ListManager.component.js
```diff
@@ -1,19 +1,28 @@
...
+import { sortCollection } from './helpers';
 
 function Manager(props) {
 	const [displayMode, setDisplayMode] = useState();
+	const [sortParams, setSortParams] = useState({});
+
+	let collection = props.collection;
+
+	// Sort items
+	collection = useMemo(() => sortCollection(collection, sortParams), [collection, sortParams]);
```
romainseb: May we use the props here & remove the let at line 14 ?
```suggestion
	const collection = useMemo(() => sortCollection(props.collection, sortParams), [collection, sortParams]);
```

sgangnant: I used `let` because `collection` will be overwritten with the filter output (in an other PR), it was this way in the previous main pull request. I can change it back but we will have to change it back back later so ... 😶

romainseb: Allright :)

PR #2190, opened 2019-05-10
fix(TMC-17307): remove static height for sidepanel list item

VolodymyrKovalM:
https://jira.talendforge.org/browse/TMC-17307

Not sure why this height has been set here, if this is important let me know in comments, we will leave it like that and we will overwrite this height on our side (TMC)

But I think it's not very good to add static height here for list item, since Actions that are inside li are rendered depending on conditions, or features available, like we have in TMC, and so - there is just empty space if tab is not rendered


output/components.sasslint.txt
```diff
@@ -19,12 +19,15 @@ src/List/Toolbar/Toolbar.scss
...
+
 src/Typeahead/Typeahead.scss
   34:5  error  Mixins should come before declarations  mixins-before-declarations
   91:5  error  Nesting depth 4 greater than max of 3   nesting-depth
 
 src/VirtualizedList/RowCollapsiblePanel/RowCollapsiblePanel.scss
   2:24  error  Strings must use single quotes  quotes
 
-✖ 11 problems (9 errors, 2 warnings)
+✖ 12 problems (10 errors, 2 warnings)
```
smouillour: could you fix that? please

smouillour:
seems there is no specification of that in frontify: https://company-57688.frontify.com/document/92132#/navigation-layout/side-panel-main-menu-1528208484

jmfrancois:
This is a very long debate but to sum it up:

SidePanel has the responsabilities of the ul - li
so if you hide action only you add empty `li` which is not wanted.

This is the exact same topic has the selected item in the sidepanel.
The sidepanel must know which li is the current activated menu.

Note: I dont take any position here just sum up the situation and notice an issue in TMC.

PR #2192, opened 2019-05-10
feat: intercom integration

jsomsanith:
We need a button that will open intercom solution in the header bar.

* Create a component that links a button to intercom js api
* Integrate this button in HeaderBar


packages/components/src/ActionIntercom/Intercom.component.js
```diff
@@ -0,0 +1,83 @@
...
+	const label = show
+		? t('TC_INTERCOM_CLOSE', { defaultValue: 'Close support messenger.' })
+		: t('TC_INTERCOM_OPEN', { defaultValue: 'Open support messenger.' });
+
+	const icon = show ? 'talend-cross' : 'talend-bubbles';
+	return (
+		<TooltipTrigger label={label} tooltipPlacement="bottom">
+			<button
+				ref={ref}
+				key="only"
```
romainseb: what is this key for ?

packages/components/src/ActionIntercom/Intercom.component.js
```diff
@@ -0,0 +1,83 @@
...
+				id={id}
+				className={classnames('btn', 'btn-link', 'tc-intercom', theme['tc-intercom'], className, {
+					[theme.open]: show,
```
romainseb: Shouldn't it be more like 
```suggestion
					[theme['tc-intercom-open']]: show,
```

packages/components/src/ActionIntercom/Intercom.scss
```diff
@@ -0,0 +1,14 @@
+.tc-intercom {
+	position: relative;
+
+	&.open:after {
```
romainseb: could be replaced by 
```suggestion
	&-open:after {
```
with the modification of the class

packages/components/src/HeaderBar/HeaderBar.component.js
```diff
@@ -304,16 +326,20 @@ function HeaderBar(props) {
 						t={props.t}
 					/>
 				)}
+				{props.intercom && (
```
romainseb: Ow, a new prop again 😞 
But sure, as it's done, we can't do it another way

jsomsanith: Yes, we need to make it modular as the list composition.
But for now, that's not the scope

packages/components/src/ActionIntercom/Intercom.component.js
jsomsanith: Not necessarily.
This is really a scoped class, and adding the `tc-intercom` context in the name is not really useful. Moreover, it will be translated into a css class like `tc-intercom__tc-intercom-open__xa4e`. With the current state, the css class will be something like `tc-intercom__open__xa4e`

packages/components/src/ActionIntercom/Intercom.component.js
jsomsanith: Oh, it was an experiment with something I tried with tooltip trigger. I remove it.

packages/components/src/HeaderBar/HeaderBar.component.js
```diff
@@ -6,11 +6,13 @@ import { translate } from 'react-i18next';
 
 import Inject from '../Inject';
 import Action from '../Actions/Action';
+import ActionIntercom from '../ActionIntercom';
 import ActionDropdown from '../Actions/ActionDropdown';
 import Typeahead from '../Typeahead';
 import theme from './HeaderBar.scss';
 import I18N_DOMAIN_COMPONENTS from '../constants';
 import '../translate';
```
annilsson:
```suggestion

```
That looks strange

packages/components/src/ActionIntercom/Intercom.component.js
```diff
@@ -0,0 +1,76 @@
...
+	const ref = useRef(null);
+
+	// init intercom messenger, attaching it to the custom button
+	useEffect(() => {
+		IntercomService.boot(`#${id}`, config);
+		IntercomService.onShow(setShow.bind(null, true));
+		IntercomService.onHide(setShow.bind(null, false));
+
+		return IntercomService.shutdown;
+	}, [config, setShow]);
```
annilsson: This is not my strongest area, but I'm trying to think when `setShow` would change? 
Would you not also want to pass `id` here?

sgangnant: Same remark as @annilsson unless I'm missing something, extra params passed to `useEffect` seems to be useful for performance tweaks and avoid useless recomputation (a bit like `useMemo` does) (ref here https://reactjs.org/docs/hooks-effect.html @ "Tip: Optimiing Performance by Skipping Effects")

jsomsanith: It's possible that `setShow` doesn't change. But in this case, having it in the hook dependencies has no impact. In all cases, that makes the hook dependencies correct.
But agree with you, id is a missing dependency. I add it.

PR #2194, opened 2019-05-13
feat(Typeahead): Support custom renders

yyanwang:
I want to render a custom element in Typeahead's dropdown, but currently, Typeahead is only open for custom `noResultText`/`loadingText`/`searchingText`.
typeahead

support custom renders for results/no result/loading/searching.


packages/components/src/Typeahead/__snapshots__/Typeahead.snapshot.test.js.snap
```diff
@@ -1,5 +1,167 @@
...
+  </div>
+  <div
+    className="theme-items-container tc-typeahead-items-container theme-items-container theme-container-open"
+    id="react-autowhatever-my-search"
+    role="listbox"
+  >
+    <div
+      className="theme-is-loading is-loading"
+    >
+      Loading...
```
jsomsanith: The snapshot doesn't match the test.
Instead of snapshoting, can you add an id or a classname to your custom renderers, and select the element, either snapshoting only those, or checking that they are present.
It's quite hard the see in the whole snapshots what you need to check.

In general, snapshots are good to check the whole structure. But when you test a specific case that will change a specific part, avoid snapshoting the whole component, try to pinpoint what you need.

yyanwang: Hi @jsomsanith ,
I've replaced snapshot tests with pinpoint test.

packages/components/src/Typeahead/Typeahead.component.renderers.js
```diff
@@ -119,6 +120,16 @@ export function renderItemsContainerFactory(
...
+	if (searching && renders.renderSearching) {
+		return renders.renderSearching;
+	} else if (noResult && loading && renders.renderLoading) {
+		return renders.renderLoading;
+	} else if (noResult && renders.renderNoResult) {
+		return renders.renderNoResult;
+	} else if (!noResult && !searching && !loading && renders.renderResults) {
+		return renders.renderResults;
+	}
+
```
jmfrancois: Note when you want to customize you have two options: customize all the container (drop all existing markup) or just let the control over the content.
I just want to be sure you want that.
The other options is to replace `content` in ItemsContainerComponent.

yyanwang: @jmfrancois you're right. it can be achieved by passing content to Typeahead.
thanks for pointing this out. I will close this one.

PR #2200, opened 2019-05-16
feat(TDI-42337) : add generated icon types enum to icons jar

akhabali:
https://jira.talendforge.org/browse/TDI-42337

We generate 3 jars (api / svg / png)
**api** : contains a java annotation that will be used by component to define an icon. the list of the available icons are generated from the svg folder as a java enum
**svg** : jar contains all the svg icons
**png** : jar contains all the png icons.

The default jar contains the api, svg and png files


packages/icons/src/main/java/org/talend/icons/Type.java
```diff
@@ -0,0 +1,279 @@
+/*
```
jmfrancois: I m not sure there is any value to versionned this file.
It will be generated by the CI so never been updated by us.
please remove it.

PR #2202, opened 2019-05-17
feat(forms): add hover handlers for the submit button

ncomont:

We need to detect hover on the UIForm submit button (this trigger a preview on TDP).

Add `onSubmitEnter` and `onSubmitLeave` props.

### DIFF
--- a/packages/forms/src/UIForm/UIForm.component.js
+++ b/packages/forms/src/UIForm/UIForm.component.js
@@ -36,6 +36,7 @@ export class UIFormComponent extends React.Component {
 		this.onSubmit = this.onSubmit.bind(this);
 		this.onTrigger = this.onTrigger.bind(this);
 		this.onActionClick = this.onActionClick.bind(this);
+		this.onSubmitHover = this.onSubmitHover.bind(this);
 		this.focusFirstError = this.focusFirstError.bind(this);
 		this.setFormRef = this.setFormRef.bind(this);
 		// control the tv4 language here.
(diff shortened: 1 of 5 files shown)


packages/forms/src/UIForm/UIForm.component.js
```diff
@@ -249,6 +250,15 @@ export class UIFormComponent extends React.Component {
 		return isValid;
 	}
 
+	onSubmitHover(event) {
+		const { properties } = this.props;
+		if (this.props.moz) {
```
jsomsanith: I don't think we need that anymore. It was introduced to make UIForm compatible with react-jsonschema-form lib api.
But since it's a new feature, let's stick to only 1 clean api.

packages/forms/src/UIForm/UIForm.component.js
```diff
@@ -277,16 +272,16 @@ export class UIFormComponent extends React.Component {
...
 		const actions = this.props.actions || [
 			{
 				bsStyle: 'primary',
 				label: 'Submit',
 				type: 'submit',
 				widget: 'button',
 				position: 'right',
-				onMouseEnter: onEnterSubmit && this.onSubmitHover,
-				onMouseLeave: onLeaveSubmit,
+				onMouseEnter: onSubmitEnter && this.onSubmitHover,
```
ncomont: `const { properties, onSubmitEnter, onSubmitLeave } = this.props;`

`onSubmitEnter && event => onSubmitHover(event, properties)`

packages/forms/src/UIForm/UIForm.component.js
```diff
@@ -266,13 +266,16 @@ export class UIFormComponent extends React.Component {
...
 	render() {
+		const { onSubmitEnter, onSubmitLeave, properties } = this.props;
 		const actions = this.props.actions || [
 			{
 				bsStyle: 'primary',
 				label: 'Submit',
 				type: 'submit',
 				widget: 'button',
 				position: 'right',
+				onMouseEnter: onSubmitEnter ? event => onSubmitEnter(event, properties) : undefined,
```
jsomsanith: to avoid passing writing the ternary with `undefined`
```suggestion
				onMouseEnter: onSubmitEnter && event => onSubmitEnter(event, properties),
```

frassinier: Good catch :+1:

PR #2203, opened 2019-05-17
feat(component): TMC-13918 show full date in a tooltip when base date is relative in a column cell

yonmey:
When a datetime cell shows a relative date it'd be useful to show the full date inside a tooltip.

Wrap the tooltip in the relative date div.

image

### DIFF
--- a/packages/components/src/VirtualizedList/CellDatetime/CellDatetime.test.js
+++ b/packages/components/src/VirtualizedList/CellDatetime/CellDatetime.test.js
@@ -69,7 +69,6 @@ describe('CellDatetime', () => {
 		// when
 		const columnData = {
 			mode: 'ago',
-			fullDateInTooltip: true,
 		};
 
 		const wrapper = shallow(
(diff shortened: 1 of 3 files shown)


packages/components/src/VirtualizedList/CellDatetime/CellDatetime.test.js
```diff
@@ -64,4 +64,31 @@ describe('CellDatetime', () => {
 		expect(computedStrOffset).toEqual(expectedStrDate);
 		expect(format).toHaveBeenCalledWith(cellDataWithOffset, columnData.pattern);
 	});
+
+	it('should test CellDatetime render with tooltip', () => {
+		// when
+		const columnData = {
+			mode: 'ago',
+			fullDateInTooltip: true,
```
jmfrancois: I would go to always have the full date in tooltip.
Why make it an option ?

yonmey: It makes sense, I removed the option.

yonmey:
> Great addition 👍 Just a thought about i18n: Do we normally format dates according to current language? (instead of using a fixed `DATE_TIME_FORMAT`). If there's already a standard way in UI I think we should use it, but maybe there isn't.

What I see it's that there's always a pattern, then, as 3rd parameter sometimes a `locale` option is passed with a custom logic from this file: https://github.com/Talend/ui/blob/master/packages/components/src/DateFnsLocale/locale.js but not sure it's working properly as I tested it without success.
We can look into this and add it to another PR if so, thank you for your comment.

PR #2204, opened 2019-05-20
fix(forms): filtered properties should not be submitted

ncomont:

If a field is hidden by a condition, it will be sent anyway.

Filter properties before calling `props.onSubmit`.

### DIFF
--- a/packages/forms/src/UIForm/utils/condition.js
+++ b/packages/forms/src/UIForm/utils/condition.js
@@ -1,5 +1,6 @@
 /* eslint-disable no-use-before-define */
 import jsonLogic from 'json-logic-js';
+import omit from 'lodash/omit';
 
 function lowercase(a) {
 	return a.toLowerCase();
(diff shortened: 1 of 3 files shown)


packages/forms/src/UIForm/utils/condition.js
```diff
@@ -78,6 +79,19 @@ function resolveArrayNotation(condition, key) {
 	return acc;
 }
 
+export function filterProperties(schema, properties) {
+	function getKeys(items) {
```
frassinier: You can extract this function

packages/forms/src/UIForm/utils/condition.js
```diff
@@ -78,6 +79,19 @@ function resolveArrayNotation(condition, key) {
...
+		return items.reduce((acc, curr) => {
+			if (curr.items) {
```
sgendre: if for any reason  `curr` is undefined or null, that will crash the entire app

PR #2207, opened 2019-05-22
fix(DatePicker): picker doesn't close + can't open picker when focused

lmaillet:
Using date picker in form mode

**1- Picker doesn't close in IE**
When clicking on Done button, `onChange` is triggered in the input field (new selected date), the field get focus (line [116](https://github.com/Talend/ui/blob/6169f6e16c9b5997b42ddd29406bf236742842f9/packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js#L116)) and this trigger openPicker (line [107](https://github.com/Talend/ui/blob/6169f6e16c9b5997b42ddd29406bf236742842f9/packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js#L107)).
The state doesn't change, because picker is already open.
Then `closePicker` is triggered.

But IE works strangely, `closePicker` is triggered before setting focus. So the datePicker remains open (firstly closed, then opened in onFocus).

**2- Can't open the picker when focus is on the input field**
After selecting a date, focus is in the input field. It's then impossible to open again the picker.

1- Avoid using `focus` to know if the date picker must be shown or hidden, because after selecting a date, the input field is focused, but the picker is hidden.
Keep in state if a date has been `picked`, and on focus, only open the date picker if no date has been picked. Reset `picked` on blur.
2- Add onClick event in the input field.

### DIFF
--- a/packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
+++ b/packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
@@ -97,14 +97,16 @@ class InputDateTimePicker extends React.Component {
 
 	onBlur(event, { onReset }) {
 		onReset();
-		this.closePicker();
+		this.closePicker({ picked: false });
 		if (this.props.onBlur) {
 			this.props.onBlur(event);
 		}
 	}
 
 	onFocus() {
-		this.openPicker();
+		if (!this.state.picked) {
+			this.openPicker();
+		}
 	}
 
 	onChange(event, payload) {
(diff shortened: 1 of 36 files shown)


packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -113,7 +114,7 @@ class InputDateTimePicker extends React.Component {
 			(!this.props.formMode && !this.props.useTime && payload.origin !== 'INPUT')
 		) {
 			this.inputRef.focus();
-			this.closePicker();
+			setTimeout(() => this.closePicker());
```
jsomsanith: Not sure to understand well.
The focus is done before, and it triggers `openPicker` which will not change the state because it's already open.
Then we have the closePicker.

lmaillet: But it does not work in IE...

lmaillet: IE_datepicker

jmfrancois:
adding logs here is what I have:
onChange
Warning: This synthetic event is reused for performance reasons. If you're seeing this, you're accessing the property `nativeEvent` on a released/nullified synthetic event. This is set to null. If you must keep the original synthetic event around, use event.persist(). See https://fb.me/react-event-pooling for more information. << this one 10 times
InputDateTimePicker.component.js:107 onFocus
InputDateTimePicker.component.js:124 setPickerVisibility true undefined
InputDateTimePicker.component.js:124 setPickerVisibility false undefined
InputDateTimePicker.component.js:130 setPickerVisibility.setState return null ? true
InputDateTimePicker.component.js:130 setPickerVisibility.setState return null ? false
InputDateTimePicker.component.js:139 render

PR #2209, opened 2019-05-22
fix(Datepicker): error translation are broken

jlame:
date picker crashes when selecting a date in time mode

make sure that the returned value is always an array


packages/components/src/DateTimePickers/DateTime/error-messages.js
```diff
@@ -1,37 +1,37 @@
-import { t } from 'i18next';
+import i18next from 'i18next';
 
 export default function getErrorMessage(key) {
```
smouillour: seems better to add `t` as parameter

smouillour: after discussion with @jsomsanith , this will be done in another PR

PR #2214, opened 2019-05-24
feat(i18n): Upgrade i18next

annilsson:
Using the react-i18next `translate` HOC comes with some downsides, like:

- In unit tests if you import the default exported component you're not testing _your_ component, but the i18next component (HOC). Snapshots become huge with i18next object attached. You can get around it by importing the component without the HOC, or using `MyComponent.WrappedComponent`, but it's easy to forget because not all components use the HOC.
- When using the `t` function you need to set defaultProp or it will throw errors in unit tests
- Component displaynames will become things like `Translate(MyComponent)` often adding confusion. 

Upgrade i18next and react-i18next so we can use the `useTranslation` hook ([https://react.i18next.com/latest/usetranslation-hook](https://react.i18next.com/latest/usetranslation-hook)). The exported component and its displaynames remains the same always, and `t` is no longer a prop, so no need for `defaultProps`. 

You don't **have** to use the hook if you don't want to. You can keep using the HOC, _but it has been renamed_ from `translate` to `withTranslation`. 

There is a codemod to change `translate()` import and use
```bash
npx jcodeshift <talend_ui_path>/codemodes/i18next/from_9_to_15.js <your_app_src>
```

You will likely need to update your tests
* snapshots: displayName has changed : `Translate(MyComponent)` --> `withI18nextTranslation (MyComponent)` and inserted `<I18n>` component has been removed from component tree.
* selectors: if you selected `Translate(MyComponent)` you now need to select `withI18nextTranslation (MyComponent)`

If you attach sub components to your main components (like `List.Item` for example) you need to pay attention to how you export them:

https://react.i18next.com/latest/withtranslation-hoc#hoist-non-react-statics
```js
// Wrong, this will result in having to import `MyComponent.WrappedComponent.Item`
MyComponent.Item = SomeOtherComponent;
export default withTranslation()(MyComponent);

// Correct, here you can import `MyComponent.Item`
const TranslatedMyComponent = withTranslation()(MyComponent);
TranslatedMyComponent.Item = SomeOtherComponent;
export default TranslatedMyComponent;
```


codemodes/i18next/codemod.js
```diff
@@ -0,0 +1,51 @@
...
+	return foundAndReplaced ? result : null;
+}
+
+function replaceTranslateCalls(file, j) {
+	return j(file)
+		.find(j.CallExpression, { callee: { name: 'translate' } })
+		.forEach(nodePath => {
+			nodePath.value.callee = j.identifier('withTranslation');
+		})
+		.toSource();
```
jmfrancois: In other code mode we call toSource with the following options:
```
toSource({ useTabs: true, quote: 'single', trailingComma: true })
```

why this is not the case here ?

codemodes/i18next/codemod.js
```diff
@@ -0,0 +1,51 @@
...
+export default function transformer(file, api) {
+	/*
+		return j(file.source)
+			.find(j.Identifier)
+			.filter(nodePath => nodePath.node.name === 'translate')
+			.forEach(nodePath => {
+				j(nodePath).replaceWith(j.identifier("withTranslation"));
+		  })
+			.toSource();
+		  */
```
jmfrancois: dead code please remove

codemodes/i18next/codemod.js
```diff
@@ -0,0 +1,51 @@
+/* eslint-disable no-param-reassign */
```
jmfrancois: please rename this file in sth more scriptish : `from_nine_to_fifteen.js` ?

codemodes/i18next/codemod.js
jsomsanith-tlnd: Oh yes thanks !

codemodes/i18next/codemod.js
jsomsanith-tlnd: Yep, just saw it / removed it :)

PR #2215, opened 2019-05-27
fix(Form/Array): Disable add/delete button when Array is disabled

yyanwang:
Currently add/delete button is always enabled even when disabled=true.
Use prop `disabled` to check if add/delete button should be disabled.


packages/forms/src/UIForm/fieldsets/Array/Array.component.js
```diff
@@ -139,6 +139,7 @@ export default class ArrayWidget extends React.Component {
 				onRemove={this.onRemove}
 				renderItem={this.renderItem}
 				isCloseable={this.isCloseable()}
+				disabled={schema.disabled}
```
jsomsanith: The schema seems to be passed already. Why not use props.schema instead of passing props.schema.disabled in a new props ?

yyanwang: yes, right. changed to `schema.disabled` instead.

PR #2222, opened 2019-06-04
chore(TUI-385): DEPRECATE components/src/index file

jmfrancois:

components is now a huge collections of components with their dependencies.
The bunde size is just huge and can t be optimized if every project rely on the index.

deprecate the index (it will stay here for a long time) but we should not add any new component in it.

Project should now **pick** the component they need from here.

```diff
- import { List } from '@talend/react-components';
+ import List from '@talend/react-components/lib/List';
```

This will also deprecate the containers cmfModule. So project should build their own dependencies from both container and components.


packages/components/src/index.js
```diff
@@ -60,6 +60,14 @@ import Inject from './Inject';
 import RichLayout from './RichTooltip/RichLayout';
 import RichError from './RichTooltip/RichError';
 
+// eslint-disable-next-line no-console
+console.warn(`DEPRECATED:
+  Please consider '@talend/react-components' global import as deprecated
+  You should not anymore rely on it.
+  It increase the bundle size for nothing.
+  You should pick the component you need on by one as with lodash:
+  import List from @talend/react-components';`);
```
jsomsanith-tlnd:
```suggestion
  import List from @talend/react-components/lib/List';`);
```

PR #2223, opened 2019-06-04
feat(components/VirtualizedList): RowLarge supporting unloaded data

sgangnant:
"Large" mode of lists currently does not support having unloaded data in a row (it shows an empty element).

As for the "table" mode, assume that if data is missing, it is being loaded (used for infinite scroll feature) and show its loading state (built with skeletons).
See guidelines here : https://company-57688.frontify.com/document/92132#/navigation-layout/list/empty-and-loading-list (_Loading list - Large view mode_)

To test it, change `type` prop to `"LARGE"` in `packages/components/stories/ListComposition/ListComposition.js`@329, run story book and find story called "Infinite scroll" under the "List composition" section


packages/components/src/VirtualizedList/RowLarge/RowLarge.component.js
```diff
@@ -22,6 +26,35 @@ import I18N_DOMAIN_COMPONENTS from '../../constants';
 
 const { LARGE } = listTypes;
 
+const SKELETON_SIZES = [
+	Skeleton.SIZES.xlarge,
+	Skeleton.SIZES.large,
+	Skeleton.SIZES.medium,
+];
```
sgangnant: Set of usable sizes defined in guidelines

packages/components/src/VirtualizedList/RowLarge/RowLarge.component.js
```diff
@@ -99,13 +128,19 @@ class RowLarge extends React.Component {
...
 				<div className={`tc-list-large-inner-box ${theme['inner-box']}`} key="inner-box">
-					<div className={theme.header} key="header">
-						{titleCell}
-						{selectionCell}
-					</div>
-					<dl className={`tc-list-large-content ${theme.content}`} key="content">
-						{otherFields.map(this.renderKeyValue)}
-					</dl>
+					{isEmpty(rowData) ? (
+						<LargeInnerRowLoading columns={3} rows={3} />
```
jsomsanith: You can guess the rows number with otherFields.
You can keep 3 cols by default, then the number of row is otherFields / 3 (be careful 10 other fields = 4 row)

But I think a safer way to do that would be to apply the same css as with data, and you just generate a skeleton for each `otherFields`

sgangnant: After looking at the current state of the code, I think it's a cleaner and simpler solution to compute the number of rows instead of using current rendering mechanism to show skeletons when we have a row with no data.

packages/components/src/VirtualizedList/RowLarge/RowLarge.component.js
```diff
@@ -99,13 +129,22 @@ class RowLarge extends React.Component {
...
 				<div className={`tc-list-large-inner-box ${theme['inner-box']}`} key="inner-box">
-					<div className={theme.header} key="header">
-						{titleCell}
-						{selectionCell}
-					</div>
-					<dl className={`tc-list-large-content ${theme.content}`} key="content">
-						{otherFields.map(this.renderKeyValue)}
-					</dl>
+					{isEmpty(rowData) ? (
+						<LargeInnerRowLoading
```
jsomsanith: RV pass a prop `isScrolling` so everytime you scroll, the row is rendered. You use a random to get the size of each skeleton. The combination of that is that on scroll, the size of the skeletons change.
To avoid that, you can use React.memo:
```diff
+const MemoLargeLoading = React.memo(LargeInnerRowLoading);

...
-<LargeInnerRowLoading />
+<MemoLargeLoading />
```

PR #2227, opened 2019-06-07
feat(TUI-317): support external icons bundle

jmfrancois:

We try to solve two issues at the same time:
- optimization of the bundle: a good parts of the bundle is taken by icons.
- we need to support external icons

IconsProvider will own a function to resolve HREF of the svg icon.
talend icons expose one function to resolve it also.

So when you put `<Icon name="talend-foo" />`
Icon.component call getIconHref from IconsProvider
getIconHref call IconsProvider props (app wide function to resolve the name)
if nothing is returns we use default implementation (find in which bundle the icon is.)

⚠️  LOTS OF FILES ⚠️ 

So to review this pr. first filter on packages/icons
Then look at components/src/IconsProvider
Then look at components/src/Icon


output/components.eslint.txt
```diff
@@ -9,6 +9,9 @@ The react/require-extension rule is deprecated. Please use the import/extensions
 /home/travis/build/Talend/ui/packages/components/src/HeaderBar/HeaderBar.component.js
   129:37  error  Elements with ARIA roles must use a valid, non-abstract ARIA role  jsx-a11y/aria-role
 
+/home/travis/build/Talend/ui/packages/components/src/IconsProvider/IconsProvider.component.js
```
smouillour: bad

packages/components/src/IconsProvider/IconsProvider.component.js
```diff
@@ -11,35 +40,42 @@ import talendIcons, { filters } from '@talend/icons/dist/react';
...
-					{iconset[id]}
-				</symbol>
-			))}
-			{Object.keys(filters).map((id, index) => (
-				<svg key={`svg-filter-${index}`} id={id}>
-					{filters[id]}
-				</svg>
-			))}
-		</svg>
+		<React.Fragment>
```
smouillour: why you need to have a react.fragment?

output/components.eslint.txt
jmfrancois: yep I don t have the choice here :/ do you want me to ignore the lint ? or we keep it so we know there is a hack.
It will be removed once all project has move to the new bundle

smouillour: ok, if we are sure that will be remove we can keep it for the moment 👍

packages/components/src/IconsProvider/IconsProvider.component.js
```diff
@@ -11,9 +40,13 @@ import talendIcons, { filters } from '@talend/icons/dist/react';
  * @example
 <IconsProvider />
  */
-function IconsProvider({ defaultIcons, icons }) {
+function IconsProvider({ defaultIcons, icons, getIconHref }) {
 	const iconset = Object.assign({}, defaultIcons, icons);
 	const ids = Object.keys(iconset);
+	context.ids = ids;
+	if (getIconHref && context.get !== getIconHREF) {
```
jsomsanith: I don't get this part.
Why do we configure anything in the IconProvider render ? And here you set `context.get = getIconHREF;` but in `getIconHREF` you call `context.get`. Infinite loop ?

Let's discuss about this part

packages/components/src/IconsProvider/IconsProvider.component.js
```diff
@@ -31,15 +64,16 @@ function IconsProvider({ defaultIcons, icons }) {
 }
 
 IconsProvider.displayName = 'IconsProvider';
-
+IconsProvider.context = context;
 IconsProvider.propTypes = {
 	defaultIcons: PropTypes.object, // eslint-disable-line react/forbid-prop-types
 	icons: PropTypes.object, // eslint-disable-line react/forbid-prop-types
+	getIconHref: PropTypes.func, // eslint-disable-line react/forbid-prop-types
```
jsomsanith: You take a function but it is used as boolean only.

packages/components/src/IconsProvider/README.md
```diff
@@ -0,0 +1,83 @@
...
+        ...
+		new webpack.DefinePlugin({
+			'process.env.ICON_BUNDLE': "true",
+		}),
+    ];
+```
+
+## Roadmap
+
+Bundle will becomes the default behavior.
```
jsomsanith:
```suggestion
Bundle will become the default behavior.
```

packages/components/src/IconsProvider/README.md
```diff
@@ -0,0 +1,83 @@
...
+| name | type | default | description |
+| -- | -- | -- | -- |
+| defaultIcons | object | all talend icons | the default icons provided |
+| icons | object | {} | use to add icons to the default ones |
+| getIconHref | function| noop | a function responsible to return a string used as href in svg use attribute |
+
+
+## getIconHref or How to activate bundles
+
+bundles are available through a feature flag at compile time.
```
jsomsanith:
```suggestion
Bundles are available through a feature flag at compile time.
```

packages/components/src/IconsProvider/IconsProvider.component.js
jmfrancois: I don t get the point here. It is an optional function. If I put boolean it will be always wrong

packages/icons/src/extract.js
```diff
@@ -2,20 +2,64 @@
...
-		);
+	return fs.readdirSync(dir).sort((a, b) => a.localeCompare(b, { sensitivity: 'base' }));
+}
+
+function getIconId(file) {
+	return file.split('.svg')[0];
+}
+
+function assertUnique(files, acc) {
+	const alreadyExists = Object.keys(files).filter(file => {
```
jsomsanith: Instead of going over the whole list, you can use Array.some()
```javascript
const hasDuplicate = Object.keys(files).some(file =>  acc[getIconId(file)]);
if (hasDuplicate) {
    // throw
}
```

packages/icons/src/extract.js
```diff
@@ -2,20 +2,64 @@
...
+	const dir = getAbsolutePath(folder);
+	return getFiles(folder).reduce((acc, file) => {
+		const p = path.resolve(dir, file);
+		if (fs.lstatSync(p).isDirectory()) {
+			const files = extractFiles(p);
+			// check if the same file exists in another bundle
+			assertUnique(files, acc);
+			return Object.assign(acc, files);
+		}
+		const iconId = getIconId(file);
```
jsomsanith: This code works if the folders are listed after the files.
But if they are mixed, it checks if the icons in subfolders are duplicates but not the icon files.

If you have a folder structure like that, it's ok
```
root
    |_ lol.svg
    |_ subFolder/
        |_ lol.svg
```

But this is ko
```
root
    |_ subFolder/
        |_ wtf.svg
    |_ wtf.svg
```
It will go over the subFolder, check if there are duplicates, then integrate the root > wtf.svg file, without checking.

packages/icons/src/extract.js
```diff
@@ -2,20 +2,64 @@
...
+	const dir = getAbsolutePath(folder);
+	return getFiles(folder).reduce((acc, file) => {
+		const p = path.resolve(dir, file);
+		if (fs.lstatSync(p).isDirectory()) {
+			const infos = extractInfo(p, file);
+			assertUnique(infos, acc);
+			return Object.assign(acc, infos);
+		}
+		// check if the same file exists in another bundle
+		const iconId = getIconId(file);
```
jsomsanith: Same remark here for unicity check.

packages/icons/src/extract.js
```diff
@@ -2,20 +2,64 @@
...
+function extractInfo(folder, parent) {
+	const dir = getAbsolutePath(folder);
+	return getFiles(folder).reduce((acc, file) => {
+		const p = path.resolve(dir, file);
+		if (fs.lstatSync(p).isDirectory()) {
+			const infos = extractInfo(p, file);
+			assertUnique(infos, acc);
+			return Object.assign(acc, infos);
+		}
+		// check if the same file exists in another bundle
```
jsomsanith: You can remove this comment that is not at the right place.

packages/components/src/IconsProvider/IconsProvider.component.js
```diff
@@ -1,6 +1,36 @@
 import PropTypes from 'prop-types';
 import React from 'react';
-import talendIcons, { filters } from '@talend/icons/dist/react';
+import { filters } from '@talend/icons/dist/react';
+import { getIconHref as getIcon } from '@talend/icons/dist/info';
+
+let talendIcons = {};
+
+if (!process.env.ICON_BUNDLE) {
+	console.log('WARNING ICON_BUNDLE will be the default in the next major release');
```
jsomsanith:
```suggestion
	console.warn('WARNING ICON_BUNDLE will be the default in the next major release');
```

PR #2238, opened 2019-06-11
fix(forms/TUI-382): support optional object

jmfrancois:

If you put an optional object with required field in it, you will always have the validation in error on the required field

during validation, if the value is undefined do not apply deep validation.


packages/forms/src/UIForm/utils/validation.js
```diff
@@ -116,7 +116,7 @@ export function validateSimple(
 
 	results[key] = validateValue(mergedSchema, value, properties, customValidationFn);
 
-	if (deepValidation && items) {
+	if (deepValidation && items && value) {
```
sgendre: what if value is `false` ?

jsomsanith: We are talking about deep validation. You can't deep validate a boolean

PR #2240, opened 2019-06-12
fix(forms/Date): issue when emptying picker in other case than string

jlame:
When emptying the date, the returned value is undefined and so we end up with an exception undefined is not an object.
The issue which should have been corrected by #2209 is still there, by changing the import it avoids the error. But i didn't take time to figure out what was the real issue, but it's a blocking issue for other teams.
checking if the value is truthy before making any manipulation on it.


packages/components/src/DateTimePickers/DateTime/error-messages.js
```diff
@@ -1,37 +1,37 @@
-import { t } from 'i18next';
+import i18next from 'i18next';
```
frassinier: It deserves a lint comment
```suggestion
// eslint-disable-next-line import/no-named-as-default-member
import i18next from 'i18next';
```

PR #2244, opened 2019-06-12
feat(components/notification): pause/resume when mouse is entering/leaving

frassinier:
_It refers to a TCCUI topic._

Notifications have a timer (except errors). At the end of the timer, the notification disappears. But when the user hovers on it, the timer stops.

On mouse/focus out, the timer doesn't start again.

On mouse/focus out, the timer should start from the paused state, closing the toaster when time is up.

Note that errors notifications are out of this rule if the notification center is not in use.

### DIFF
--- a/packages/components/src/Notification/Notification.component.js
+++ b/packages/components/src/Notification/Notification.component.js
@@ -134,38 +134,44 @@ function Timer(callback, delay) {
 	this.resume();
 }
 
+function Registry() {
+	const timerRegistry = {};
+
+	this.register = (notification, timer) => {
+		timerRegistry[notification.id] = timer;
+	};
+
+	this.isRegistered = notification => !!timerRegistry[notification.id];
+
+	this.pause = notification => {
+		if (this.isRegistered(notification)) {
+			timerRegistry[notification.id].pause();
+		}
+	};
+
+	this.resume = notification => {
+		if (this.isRegistered(notification)) {
+			timerRegistry[notification.id].resume();
+		}
+	};
+
+	this.cancel = notification => {
+		if (this.isRegistered(notification)) {
+			timerRegistry[notification.id].cancel();
+		}
+	};
+}
+
 class NotificationsContainer extends React.Component {
 	constructor(props) {
 		super(props);
 		this.state = {};
+		this.registry = new Registry();
 		this.onMouseEnter = this.onMouseEnter.bind(this);
 		this.onMouseOut = this.onMouseOut.bind(this);
 		this.onClick = this.onClick.bind(this);
 		this.onClose = this.onClose.bind(this);
 		this.register = this.register.bind(this);
-		this.timerRegistry = {};
-		const self = this;
-		this.registry = {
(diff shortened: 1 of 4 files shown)


output/components.eslint.txt
```diff
@@ -41,6 +44,6 @@ The react/require-extension rule is deprecated. Please use the import/extensions
 /home/travis/build/Talend/ui/packages/components/src/VirtualizedList/utils/tablerow.js
   42:3  warning  Unexpected console statement  no-console
 
-✖ 21 problems (19 errors, 2 warnings)
+✖ 22 problems (20 errors, 2 warnings)
```
smouillour: could you fix that please

packages/components/src/Notification/Notification.component.js
```diff
@@ -127,9 +150,19 @@ class NotificationsContainer extends React.Component {
 				self.timerRegistry[notification.id] = timer;
 			},
 			isRegistered: notification => !!self.timerRegistry[notification.id],
+			pause(notification) {
+				if (this.isRegistered(notification)) {
+					self.timerRegistry[notification.id].pause();
```
sgangnant: Can't `self` usages be removed ?

packages/components/src/Notification/Notification.scss
```diff
@@ -117,8 +117,7 @@ $tc-notification-icon-size: $svg-md-size !default;
 	}
 
 	&:hover > &-timer-bar {
-		transform: translateX(100%);
-		animation: none;
+		animation-play-state: paused;
```
jsomsanith: oh, nice, I didn't know this one 👍

packages/components/src/Notification/Notification.test.js
```diff
@@ -114,17 +114,24 @@ describe('Notification', () => {
...
+			const mockTimer = {
+				pause: jest.fn(),
+				resume: jest.fn(),
+				cancel: jest.fn(),
+			};
 			const instance = new NotificationsContainer({ notifications: [] });
 			instance.registry.register(notification, mockTimer);
-			expect(instance.timerRegistry[notification.id]).toEqual(mockTimer);
 			expect(instance.registry.isRegistered(notification)).toEqual(true);
+			instance.registry.pause(notification);
```
jsomsanith: I won't block the PR for that because it was already like that and refactoring the tests would be out of scope, but we should not test internal functions like that anymore. That makes the test fragile

packages/components/src/Notification/Notification.component.js
```diff
@@ -111,28 +111,67 @@ export function Notification({ notification, leaveFn, ...props }) {
 	);
 }
 
+function Timer(callback, delay) {
```
ncomont: `Timer` and `Registry` are used as classes ; why are you using functions instead of classes ?
I would prefer classes for readability reasons, and I suppose that Babel will do the job anyway.

frassinier: Done

PR #2247, opened 2019-06-13
fix(GridLayout): verticalCompact as props

jlame:
passing vertical compact as props so we can choose whether we want it or not

### DIFF
--- a/packages/components/src/GridLayout/Grid.component.js
+++ b/packages/components/src/GridLayout/Grid.component.js
@@ -36,6 +36,7 @@ function Grid({
 	isLoading = false,
 	skeletonConfiguration,
 	compactType = 'vertical',
+	verticalCompact = true,
 }) {
 	return (
 		<ResponsiveGridLayout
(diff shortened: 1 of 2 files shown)


packages/components/src/GridLayout/Grid.component.js
```diff
@@ -51,7 +52,7 @@ function Grid({
 			compactType={compactType}
 			isResizable={isResizable}
 			useCSSTransforms={false}
-			verticalCompact={false}
+			verticalCompact={verticalCompact}
```
frassinier: You could use shortcut I guess
Dunno if Prettier will not doing it for you in fact
```suggestion
			verticalCompact
```

packages/components/src/GridLayout/Grid.component.js
```diff
@@ -36,6 +36,7 @@ function Grid({
 	isLoading = false,
 	skeletonConfiguration,
 	compactType = 'vertical',
+	verticalCompact = true,
```
frassinier: defaultProps work well too :)

jmfrancois: it will be removed soon

PR #2249, opened 2019-06-13
feat(TDP-6894): datalist selected item indicator

ncomont:

In datalist component, there is no visual indication that an item is selected. We only have the text bold weight that indicates the part that matches the filter.
It can be confusing from user point of view.

http://guidelines.talend.com/document/92132#/ui-controls/dropdowns


packages/components/src/Typeahead/Typeahead.scss
```diff
@@ -176,6 +180,10 @@ $tc-typeahead-items-font-size: 1.5rem !default;
 			background: none;
 		}
 	}
+
+	&.selected {
+		background-color: rgba($scooter, 0.2);
```
jmfrancois: Warning rgba($scooter, 0.2) !== opacity 0.2 from the spec

ncomont: fixed, thanks

PR #2253, opened 2019-06-17
fix(components/ListView): Automatic rows sizing

sgangnant:
https://jira.talendforge.org/browse/TUI-334
There is an issue with the computation of rows' heights in `ListView` widgets when there are nested values, this makes a blank space appear at the bottom of each nested elements (depends on the number of children elements).

- In the case where the row height calculation prop (`getItemHeight`) is not provided, let the component compute autonoumously a row (prop used to be mandatory otherwise execution would break, so no breaking change)
- Improve the height calculation logic (by using variables from the theme and including margins in the computation)
- If the calculation prop is provided, old computation code is called (no breaking change)
- `NestedListView` widget from forms accepts an boolean `autosize` attribute in its schema so that it let the component handle the row height calculation (defaults to `false` so no breaking change)

This fix is not ideal but the ListView widget is deprecated UI wise, this addresses the issue without introducing breaking changes.


packages/components/src/ListView/Items/Items.component.js
```diff
@@ -13,6 +13,27 @@ const listClasses = classNames(theme['tc-list-items'], 'tc-list-items');
...
+ * @returns {Number}
+ */
+function pxToInt(sizeInPixels) {
+	if (!sizeInPixels) {
+		return 0;
+	}
+
+	const parsed = parseFloat(sizeInPixels.replace('px', ''));
+
+	return isNaN(parsed) ? 0 : parsed;
```
ncomont: WDYT about `return parseFloat(sizeInPixels.replace('px', '')) || 0;` ?

ncomont:
```javascript
function pxToInt(sizeInPixels = 0) {
 	return parseFloat(sizeInPixels.replace('px', '')) || 0;
}
```

packages/components/src/ListView/Items/Items.component.js
```diff
@@ -13,6 +13,27 @@ const listClasses = classNames(theme['tc-list-items'], 'tc-list-items');
...
+}
+
+const TOGGLE_ALL_ROW_HEIGHT = 40;
+const ROW_LINE_HEIGHT = 13;
+const ROW_MARGIN = pxToInt(theme.rowMargin);
+const ROW_INNER_PADDINGS = pxToInt(theme.rowInnerPaddings);
+const ROW_HEIGHT = ROW_LINE_HEIGHT + ROW_MARGIN;
```
ncomont: Could you move all of these consts on the top of the file, with others consts ?

packages/components/src/ListView/Items/Items.component.js
```diff
@@ -25,7 +46,7 @@ export class ItemsComponent extends React.PureComponent {
 
 		this.hasToggleAll = this.props.showToggleAll && this.props.items.length > 1;
 		this.cache = new CellMeasurerCache({
-			defaultHeight: props.getItemHeight(),
+			defaultHeight: props.getItemHeight ? props.getItemHeight() : ROW_HEIGHT,
```
ncomont: I would prefer a `defaultProps` `getItemHeight: () => ROW_HEIGHT;` ; WDYT ?

packages/components/src/ListView/Items/Items.component.js
```diff
@@ -55,6 +84,20 @@ export class ItemsComponent extends React.PureComponent {
 		return this.props.items.length;
 	}
 
+	oldGetRowHeight({ index }) {
+		if (this.hasToggleAll && index === 0) {
+			return 40;
```
ncomont: `TOGGLE_ALL_ROW_HEIGHT `

packages/components/src/ListView/Items/Items.component.js
sgangnant: Can't do because 2 of them use `pxToInt`

packages/components/src/ListView/Items/Items.component.js
sgangnant: If a default value is provided for getItemHeight, then old computation method will be used. I want to let the component autonomous regarding height computation when there is no method provided (as mentionned in the description, to avoid breaking changes).

packages/components/src/ListView/Items/Items.component.js
sgangnant: If the stylesheet happens to change, `theme` properties may be undefined, so would `sizeInPixels`, `.replace` on `undefined` === 💥

packages/components/src/ListView/Items/Items.component.js
sgangnant: Left old code in place, let me change that

packages/components/src/ListView/Items/Items.component.js
```diff
@@ -35,17 +56,25 @@ export class ItemsComponent extends React.PureComponent {
...
-		let extraHeight = 0;
 		const currentItem = this.getItemByIndex(index);
+
 		if (currentItem && currentItem.children && currentItem.expanded) {
-			extraHeight = currentItem.children.length * this.props.getItemHeight();
+			return (currentItem.children.length + 1) * ROW_HEIGHT + ROW_INNER_PADDINGS;
 		}
 
-		return this.props.getItemHeight() + extraHeight;
+		const isLastItem = index === this.props.items.length - Number(!this.hasToggleAll);
```
annilsson: Could you please explain this one a bit more. I get confused with `!` together with `Number()`. You take the total number of items and subtract Number(boolean)? 🤔 Is there another way to express this one?

sgangnant: I wanted to use the same logic as used above (in the code). In the case where ther is a "toggle all" item, the last item's index will equal `items.length`, it would be `items.length` - 1 otherwise

annilsson: I see, if it's already used elsewhere that's fine I guess. 

I think it may still be easier to follow if it's seen as two conditions, similar to the if condition just before this one 
```
if (this.hasToggleAll && index === 0)
``` 

So in this case, what do you think of this (seeing as it also has some previous code like it):
```
const isLastItem = index === this.props.items.length && this.hasToggleAll;
```
(I don't manage to convert it on the spot, not sure if it should be `this.hasToggleAll` or `!this.hasToggleAll`, and if it should match the length or be `this.props.items.length - 1`) 😄

packages/components/src/ListView/Items/Items.component.js
ncomont: `signInPixel` has a default value in the function signature, so this should not happen.

sgangnant: Didn't see it, will change the PR (has to be a string though, `.replace` does not exist on numbers)

PR #2258, opened 2019-06-18
feat(Containers/DeleteResource): pass action property to sucessful action

jare-talend:
when the webapp catches the action DELETE_SUCESSFUl, we haven't no information about the model

pass the model to the action

### DIFF
--- a/packages/containers/src/DeleteResource/sagas.js
+++ b/packages/containers/src/DeleteResource/sagas.js
@@ -75,6 +75,7 @@ export function* deleteResourceValidate(
 			yield put({
 				type: deleteResourceConst.DIALOG_BOX_DELETE_RESOURCE_SUCCESS,
 				model: {
+					...get(action, 'data.model', {}),
 					id: safeId,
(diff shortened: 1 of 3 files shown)


packages/containers/src/DeleteResource/sagas.js
```diff
@@ -75,6 +75,7 @@ export function* deleteResourceValidate(
 			yield put({
 				type: deleteResourceConst.DIALOG_BOX_DELETE_RESOURCE_SUCCESS,
 				model: {
+					...get(action, 'data.model', {}),
```
romainseb: Don't you want to put the model in some attribute like 'model' in order to not have override of the id / labelResource if the entity has those fields ?
```suggestion
					model: get(action, 'data.model', {}),
```

jare-talend: If you do that, we have a model inception: model.model.{}. It's weird

jare-talend: I prefer spread all model properties in model in model and override id, labelResource to avoid a breaking change

PR #2266, opened 2019-06-24
feat(datagrid): update ag-grid

jare-talend:
we use ag-grid#20

upgrade ag-grid to 20.2

Before, we use forceRedrawRows to refresh the grid after props changed on componentDidUpdate. But, this method refresh the alls grids.
Now, it is not necessary, ag-grid detects if a cell have to be refreshed when the data in rowData changes their reference.
Optimize the datagrid with pass your rowData to Ag-grid to avoid a long parsing with the				method getRowData very cost with a lot of rows and some loops in addation. If you don't use rowData, you have one loop in your application, again a loop with the getRowData and a loop in ag-grid to detect the changes. With rowData, we can avoid the loop with the getRowData

**Before**: 
`
<Datagrid forceRedrawRows={() => true} getRowDataFn={()=>rowsData} />
`
**After:**
`
<Datagrid getRowDataFn={()=>rowsData}/>
`

[ ] **This PR introduces a breaking change**


yarn.lock
```diff
@@ -11905,15 +11905,6 @@ prop-types-extra@^1.0.1:
     react-is "^16.3.2"
     warning "^3.0.0"
 
-prop-types@15.6.0:
```
jmfrancois: 👍

packages/datagrid/src/components/DefaultCellRenderer/DefaultCellRenderer.component.js
```diff
@@ -12,18 +12,23 @@ import theme from './DefaultCell.scss';
 
 export const CELL_RENDERER_COMPONENT = 'cellRenderer';
 
-export default function DefaultCellRenderer({ avroRenderer, colDef, value, getComponent, data }) {
+function DefaultCellRenderer({ avroRenderer, colDef, value, getComponent, data }) {
 	let content;
 
+	let plainValue = value;
+	if (value.toJS) {
+		plainValue = value.toJS();
```
jsomsanith-tlnd: Please memoize that

packages/datagrid/stories/datagrid.component.js
```diff
@@ -175,4 +177,41 @@ storiesOf('Component Datagrid', module)
...
 	})
-	.add('dynamic change data', () => <DynamicDataGrid />);
+	.add('@deprecated - dynamic change data with forceRedrawRows', () => (
+		<DynamicDataGrid forceRedraw />
+	))
+	.add('dynamic change data', () => <DynamicDataGrid />)
+	.add('faster datagrid', () => (
+		<div style={{ height: '100vh' }}>
+			<IconsProvider />
+			<p>
```
jsomsanith-tlnd: All this description is an explanation of the implementation. Remove that from the story please, and add it to the PR description please

packages/datagrid/src/components/DefaultCellRenderer/DefaultCellRenderer.component.js
```diff
@@ -12,13 +13,20 @@ import theme from './DefaultCell.scss';
...
 
+function convertValue(value) {
+	if (!value.toJS) {
+		return value;
+	}
+
+	return value.toJS();
+}
+
+const memoizedConvertValue = memoize(convertValue);
```
jsomsanith-tlnd: You never clear the cache, I think you you use a memoize-one solution.
Here it's a functional component, you can't attach a memoized function by instance, but there is the React.useMemo hook 🎉

jare-talend: We have decided that use react.memo isn't interesting. ag-grid use a virtualization so the row is destroy or add when the user scrolls. So, we lost the benefit of react.memo when the row is destroyed.

PR #2272, opened 2019-06-27
fix(components/Datalist): restricted mode

sgendre:
Restricted props prevents onBlur matching between inputted string and available items

Skip that unrelated condition.


packages/forms/src/UIForm/fields/Datalist/__snapshots__/Datalist.component.test.js.snap
```diff
@@ -27,7 +27,7 @@ exports[`Datalist component should render 1`] = `
     onFocus={[Function]}
     placeholder="Type here"
     readOnly={false}
-    restricted={false}
+    restricted={true}
```
ncomont: seems like the default value assigned in `get(this.props, 'schema.restricted', false)` is not taken into account, right ?

sgendre: I think you might look at the wrong place, both schema and schemaMultiSection do have `restricted: true`

ncomont: oh you're right, my bad !

packages/forms/src/UIForm/fields/Datalist/Datalist.component.js
```diff
@@ -165,6 +165,7 @@ class Datalist extends Component {
 						'aria-required': this.props.schema.required,
 						'aria-describedby': `${descriptionId} ${errorId}`,
 					}}
+					restricted={get(this.props, 'schema.restricted', false)}
```
jmfrancois: this is supposed to be in options.

packages/components/src/Datalist/Datalist.component.js
```diff
@@ -308,6 +308,13 @@ class Datalist extends Component {
 		});
 	}
 
+	clearValue() {
```
jsomsanith-tlnd: You can clear that

PR #2273, opened 2019-07-02
feat(cmf): mutating nested collections

VolodymyrKovalM:

Currently our reducers allow us to perform CRUD operations only on "top level" collections,
so if we need to change something in nested collection, we can't use advantage of our reducers, need to take nested collection from store first, modify it, and put back

The proposal is to have possibility do the same CRUD operations with data on any nested level
So we could just pass a path to collection, represented either by string separated by dots, or by array of strings

The PR may introduce breaking change, but if all those CRUD actions (COLLECTION_ADD_OR_REPLACE, CONSTANTS.COLLECTION_REMOVE, CONSTANTS.COLLECTION_MUTATE) are always dispatched using our action creators, then everything should work as before
So, if you have cases that this PR may break, please let me know


packages/cmf/src/actions/collectionsActions.js
```diff
@@ -10,42 +10,42 @@ export const { COLLECTION_ADD_OR_REPLACE, COLLECTION_REMOVE, COLLECTION_MUTATE }
...
  * curried function
- * @param {string} collectionId identifier
+ * @param {string} path - path to collection
  * @param {any} data element that represent business data
  */
-export const addOrReplace = curry((collectionId, data) => ({
+export const addOrReplace = curry((path, data) => ({
 	type: CONSTANTS.COLLECTION_ADD_OR_REPLACE,
-	collectionId,
+	path: Array.isArray(path) ? path : path.split('.'),
```
sgendre: what about regular collection name that are string ?

PR #2278, opened 2019-07-03
fix(Badge): attach the enum to the translated component

jsomsanith-tlnd:
Badge component has a SIZE enum. With the upgrade of react-i18next, the `withTranslation` HOC doesn't port those static elements anymore to the translated component.

Attache the SIZE enum to the translated component


packages/components/src/Badge/Badge.component.js
```diff
@@ -116,7 +114,7 @@ Badge.propTypes = {
 	t: PropTypes.func.isRequired,
 	style: PropTypes.object, // eslint-disable-line react/forbid-prop-types
 	className: PropTypes.string,
-	display: PropTypes.oneOf([Badge.SIZES.small, Badge.SIZES.large]),
+	display: PropTypes.oneOf([SIZES.small, SIZES.large]),
```
vbocquet:
```suggestion
	display: PropTypes.oneOf(values(SIZES.small)),
```
with a 
```
import { values } from 'lodash';
```
Just a suggestion, to make it more evolutive if a badge size is remove / add

jmfrancois: I would prefer Object.values in that case

PR #2280, opened 2019-07-03
fix(forms): Add translations for tooltips in arrayitem/collapsiblefieldset

annilsson:
Translations are missing in tooltip of fieldset components (delete and collapse). And the "collapse" button tooltip is always "collapse" even when it will "expand". 

Screen Shot 2019-07-12 at 10 53 17

Screen Shot 2019-07-12 at 11 15 07

Add missing translations, and use both "collapse" and "expand". 

Also use `useTranslation` hook to avoid issue with properties on the component (`CollapsibleFieldset.isCloseable`) together with the HOC (https://github.com/Talend/ui/wiki/Migration-path:-3.x-to-4.x)

Before: http://talend.surge.sh/forms/?selectedKind=Core%20fieldsets&selectedStory=Core-arrays&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

After: http://2280.talend.surge.sh/forms/?selectedKind=Core%20fieldsets&selectedStory=Core-arrays&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel


packages/forms/src/UIForm/fieldsets/CollapsibleFieldset/CollapsibleFieldset.component.js
```diff
@@ -12,74 +14,70 @@ function defaultTitle(_, schema) {
...
+				<div
+					onDoubleClick={toggle}
+					id={id && `${id}__title_bar`}
+					role="button"
+					className={theme['title-bar']}
 				>
-					<div
-						onDoubleClick={this.toggle}
-						id={id && `${id}__title_bar`}
+					<div // eslint-disable-line jsx-a11y/no-static-element-interactions
```
sgendre: div -> button

annilsson: I agree but I'd rather not touch the markup in this PR, especially since someone disabled this rule on purpose

jsomsanith-tlnd: This is ok, I even think we can remove it, because the accessible way to toggle is the toggle button. This is just a additional way with mouse.
But this is out of this PR scope

PR #2281, opened 2019-07-03
fix: stories with i18n, removing provider

jsomsanith-tlnd:
There is an issue with react-i18next, described here : https://github.com/i18next/react-i18next/issues/750#issuecomment-508092824

Some stories use i18next provider, others don't, which introduce a bug, trying to destructure undefined.

Remove react-i18next provider from stories, and configure i18next, exposing the configuration.

### DIFF
--- a/packages/components/stories/GuidedTour.js
+++ b/packages/components/stories/GuidedTour.js
@@ -2,8 +2,8 @@ import React from 'react';
 import { storiesOf } from '@storybook/react';
 import { action } from '@storybook/addon-actions';
 
-import { I18nextProvider, withTranslation } from 'react-i18next';
-import i18n, { LanguageSwitcher } from './config/i18n';
+import { withTranslation } from 'react-i18next';
+import { LanguageSwitcher } from './config/i18n';
 
 import { GuidedTour } from '../src/index';
 import I18N_DOMAIN_COMPONENTS from '../src/constants';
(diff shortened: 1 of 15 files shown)


packages/components/stories/GuidedTour.js
```diff
@@ -277,12 +277,10 @@ storiesOf('GuidedTour', module)
 	.addDecorator(story => (
 		<React.Fragment>
 			<LanguageSwitcher />
-			<I18nextProvider i18n={i18n}>
-				<React.Fragment>
-					{story()}
-					{getLayoutWithLoremIpsum()}
-				</React.Fragment>
-			</I18nextProvider>
+			<React.Fragment>
```
vbocquet: No need to use this `<Fragment />` anymore?

packages/datagrid/stories/datagrid.component.js
```diff
@@ -2,6 +2,7 @@ import React from 'react';
 import { storiesOf } from '@storybook/react';
 import { IconsProvider } from '@talend/react-components';
 
+import './config/i18n';
```
vbocquet: Why you need to import this now? Don't understand, stories don't seem to be broke on datagrid

vbocquet:
I just check all stories on surge, everything seem to be fix 👍

packages/datagrid/stories/datagrid.component.js
annilsson: I think it's needed to make the `i18next.init()` code run, it's not exported but just run in the global scope so it's enough to do an import. Correct me if I'm wrong

PR #2283, opened 2019-07-05
fix(components): fix warnings

yyanwang:
Attempt to reduce some warnings


packages/components/src/Layout/OneColumn/OneColumn.component.js
```diff
@@ -20,7 +21,13 @@ function OneColumn({ drawers, children, tabs, ...props }) {
...
 	};
 	return (
-		<div role="main" id="tc-layout-main" tabIndex="-1" className={container} {...props}>
+		<div
+			role="main"
+			id="tc-layout-main"
+			tabIndex="-1"
+			className={container}
+			{...omit(props, 'getComponents')}
+		>
```
yyanwang: This is to reduce console warnings `Warning: React does not recognize the `getComponent` prop on a DOM element.`

packages/forms/src/UIForm/fieldsets/Array/ArrayItem.component.js
```diff
@@ -136,7 +136,7 @@ if (process.env.NODE_ENV !== 'production') {
 		index: PropTypes.number.isRequired,
 		isClosed: PropTypes.bool,
 		onRemove: PropTypes.func.isRequired,
-		onReorder: PropTypes.func.isRequired,
+		onReorder: PropTypes.oneOfType([PropTypes.func, PropTypes.bool]).isRequired,
```
vbocquet: In some case, boolean can be passed as a callback?
Because on line 22 and 34, we call `onReorder()`
https://github.com/Talend/ui/blob/af7de6bf661e06b7195d5059d6fac74725aa6719/packages/forms/src/UIForm/fieldsets/Array/ArrayItem.component.js#L22 
https://github.com/Talend/ui/blob/af7de6bf661e06b7195d5059d6fac74725aa6719/packages/forms/src/UIForm/fieldsets/Array/ArrayItem.component.js#L34

yyanwang: right, should fix this in another way

packages/forms/src/UIForm/fieldsets/Array/DefaultArrayTemplate.component.js
```diff
@@ -69,7 +69,7 @@ function DefaultArrayTemplate(props) {
 							id={id && `${id}-control-${index}`}
 							index={index}
 							onRemove={onRemove}
-							onReorder={canReorder && onReorder}
+							onReorder={canReorder ? onReorder : undefined}
```
yyanwang: here, before fix, onReorder can be false and trigger proptype warning since it should be a function

vbocquet:
```suggestion
							onReorder={canReorder ? onReorder : undefined}
```

Understand! Last suggestion, maybe a noop from lodash? Instead, you can have some TypeError like `undefined is not a function`

yyanwang: Actually if onReorder is undefined, it won't get called because:
https://github.com/Talend/ui/blob/master/packages/forms/src/UIForm/fieldsets/Array/ArrayItem.component.js#L108
it won't bother to render a reorder button.
But if I pass a noop, it will show a reorder button and click it won't do anything.
I prefer not to show a reorder button at all. WDYT? :)

vbocquet: You'r right, thanks for the explanation 👍

PR #2284, opened 2019-07-05
fix(drawer): do not override theme for cancel Action

lmaillet:
On drawer, cancel button style also applies to the cross in the header.
And if you set a className on the cancel Action, it will override the default className from the Action in the header, and theme will not be applied.

Do not override theme in the Cancel action component, even if user add className in the CancelAction.

### DIFF
--- a/packages/components/src/Drawer/Drawer.component.js
+++ b/packages/components/src/Drawer/Drawer.component.js
@@ -97,7 +97,7 @@ export function cancelActionComponent(onCancelAction, getComponent) {
 		},
 		onCancelAction,
 	);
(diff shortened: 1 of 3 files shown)


yyanwang:
Can you please add a test for className passed through cancel button props will not be overridden?
I mean something like:
```
	it('should render cancelActionComponent with passed in className', () => {
		const wrapper = mount(cancelActionComponent({ className: 'btn-inverse' }));
		expect(wrapper.find('Action').prop('className')).toEqual('theme-tc-drawer-close-action btn-inverse');
	});
```

lmaillet:
UT are added.
Thanks @yyanwang for the sample.

PR #2289, opened 2019-07-10
feat(forms/checkbox): data-feature

frassinier:
UIForm does not deal with data-feature for each field

Be able to render data-feature attribute for Checkbox from `uiSchema`

### DIFF
--- a/packages/forms/src/UIForm/fields/CheckBox/SimpleCheckBox.component.js
+++ b/packages/forms/src/UIForm/fields/CheckBox/SimpleCheckBox.component.js
@@ -17,7 +17,7 @@ export default function SimpleCheckBox({
 	const { autoFocus } = schema;
 	return (
 		<div className={classnames('checkbox', { disabled })}>
-			<label>
+			<label data-feature={schema['data-feature']}>
 				<input
 					id={id}
 					autoFocus={autoFocus}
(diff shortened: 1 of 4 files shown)


packages/forms/src/UIForm/fields/CheckBox/SimpleCheckBox.component.js
```diff
@@ -17,7 +17,7 @@ export default function SimpleCheckBox({
 	const { autoFocus } = schema;
 	return (
 		<div className={classnames('checkbox', { disabled })}>
-			<label>
+			<label data-feature={schema['data-feature']}>
```
vbocquet: Don't know much yet schema, but are you sure that you have property `data-feature` in schema?
Maybe a 
```
import { get } from 'lodash';
get(schema, 'data-feature', '');
```
or equivalent to prevent Error ?

frassinier: Thank you @vbocquet for your comment but, no, because if `data-feature` is empty, it will generate en extra empty attribute. `schema` can't be null in that case :)

vbocquet: Perfect, thanks for information 👍

PR #2293, opened 2019-07-16
feat(components/datepicker): fixed position

frassinier:
Datepicker has `position: absolute` so it's quickly hidden when its parents have fixed height

Choose `position: fixed` to deal with parents overflow

To do so, I added a new dependency : [Popper.js](https://popper.js.org)
https://github.com/FezVrasta/react-popper
https://bundlephobia.com/result?p=react-popper@1.3.3

### DIFF
--- a/packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
+++ b/packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
@@ -2,9 +2,9 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import omit from 'lodash/omit';
-import { Overlay, Popover } from 'react-bootstrap';
 import keycode from 'keycode';
 import uuid from 'uuid';
+import { Popper } from 'react-popper';
 
 import FocusManager from '../../FocusManager';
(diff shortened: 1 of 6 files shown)


packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -133,7 +133,11 @@ class InputDateTimePicker extends React.Component {
 			if (showPicker === isShown) {
 				return extra;
 			}
-			return { showPicker: isShown, ...extra };
+			return {
+				showPicker: isShown,
+				style: this.inputRef ? { width: this.inputRef.getBoundingClientRect().width } : {},
```
jsomsanith-tlnd: Instead of getting the width of the input, requesting the DOM, can you just set a fixed size ?
I think you added that to fit the current behavior but a spec v2 will come soon, you can remove that.

frassinier:
@jmfrancois it will be hard to deal with since your viewport is smaller than our datepicker in that case.
The issue here is that our datepicker is now with a fixed position so it will be out of HTML flow.
I can hope this situation will never happen for now.

PR #2296, opened 2019-07-17
fix(ListComposition): Rename InfiniteScrollList to LazyLoadingList

lmaillet:
Current `InfiniteScrollList` component just makes lazy loading.

Rename the component form `InfiniteScrollList` to `LazyLoadingList`.


packages/components/src/List/ListComposition/index.js
```diff
@@ -8,7 +8,7 @@ import VList from './VList';
 
 export default {
 	DisplayMode,
-	InfiniteScrollList,
+	LazyLoadingList,
```
jmfrancois: AFAIK this is a breaking change

jsomsanith-tlnd:
Just for next reviewers: this is a breaking change, but on a component that we can consider as under development. So this is acceptable IMO.

packages/components/src/List/ListComposition/index.js
```diff
@@ -8,7 +8,8 @@ import VList from './VList';
...
+	InfiniteScrollList: LazyLoadingList,
```
lmaillet: Following @jmfrancois advice, also export InfiniteScrollList to avoid a breaking change.

PR #2312, opened 2019-08-01
fix(TMC-18148): fix centering datalist caret in ie11

VolodymyrKovalM:
https://jira.talendforge.org/browse/TMC-18148


packages/components/src/Typeahead/Typeahead.scss
```diff
@@ -95,6 +95,7 @@ $tc-typeahead-items-font-size: 1.5rem !default;
 			position: absolute;
 			margin: 0;
 			right: 0;
+			bottom: 0.6rem;
```
frassinier: it seems to be a magic number, could you explain why 6px is needed here, please?

VolodymyrKovalM: no any magic here

it's just a bottom position, which needs to be specified for IE11
just "right" position seems not enough for IE11 and it sets the bottom to 0 by default, that's why it's positioned at the bottom of container with position relative, and not centered

0.6rem is the value that centers it vetically in Typeahead container, decided to specify it in rems since the width and height for icon are also in rems

(tested in all browsers)

jsomsanith-tlnd: By magic number, he means that 0.6rem is not normalized and is really specific to a context. So if we use this component in another context, where we want it to have another height for example, it doesn't work. For example if we want to put that in a headerbar taking all the height.
Center things vertically should nnot be done with magic number

VolodymyrKovalM: with the last changes the caret will be centered vertically, regardless of height

although it won't  be if we are going to change the caret size in some contexts, but if are going to - then the caret width and height will need to be overwritten anyway, and so margin can be adjusted

packages/components/src/Typeahead/Typeahead.scss
```diff
@@ -93,8 +93,9 @@ $tc-typeahead-items-font-size: 1.5rem !default;
 
 		.icon-caret {
 			position: absolute;
-			margin: 0;
 			right: 0;
+			bottom: 50%;
+			margin: 0 0 -10px 0;
```
jsomsanith-tlnd: 10px is still a magic number.
(and BTW you should not use px, but rem).

Doesn't it work with flexbox ?

frassinier: Yes, you probably deal with that weird situation by using:

```diff
.icon-caret {
  position: absolute;
+  margin: 0;
+ top: 0;
  right: 0;
+ bottom: 0;
-  bottom: 50%;
-  margin: 0 0 -10px 0;
+ display: flex;
+ align-items: center;
+ justify-content: center;
}
```

VolodymyrKovalM: Thanks Guys, I've made using flexbox
even more weird that just adding display: flex; is enough

annilsson: Just adding one more alternative 😄
If you want to avoid flexbox altogether you could also do this:
```
.icon-caret {
  position: absolute;
  right: 0;
  bottom: 50%;
  transform: translateY(50%);
  margin: 0;
}
```

PR #2322, opened 2019-08-08
feat(TimePicker): update input component

yyanwang:
Update `Input.component.js` with new prop 'type', so it can render a date input or time input conditionally.

Next steps:
Add InputTimePicker.
Update context Manager to handle input change and picker change for date/time.

### DIFF
--- a/packages/components/src/DateTimePickers/DateTime/Input/Input.component.js
+++ b/packages/components/src/DateTimePickers/DateTime/Input/Input.component.js
@@ -3,10 +3,16 @@ import PropTypes from 'prop-types';
 import DebounceInput from 'react-debounce-input';
 
 import { DateTimeContext } from '../Context';
-import { extractPartsFromDateTimeFormat } from '../date-extraction';
 
 export default function Input(props) {
-	const { datetime, inputManagement, errorManagement } = useContext(DateTimeContext);
+	const {
+		datetime,
+		inputManagement,
+		dateInputManagement,
+		timeInputManagement,
+		errorManagement,
+	} = useContext(DateTimeContext);
+	const partInputManagement = props.part === 'date' ? dateInputManagement : timeInputManagement;
 
 	return (
 		<DebounceInput
@@ -15,21 +21,21 @@ export default function Input(props) {
 			className="form-control"
 			debounceTimeout={300}
 			onFocus={errorManagement.onInputFocus}
+			type="text"
 			value={datetime.textInput}
 			{...inputManagement}
+			{...partInputManagement}
 			{...props}
-			type="text"
-			placeholder={extractPartsFromDateTimeFormat(inputManagement.placeholder, props.type)}
 		/>
 	);
 }
 
 Input.defaultProps = {
-	type: 'date',
+	part: 'date',
 };
 
 Input.propTypes = {
-	type: PropTypes.oneOf(['date', 'time']).isRequired,
+	part: PropTypes.oneOf(['date', 'time']).isRequired,
 };
 
 Input.displayName = 'DateTime.Input';
--- a/packages/components/src/DateTimePickers/DateTime/Input/Input.component.js
+++ b/packages/components/src/DateTimePickers/DateTime/Input/Input.component.js
@@ -24,6 +24,10 @@ export default function Input(props) {
 	);
 }
 
+Input.defaultProps = {
+	type: 'date',
+};
+
 Input.propTypes = {
 	type: PropTypes.oneOf(['date', 'time']).isRequired,
 };
--- a/packages/components/src/DateTimePickers/DateTime/Input/Input.component.js
+++ b/packages/components/src/DateTimePickers/DateTime/Input/Input.component.js
@@ -3,9 +3,11 @@ import PropTypes from 'prop-types';
 import DebounceInput from 'react-debounce-input';
 
 import { DateTimeContext } from '../Context';
+import { extractPartsFromDateTimeFormat } from '../date-extraction';
 
 export default function Input(props) {
 	const { datetime, inputManagement, errorManagement } = useContext(DateTimeContext);
+
 	return (
 		<DebounceInput
 			aria-describedby={errorManagement.inputErrorId}
@@ -17,6 +19,7 @@ export default function Input(props) {
 			{...inputManagement}
 			{...props}
 			type="text"
+			placeholder={extractPartsFromDateTimeFormat(inputManagement.placeholder, props.type)}
 		/>
 	);
 }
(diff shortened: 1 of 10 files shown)


packages/components/src/DateTimePickers/DateTime/date-extraction.js
```diff
@@ -423,6 +423,21 @@ function extractPartsFromDateTime(datetime, options) {
 	};
 }
 
+/**
+ * extract date/time part from datetimeformat
+ * @param {string} datetimeFormat
+ * @param {string} type
+ */
+function extractPartsFromDateTimeFormat(datetimeFormat, type) {
```
jsomsanith-tlnd: Can you rename that please ?
The problem is that we have other functions named `extractPartsFromXXX()` and they all extract the same thing : date, time, datetime, textInput, errors. But this function does something else, it extract the date or time within a string

Or you can make 2 functions `extractDateFormat()` / `extractTimeFormat()`

packages/components/src/DateTimePickers/DateTime/date-extraction.js
```diff
@@ -423,6 +423,21 @@ function extractPartsFromDateTime(datetime, options) {
...
+	const formats = datetimeFormat.split(' ');
+	if (type === 'date') {
+		return formats[0];
+	} else if (formats.length === 2) {
```
jsomsanith-tlnd: This doesn't work if user add multiple blanks.
I think we can do a little better: 
```javascript
const formatParts = dateTimeFormat
    .split(/\s/g) // split from blank
    .filter(part => part); // remove empty elements

if (type === 'date') {
     return formats[0] || ''; 
}
return formats[1] || '';
```
Then you can return the first or second element.

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -161,6 +161,7 @@ class InputDateTimePicker extends React.Component {
 				inputRef={ref => {
 					this.inputRef = ref;
 				}}
+				type="date"
```
jsomsanith-tlnd: This is not what we want. This will make some browser render the native date picker. We are developing a custom picker, let's remove that and clean the tests and other files that introduce that

packages/components/src/DateTimePickers/DateTime/Input/Input.component.js
```diff
@@ -1,22 +1,35 @@
...
 			{...inputManagement}
 			{...props}
+			type="text"
+			placeholder={extractPartsFromDateTimeFormat(inputManagement.placeholder, props.type)}
 		/>
 	);
 }
+
+Input.defaultProps = {
+	type: 'date',
```
jsomsanith-tlnd: This is not what we as that would make some browser render their native picker. Moreover it is never pass to the input, as you changed the props order, overriding any type attribute by "text"

packages/components/src/DateTimePickers/DateTime/date-extraction.js
yyanwang: OK. I've renamed the function as `extractFormatByPart`

packages/components/src/DateTimePickers/DateTime/date-extraction.js
yyanwang: changed as suggested.

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
yyanwang: I am not sure if you are against with the naming `type` itself (which will make some browser render the native date picker) or the approach that I want to render date/time Input conditionally.
For now I renamed `type` to `part`, because I need a prop to flag whether it's the input for date or it's the input for time.

packages/components/src/DateTimePickers/DateTime/date-extraction.js
```diff
@@ -423,6 +423,25 @@ function extractPartsFromDateTime(datetime, options) {
...
+function extractFormatByPart({ dateFormat, useTime, useSeconds }, part = 'date') {
+	const formatParts = dateFormat
+		.split(/\s/g) // split from blank
+		.filter(formatPart => formatPart); // remove empty elements
+
+	if (part === 'date') {
+		return formatParts[0] || '';
+	}
+	if (useTime) {
+		return useSeconds ? 'HH:mm:ss' : 'HH:mm';
```
jsomsanith-tlnd: It seems that in this new change, you only have date in « dateFormat ». So I don’t see the point in extracting the date part.

Let’s see the need again, you need to get the date format and time format for placeholders.
- date : just use the dateFormat from props
- time: you can have a getTimeFormat() that generate the format depending on useSeconds. 

No need of the extractFormatByPart anymore.

yyanwang: Actually `dateFormat` can be `YYYY-MM-DD HH:mm` which includes time part.
So I still need this extractFormatByPart for get date part format.
maybe change the naming `dateFormat` to `datetimeFormat`?

packages/components/src/DateTimePickers/DateTime/Input/Input.component.test.js
```diff
@@ -35,6 +35,9 @@ describe('DateTime.Input', () => {
 	it('should call manager focus callback in input focus', () => {
 		// given
 		const managerValue = {
+			inputManagement: {
```
jsomsanith-tlnd: You added dateInputManagement or timeInputManagement, not inputManagement

packages/components/src/DateTimePickers/DateTime/Manager/Manager.component.js
```diff
@@ -12,7 +12,8 @@ import {
 	extractParts,
 	extractPartsFromDateAndTime,
 	extractPartsFromTextInput,
-	getFullDateFormat,
+	extractFormatByPart,
```
jsomsanith-tlnd: You can remove this

packages/components/src/DateTimePickers/DateTime/date-extraction.js
```diff
@@ -423,6 +423,14 @@ function extractPartsFromDateTime(datetime, options) {
 	};
 }
 
+/**
+ * get time format base on useSeconds
+ * @param {string} useSeconds
+ */
+function getTimeFormat(useSeconds) {
```
jsomsanith-tlnd: Can you add tests please ?

packages/components/src/DateTimePickers/DateTime/Input/Input.component.test.js
yyanwang: removed this change since it's not necessary. test is about input focus

packages/components/src/DateTimePickers/DateTime/Manager/Manager.component.js
yyanwang: sorry, removed.

packages/components/src/DateTimePickers/DateTime/date-extraction.js
yyanwang: Test added

PR #2329, opened 2019-08-09
feat(component/LazyLoadingList): Debounce support for callback

sgangnant:
LazyLoadingList loading callback is forwarded as-is.
It's most likely that every project using this component will need to implement a debouncing on this callback so that it is not called excessively.

Accept a `debounceDelay` prop so that parent can specify a delay for the callbacks debouncing. Default is set to `0` (meaning no debouncing) and will not affect current behavior.


packages/components/src/List/ListComposition/LazyLoadingList/LazyLoadingList.component.js
```diff
@@ -1,19 +1,26 @@
...
+const DEFAULT_DEBOUNCE_DELAY = 0;
 
 function LazyLoadingList(props) {
-	const { rowCount, threshold, minimumBatchSize, loadMoreRows, ...restProps } = props;
+	const { rowCount, threshold, minimumBatchSize, loadMoreRows, debounceDelay, ...rest } = props;
 	const { collection } = useListContext();
 
 	const isRowLoaded = ({ index }) => collection && collection[index];
 
+	let loadMoreRowsCallback = loadMoreRows;
```
annilsson: Should `loadMoreRowsCallback` not be passed to `<InfiniteLoader` `loadMoreRows` props?

sgangnant: You're right, went to quick on that 🙃

packages/components/src/List/ListComposition/LazyLoadingList/LazyLoadingList.component.js
```diff
@@ -1,23 +1,30 @@
...
+	if (loadMoreRowsCallback && debounceDelay) {
+		loadMoreRowsCallback = debounce(loadMoreRowsCallback, debounceDelay);
```
frassinier: You never cancel that debounce, dunno if it's ok with high `debounceDelay`

PR #2332, opened 2019-08-13
fix(component): Datalist icon

smouillour:
The datalist use only the name in the titlemap to display the icon.

In my case, i need to find the icon with the value.

i added my case without adding breaking change or impact
if the first check via the name is undefined, we do the same with the value

ezgif-2-bd54d6e283bf


packages/components/src/Datalist/Datalist.component.js
```diff
@@ -220,7 +220,11 @@ class Datalist extends Component {
 				);
 			}
 
-			return get(this.props.titleMap.find(titleMap => titleMap.name === this.state.value), 'icon');
+			let item = this.props.titleMap.find(titleMap => titleMap.name === this.state.value);
+			if (!item) {
+				item = this.props.titleMap.find(titleMap => titleMap.value === this.state.value);
+			}
+			if (item) return get(item, 'icon');
```
lmaillet:
```suggestion
			if (item) {
				return get(item, 'icon');
			}
```

PR #2347, opened 2019-08-21
feat(List composition): expose sort hook

jsomsanith-tlnd:
_(This is part of a series of PR to improve api, and help to switch from List legacy component to List Composition)._

Sort on columns works out of the box with a default sort. But when we want to add some custom sorting, we have to switch to controlled mode, managing the sort state, and the sorting of the collection.

Expose a sorting hook, that is used in the List manager.
So once you want to add custom sort functions
- switch to controlled mode
- use the sort hook, passing a sort function dictionnary

```javascript
const sortFunctions = {
    favorite: sortParams => (a, b) => { /* sort function on "favorite" data, depending on the sortParams */ }
    quality= sortParams => (a, b) => { /* sort function on "quality" data, depending on the sortParams */ }
};

const {
    sortedCollection,
    sortParams,
    setSortParams,
} = useCollectionSort(collection, initialSortParams, sortFunctions);
```


packages/components/src/List/ListComposition/Manager/ListManager.component.js
```diff
@@ -5,18 +5,18 @@ import { withTranslation } from 'react-i18next';
...
+
 	const [displayMode, setDisplayMode] = useState();
-	const [sortParams, setSortParams] = useState({});
 	const [textFilter, setTextFilter] = useState();
 
-	let collection = props.collection;
-
 	// Sort items
-	collection = useMemo(() => sortCollection(collection, sortParams), [collection, sortParams]);
+	const { sortedCollection, sortParams, setSortParams } = useCollectionSort(props.collection);
```
jlame: Sort parameter and the sort function passed to the useCollection are missing here ?

jlame:
> switch to controlled mode
is this needed ? in fact it's working in uncontrolled mode in surge too

packages/components/src/List/ListComposition/Manager/sort.hook.js
```diff
@@ -0,0 +1,45 @@
...
+function getDefaultSortFunction({ sortBy, isDescending }) {
+	const direction = isDescending ? -1 : 1;
+
+	return function defaultSort(a, b) {
+		const valueA = a && a[sortBy];
+		const valueB = b && b[sortBy];
+
+		const result =
+			isNaN(valueA) || isNaN(valueB)
+				? valueA.toString().localeCompare(valueB.toString())
```
smouillour: will have an error if valueA or valueB is undefined or null

packages/components/src/List/ListComposition/Manager/ListManager.component.js
jsomsanith-tlnd: No, they are not missing, there is no custom ones. This usage in ListManager is the uncontrolled one. In uncontrolled mode, we use the default.
The initial sort params are managed by SortBy component `initialValue` props. The purpose of the manager is not to manage all of that, but to init empty states, and propagate the value et setters to the children via context.

jsomsanith-tlnd:
@jlame 

> is this needed ? in fact it's working in uncontrolled mode in surge too
It's working with the default sort yes, but not for all use cases. Let's take a an example. In a project you have a list with a certification column. This column has 3 states : no certification, pending certification, certified. This 3-state value is not managed by the default sort, it's project specific, with business values. So you want to add a sort function for this column.
That's why you will have to switch to controlled mode.

This PR is to ease the usage, having the default sort, except for the ones you need to specify, but you don't want to reimplement the default sort for the other columns.
The exposed hook allows you in a few line of code to
- manage the react state
- insert your custom sort functions
- apply the sorting to the collection

jlame:
lgtm, after talking i realise i misunderstood the controlled mode with the controlled mode of the sort by component.

PR #2350, opened 2019-08-22
feat/date-picker-style

jillyan:
TCCUI-65 update the date picker margins and underline.
TCCUI-65 update the date picker margins and underline.


jillyan:
Thanks @jsomsanith-tlnd .
> You should style the picker after the split date/time as the markup will change.
I checked with @yyanwang , no cross files are impacted.
> Some dev guideline when you do this after the split :
> 
> * don't set the containers height in css. They will be set automatically depending on the margin/padding that are specified in the ux guideline.
I removed the footer height.
but for the table header, I am still checking. since some margin/padding doesn't work via table, I have to make tricky to set up the margin for the underline.
> * ux guidelines are supposed to use only the defined paddings values, so you should not set custom ones (ex: 0.5rem), but use the sass variables
I changed them and use the variables.

jillyan:
@jsomsanith-tlnd  
sorry, seems I mis-click the request button. I thought it is a refresh.

PR #2352, opened 2019-08-23
feat(TimePicker): Add TimePicker component

yyanwang:
Add TimePicker component
guidelines: 
http://guidelines.talend.com/document/92132#/ui-controls/date-time-picker-1561102408/time-picker


packages/components/src/DateTimePickers/pickers/TimePicker/TimePicker.scss
```diff
@@ -0,0 +1,26 @@
+.container {
+    padding: $padding-smaller 0;
+    background-color: $white;
+
+    button {
+        margin: 1px;
```
yyanwang: add margin 1px so there's space to show focus outline

packages/components/src/DateTimePickers/pickers/TimePicker/TimePicker.component.js
```diff
@@ -0,0 +1,143 @@
...
+						block: 'center',
+					});
+				}
+				if (found !== this.state.hightlightedItemIndex) {
+					this.updateHighlightIndex(found);
+				}
+			}
+		}
+	}
+	onMouseOver(event, index) {
```
jsomsanith-tlnd: This is not the expected behavior.
The active one is the one that fit the props value. It has a specific style. But on hover, it should be just css, to apply a grey background, which is different from the active background.
I think you can remove that and just add some css rules. Don't forget to include focus that should have the same style as hover

packages/components/src/DateTimePickers/pickers/TimePicker/TimePicker.component.js
```diff
@@ -0,0 +1,143 @@
...
+		});
+	}
+	updateHighlightIndex(index) {
+		this.setState(({ hightlightedItemIndex }) => {
+			if (hightlightedItemIndex !== index) {
+				return {
+					hightlightedItemIndex: index,
+				};
+			}
+			return {};
```
jsomsanith-tlnd:
```suggestion
			return null;
```
This will avoid a re-render

packages/components/src/DateTimePickers/pickers/TimePicker/TimePicker.component.js
```diff
@@ -0,0 +1,143 @@
...
+					this.updateHighlightIndex(found);
+				}
+			}
+		}
+	}
+	onMouseOver(event, index) {
+		this.updateHighlightIndex(index);
+	}
+	onSelect(event, time) {
+		event.stopPropagation();
```
jsomsanith-tlnd: Why do you stop propagation ?

packages/components/src/DateTimePickers/pickers/TimePicker/TimePicker.component.js
```diff
@@ -0,0 +1,143 @@
...
+		this.props.onSubmit(event, {
+			time: strToTime(time, this.props.useSeconds),
```
jsomsanith-tlnd: I think you should propagate the time and the textInput since you have it. You will need it to sync with input, avoiding another conversion

packages/components/src/DateTimePickers/pickers/TimePicker/TimePicker.component.test.js
```diff
@@ -0,0 +1,113 @@
...
+import { TimePicker, isBefore, addInterval, getOptions } from './TimePicker.component';
+
+describe('TimePicker component', () => {
+	it('should render', () => {
+		const wrapper = shallow(<TimePicker onSubmit={jest.fn()} />);
+
+		expect(wrapper.getElement()).toMatchSnapshot();
+	});
+	describe('functions', () => {
+		describe('function - isBefore', () => {
```
jsomsanith-tlnd: You should not expose this internal function. This will be tested with the component.
The reason: if we refacto an internal implementation, changing its internal api, your tests will fail even if it doesn't change your component

packages/components/src/DateTimePickers/pickers/TimePicker/TimePicker.component.test.js
```diff
@@ -0,0 +1,113 @@
...
+				expect(isBefore(time1, time2)).toBe(false);
+			});
+			it('should return false when time1 is hours after time2', () => {
+				const time1 = { hours: 11, minutes: 30, seconds: 1 };
+				const time2 = { hours: 10, minutes: 30, seconds: 1 };
+
+				expect(isBefore(time1, time2)).toBe(false);
+			});
+		});
+		describe('function - addInterval', () => {
```
jsomsanith-tlnd: Same here

packages/components/src/DateTimePickers/pickers/TimePicker/TimePicker.component.test.js
```diff
@@ -0,0 +1,113 @@
...
+				const base = { hours: 10, minutes: 20 };
+				expect(addInterval(base)).toEqual({ hours: 11, minutes: 20 });
+			});
+			it('should add minutes to time specified by interval', () => {
+				const base = { hours: 10, minutes: 20 };
+				const interval = 45;
+				expect(addInterval(base, interval)).toEqual({ hours: 11, minutes: 5 });
+			});
+		});
+		describe('function - getOptions', () => {
```
jsomsanith-tlnd: Same here

packages/components/src/DateTimePickers/pickers/TimePicker/TimePicker.component.test.js
```diff
@@ -0,0 +1,113 @@
...
+			// when
+			wrapper.find('button').at(3).simulate('click');
+			// then
+			expect(onSubmit).toBeCalledWith(event, { time: { hours: '03', minutes: '00', seconds: '00' } });
+		});
+		it('should hightlight item matches user input', () => {
+			// when
+			const wrapper = mount(<TimePicker onSubmit={jest.fn()} textInput="12:00" />);
+			// then
+			expect(wrapper.state().hightlightedItemIndex).toBe(12);
```
jsomsanith-tlnd: Avoid testing `state()`, inspect the element, checking if the right one has the classname

packages/components/src/DateTimePickers/pickers/TimePicker/TimePicker.component.test.js
```diff
@@ -0,0 +1,113 @@
...
+			window.HTMLElement.prototype.scrollIntoView = scrollIntoViewMock;
+			const wrapper = mount(<TimePicker onSubmit={onSubmit} />);
+
+			// when
+			wrapper.setProps({ textInput: '20' });
+			wrapper.update();
+
+			// then
+			expect(scrollIntoViewMock).toBeCalledWith({ behavior: 'smooth', block: 'center' });
+			expect(wrapper.state().hightlightedItemIndex).toBe(20);
```
jsomsanith-tlnd: Same here

packages/components/src/DateTimePickers/pickers/TimePicker/TimePicker.component.test.js
```diff
@@ -0,0 +1,113 @@
...
+			// then
+			expect(onSubmit).toBeCalledWith(event, { time: { hours: '03', minutes: '00', seconds: '00' } });
+		});
+		it('should hightlight item matches user input', () => {
+			// when
+			const wrapper = mount(<TimePicker onSubmit={jest.fn()} textInput="12:00" />);
+			// then
+			expect(wrapper.state().hightlightedItemIndex).toBe(12);
+		});
+		it('should sroll the first match into view when user inputs', () => {
```
jsomsanith-tlnd:
```suggestion
		it('should scroll the first match into view when user inputs', () => {
```

packages/components/src/DateTimePickers/pickers/TimePicker/TimePicker.scss
jsomsanith-tlnd: You can remove the outline as its replaced by the background. And remove the margin.

packages/components/src/DateTimePickers/pickers/TimePicker/TimePicker.component.test.js
yyanwang: so I just not export these functions and remove the test cases?

yyanwang:
thanks Jimmy, I will address the problems in another pr after splitting datetime picker into 2 fields.

PR #2353, opened 2019-08-23
fix(cmf): display loading screen during settings download

jmfrancois:

if a component is rendered before settings are loaded then it will never get its props mapped from it.

Double protection:
* add settings initialized status in the cache key (so when settings are loaded the key change.
* display loading screen while settings are not loaded


packages/cmf/src/bootstrap.js
```diff
@@ -150,9 +154,9 @@ export default function bootstrap(appOptions = {}) {
 	const store = bootstrapRedux(options, saga.middleware);
 	onError.bootstrap(options, store);
 	saga.run();
-	const RootComponent = options.RootComponent;
+	const RootComponent = options.RootComponent || DefaultRootComponent;
```
jmfrancois: I have added that default because of the too many error during test of this module.

PR #2357, opened 2019-08-26
feat(ListComposition): useLocalStorage

jsomsanith-tlnd:
_(This is part of a series of PR to improve api, and help to switch from List legacy component to List Composition)._
In the list, we had a thinking with @jmfrancois and UX team about 
- user preferred state: from user preference, that will reflect on the UI. Ex: default sort. We don't want to persist any change, so next time the user comes back, he will see the preferred sort and not the live changed one.
- live state: when user set a state, it should stay next time the user comes back. Ex: list display mode

We don't have a service to persist that for now, but to enforce the behavior, let's save the live state in local storage.

Create a localStorage hook, use it on display mode.
Expose a useDisplayMode hook for controlled mode.


packages/components/src/List/ListComposition/Manager/hooks/useLocalStorage.hook.js
```diff
@@ -0,0 +1,12 @@
+import { useState } from 'react';
+
+export default function useLocalStorage(localStorageKey, initialValue) {
+	const [value, setValue] = useState(localStorage[localStorageKey] || initialValue);
```
sgendre: localStorage is signifcantly slower than in memory and also DOM blocking
what about having an intermediate in memory object and async access to localstorage ?

packages/components/src/List/ListComposition/Manager/ListManager.component.js
```diff
@@ -41,6 +42,7 @@ Manager.defaultProps = {
 	t: getDefaultT(),
 };
 Manager.propTypes = {
+	id: PropTypes.string,
```
frassinier: If it's not mandatory, what will happen if it's not set?

packages/components/src/List/ListComposition/Manager/hooks/useLocalStorage.hook.js
```diff
@@ -0,0 +1,12 @@
...
+	const onChange = newValue => {
+		setValue(newValue);
```
sgendre: the intermediate in memory object maybe useful to check that newValue !== oldValue and be sure to  prevent writting it if not needed

packages/components/src/List/ListComposition/index.js
```diff
@@ -18,4 +19,7 @@ export default {
 	VList,
 };
 
-export { useCollectionSort };
+export const hooks = {
+	useCollectionSort,
+	useDisplayMode: useLocalStorage,
```
frassinier: What's the purpose to export it?

packages/components/src/List/ListComposition/Manager/hooks/useLocalStorage.hook.js
jsomsanith-tlnd: You need to temper that. It's not a problem, there is no perf issue at all. This is done at user action, so this is not very costly.

packages/components/src/List/ListComposition/Manager/hooks/useLocalStorage.hook.js
jsomsanith-tlnd: I changed the implementation to use react-use instead of implementing it, as @frassinier suggested

packages/components/src/List/ListComposition/index.js
jsomsanith-tlnd: To enforce the local storage behavior on display mode.
In project, we will use this hook if we want to control the display mode, adding automatically the local storage behavior

packages/components/src/List/ListComposition/Manager/ListManager.component.js
jsomsanith-tlnd: Good point, now it's required

PR #2360, opened 2019-08-27
feat(date picker): move year picker as a dropdown

jillyan:
TCCUI-65, move the year picker and make it as a dropdown in the year action.


packages/components/src/DateTimePickers/views/HeaderTitle/HeaderTitle.component.js
```diff
@@ -21,6 +21,28 @@ function HeaderTitle(props) {
...
+				return YearPicker;
+			}
+			throw new Error('Component not found');
+		},
+		components: {
+			itemsDropdown: [
+				{
+					id: 'year-picker-in-datetime-header',
+					component: 'YearPicker',
+					label: 'First item',
```
yyanwang: is this label necessary?

packages/components/src/Actions/ActionDropdown/ActionDropdown.component.js
```diff
@@ -43,7 +43,7 @@ function InjectDropdownMenuItem({
 	}
 	return (
 		<li role="presentation" key={key} {...liProps}>
-			<Inject component={component} getComponent={getComponent} {...rest} />
+			<Inject component={component} getComponent={getComponent} onSelect={onSelect} {...rest} />
```
yyanwang: `onSelect` is for menuItem, can you change the handler name as `onSelectItem` so it will be passed to YearPicker with `{...rest}`

packages/components/src/DateTimePickers/pickers/DateTimePicker/DateTimePicker.component.js
```diff
@@ -189,6 +189,7 @@ class DateTimePicker extends React.Component {
 				{viewElement}
 				<div className={theme.footer}>
 					<Action
+						id="date-picker-today-btn"
```
jmfrancois: this mean I can t have two datetimepicker in the same page ☝️ .
please use className instaed

packages/components/src/DateTimePickers/views/HeaderTitle/HeaderTitle.component.js
```diff
@@ -21,6 +21,27 @@ function HeaderTitle(props) {
...
+			itemsDropdown: [
+				{
+					id: 'year-picker-in-datetime-header',
+					component: 'YearPicker',
+					'data-feature': 'actiondropdown.items',
+					onSelect: props.onSelectYear,
+				},
+			],
+		},
+	};
```
jmfrancois: Oh nice use of the injection API but I guess with the API change happening in UI we should open children for that kind of business.
So I would add children in ActionDropdown.

packages/components/src/Actions/ActionDropdown/ActionDropdown.component.js
jillyan: hi, change to be onSelectItem will have a side effect which make the actionDropDown keep open even after select a year in the list.

packages/components/src/DateTimePickers/views/HeaderTitle/HeaderTitle.scss
```diff
@@ -6,6 +6,14 @@
 	align-items: center;
 	text-transform: uppercase;
 	font-weight: 600;
+
+	:global(.dropdown-menu) {
+		min-width: 63px;
```
jmfrancois: why 63 px ?

jillyan: there is no width defined for this dropdown list in guidline. and UX said should be the same to the ActionDropDown. And for now ActionDropDown is a little more than 62px. so I set it 63 here.

packages/components/src/DateTimePickers/views/HeaderTitle/HeaderTitle.component.js
jillyan: sorry, I don't understand.  🤣

packages/components/src/DateTimePickers/pickers/DateTimePicker/DateTimePicker.component.js
jillyan: moved.

packages/components/src/DateTimePickers/views/HeaderTitle/HeaderTitle.component.js
jillyan: Added , thanks for your patch 😄

jillyan:
> please remove the no option in the dropdown
> Screenshot 2019-08-28 at 15 10 40

Removed. Please take a look. thx. 😃

PR #2362, opened 2019-08-27
feat(TimePicker): Add InputTimePicker

yyanwang:
Add InputTimePicker which can be used standalone.

### DIFF
--- a/packages/components/src/DateTimePickers/DateTime/Manager/Manager.component.js
+++ b/packages/components/src/DateTimePickers/DateTime/Manager/Manager.component.js
@@ -50,7 +50,9 @@ class ContextualManager extends React.Component {
 	constructor(props) {
 		super(props);
 
-		checkSupportedDateFormat(props.dateFormat);
+		if (props.dateFormat) {
+			checkSupportedDateFormat(props.dateFormat);
+		}
 		this.initialState = extractParts(props.selectedDateTime, this.getDateOptions());
 		this.state = {
 			...this.initialState,
--- a/packages/components/src/DateTimePickers/DateTime/Manager/Manager.component.js
+++ b/packages/components/src/DateTimePickers/DateTime/Manager/Manager.component.js
@@ -50,9 +50,7 @@ class ContextualManager extends React.Component {
 	constructor(props) {
 		super(props);
 
-		if (props.dateFormat) {
-			checkSupportedDateFormat(props.dateFormat);
-		}
+		checkSupportedDateFormat(props.dateFormat);
 		this.initialState = extractParts(props.selectedDateTime, this.getDateOptions());
 		this.state = {
 			...this.initialState,
--- a/packages/components/src/DateTimePickers/hooks/useInputPickerHandlers.js
+++ b/packages/components/src/DateTimePickers/hooks/useInputPickerHandlers.js
@@ -1,0 +1,82 @@
+import { useState } from 'react';
+import keycode from 'keycode';
+
+
+export default function useInputPikcerHandlers(props) {
+	const [showPicker, setPickerVisibility] = useState(false);
+	const [picked, setPicked] = useState(false);
+
+	function openPicker() {
+		setPickerVisibility(true);
+	}
+	function closePicker() {
+		setPickerVisibility(false);
+	}
+	function onBlur(event) {
+		setPicked(false);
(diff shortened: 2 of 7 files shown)


packages/components/stories/DatePicker.js
```diff
@@ -285,4 +286,13 @@ storiesOf('DatePicker', module)
 				</div>
 			</div>
 		);
+	})
+	.add('Time picker - with input', () => {
```
jsomsanith-tlnd: A little linter error here
```
.add('Time picker - with input', () => (
	<form>
		<InputTimePicker onChange={action('onChange')} />
	</form>
));
```

packages/components/src/DateTimePickers/DateTime/Manager/Manager.component.js
```diff
@@ -50,7 +50,9 @@ class ContextualManager extends React.Component {
 	constructor(props) {
 		super(props);
 
-		checkSupportedDateFormat(props.dateFormat);
+		if (props.dateFormat) {
```
jsomsanith-tlnd: You can revert that, we always have a dateFormat, if not provided, we have a default one

packages/components/src/DateTimePickers/pickers/TimePicker/TimePicker.component.js
```diff
@@ -84,6 +84,7 @@ export class TimePicker extends React.Component {
 		this.props.onSubmit(event, {
 			textInput: option.label,
 			time: option.value,
+			origin: 'PICKER',
```
jsomsanith-tlnd: Can you set that in `InputTimePicker` please ?
(And it's a little out of this PR scope, but could you rename the onSubmit name to onChange ?)

packages/components/src/DateTimePickers/hooks/useInputPickerHandlers.js
```diff
@@ -0,0 +1,82 @@
+import { useState } from 'react';
+import keycode from 'keycode';
+
+
+export default function useInputPikcerHandlers(props) {
```
jsomsanith-tlnd: Typo: `useInputPickerHandlers`

jsomsanith-tlnd: Don't pass all the props, only what you need: `onBlur, onChange`
And doing so, you'll fix linter errors, because you defined propTypes for onBlur and onChange, but never use them

packages/components/src/DateTimePickers/hooks/useInputPickerHandlers.js
```diff
@@ -0,0 +1,82 @@
...
+	function onClick() {
+		openPicker();
+	}
+
+	function onFocus() {
+		if (!picked) {
+			openPicker();
+		}
+	}
+	function onKeyDown(event, inputRef, containerRef) {
```
jsomsanith-tlnd: containerRef is not used

packages/components/src/DateTimePickers/hooks/useInputPickerHandlers.js
```diff
@@ -0,0 +1,82 @@
...
+					// TODO: focuseOnTime();
+				} else {
+					openPicker();
+				}
+				break;
+			default:
+				break;
+		}
+	}
+	function getPopperPlacement(input) {
```
jsomsanith-tlnd: Not sure, but I think popper.js manages that already.
Can you remove that and check if the placement is ok ?

yyanwang: Still need this for InputDatePicker.
if removed, it will fall to default placement 'bottom', which align the picker to middle of input.
and we need 'bottom-start' or 'bottom-end' which align picker to left/right of input.
placement

jsomsanith-tlnd: In fact what I meant is to set the placement to what we want as default (here bottom-start) and see if popper manages the placement depending on the space

yyanwang: I just tried, and popper doesn't auto adjust placement base on space. If I set placement="bottom-start", it always align start of input. even for the Input on the right top corner in picture above.

PR #2363, opened 2019-08-28
fix(EditableText): disable submit on error

lmaillet:
Currently, the submit button of the `EditableText` is disabled only if it is not filled.
But it's not disabled if the component is not valid.

Disable the submit button if `EditableText` is not valid.


packages/components/src/EditableText/InlineForm.component.js
```diff
@@ -121,7 +121,7 @@ class InlineForm extends React.Component {
 							theme['tc-editable-text-form-buttons-submit'],
 							'tc-editable-text-form-buttons-submit',
 						)}
-						disabled={notFilled}
+						disabled={notValid}
```
mkondzera: :+1:

lmaillet:
Oops, that's right...

PR #2373, opened 2019-08-30
feat(theme): checkbox and radio

frassinier:
Checkbox and Radio styles are not up to date anymore

According to this [article](https://css-tricks.com/indeterminate-checkboxes),
Impl guidelines https://company-57688.frontify.com/document/92132#/ui-controls/checkbox


packages/theme/src/theme/_forms.checkbox.scss
```diff
@@ -47,48 +31,64 @@ input[type='checkbox'],
...
+			// border-color: $tt-checkbox-focus-border-color;
+			@include checkbox-outline;
 		}
 	}
 
+	// Indeterminate Checkboxes style
+	&[data-checked='1']:before,
+	&[data-checked='1'] + *:before {
+		background-color: $tt-checkbox-checked-background-color;
+		border-color: $tt-checkbox-checked-border-color;
```
jlame: is it normal that the border is different in indeterminate and checked case ? 
image

if yes, seems like the same variable is used for both whereas guidelines says it should be shade in one case and tint in another

packages/theme/src/theme/_forms.radio.scss
```diff
@@ -0,0 +1,74 @@
...
+
+	&:focus + span:after,
+	&:checked:after,
+	&:checked + *:after {
+		border-color: $tt-radio-checked-border-color;
+	}
+
+	&:focus:after,
+	&:focus + span:after {
+		@include checkbox-outline;
```
jlame: same variable here concerning the outline color which should be scooter in guideline (but doesn't look like it)

packages/theme/src/theme/_forms.checkbox.scss
mhuet: Hello @jlame — there is a mistake on the frontify page we should only use "Tint" instead of "darker" values. I'll update the guidelines ASAP. Thank you :)

PR #2374, opened 2019-09-02
feat(TimePicker): handle key pressing

yyanwang:
Some changes of Time Picker gesture:
- each element should NOT be focusable in the normal flow through tab key
- focus on the next/previous element on up/down key

### DIFF
--- a/packages/components/src/Gesture/withListGesture.js
+++ b/packages/components/src/Gesture/withListGesture.js
@@ -58,10 +58,12 @@ function onKeyDown(event, ref) {
 	switch (event.keyCode) {
 		case keycode.codes.down:
 			event.stopPropagation();
+			event.preventDefault();
 			focusOn(event, getNextItem(ref));
 			break;
 		case keycode.codes.up:
 			event.stopPropagation();
+			event.preventDefault();
 			focusOn(event, getPreviousItem(ref));
 			break;
 		default:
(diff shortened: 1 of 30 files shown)


packages/components/src/Gesture/withListGesture.js
```diff
@@ -58,10 +58,12 @@ function onKeyDown(event, ref) {
 	switch (event.keyCode) {
 		case keycode.codes.down:
 			event.stopPropagation();
+			event.preventDefault();
```
yyanwang: add `preventDefault()` to prevent time picker container from scrolling on every Up/Down pressing

packages/components/src/Gesture/withListGesture.js
```diff
@@ -71,9 +85,16 @@ function onKeyDown(event, ref) {
 
 export default function withListGesture(WrappedComponent) {
 	function ListGesture(props) {
-		return <WrappedComponent {...props} onKeyDown={onKeyDown} />;
+		const { loop } = props;
```
jsomsanith-tlnd: This must not be in a prop, as it must not change in time. A component would want he loop or not.
Take it in `withListGesture` arguments `withListGesture(WrappedComponent, loop)`

yyanwang: fixed as suggested.

PR #2375, opened 2019-09-03
fix(TMC-18149) wrong dropdown arrow

VolodymyrKovalM:
https://jira.talendforge.org/browse/TMC-18149

Split dropdown action is not possible to change now (without modifying react-bootstrap ), as there is no such possibility to pass custom icon renderer

As discussed in Jira comments, it's ok to change only dropdown action, since we don't plan to use Split dropdown a lot 

### DIFF
--- a/packages/components/src/Actions/ActionDropdown/ActionDropdown.scss
+++ b/packages/components/src/Actions/ActionDropdown/ActionDropdown.scss
@@ -9,6 +9,21 @@ $tc-dropdown-loader-padding: $padding-small !default;
 		}
 	}
 
+	&-button {
+		padding-right: 8px;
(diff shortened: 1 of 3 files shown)


packages/components/src/Actions/ActionDropdown/ActionDropdown.scss
```diff
@@ -1,4 +1,5 @@
 $tc-dropdown-loader-padding: $padding-small !default;
+$tc-dropdown-button-right-padding: 8px;
```
vbocquet:
```suggestion
$tc-dropdown-button-right-padding: 0.8rem;
```

PR #2384, opened 2019-09-05
fix(TMC-18455): data list positioning

VolodymyrKovalM:
https://jira.talendforge.org/browse/TMC-18455

After using popper, we have an issue with datalist positioning, if there are too many items that do not fit to screen, popper sets top position, although there are much more available space at the bottom (see on video attached to the jira)

The idea is to compute where is more available space and position items container accordingly, setting also max height so it won't go out from screen boundaries

### DIFF
--- a/packages/components/src/Typeahead/Typeahead.component.renderers.js
+++ b/packages/components/src/Typeahead/Typeahead.component.renderers.js
@@ -145,16 +145,17 @@ export function renderItemsContainerFactory(
 					computePosition: {
 						enabled: true,
 						fn: data => {
-							const OFFSET_FROM_SCREEN_BOUNDARIES = 15;
+							const GAP = 15; // the offset between the end of items container and screen boundaries
 							const inputDimensions = data.offsets.reference;
-							const offsetFromTop = inputDimensions.top - OFFSET_FROM_SCREEN_BOUNDARIES;
-							const offsetFromBottom = window.innerHeight - inputDimensions.top - inputDimensions.height - OFFSET_FROM_SCREEN_BOUNDARIES;
+							const { top, height } = inputDimensions;
+							const offsetTop = top - GAP;
+							const offsetBottom = window.innerHeight - top - height - GAP;
 							const placements = data.placement.split('-');
 							let newPlacement = data.placement;
-							if (placements[0] === 'top' && offsetFromBottom > offsetFromTop) {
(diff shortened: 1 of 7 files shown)


packages/components/src/Typeahead/Typeahead.component.renderers.js
```diff
@@ -150,6 +142,32 @@ export function renderItemsContainerFactory(
...
+								...data,
+								placement: newPlacement,
+								styles: {
+									...data.styles,
+									width: inputDimensions.width,
+									maxHeight,
+								},
+							};
+						},
+					},
```
jmfrancois: this is an almost pure function (depends on window) so please put it outside as a function

packages/components/src/Typeahead/Typeahead.component.renderers.js
```diff
@@ -150,6 +142,32 @@ export function renderItemsContainerFactory(
 					shift: {
 						enabled: false,
 					},
+					computePosition: {
+						enabled: true,
+						fn: data => {
+							const GAP = 15; // the offset between the end of items container and screen boundaries
+							const inputDimensions = data.offsets.reference;
```
VolodymyrKovalM: this gap still can be a point to discus, maybe we should make it bigger, or not put at all

packages/theme/src/theme/_variables.scss
```diff
@@ -292,6 +292,7 @@ $dropdown-header-color: $gray-light !default;
 $zindex-navbar: 1000 !default;
 $zindex-dropdown: 1000 !default;
 $zindex-navbar-fixed: 1030 !default;
+$zindex-typeahead-items: 1035 !default;
```
jmfrancois: you can move this one directly in typeahead.scss, this is not part of the theme.

packages/components/src/Typeahead/Typeahead.component.renderers.js
VolodymyrKovalM: done

packages/theme/src/theme/_variables.scss
VolodymyrKovalM: done, also deleted defining z-index in multiselect tag https://github.com/Talend/ui/pull/2384/files#diff-d44204ff36a8e26022441648ac114328L56 , so it will be the same as typeahead's zindex

PR #2389, opened 2019-09-09
chore(Component): Tweak the list display mode api

nicomaligne:
Make api simpler.
Removing local storage from the list manager
Moving the initialDisplayMode props from ListDisplayMode to ListManager.


packages/components/src/List/ListComposition/DisplayMode/ListDisplayMode.component.js
```diff
@@ -93,16 +86,18 @@ function ListDisplayMode({
 	);
 }
 
+
+export const displayModesOptions = ['table', 'large'];
```
smouillour: there is const for both in constants.js, please use them

PR #2410, opened 2019-09-13
fix(UIForm/Array): For single select, set default value as first enum item

vbocquet:
If a form schema have an array of select, we set default value as `''`. Problem, we have trouble with validation because we set a value that is not in enum.

Schema example:
```
{
   "jsonSchema":{
      "type":"object",
      "properties":{
         "Color":{
            "type":"array",
            "minItems": 1,
            "items":{
               "type":"string",
               "enum":[
                  "White",
                  "Lemon",
                  "Khaki"
               ]
            }
         }
      }
   },
   "uiSchema":[
      {
         "key":"Color",
         "type":"array",
         "items":[
            {
               "key":"Color[]",
               "type":"select"
            }
         ],
      }
   ],
   "properties":{
      "Color":[]
   }
}
```

When we click on Add element button, we set the first enum item as a value, if we have a single select.

### DIFF
--- a/packages/forms/src/UIForm/fieldsets/Array/Array.component.js
+++ b/packages/forms/src/UIForm/fieldsets/Array/Array.component.js
@@ -1,5 +1,6 @@
 import PropTypes from 'prop-types';
 import React from 'react';
+import { head, get } from 'lodash';
 import Widget from '../../Widget';
 import { shiftArrayErrorsKeys } from '../../utils/validation';
 import defaultTemplates from '../../utils/templates';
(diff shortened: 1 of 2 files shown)


vbocquet:
> I'm not sure about that. For optional selects, this will take the first enum value by default.

@jsomsanith-tlnd I understand your point of view, i have the same first.

The fact is, if you have optional select, you also send a empty string as a value. So this will not pass the validation, you will have a error message like `'' is not part as the enum`.

For me, if select is optional and you don't want a value, you don't click on `Add a element` button.

What do you think?

jsomsanith-tlnd:
Oh yeah you're right, it's in an array, so it's ok. My bad

vbocquet:
No problem, thx for review 👍

vbocquet:
@sgendre Thanks for notice that, you're agree, i did not confirm with UX.
So this morning, I had a meeting with Maxime from UX Designer to show him the problem and the result of this PR. 
He is agree with the solution, no need for him to do a guideline for that, especially because it's a bugfix.
Is it ok for you?

sgendre:
ok agree even if I would have like this default init value to be driven by the defintion

PR #2411, opened 2019-09-16
fix(components): fix errors on proptypes

jare-talend:

fix errors in the console when we use ResourcePicker

Fix the proptype defintion 


packages/components/src/RichTooltip/RichLayout/RichLayout.component.js
```diff
@@ -5,7 +5,7 @@ import Inject from '../../Inject';
 import theme from './RichLayout.scss';
 
 const TooltipPropTypes = {
-	id: PropTypes.string.isRequired,
+	id: PropTypes.string,
```
jsomsanith-tlnd: So you will have `undefined-content` as id sometimes `id={`${id}-content`}`

jare-talend: For Body yes but Header and Footer no
And using class undefined-content is dirty. Id should be uniq

packages/components/src/RichTooltip/RichLayout/RichLayout.component.js
```diff
@@ -60,4 +64,5 @@ RichLayout.propTypes = {
 	Header: Inject.getReactElement.propTypes,
 	Footer: Inject.getReactElement.propTypes,
 	text: PropTypes.string,
+	prefixId: PropTypes.string.isRequired,
```
frassinier: WDYT? to be consistent!
```suggestion
	id: PropTypes.string.isRequired,
```

jare-talend: I expect that if i name the variable "id", i do id={id} in the component. In my case, I use a prefix to create the id used by the component, so, why i name it prefixId. However, i don't disturb me to change to id if you prefer.

jsomsanith-tlnd: If it bothers you, you can implement the behavior you just described, adding the id on the main element wrapper, and still use this id to generate ids on sub parts

frassinier: +1 yes it sounds great

PR #2416, opened 2019-09-19
fix(forms/fieldset): markup with empty title or items

frassinier:
Fieldset title and its items can be empty.

Adapt, improvise, overcome its markup.

### DIFF
--- a/packages/forms/src/UIForm/fieldsets/Fieldset/Fieldset.component.js
+++ b/packages/forms/src/UIForm/fieldsets/Fieldset/Fieldset.component.js
@@ -5,16 +5,18 @@ import Widget from '../../Widget';
 
 export default function Fieldset(props) {
 	const { schema, ...restProps } = props;
-	const { title, items, options } = schema;
+	const { title, items = [], options } = schema;
 
-	return (
+	return items.length ? (
 		<fieldset className="form-group">
-			<legend className={classnames({ 'sr-only': options && options.hideTitle })}>{title}</legend>
+			{title && (
+				<legend className={classnames({ 'sr-only': options && options.hideTitle })}>{title}</legend>
+			)}
 			{items.map((itemSchema, index) => (
 				<Widget {...restProps} key={index} schema={itemSchema} />
 			))}
 		</fieldset>
-	);
+	) : null;
 }
 
 if (process.env.NODE_ENV !== 'production') {
--- a/packages/forms/src/UIForm/fieldsets/Fieldset/Fieldset.component.js
+++ b/packages/forms/src/UIForm/fieldsets/Fieldset/Fieldset.component.js
@@ -5,7 +5,7 @@ import Widget from '../../Widget';
 
 export default function Fieldset(props) {
 	const { schema, ...restProps } = props;
-	const { title, items = [], options } = schema;
+	const { title, items, options } = schema;
 
 	return items.length ? (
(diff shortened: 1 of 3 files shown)


packages/forms/src/UIForm/fieldsets/Fieldset/Fieldset.component.test.js
```diff
@@ -29,6 +29,44 @@ describe('Fieldset widget', () => {
 		expect(wrapper.getElement()).toMatchSnapshot();
 	});
 
+	it('should not render fieldset with empty items', () => {
```
vbocquet: Sure about this test? You have items in your schema here (so you have `fieldset` in the snapshot)

packages/forms/src/UIForm/fieldsets/Fieldset/Fieldset.component.test.js
```diff
@@ -29,6 +29,44 @@ describe('Fieldset widget', () => {
...
+		// when
+		const wrapper = shallow(<Fieldset schema={schema} />);
+
+		// then
+		expect(wrapper.getElement()).toMatchSnapshot();
+	});
+
+	it('should not render fieldset legend without any title', () => {
+		// given
+		const schema = {
```
vbocquet: Maybe this test need items, to focus of shallow a render with fieldset, but no title?

packages/forms/src/UIForm/fieldsets/Fieldset/Fieldset.component.js
```diff
@@ -5,16 +5,18 @@ import Widget from '../../Widget';
 
 export default function Fieldset(props) {
 	const { schema, ...restProps } = props;
-	const { title, items, options } = schema;
+	const { title, items = [], options } = schema;
 
-	return (
+	return items.length ? (
```
jmfrancois: This make schema.items optional which is not the case in proptypes.
I would like to know how you can have that use case ?

frassinier: Actually I don't know if we have this case but it can happen

packages/forms/src/UIForm/fieldsets/Fieldset/Fieldset.component.test.js
```diff
@@ -29,6 +29,44 @@ describe('Fieldset widget', () => {
...
+					schema: { type: 'string' },
+				},
+			],
+		};
+
+		// when
+		const wrapper = shallow(<Fieldset schema={schema} />);
+
+		// then
+		expect(wrapper.getElement()).toMatchSnapshot();
```
jmfrancois: please don t use snapshot that way.
you should find(legend) and assert there is no.

packages/forms/src/UIForm/fieldsets/Fieldset/Fieldset.component.test.js
```diff
@@ -29,6 +29,44 @@ describe('Fieldset widget', () => {
...
+		const schema = {
+			title: 'My fieldset',
+			items: [],
+		};
+
+		// when
+		const wrapper = shallow(<Fieldset schema={schema} />);
+
+		// then
+		expect(wrapper.getElement()).toMatchSnapshot();
```
jmfrancois: same here, just expect null

PR #2418, opened 2019-09-19
fix(cmf): trigger error on console when the value on the setting is null

jare-talend:

We have some errors on the console when the value on the settings is null

return empty string if the value is null

* [] Tests for the changes have been added (for bug fixes / features) And [non reg](./screenshots.md) done before need review

### DIFF
--- a/packages/cmf/__tests__/onError.test.js
+++ b/packages/cmf/__tests__/onError.test.js
@@ -42,14 +42,16 @@ describe('onError', () => {
 	});
 	describe('addAction', () => {
 		it('should add action in singleton', () => {
-			onError.addAction({ type: 'FOO', password: 'secret' });
+			onError.addAction({ type: 'FOO', password: 'secret', value: null, other: false, number: 0 });
 			const info = onError.getReportInfo(new Error('my'));
 			expect(info.actions.length).toBe(1);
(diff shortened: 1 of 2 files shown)


packages/cmf/src/onError.js
```diff
@@ -82,7 +82,12 @@ function anon(value, key) {
  * @return {Object} friendly with JSON.stringify
  */
 function prepareObject(originalState) {
-	const state = originalState.toJS ? originalState.toJS() : originalState;
+	const state = originalState && originalState.toJS ? originalState.toJS() : originalState;
+
+	if (state === undefined || state === null) {
+		return '';
```
jmfrancois: the function is supposed to return an object.

packages/cmf/__tests__/onError.test.js
```diff
@@ -42,14 +42,16 @@ describe('onError', () => {
...
 			expect(info.actions.length).toBe(1);
-			expect(info.actions[0]).toMatchObject({
+			expect(info.actions[0]).toEqual({
 				type: 'FOO',
 				password: expect.anything(),
+				value: null,
+				other: false,
+				number: 0,
 			});
-			expect(info.actions[0].password).not.toBe('secret');
```
jmfrancois: please add it this one back

PR #2419, opened 2019-09-19
feat(components/sidepanel): side panel updates

ncomont:
https://jira.talendforge.org/browse/TDP-7557


packages/components/stories/SidePanel.js
```diff
@@ -164,7 +192,7 @@ stories
 				]);
 				const panel = (
 					<SidePanel
-						actions={panelItems}
+						actions={_.shuffle([...items, ...other, ...other, ...other])}
```
frassinier: import shuffle instead of using lodash for some tree shaking reasons

packages/components/stories/SidePanel.js
```diff
@@ -107,43 +148,30 @@ stories
...
-			large
+			minimised
 			tooltipPlacement="top"
 		/>
 	))
-	.add('reverse', () => (
+	.add('with a large amount of items', () => (
 		<SidePanel
-			actions={items}
+			actions={_.shuffle([...items, ...other, ...other, ...other])}
```
frassinier: import shuffle from lodash instead

packages/components/src/SidePanel/SidePanel.scss
```diff
@@ -55,7 +55,13 @@ $toggle-button-padding: $padding-small;
 		padding-top: 0;
 		padding-bottom: 0;
 		padding-left: $toggle-button-padding;
+		background: $list-item-reverse-color;
 		text-align: left;
+		position: sticky;
+		top: 0;
+		left: 0;
+		right: 0;
+		z-index: 10;
```
frassinier: why do you need z-index?

packages/components/src/SidePanel/SidePanel.component.js
```diff
@@ -99,6 +109,7 @@ SidePanel.defaultProps = {
 	reverse: false,
 	large: false,
 	dockable: true,
+	minimised: false,
```
frassinier: why do you not use !large instead?

packages/components/stories/SidePanel.js
ncomont: I finally removed the use of shuffle.

packages/components/stories/SidePanel.js
ncomont: I finally removed the use of shuffle.

packages/components/src/SidePanel/SidePanel.scss
ncomont: image
because the toggle button is before the action list

packages/components/src/SidePanel/SidePanel.component.js
ncomont: image
large mode does not have the same effet than the new minimised mode

yarn.lock
```diff
@@ -14156,6 +14164,18 @@ simplebar@^4.0.0:
...
+simplebar@^4.2.2:
+  version "4.2.2"
+  resolved "https://registry.yarnpkg.com/simplebar/-/simplebar-4.2.2.tgz#3d8144091274f0daa483179cd54708818f39b2ed"
+  integrity sha512-OWHcnifIbGhB7RBdYS57xsqOUIp9GOKCJw5dOJCuKsV83+1kdKcbIaXaqb6A9zAOIRqhJIdN2qOBrsd4Axe84A==
+  dependencies:
+    can-use-dom "^0.1.0"
+    core-js "^3.0.1"
+    lodash.debounce "^4.0.8"
+    lodash.memoize "^4.1.2"
+    lodash.throttle "^4.1.1"
```
jmfrancois: those lodash.* deps are out of date and should not be used

jmfrancois: image

packages/components/src/SidePanel/SidePanel.component.js
frassinier: ok so actually `minimised` is the same as the combination of `docked={true}` and `dockable={false}`

packages/components/src/SidePanel/SidePanel.scss
frassinier: Oh no ok, I saw that it's because `<ul>` has an absolute position with simplebar :+1:

packages/components/src/SidePanel/SidePanel.component.js
frassinier: Maybe you can just remove this props?

ncomont: Which props would you like to remove ?

frassinier: It's ok as it, it will be more understandable to apply guidelines

yarn.lock
frassinier: The version resolved of lodash is [4.17.11](https://github.com/Talend/ui/blob/e08f84462b2899c995fb1d49a712d26a4d3e1619/yarn.lock#L9889) so we don't care?
No related dependencies are new here actually so I don't see any problems?

romainseb:
Testing it, sometimes, the lib seems to stuck while scrolling.
Could we use native scrollbar instead of a new lib for that ? (also for the bundle size)

Not using this exact solution but with the idea :
https://stackoverflow.com/questions/43737303/only-show-scrollbar-when-page-scrolls-in-css?rq=1

smouillour:
seems there is a braking change (visible in story layout - two column)
image

yarn.lock
jmfrancois: lodash.* package do not follow 4.17. They stop at 4.4 so you have lodash twice.

The fact they are not new comes from different reason: they are used in dev deps (so not a bundle pb) + some other components like `reactour ` in fact.

So I just alert on the fact every project will take 50k for a scrollbar.

ncomont:
@romainseb the goal is to show a scrollbar that looks like to OSX ones on every browsers / OS.
@smouillour I think that there is no better way to implement the specifications, butI will ask UI teams.

packages/components/src/ActionList/ActionList.scss
```diff
@@ -19,7 +19,7 @@ $tc-action-list-item-border-size: 2px !default;
 			.btn.btn-link {
 				padding: $padding-normal;
 				text-overflow: inherit;
-				text-transform: uppercase;
+				text-transform: inherit;
```
smouillour: i had a discussion with Stéphanie B. about this change and i'm not sure we need to do it now. (need before to fix existing translation in our projects). please see with her before merge your PR

ncomont: As discussed with Stephanie, everything is OK. Screenshots will be made on the INT stack.

output/components.sasslint.txt
```diff
@@ -23,8 +23,11 @@ src/List/Toolbar/Toolbar.scss
   80:6   error  Nesting depth 4 greater than max of 3                    nesting-depth
   93:12  error  Qualifying elements are not allowed for class selectors  no-qualifying-elements
 
+src/SidePanel/SidePanel.scss
+  45:6  error  Nesting depth 4 greater than max of 3  nesting-depth
```
smouillour: this is possible to remove this error? please

PR #2422, opened 2019-09-20
feat(Time Picker): timezone info

yyanwang:
Add timezone info icon according to guildines:
http://guidelines.talend.com/document/92132#/ui-controls/date-time-picker-1561102408/time-picker

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -39,6 +39,7 @@
     "classnames": "^2.2.5",
     "d3-shape": "1.2.0",
     "date-fns": "^1.27.2",
+    "date-fns-timezone": "^0.1.4",
     "focus-outline-manager": "^1.0.2",
     "immutable": "^3.8.1",
     "invariant": "^2.2.2",
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -63,6 +63,7 @@
     "reactour": "^1.13.4",
     "simulant": "^0.2.2",
     "styled-components": "^4.1.3",
+    "timezone-support": "^2.0.2",
     "uuid": "^3.0.1"
   },
   "devDependencies": {
--- a/packages/components/src/DateTimePickers/Date/Input/Input.component.js
+++ b/packages/components/src/DateTimePickers/Date/Input/Input.component.js
@@ -8,7 +8,10 @@ export default function Input(props) {
 	const { value, inputManagement } = useContext(DateContext);
 
 	return (
-		<InputSizer placeholder={inputManagement.placeholder} inputText={value.textInput}>
+		<InputSizer
+			placeholder={inputManagement.placeholder}
+			inputText={value.date ? value.textInput : ''}
+		>
 			{width => (
 				<DebounceInput
 					autoComplete="off"
--- a/packages/components/src/DateTimePickers/Date/Manager/Manager.component.js
+++ b/packages/components/src/DateTimePickers/Date/Manager/Manager.component.js
@@ -41,8 +41,8 @@ function ContextualManager(props) {
 		if (!props.onChange) {
 			return;
 		}
-		const { errorMessage, date, textInput, errors } = nextState;
-		props.onChange(event, { errors, errorMessage, date, textInput, origin });
+		const { errorMessage, date, localDate, textInput, errors } = nextState;
+		props.onChange(event, { errors, errorMessage, date, localDate, textInput, origin });
 	}
 
 	function onInputChange(event) {
--- a/packages/components/src/DateTimePickers/Date/Manager/Manager.component.js
+++ b/packages/components/src/DateTimePickers/Date/Manager/Manager.component.js
@@ -5,7 +5,7 @@ import { DateContext } from '../Context';
 import {
 	checkSupportedDateFormat,
 	extractDate,
-	extractDateFromDate,
+	extractFromDate,
 	extractDateFromTextInput,
 } from '../date-extraction';
 
@@ -47,7 +47,7 @@ function ContextualManager(props) {
 	}
 
 	function onPickerChange(event, { date }) {
-		const nextState = extractDateFromDate(date, getDateOptions());
+		const nextState = extractFromDate(date, getDateOptions());
 		setState(nextState);
 		onChange(event, 'PICKER', nextState);
 	}
--- a/packages/components/src/DateTimePickers/Date/Manager/Manager.component.js
+++ b/packages/components/src/DateTimePickers/Date/Manager/Manager.component.js
@@ -6,7 +6,6 @@ import {
 	checkSupportedDateFormat,
 	extractDate,
 	extractDateFromDate,
-	extractPartsFromDate,
 	extractDateFromTextInput,
 } from '../date-extraction';
 
@@ -36,8 +35,8 @@ function ContextualManager(props) {
 		if (!props.onChange) {
 			return;
 		}
-		const { errorMessage, dateWithTimezone, textInput, errors } = nextState;
-		props.onChange(event, { errors, errorMessage, date: dateWithTimezone, textInput, origin });
+		const { errorMessage, date, textInput, errors } = nextState;
+		props.onChange(event, { errors, errorMessage, date, textInput, origin });
 	}
 
 	function onInputChange(event) {
@@ -58,7 +57,7 @@ function ContextualManager(props) {
 			value={{
 				value: {
 					textInput: state.textInput,
-					date: state.date,
+					date: state.localDate,
 				},
 
 				inputManagement: {
--- a/packages/components/src/DateTimePickers/Date/Manager/Manager.component.js
+++ b/packages/components/src/DateTimePickers/Date/Manager/Manager.component.js
@@ -4,6 +4,7 @@ import PropTypes from 'prop-types';
 import { DateContext } from '../Context';
 import {
 	checkSupportedDateFormat,
+	checkSupportedTimezone,
 	extractDate,
 	extractFromDate,
 	extractDateFromTextInput,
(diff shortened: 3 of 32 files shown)


packages/components/src/DateTimePickers/Date/Input/Input.component.js
```diff
@@ -8,7 +8,10 @@ export default function Input(props) {
 	const { value, inputManagement } = useContext(DateContext);
 
 	return (
-		<InputSizer placeholder={inputManagement.placeholder} inputText={value.textInput}>
+		<InputSizer
+			placeholder={inputManagement.placeholder}
+			inputText={value.date ? value.textInput : ''}
```
yyanwang: when input of date is valid, then take the width of input. if date is not valid, take width of placeholder.

packages/components/src/DateTimePickers/InputDateTimePicker/__snapshots__/InputDateTimePicker.snapshot.test.js.snap
```diff
@@ -186,19 +186,23 @@ exports[`InputDateTimePicker should render 1`] = `
...
+                      style="width: 47px;"
+                      type="text"
+                      value="15:27"
+                    />
+                  </div>
                   <span
                     style="font-weight: 400; font-size: 1.6rem; visibility: hidden;"
                   >
-                    04:08:29
+                    15:27
```
jsomsanith-tlnd: In the test, you pass useSecond, why isn't there the seconds ?

packages/components/src/DateTimePickers/Time/Input/Input.scss
```diff
@@ -0,0 +1,14 @@
...
+	align-items: flex-end;
+
+	:global(.form-control) {
+		margin-right: $padding-small;
+	}
+
+	svg {
+		width: 1.4rem;
+		height: 1.4rem;
+		margin-bottom: 0.2rem;
```
jsomsanith-tlnd: The icon is not well aligned for every font size (user can change the font size with accessibility features.
What you need to do: 
- .time-picker-input : align items to `center`
- svg : remove the margin
- on the tooltip div wrapper: display block

packages/components/package.json
```diff
@@ -62,6 +63,7 @@
     "reactour": "^1.13.4",
     "simulant": "^0.2.2",
     "styled-components": "^4.1.3",
+    "timezone-support": "^2.0.2",
```
yyanwang: This give us list of supported timezones, so we can check if specified timezone is valid.

packages/components/src/DateTimePickers/Date/Manager/Manager.component.js
```diff
@@ -56,7 +63,7 @@ function ContextualManager(props) {
 			value={{
 				value: {
 					textInput: state.textInput,
-					date: state.date,
+					date: state.localDate,
```
yyanwang: `state.localDate` stores date in local timezone,
`state.date` stores date in target timezone.

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -31,14 +39,20 @@ function InputDateTimePicker(props) {
 							id={`${props.id}-date-picker`}
 							onChange={onDateChange}
 							dateFormat={props.dateFormat}
+							useUTC={props.useUTC}
+							timezone={props.timezone}
 							value={date}
+							hideTimezone
```
jsomsanith-tlnd: I don't see anything about timezone indication in date picker, it seems to be only in time picker.

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -31,14 +39,20 @@ function InputDateTimePicker(props) {
...
 						/>
 						<InputTimePicker
 							id={`${props.id}-time-picker`}
 							onChange={onTimeChange}
 							useSeconds={props.useSeconds}
 							value={time}
 						/>
+						{props.timezone && (
+							<TimeZone timezone={props.timezone} />
+						)}
```
jsomsanith-tlnd: Related to my previous comment: shouldn't it be rendered on InputTimePicker then ?
In the current situation, we have the timezone info only in DateTime picker, but if we want to use a Time picker alone, it should have the info.
I suggest to move it.

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -16,12 +17,19 @@ function InputDateTimePicker(props) {
 			'Warning: "selectedDateTime" is deprecated and will be removed in the next major version. Use "value" instead please.',
 		);
 	}
+	if (props.useUTC) {
+		// eslint-disable-next-line no-console
+		console.warn(
+			'Warning: "useUTC" is deprecated and will be removed in the next major version. Set "timezone" instead please.',
+		);
```
jsomsanith-tlnd: I didn't know that we have to deprecate the utc mode. Is it a request from UX ?

packages/components/src/DateTimePickers/InputDatePicker/InputDatePicker.component.js
```diff
@@ -60,13 +65,14 @@ export default function InputDatePicker(props) {
 		),
 	].filter(Boolean);
 	return (
-		<div className="date-picker">
+		<div className={classNames(theme['date-picker'], 'date-picker')}>
```
jsomsanith-tlnd: To validate with UX, but if we never have the timezone info on date, you can remove this extra class

packages/components/src/DateTimePickers/DateTime/datetime-extraction.js
```diff
@@ -12,40 +15,28 @@ export function DatePickerException(code, message) {
...
 	if (useUTC) {
 		hours = date.getUTCHours();
 		minutes = date.getUTCMinutes();
 		seconds = date.getUTCSeconds();
-	} else {
-		hours = date.getHours();
-		minutes = date.getMinutes();
-		seconds = date.getSeconds();
+	}
+	if (timezone) {
```
jsomsanith-tlnd:
```suggestion
	else if (timezone) {
```

packages/components/src/DateTimePickers/DateTime/datetime-extraction.js
```diff
@@ -88,19 +62,21 @@ function timeToSeconds(hours, minutes, seconds) {
...
 
 	try {
+		if (typeof time === 'string') {
+			// eslint-disable-next-line no-param-reassign
+			time = strToTime(time);
+		}
 		const { hours, minutes, seconds } = time;
 		const timeInSeconds = timeToSeconds(hours, minutes, seconds);
-		const localTimezoneDate = setSeconds(date, timeInSeconds);
-		return useUTC ? convertToUTC(localTimezoneDate) : localTimezoneDate;
```
jsomsanith-tlnd: Why have you removed the utc thing ? 
Even if you deprecate it, you must maintain the behavior, or it's a breaking change. Then when we want to remove it in a major version, we can remove the behavior

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
yyanwang: No, it's not request from UX, just I supposed it. because `useUTC` is equivalent with `timezone='UTC'`.

packages/components/src/DateTimePickers/Date/date-extraction.js
```diff
@@ -128,7 +161,7 @@ function checkSupportedDateFormat(dateFormat) {
  *		textInput: string
  * 	}}
  */
-function extractPartsFromDate(date, options) {
+function extractDateFromDate(date, options) {
```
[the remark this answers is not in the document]
yyanwang: yes, I wanted to align the naming with `extractDateFromTextInput`.
methods renamed.

packages/components/src/DateTimePickers/DateTime/datetime-extraction.js
yyanwang: I thought date is already converted to utc when pass from date picker to datetime picker, so I just add seconds base on date.
But you're right, better to handle utc/timezone in datetime manager, since it may result in different date for local and target timezone.

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
jsomsanith-tlnd: Oh ok I didn't know that there was an `UTC` value in timezone.
Then, we can deprecate it, but in the pickers, if we have the useUTC props, just set the timezone to `UTC`, and then you can remove all the `if(useUTC)` code.
With that we still manage the useUTC props, adding the warning to pass `timezone="UTC"` instead, but only have the timezone code.

rmannibucau:
Hi guys,
what's the rational to deprecate useUTC - i'm relying on it and like the fact it is a typed value (boolean)?
timezone=UTC is less nice cause it is a string and not an enum so errors are way easier to do.

packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
yyanwang: @jsomsanith Oh sorry, actually `Etc/UTC` should be the timezone value to use. Then should we deprecate it? (also take @rmannibucau 's comment into account: https://github.com/Talend/ui/pull/2422#issuecomment-539973061)

yyanwang: I will merge this one to feature branch for now so I can continue work. I will author another PR to address useUTC and timezone issue.

PR #2425, opened 2019-09-23
feat(TPSVC-11298): tabs update

mkondzera:
Update TabBar styling (exluding responsive) https://jira.talendforge.org/browse/TPSVC-11298 provided by following link https://company-57688.frontify.com/document/92132#/navigation-layout/tabs-1560514405
Responsive tab bar should be changed after dropdown update in scope of this item https://jira.talendforge.org/browse/TPSVC-11545
Updating tab bar styling and adding icons for tabs.


packages/components/src/TabBar/TabBar.scss
```diff
@@ -10,13 +10,62 @@
...
+			height: $svg-sm-size;
+			font-size: $svg-sm-size;
+		}
+
+		button {
+			padding: $padding-smaller $padding-small;
+			padding-bottom: calc(#{$padding-smaller} + #{$underline-height});
+			color: $black;
+
+			max-width: 200px;
```
vbocquet: no rem?

packages/components/src/TabBar/TabBar.scss
```diff
@@ -10,13 +10,62 @@
 	background-image: url('data:image/svg+xml;utf8,<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 16 16"><path d="M8 13.6l-8-8L2.6 3 8 8.4 13.4 3 16 5.6z" fill="#{hex-to-url($color)}"/></svg>');
 }
 
+$underline-height: 2px;
```
vbocquet: rem?

packages/components/src/TabBar/TabBar.scss
```diff
@@ -10,13 +10,62 @@
...
+			text-overflow: ellipsis;
+
+			&:hover {
+				color: $lochmara;
+			}
+
+			&[aria-selected='true'] {
+				position: relative;
+
+				font-size: 14px;
```
vbocquet: `$font-size-base` ?

packages/components/src/TabBar/TabBar.scss
```diff
@@ -10,13 +10,62 @@
...
 		display: inline-block;
+
+		&-icon {
+			margin-right: $padding-smaller;
+			width: $svg-sm-size;
+			height: $svg-sm-size;
+			font-size: $svg-sm-size;
+		}
+
+		button {
```
vbocquet: Seems that need to be in `text-transform: capitalize;` in frontify no? There is a style that need to be override.

packages/components/src/TabBar/TabBar.scss
frassinier: And I would suggest using a better const name like
```suggestion
$tc-tabbar-active-border-size: .2rem !default;
```

packages/components/src/TabBar/TabBar.scss
frassinier: Can you try not using element-based selectors, but using class instead

packages/components/src/TabBar/TabBar.scss
```diff
@@ -10,13 +10,62 @@
...
+			padding: $padding-smaller $padding-small;
+			padding-bottom: calc(#{$padding-smaller} + #{$underline-height});
+			color: $black;
+
+			max-width: 200px;
+			white-space: nowrap;
+			overflow: hidden;
+			text-overflow: ellipsis;
+
+			&:hover {
```
frassinier: and `&:active`, no?

packages/components/src/TabBar/TabBar.scss
```diff
@@ -10,13 +10,62 @@
...
+
+				&:after {
+					position: absolute;
+					content: '';
+
+					left: $padding-small;
+					right: $padding-small;
+					bottom: 0;
+
+					border-bottom: solid $underline-height $lochmara;
```
frassinier: `$lochmara` is used 3 times, use variable instead like `$tc-tabbar-active-color`

lmaillet:
Don't know if it's part of your PR, but long tab headers are not truncated when used in a dropdown.
See http://2425.talend.surge.sh/components/?path=/story/tabbar--default
image

packages/components/src/TabBar/TabBar.scss
mkondzera: This one is extra, 14px is already inherited, so I removed it.

packages/components/src/TabBar/TabBar.scss
mkondzera: But $lochmara is already a global variable, why we need to reasign it to another local one?

packages/components/src/TabBar/TabBar.scss
mkondzera: It's not clear enought for me why using rem right over here for width size, such as rem depends on font-size. And in fontify is max-width provided with exact value 200px. In this case if font-size will change, max-width size will also change.

packages/components/src/TabBar/TabBar.scss
mkondzera: active is already has this color. It's defined on line 49.
By the way it won't work here, such as button doesn't have active class. But it will work with &[aria-selected='true']

packages/components/src/TabBar/TabBar.scss
mkondzera: It's better to use text-transform none here, for example truncation section from fontify has lowercase words
image

mkondzera: Element is generated by NavItem bottstrap component. I tried to apply class for this component, but didn't find the way ho to do it https://react-bootstrap.github.io/components/navs/#nav-item-props

mkondzera: Element is generated by NavItem bottstrap component. I tried to apply class for this component, but didn't find the way ho to do it https://react-bootstrap.github.io/components/navs/#nav-item-props. className is applied to li element, not to button. But in my case I need button to select

mkondzera:
> Don't know if it's part of your PR, but long tab headers are not truncated when used in a dropdown.
> See http://2425.talend.surge.sh/components/?path=/story/tabbar--default
> image

This issue would be solved in scope of this item https://jira.talendforge.org/browse/TPSVC-11545

packages/components/src/TabBar/TabBar.scss
frassinier: Actually `$lockmara` is a generated name for a specific color, it does not mean that it designed for active color

packages/components/src/TabBar/TabBar.scss
frassinier: You have to check [the doc for bootstrap 3](https://5c507d49471426000887a6a7--react-bootstrap.netlify.com/components/tabs/#tabs-props-pane) instead

packages/components/src/TabBar/TabBar.scss
frassinier: +1 with @vbocquet 
Guidelines do not respect UX recommendations for A11Y issues imho.
By default using rem unit will not block you UI when you will zoom in. 
Mixing rem and px is not a good idea at all, 1rem is designed to be an equivalent of 10px so i's easy to not block zoomed viewport.

packages/components/src/ActionIntercom/Intercom.component.js
```diff
@@ -39,7 +39,7 @@ function Intercom({ id, className, config, t, tooltipPlacement }) {
 		: t('TC_INTERCOM_OPEN', { defaultValue: 'Chat with Talend Support' });
 
 	return (
-		<TooltipTrigger label={label} tooltipPlacement={tooltipPlacement}>
+		<TooltipTrigger label={label} tooltipPlacement={'top-right'}>
```
frassinier: why do you need to change it in your PR?

packages/components/src/TabBar/TabBar.scss
mkondzera: I checked it, but didn't find styling prop for NavItem child component. For now button style is already applied (tc-tab-bar-item) but not for its button child component
image

frassinier: Did you just try passing `className` ? :D

packages/theme/src/theme/_navs.scss
```diff
@@ -92,41 +95,52 @@
...
 			}
 
 			&.active {
 				> a,
 				> button {
-					box-shadow: 0 -2px 0 $brand-primary inset;
-					border: none;
-					color: $brand-primary;
+					color: $lochmara;
+					font-weight: semi-bold;
```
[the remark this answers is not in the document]
mkondzera: forgot $ sign while using variable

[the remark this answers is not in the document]
mkondzera: Fount that this variable should be used $font-weight-semi-bold

PR #2431, opened 2019-09-24
feat(TCCUI65/Calendar) add underline in month picker

jillyan:
http://guidelines.talend.com/document/92132#/ui-controls/date-time-picker-1561102408
Add underline in month picker, also refactor code.
underline


output/components.eslint.txt
```diff
@@ -19,10 +19,6 @@ The react/require-extension rule is deprecated. Please use the import/extensions
   33:11  error  Caution: `i18next` also has a named export `t`. Check if you meant to write `import {t} from 'i18next'` instead  import/no-named-as-default-member
   35:11  error  Caution: `i18next` also has a named export `t`. Check if you meant to write `import {t} from 'i18next'` instead  import/no-named-as-default-member
 
-/home/travis/build/Talend/ui/packages/components/src/DateTimePickers/generator.js
-  3:8  error  'differenceInCalendarWeeks' is defined but never used  no-unused-vars
-  5:8  error  'endOfMonth' is defined but never used                 no-unused-vars
-
```
jillyan: these 2 methods have been deleted already, but this components.eslint.txt is not updated.

packages/components/src/DateTimePickers/pickers/DateTimePicker/DateTimePicker.component.js
```diff
@@ -187,7 +188,11 @@ class DateTimePicker extends React.Component {
 				aria-label="Date picker"
 			>
 				{viewElement}
-				<div className={theme.footer}>
+				<div
+					className={classNames(theme.footer, {
+						[theme['date-padding']]: this.state.isDateTimeView,
```
jsomsanith-tlnd: Time picker within dateTimePicker will be deprecated soon. I don't think we should bother with that.
I suggest you to set the line within date picker sass file only to set the line above and under the calendar, so it's in the same file, managed the same way, and it will survive the current date picker refacto where we split date and time.
It's not a problem that the time part doesn't have the line, it will be removed.

jillyan: if  i set only in date picker sass, then there will be no underline in the month picker.
For now the DateTimePicker has a `switcher` between date picker and month picker, so it is best place for the underline.
And after the DateTimePicker deprecated, it is easy to move the footer(with the Today button) and the underline to the new `switcher`. I will work with Wang Yan that time for it.

jsomsanith-tlnd: Ok let's reset: what bothers me the most is that `.date-padding`that add a padding top in one case and not in the other.
Can you move that to have this space managed by the date and the month part to have a margin-bottom specifically, instead of managing it at this level ?
Then you can remove that classname

jillyan: Not only you, the underline bothered me for long time since html table doesn't support margin things.
I moved the margin for the underline several times already. From viewlayout to footer at last.
For now the default margin-top is $padding-normal witch is designed for month picker, and the `.date-padding` will override it in date picker.

PR #2443, opened 2019-09-27
fix(component): ListComposition -error with filter hook when a value is null

smouillour:
There is an error when a value in the list is null (not undefined)
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters
Manage the both case null and undefined with isNil

### DIFF
--- a/packages/components/src/List/ListComposition/Manager/hooks/useCollectionFilter.hook.js
+++ b/packages/components/src/List/ListComposition/Manager/hooks/useCollectionFilter.hook.js
@@ -1,7 +1,9 @@
 import { useMemo, useState } from 'react';
+import isNil from 'lodash/isNil';
 
-function defaultFilterFunction(value = '', textFilter) {
-	return value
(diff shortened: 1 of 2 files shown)


packages/components/src/List/ListComposition/Manager/hooks/useCollectionFilter.hook.js
```diff
@@ -1,7 +1,9 @@
 import { useMemo, useState } from 'react';
+import isNil from 'lodash/isNil';
 
-function defaultFilterFunction(value = '', textFilter) {
-	return value
+function defaultFilterFunction(value, textFilter) {
+	const filteredValue = isNil(value) ? '' : value;
```
ncomont: Could not it be replaced by `value == null` ?

smouillour: isNill is used in the hook for the sort also, so i did the same ¯\_(ツ)_/¯

PR #2449, opened 2019-10-03
feat(components/CollapsiblePanel): add skeleton instead of status text

vyshkov:

In order to finish Task Execution live updates feature in TMC (updates come from server side),
we need a way to set execution status with a skeleton, as we don't know actual status at the moment.

Please take a look to mock-up:
https://jira.talendforge.org/secure/attachment/203492/image.png

I added 'skeleton' status to header props.

### DIFF
--- a/packages/components/src/Status/Status.component.js
+++ b/packages/components/src/Status/Status.component.js
@@ -75,9 +75,9 @@ function renderIcon(status, icon, progress) {
 					size={Skeleton.SIZES.small}
 				/>
 			);
+		default:
+			return icon && <Icon name={icon} />;
 	}
-
-	return icon && <Icon name={icon} />;
 }
 
 function renderLabel(status, label) {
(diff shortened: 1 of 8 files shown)


packages/components/src/Status/Status.component.js
```diff
@@ -50,19 +52,46 @@ export function getbsStyleFromStatus(status) {
 			return 'muted';
 		case STATUS.WARNING:
 			return 'warning';
+		case STATUS.SKELETON:
+			return 'skeleton';
```
jmfrancois: this is strange to declare constant to the write strings ... but consistency FTW :D

packages/components/src/Status/Status.scss
```diff
@@ -58,3 +62,8 @@ $tc-status-space-btn-icon-label: 5px !default;
 		}
 	}
 }
+
+.tc-status-skeleton-item {
+	margin-right: 10px;
```
jmfrancois: please use $padding variable

packages/components/src/Status/Status.snapshot.test.js
```diff
@@ -64,6 +64,10 @@ const inProgressStatusWithPercent = {
 	],
 };
 
+const skeletonStatusProps = {
+	status: 'skeleton',
```
jmfrancois: please use the constant here

packages/components/src/Status/Status.component.js
vyshkov: :D

PR #2452, opened 2019-10-07
fix(TPSVC-11545): add missing icon for selected element

mkondzera:
https://jira.talendforge.org/browse/TPSVC-11545
Icon for selected element is fixed
Add icon for selected element
image

### DIFF
--- a/packages/components/src/TabBar/TabBar.component.js
+++ b/packages/components/src/TabBar/TabBar.component.js
@@ -117,11 +117,13 @@ function TabBar(props) {
 		</Tab.Content>
 	);
 	if (responsive && showDropdown) {
+		const selectedItem = items[selectedKey - 1];
 		return (
 			<React.Fragment>
 				<ActionDropdown
 					className={classnames(theme['tc-tab-bar-dropdown'], 'tc-tab-bar-dropdown')}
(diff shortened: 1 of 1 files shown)


packages/components/src/TabBar/TabBar.component.js
```diff
@@ -117,11 +117,13 @@ function TabBar(props) {
...
 	);
 	if (responsive && showDropdown) {
+		const selectedItem = items[selectedKey - 1];
 		return (
 			<React.Fragment>
 				<ActionDropdown
 					className={classnames(theme['tc-tab-bar-dropdown'], 'tc-tab-bar-dropdown')}
-					label={items[selectedKey - 1].label}
+					label={selectedItem.label}
+					icon={selectedItem.icon && selectedItem.icon.name}
```
vyshkov: pls add "icon" to items prop-types shape. Test also would be nice to have.

mkondzera: I added missed prop-type.
About test, I can't catch whith responsive behaviour even after using mount and adding wrapper with fixed width. It seems because of logic about calculation which depends on window and dom tree.
image

PR #2456, opened 2019-10-08
chore(eslint-config) update eslint api config to remove globals for lodash and angular

Kporal:

In the API Designer we reworked the way we imported our deps and we don't want to have angular and lodash as global from eslint anymore.

Remove lodash and angular as globals on api eslint config file.

* [] The PR commit message follows our [guidelines](https://github.com/talend/tools/blob/master/tools-root-github/CONTRIBUTING.md)

### DIFF
--- a/packages/eslint-config/api-app.yaml
+++ b/packages/eslint-config/api-app.yaml
@@ -1,6 +1,7 @@
 extends: ./api/eslintrc.base.yaml
 env:
   browser: true
+  jquery: true
 rules:
   import/no-commonjs: error
   import/no-nodejs-modules: error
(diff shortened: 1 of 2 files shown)


packages/eslint-config/api-app.yaml
```diff
@@ -1,6 +1,7 @@
 extends: ./api/eslintrc.base.yaml
 env:
   browser: true
+  jquery: true
```
quilicicf: Adding jquery for all apps? :scream: 
Can't you update the `api-app-angular` file instead?

Kporal: Well I thought this file was only use by the API team and we have jquery on both...

PR #2463, opened 2019-10-11
fix(EditableText): stop push content with elipsis

romainseb:
The current editable text push the content when we over it to show the pencil icon

We have to :
- Ensure we can focus the pencil with the keyboard
- Keep some space in the elipsis to place the icon
- The button has not to be sr-only as it's not a content for the screen reader only

### DIFF
--- a/packages/components/src/Drawer/Drawer.scss
+++ b/packages/components/src/Drawer/Drawer.scss
@@ -69,18 +69,13 @@ $tc-drawer-z-index: $drawer-z-index !default;
 }
 
 .tc-drawer-header {
-	:global(.tc-editable-text-pencil) {
-		color: $gray;
-		opacity: 0;
-	}
-
-	&:hover {
-		:global(.tc-editable-text-pencil) {
-			opacity: 1;
-			&:hover {
-				color: $scooter;
-			}
+	:global {
+		.tc-editable-text {
+			width: 90%;
 		}
+	}
+	:global(.tc-editable-text-pencil) {
+		color: $white;
 	}
 	background-color: $brand-primary-lighter;
 
--- a/packages/components/src/Drawer/Drawer.scss
+++ b/packages/components/src/Drawer/Drawer.scss
@@ -73,9 +73,9 @@ $tc-drawer-z-index: $drawer-z-index !default;
 		.tc-editable-text {
 			min-width: 0;
 		}
-	}
-	:global(.tc-editable-text-pencil) {
-		color: $white;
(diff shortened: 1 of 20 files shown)


packages/components/stories/Drawer.js
```diff
@@ -217,14 +217,39 @@ const editableDrawers = [
...
 ];
 
+const longEditableDrawers = [
+	<Drawer
+		stacked
+		title="Im stacked drawer 1"
+		footerActions={Object.assign({}, basicProps, { selected: 0 })}
+	>
+		<h1>Hello drawer 1</h1>
+		<p>{"You should not being able to read this because I'm first"}</p>
```
jare-talend:
```suggestion
		<p>You should not being able to read this because I'm first</p>
```

jare-talend:
cross
The button delete move slightly

packages/components/src/Drawer/Drawer.scss
```diff
@@ -69,19 +69,14 @@ $tc-drawer-z-index: $drawer-z-index !default;
...
-
-	&:hover {
-		:global(.tc-editable-text-pencil) {
-			opacity: 1;
-			&:hover {
-				color: $scooter;
-			}
+	:global {
+		.tc-editable-text {
+			width: 90%;
```
jare-talend:
```suggestion
			min-width: 0;
```

jare-talend: You use flexibility. The button close don't move with that.

romainseb:
> I don't understand why you remove the sr-only. This class is a way to mask accessibility an element. For every one, the element gone away but for the people using a screen reader, it can see it.

It's the opposite, sr-only class is placed on elements we don't want to show to non screen reader users but when we need to give some information to *screen reader only*.
In this case, we want the button for the users using a screen reader or not 😄

packages/components/src/Drawer/Drawer.scss
```diff
@@ -69,19 +69,14 @@ $tc-drawer-z-index: $drawer-z-index !default;
...
-				color: $scooter;
-			}
+	:global {
+		.tc-editable-text {
+			min-width: 0;
 		}
 	}
+	:global(.tc-editable-text-pencil) {
+		color: $white;
+	}
```
ncomont:
```
	:global {
 		.tc-editable-text {
 			min-width: 0;
 		}
 	        .tc-editable-text-pencil) {
 		        color: $white;
 	}
```

PR #2469, opened 2019-10-15
(locale) add "about"

imba-tjd:
It's simple so I'm not using the template.

This fix is referred from the context.

Change the title if you want


jsomsanith-tlnd:
Locales have changed a lot, the destination branch doesn't exists anymore.
Locales are now published in a separated package: [@talend/locales-tui](https://www.npmjs.com/package/@talend/locales-tui)

PR #2475, opened 2019-10-16
fix(Components): badge proptypes

nicomaligne:
The badge components emits proptypes error when used with children.
This is due to the creation of the defaultContent var.
Reorganize the code a little bit, to create a default badge only when needed.


packages/components/src/Badge/Badge.component.js
```diff
@@ -16,6 +16,47 @@ const SIZES = {
...
+const BadgeType = ({ disabled, onSelect, children, ...rest }) =>
+	onSelect ? (
+		<button {...rest} key="button" type="button" disabled={disabled} onClick={onSelect}>
+			{children}
+		</button>
+	) : (
+		<div {...rest} key="div">
+			{children}
+		</div>
+	);
```
ncomont: Could you write something like that in order to remove the `eslint-disable`, please ?

```javascript
const BadgeType = ({ disabled, onSelect, children, ...rest }) => {
	if (onSelect) {
		return (
			<button {...rest} key="button" type="button" disabled={disabled} onClick={onSelect}>
				{children}
			</button>
		);
	}

	return (
		<div {...rest} key="div">
			{children}
		</div>
	);
}
```

PR #2480, opened 2019-10-17
chore(forms): cleanup imports to be tree shakable

jmfrancois:

Forms import component from index. This make it not tree shaking friendly.

Use direct import to the component


packages/forms/src/FormSkeleton.js
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import { Skeleton } from '@talend/react-components';
```
sgendre: what about webpack 4 optimization 
> Thanks to the usage of the optimization.usedExports and the unused option of the UglifyJsPlugin, unnecessary code was removed
https://wanago.io/2018/08/13/webpack-4-course-part-seven-decreasing-the-bundle-size-with-tree-shaking/

jmfrancois: I can t give guaranties component do not have sideEffects.
Last time I tried this approch it just didn t work with default export and we use only this approch.
I think the import from 'lodash/get' looks more natural.
WDYT ?

sgendre: was just in cas it could avoid going through all component by just webpack config, didn't try either

PR #2485, opened 2019-10-18
fix(cmf/onError): remove uiState aka PII

jmfrancois:

prepareObject is used to serialize all actions which are dispatched.
In some case the action has an event object which contains circular dependencies.
that fails into uncaught error.

uiState contains PII which is not acceptable to be just logged on our side.

Remove UIState, keep only last actions, time, browser, error.
In Action we keep only the type.


packages/cmf/src/onError.js
```diff
@@ -85,6 +104,11 @@ function prepareObject(originalState) {
 	if (originalState === null) {
 		return null;
 	}
+	if (isEvent(originalState)) {
+		return {
+			type: originalState.type,
```
sgendre: can't we embed the just first one or two level of the event?
it might be enough to proceed to debug.

packages/containers/src/Notification/Notification.sagas.js
```diff
@@ -31,8 +34,8 @@ function* onCMFError(action) {
 	};
 	const notification = {
 		type: 'error',
-		title: `${error.name}: ${error.message}`,
-		message: error.stack,
+		title: error.name,
+		message: error.message,
```
jmfrancois: make the notification very light and do not display the stack

packages/containers/src/Notification/Notification.sagas.js
```diff
@@ -22,6 +22,9 @@ export function* onPushNotification(action) {
 }
 
 function* onCMFError(action) {
+	if (process.env.RELEASE_WITH_ERROR && process.env.NODE_ENV === 'production') {
```
sgendre: better named by RELEASE_WITHOUT_ERROR as it does a return ?

output/cmf.eslint.txt
```diff
@@ -8,12 +8,15 @@ The react/require-extension rule is deprecated. Please use the import/extensions
 /home/travis/build/Talend/ui/packages/cmf/src/Inject.component.js
   23:2  warning  Unexpected console statement  no-console
 
+/home/travis/build/Talend/ui/packages/cmf/src/onError.js
+  126:33  error  'addSensibleKeyRegexp' is not defined  no-undef
```
jare-talend: to use a function addSensibleKeyRegexp not declared in https://github.com/Talend/ui/pull/2485/files#diff-fea8ec2c9e85e497f04b0115a332439eR126

packages/containers/src/Notification/Notification.sagas.js
jare-talend: or NOTIFICATION_JS_ERROR because I think that this variable disable the errors. It is only the notification.

jare-talend: By default, it is not disabled. I prefer disable by default.
In other hand, we can activate when  process.env.NODE_ENV !== 'production' to remove the errors during the development and see them

output/cmf.eslint.txt
jmfrancois: already fixed

packages/containers/src/Notification/Notification.sagas.js
jmfrancois: What we have validated with PM is: Dev must see error notification during dev.
Because it s a temporary solution we dont want team to activate it but to deactivate it.
Either we will have same scenario at the end when we will want to remove the flag.

OK to rename the flag

jare-talend:
https://github.com/Talend/ui-scripts/pull/26 allow to disable the notifications from a environnement variable.

packages/cmf/src/onError.js
```diff
@@ -176,32 +101,10 @@ function report(error) {
  * addAction store last 20 actions to let onError.report use it.
  */
 function addAction(action) {
-	if (ref.actions.length >= 20) {
+	if (ref.actions.length >= 200) {
```
jsomsanith-tlnd: Why that many ?

jmfrancois: reverted

PR #2487, opened 2019-10-18
feat(faceted-search): add connection type badge

nicomaligne:
We need to create a new badge for the connection type.
[UX design](http://guidelines.talend.com/document/215501#/navigation-layout/filtering-search)
The connection type is a checkbox form, so a new kind of badge "checkbox" as been created.
It will handle all the checkbox need for future badge
BadgesCheckboxes is the new badge.


packages/faceted-search/src/queryClient/tql.js
```diff
@@ -23,9 +23,13 @@ const prepareBadges = flow([removeBadgesWithEmptyValue, getBadgesQueryValues]);
 const getTqlClassOperatorsDictionary = query => ({
 	contains: query.contains,
 	'=': query.equal,
+	in: query.in,
 });
 
 const formatValue = value => {
+	if (Array.isArray(value)) {
+		return value.map(val => val.id).join(',');
```
vbocquet: Here, if i get a, empty array as value, we return `''` which can be get a TQL query like, for example, `connection.type in []`.

Are we ok with that? Label in facet show a "all", be i think query return 0 element no?

nicomaligne: good point

PR #2496, opened 2019-10-23
feat(components/ListComposition): Expose setSelectedIds from hooks

sgangnant:
External events can change the items of the list. This can cause an issue for example when deleting elements from the list :

deleteallgroup

As we can see, multiSelectActions remain in place while their state should be resumed.

Therefore we need to also be able to update the item selection accordingly.

Expose state setter


sgangnant:
Not the right way to adress this issue

PR #2500, opened 2019-10-28
fix(components): VirtualizedList - resizable.js workingIndex is equal to -1

smouillour:
An error is visible when the last column is reduced at the minimum possible

ListResizableError

In fact the variable workingIndex is set to -1 and in this case there is no related value in columnsWidths

ListResizableError2


ncomont:
Hi *Sebastien Le Mouillour*,
Approved, but agree with JM, a test is missing :)

PR #2257, opened 2019-06-18
feat(components/breadcrumb): refresh style

frassinier:
**What is the problem this PR is trying to solve?**
[TCCUI](https://jira.talendforge.org/browse/TDP-6895) related changes 

**What is the chosen solution to this problem?**
Refresh breadcrumb style

**Please check if the PR fulfills these requirements**

* [x] The PR commit message follows our [guidelines](https://github.com/talend/tools/blob/master/tools-root-github/CONTRIBUTING.md)
* [x] Tests for the changes have been added (for bug fixes / features) And [non reg](./screenshots.md) done before need review
* [ ] Docs have been added / updated (for bug fixes / features)
* [x] Related design / discussions / pages (not in jira), if any, are all linked or available in the PR
https://company-57688.frontify.com/document/92132#/navigation-layout/breadcrumb-1555666347

**[x] This PR introduces a breaking change**

**Before** we used `maxItems ` to specify how many items should be always displayed after the ellipsis button.

| maxItems | example |
|--- | --- |
| 2 | `Home / Folder 1` |
| 2 | `… / Folder 1 / Folder 2` |
| 3 | `Home / Folder 1 / Folder 2` |
| 3 | `… / Folder 2 / Folder 3 / Folder 4` |

**Now**, root item is always visible, so we use `maxItems` to specify how many items should be displayed including root and the ellipsis button

| maxItems | example |
|--- | --- |
| 3 | `Home / Folder 1 / Folder 2` |
| 3 | `Home / … / Folder 3` |
| 4 | `Home / … / Folder 3 / Folder 4` |
| 4 | `Home / … / Folder 4 / Folder 5` |

So if you used
```jsx
<Breadcrumb maxItems='2'  />
``` 
you must consider changing it to 
```jsx
<Breadcrumb maxItems='4'  />
```

PR #2504, opened 2019-10-28
fix(containers/breadcrumbs): loading state

frassinier:
Not able to use loading state using CMF implemented in #2257 

Just pass loading state in the mapStateToProps


packages/containers/src/Breadcrumbs/Breadcrumbs.connect.js
```diff
@@ -9,10 +9,12 @@ export function ContainerBreadcrumbs(props) {
 	const state = props.state || DEFAULT_STATE;
 	const items = state.get('items', props.items);
 	const newProps = {
+		id: state.get('id', props.id),
```
jmfrancois: this state.get looks strange to me. the state is supposed to change ?

jsomsanith-tlnd: I agree, why do we have id in state ? Shouldn't it be only the prop ?

PR #2506, opened 2019-10-29
feat(date-range-picker): DateRange.Picker

yyanwang:
`InputDateRangePicker`'s structure will looks like this:
```
<DateRange.Manager>
    <DateRange.Inputs />
    <DateRange.Picker />
</DateRange.Manager>
```

This PR is to add `DateRangeContext` and `DateRange.Picker` component.


yyanwang:
Hi @jsomsanith 
It can be confusing since this is only part of change, but what I intend to do is that:
* when startDate/endDate get passed to calendarPicker, it suggest calendar picker work in "range mode":
1. if `startDate` and `endDate` both provided, it will highlight range
2. if only startDate provided and focus is on `from` field, it will only highlight startDate
3. if only endDate provided and focus is on `to` field, it will only highlight endDate

* maintain a `state.focusedInput` in `DateRange.Manager`.
* when user pick a date from calendar, it will set state.startDate or state.endDate base on state.focusedInput.

yyanwang:
more changes comes in https://github.com/Talend/ui/pull/2505
and story: http://2505.talend.surge.sh/components/?path=/story/datepicker--date-range-picker-with-inputs

jsomsanith-tlnd:
(not related to this PR, but the story has an issue, when you are on end input, the down arrow should focus on end selected date)
With this API, you have to pass the value and the start/end date. If you are on `from` pickler for example, you have to duplicate the same date in `value` and `startDate` props.

What I suggest in Calendar is:
- you have a props.value. It's the current value
- if you have a startDate, then you highlight between [start, value]
- else if you have an endDate, then you highlight between [value, end]

But either way, you need to have only the `value` date as selected state (that will fix the issue).

yyanwang:
@jsomsanith OK, I made changes as you suggested, please check again.

PR #2509, opened 2019-10-30
feat(faceted-search/AddFacetPopover): Filter badge by label, not attribute

vbocquet:
In faceted popover, when we need to choose a field, we have a search bar to help us.
Today, filtering is based on `attribute` string of badge definition, not on `label`.
Problem is, `attribute` can be really a technical key (like `product/name`). We need to filter by label, which is the value displayed to user in list below.

We also need to made label case insensitive because the `includes()` method is case sensitive.

- Filter by `label`, not `attribute`. 
- Make it case insensitive.
- Add some tests.

### DIFF
--- a/packages/faceted-search/src/components/AddFacetPopover/AddFacetPopver.component.test.js
+++ b/packages/faceted-search/src/components/AddFacetPopover/AddFacetPopver.component.test.js
@@ -2,7 +2,7 @@
 import React from 'react';
 import { mount } from 'enzyme';
 import { act } from 'react-dom/test-utils';
-import { AddFacetPopover, filterByLabel } from './AddFacetPopover.component';
+import { AddFacetPopover } from './AddFacetPopover.component';
 import getDefaultT from '../../translate';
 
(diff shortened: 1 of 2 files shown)


packages/faceted-search/src/components/AddFacetPopover/AddFacetPopver.component.test.js
```diff
@@ -2,7 +2,7 @@
 import React from 'react';
 import { mount } from 'enzyme';
 import { act } from 'react-dom/test-utils';
-import { AddFacetPopover } from './AddFacetPopover.component';
+import { AddFacetPopover, filterByLabel } from './AddFacetPopover.component';
```
nicomaligne: In my opinion there is no need to test the filterByLabel unitary.
It's implicitly test in an higher test.

vbocquet: Yes, i think you're right, this test is a little bit overkill.
I remove it

PR #2511, opened 2019-10-30
chore(forms/MultiSelectTag): calling onFinish when focusing out

jlame:
Trying to add a customValidation on the multitag select widget, but the onFinish is not called when blurring.

Calling onFinish when focusing out from the typeahead

### DIFF
--- a/packages/forms/src/UIForm/fields/MultiSelectTag/MultiSelectTag.component.js
+++ b/packages/forms/src/UIForm/fields/MultiSelectTag/MultiSelectTag.component.js
@@ -171,11 +171,13 @@ export default class MultiSelectTag extends React.Component {
 	/**
 	 * Remove all suggestions
 	 */
-	resetSuggestions() {
+	resetSuggestions(event) {
+		const { onFinish, schema } = this.props;
 		this.setState({
 			suggestions: undefined,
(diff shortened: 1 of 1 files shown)


jlame:
> Sorry but I don't understand this change.
> MultiSelectTag is about tags. When we add a tag, we should validate. And that's what is done in `onAddTag()`. The input has never been something to store the value as string, where the blur validate the current value.
> Can you explain the use case please ?

So the usecase was to display an error using custom validation when focusing the input and not selecting anything then go somewhere else (blurring the component)

packages/forms/src/UIForm/fields/MultiSelectTag/MultiSelectTag.component.js
```diff
@@ -171,11 +171,13 @@ export default class MultiSelectTag extends React.Component {
...
 	 * Remove all suggestions
 	 */
-	resetSuggestions() {
+	resetSuggestions(event) {
+		const { onFinish, schema } = this.props;
 		this.setState({
 			suggestions: undefined,
 			focusedItemIndex: undefined,
 		});
+		onFinish(event, { schema });
```
frassinier: Can you add a unit test for that?

PR #2517, opened 2019-11-04
feat(date range): Date Range Inputs

yyanwang:
Add `DateRange.Inputs`
btw no story specific for this change. the story will comes later with `DateRange.Manager` changes.
`InputDateRangePicker`'s structure will looks like this:
```
<DateRange.Manager>
    <DateRange.Inputs />
    <DateRange.Picker />
</DateRange.Manager>
```


packages/components/src/DateTimePickers/DateRange/Inputs/Inputs.component.js
```diff
@@ -0,0 +1,88 @@
...
+import InputSizer from '../../shared/InputSizer';
+
+import Icon from '../../../Icon';
+
+import theme from './Inputs.scss';
+
+const OMIT_PROPS_INPUT = ['startInputRef', 'endInputRef'];
+
+function Inputs(props) {
+	const { startDate, endDate, inputManagement } = useContext(DateRangeContext);
```
smouillour: maybe we need to manage the case of the user doesn't define inputManagement correctly and display a specific error in the console to explain what is missing instead something like `can not read property placeholder of undefined`  ?

smouillour: and use deconstruction for inputManagement instead of use inputManagement.something

yyanwang: Changed to deconstruction of `inputManagement`.
And `DateRange.Inputs` is not an individual component, it is supposed to always use under DateRangeContext. `DateRange.Manager` will provide the context value(changes comes in next PR), so it guarantee that inputManagement is passed. 😸

smouillour: not totally agree with that 😕. let's see what @yonmey or @jsomsanith will said about that 😃

jsomsanith-tlnd: You must not expose this component in the index, so it's considered as internal, and that's sufficient for me.

packages/components/src/DateTimePickers/DateRange/Inputs/Inputs.component.js
```diff
@@ -0,0 +1,82 @@
...
+	let endDateInputRef = useRef(null);
+
+	useEffect(() => {
+		if (startDateInputRef && focusedInput === START_DATE) {
+			startDateInputRef.focus();
+		}
+		if (endDateInputRef && focusedInput === END_DATE) {
+			endDateInputRef.focus();
+		}
+	});
```
jsomsanith-tlnd: I don't understand what this is supposed to do, and how focusedInput would be changed.
And you haven't passed any dependencies to your useEffect, which means it is executed at each render.

packages/components/src/DateTimePickers/DateRange/Inputs/Inputs.component.js
```diff
@@ -0,0 +1,82 @@
...
+
+	return [
+		<InputSizer placeholder={placeholder} inputText={startDate.value ? startDate.textInput : ''}>
+			{width => (
+				<DebounceInput
```
jsomsanith-tlnd: The labels are missing

packages/components/src/DateTimePickers/DateRange/Inputs/Inputs.scss
```diff
@@ -0,0 +1,14 @@
+$range-picker-arrow-size: 1.2rem;
```
jsomsanith-tlnd: Please use the sass var

packages/components/src/DateTimePickers/DateRange/Inputs/Inputs.component.js
yyanwang: will add the condition array.
but I need to put some words about this `focusedInput`. `focusedInput` can be 'startDate' or 'endDate', and it's part of the context value of DateRangeContext. It's used for moving focus to end date after start date is selected.

packages/components/src/DateTimePickers/DateRange/Inputs/Inputs.component.js
yyanwang: by labels do you mean placeholder or "from"/"to"?

packages/components/src/DateTimePickers/DateRange/Inputs/Inputs.component.js
jsomsanith-tlnd: I have a responsibility problem here. It's more the manager that should manage the focus between from/to, when the from is selected from picker.
Let's merge as it is for now, but when you create the `InputDateRange` component, let's merge it with this one. So `InputDateRange` will compose the manager, the 2 inputs, and the popover. You can then register the `to` input refs in the manager, and focus on it, on `from` picker selection.

packages/components/src/DateTimePickers/DateRange/Inputs/Inputs.component.js
jsomsanith-tlnd: Yes the labels should be set with the inputs, and don't forget i18n

packages/components/src/DateTimePickers/DateRange/Inputs/Inputs.scss
jsomsanith-tlnd: It's still ko, use `$svg-sm-size`. You can find the svg size variables in the bootstrap theme > _guidelines.scss

packages/components/src/DateTimePickers/DateRange/Inputs/Inputs.component.js
yyanwang: ok, noted

packages/components/src/DateTimePickers/DateRange/Inputs/Inputs.component.js
yyanwang: labels added.
I thought the labels will be provided by form schema.
So talking about form wrapper of this, we just need to wrap `InputDateRangePicker`? instead of composing 2 DateWidget?

jsomsanith-tlnd: There would be 2 kinds of labels. For example, when you define travel dates you will have some kind of "Travel", then "from/to".
We will have to manage the labels a little differently to have screen readers have some context when focusing on the inputs, let's tackle that with the wrapper.

packages/components/src/DateTimePickers/DateRange/Inputs/Inputs.component.js
```diff
@@ -0,0 +1,97 @@
...
+			startDateInputRef.focus();
+		}
+		if (endDateInputRef && focusedInput === END_DATE) {
+			endDateInputRef.focus();
+		}
+	}, [focusedInput]);
+
+	return [
+		<div>
+			<label htmlFor={`${props.id}-start-date-input`} className="control-label">{t('DATE_RANGE_FROM', { defaultValue: 'From' })}</label>
```
jsomsanith-tlnd:
```suggestion
			<label htmlFor={`${props.id}-start-date-input`} className="control-label">{t('TC_DATE_PICKER_RANGE_FROM', { defaultValue: 'From' })}</label>
```

packages/components/src/DateTimePickers/DateRange/Inputs/Inputs.component.js
```diff
@@ -0,0 +1,97 @@
...
+						{...omit(props, OMIT_PROPS_INPUT)}
+					/>
+				)}
+			</InputSizer>
+		</div>,
+		<span className={theme.arrow}>
+			<Icon name="talend-arrow-right" className={theme.icon} />
+		</span>,
+		<div>
+			<label htmlFor={`${props.id}-end-date-input`} className="control-label">{t('DATE_RANGE_TO', { defaultValue: 'To' })}</label>
```
jsomsanith-tlnd:
```suggestion
			<label htmlFor={`${props.id}-end-date-input`} className="control-label">{t('TC_DATE_PICKER__RANGE_TO', { defaultValue: 'To' })}</label>
```

PR #2525, opened 2019-11-04
fix(component/TabBar): TMC-19289 fix dropdown item crash

yonmey:
https://jira.talendforge.org/browse/TMC-19289

When resizing the screen on a TabBar component page, the app crashes due to an undefined item.

Fixed by finding the selected item with the .find() method instead of looking into a false array.


packages/components/src/TabBar/TabBar.component.js
```diff
@@ -116,8 +116,9 @@ function TabBar(props) {
 			))}
 		</Tab.Content>
 	);
-	if (responsive && showDropdown) {
-		const selectedItem = items[selectedKey - 1];
+
+	const selectedItem = items.find(item => item.key === selectedKey);
```
yyanwang: Does this mean that when no tab selected, it will not render responsive layout?
I propose when no tab selected, render responsive layout with first tab item instead.

PR #2537, opened 2019-11-08
fix(typeahead): restore id selector

romainseb:
We have an override for the id, it shows react-autowathever-{id}
We want to stay with the actual id
Change the override order for the props


romainseb:
It looks weird i agree to use the key for id as it could be a technical id for react 🤔 
I assume it's override each time right now

jsomsanith-tlnd:
`renderInputComponent` is not a component, but a function that is called by react-autowhatever. JM noticed that there are propTypes, but I guess this is because of the linter :D

PR #2545, opened 2019-11-14
fix(faceted-search/tql): Full support of operator notEqual

vbocquet:
`notEqual` operator is available in operators dictionary as a standard one. 
Problem, when we try to use it in a text badge with the `createTqlQuery` adapter, it pop an error: ` TypeError: queryFunction is not a function`
This is because a mapping is missing in the tql converter function `getTqlClassOperatorsDictionary`

Just add the key / value in the mapping function, add some tests / improve story

Also take the opportunity of this PR to rename the `=` operator in `equal`. **Be aware that it's a breaking change and that's need some changes on app usage.**

### DIFF
--- a/packages/faceted-search/src/dictionary/operator.dictionary.js
+++ b/packages/faceted-search/src/dictionary/operator.dictionary.js
@@ -1,6 +1,6 @@
 const operatorNames = {
 	contains: 'contains',
-	equal: '=',
+	equal: 'equal',
 	notEqual: 'notEqual',
 	in: 'in',
 };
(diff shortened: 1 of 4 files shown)


packages/faceted-search/src/dictionary/operator.dictionary.js
```diff
@@ -1,6 +1,6 @@
 const operatorNames = {
 	contains: 'contains',
-	equal: '=',
+	equal: 'equal',
```
frassinier: I've implemented something like that [here](https://github.com/Talend/ui/blob/master/packages/forms/src/UIForm/fields/Comparator/Comparator.component.js#L13)
We definitively capitalize somewhere else this kin of mapping!

vbocquet: Thx for show me this, we can't support all operators for now so it's good here to have only the one that we implement, but yes it's a maybe good idea. 
As faceted-search is another package that ui and can evolve in another way, maybe it's more simple to not share this object for now

frassinier: 👌 just a comment to think about it!

PR #2549, opened 2019-11-19
feat(range picker): InputDateRangePicker

yyanwang:
Add `InputDateRangePicker`.
```
<InputDateRangePicker>
    <DateRange.Manager>
        <DateRange.Input /> // start date input
        <DateRange.Input /> // end date input
        <DateRange.Picker />
    </DateRange.Manager>
</InputDateRangPicker>
```


yyanwang:
story:
http://2549.talend.surge.sh/components/?path=/story/datepicker--date-range-picker-with-inputs

packages/components/src/DateTimePickers/InputDateRangePicker/InputDateRangePicker.component.js
```diff
@@ -0,0 +1,145 @@
...
+
+const PROPS_TO_OMIT_FOR_INPUT = ['id', 'dateFormat', 'required', 'value', 'onBlur', 'onChange'];
+
+export default function InputDateRangePicker(props) {
+	const popoverId = `date-range-picker-${props.id || uuid.v4()}`;
+
+	const startDateInputRef = useRef(null);
+	const endDateInputRef = useRef(null);
+	const containerRef = useRef(null);
+	const [inputRef, setInputRef] = useState(null);
```
yyanwang: keep a record of which input is focused on, and pass it to `DateRange.Picker` so it will know which picker to render

packages/components/src/DateTimePickers/InputDateRangePicker/InputDateRangePicker.scss
```diff
@@ -0,0 +1,31 @@
...
+	width: $tc-datepicker-width;
+	background: $tc-datepicker-background;
+	box-shadow: $tc-datepicker-box-shadow;
+	z-index: $tc-datepicker-z-index;
+}
+.date-picker {
+	width: fit-content;
+	display: flex;
+	align-items: flex-end;
+}
```
jsomsanith-tlnd: Please don't duplicate that for each picker. Extract that somewhere common and import this common style in each picker to set the popper classname

packages/components/src/DateTimePickers/DateRange/Input/Input.component.js
```diff
@@ -0,0 +1,48 @@
...
+						autoComplete="off"
+						className="form-control"
+						debounceTimeout={300}
+						type="text"
+						placeholder={placeholder}
+						value={date.textInput}
+						style={{ width }}
+						onChange={onChange}
+						onFocus={onFocus}
+						inputRef={ref}
```
jsomsanith-tlnd: You don't pass/spread additional props from InputDateRangePicker. That can lead to accessiibility issue?
For example if we want to have an error div, we want to associate the inputs with this error via `aria-describedby`. And there can be tons of cases. Just spread the rest of the props from InputDateRangePicker to DateRange.Input, and from DateRange.Input to the DebounceInput please

yyanwang:
@jsomsanith glitch is caused by input resizing.
when there's no input, use the size of placeholder,
while user input, means still no valid input, use the size of placeholder
when there's valid date input, use the size of input.

so to prevent glitch, I changed to use placeholder width always.

PR #2554, opened 2019-11-21
fix(faceted-search/queryClient): Fix TQL creation function with numbers

vbocquet:
TQL creation do not works for badge numbers.
I write a wrong unit test which have value as a string instead of a number. With that, all is good.
But the badge number return a number, not a string.
The problem come from `isNotEmptyOrNull` function which do not return a boolean and not handle numbers for now.

Fix the `isNotEmptyOrNull` function and tests.

### DIFF
--- a/packages/faceted-search/src/queryClient/tql.js
+++ b/packages/faceted-search/src/queryClient/tql.js
@@ -9,7 +9,7 @@ const getBadgeQueryValues = ({ properties }) => [
 
 const getBadgesQueryValues = badges => badges.map(getBadgeQueryValues);
 
-const isNotEmptyOrNull = value => value && value.length;
+const isNotEmptyOrNull = value => !!(typeof value === 'number' || (value && value.length));
 const isObjectIdNotEmptyOrNull = value => isNotEmptyOrNull(value.id);
 
 const filterBadgeWithNoValue = ({ properties }) => {
--- a/packages/faceted-search/src/queryClient/tql.js
+++ b/packages/faceted-search/src/queryClient/tql.js
@@ -12,19 +12,19 @@ const getBadgesQueryValues = badges => badges.map(getBadgeQueryValues);
 
 const isValidValue = value => {
 	if (typeof value === 'string') {
-		return isEmpty(value);
+		return !isEmpty(value);
 	}
-	return Number.isNaN(value);
+	return !Number.isNaN(value);
 };
 
-const filterBadgeWithNoValue = ({ properties }) => {
+const keepValidValues = ({ properties }) => {
 	if (Array.isArray(properties.value)) {
-		return !properties.value.every(({ id }) => isValidValue(id));
+		return properties.value.length && properties.value.every(({ id }) => isValidValue(id));
 	}
-	return !isValidValue(properties.value);
+	return isValidValue(properties.value);
 };
 
-const removeBadgesWithEmptyValue = badges => badges.filter(filterBadgeWithNoValue);
+const removeBadgesWithEmptyValue = badges => badges.filter(keepValidValues);
 
 const prepareBadges = flow([removeBadgesWithEmptyValue, getBadgesQueryValues]);
 
--- a/packages/faceted-search/src/queryClient/tql.js
+++ b/packages/faceted-search/src/queryClient/tql.js
@@ -1,4 +1,5 @@
 import { Query } from '@talend/daikon-tql-client';
+import isEmpty from 'lodash/isEmpty';
 import flow from 'lodash/flow';
 
 const getBadgeQueryValues = ({ properties }) => [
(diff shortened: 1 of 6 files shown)


packages/faceted-search/src/queryClient/tql.js
```diff
@@ -9,7 +9,7 @@ const getBadgeQueryValues = ({ properties }) => [
 
 const getBadgesQueryValues = badges => badges.map(getBadgeQueryValues);
 
-const isNotEmptyOrNull = value => value && value.length;
+const isNotEmptyOrNull = value => !!(typeof value === 'number' || (value && value.length));
```
nicomaligne: I'm not sure about your condition...
```javascript
const isNotEmptyOrNull = value => typeof value === 'number' || (value && value.length);
```
☝️ seems to do the job, no ? I may miss a point...

vbocquet: You're right 👍  I make it too complicated, just fix that, thanks

packages/faceted-search/src/queryClient/tql.js
```diff
@@ -9,7 +9,7 @@ const getBadgeQueryValues = ({ properties }) => [
 
 const getBadgesQueryValues = badges => badges.map(getBadgeQueryValues);
 
-const isNotEmptyOrNull = value => value && value.length;
+const isNotEmptyOrNull = value => typeof value === 'number' || (value && value.length);
```
jmfrancois: do not change this one. I prefer you to add an other isNumber, add some test (using NaN for example).
This one is used to check ids, ... not only the tql value field.

vbocquet: @jmfrancois Thx, you're right, i introduced a bug. Just fix that with a split function and more tests in `packages/faceted-search/src/queryClient/tql.test.js`. Also exclude `NaN` from condition.

packages/faceted-search/src/queryClient/tql.js
```diff
@@ -10,13 +10,15 @@ const getBadgeQueryValues = ({ properties }) => [
 const getBadgesQueryValues = badges => badges.map(getBadgeQueryValues);
 
 const isNotEmptyOrNull = value => value && value.length;
+const isValidNumber = value => typeof value === 'number' && !isNaN(value);
```
jsomsanith-tlnd: Wouldn't it be easier to fix `isNotEmptyOrNull` ?
Your fix is good for your case, but here we are adding specific cases. It's equivalent to maintain a white list of rules. What about maintaining a black list instead ? And that's what I understand reading `isNotEmptyOrNull`. 
We could, instead of checking `value => value && value.length`
- check if it's not an empty string
- check if it's not undefined
- check if it's not null

vbocquet: Thx for review, `isNotEmptyOrNull` is used in two lines :
- https://github.com/Talend/ui/blob/1ad07026e673dc546b80f7ddaefde71560752e49/packages/faceted-search/src/queryClient/tql.js#L21
- https://github.com/Talend/ui/blob/1ad07026e673dc546b80f7ddaefde71560752e49/packages/faceted-search/src/queryClient/tql.js#L15
That's why i didn't fix `isNotEmptyOrNull` after @jmfrancois review.

jsomsanith-tlnd: Oh I haven't seen his review. But I still don't understand why checking the id to be different to null, undefined and '' would be ko

nicomaligne: Why not create a specific method for the id check and the badge value check : 
```suggestion
const isNotEmptyOrNull = value => value && value.length;
const isValidNumber = value => typeof value === 'number' && !isNaN(value);

const isObjectIdNotEmptyOrNull = value => isNotEmptyOrNull(value.id);
const isValidBadge = value => isValidNumber(value) || isNotEmptyOrNull(value);

const filterBadgeWithNoValue = ({ properties }) => {
	if (Array.isArray(properties.value) && properties.value.length) {
		return properties.value.every(isObjectIdNotEmptyOrNull);
	}
	return isValidBadge(properties.value);
};
```

nicomaligne: @jsomsanith the id is a necessity in the design.
The id is the technical id that will be used in the tql request.
For example : https://github.com/Talend/ui/blob/master/packages/faceted-search/src/queryClient/tql.test.js#L145

vbocquet: @nicomaligne Thx, just commit your suggestion 👍

jsomsanith-tlnd: I still don't get why checking if a value 
- is not undefined
- is not null
- is not empty string
... is not acceptable for an id.

What I wanted to say with the white list/black list thing is that tomorrow, you have a new type of value, for example a boolean. The check will fail, because the "false" value won't pass the current check.
That's because you have an equivalent of white list, each use case will need to add a new rule.
If you indicates the values that are invalid, as "not provided value", you add a use case, it will still be ok.

That's why I asked if checking the id and the value badge would be ok with the 3 cases (undefined, null, '') described above.

PR #2556, opened 2019-11-25
fix(notifications): clean registry

lmaillet:
I've noticed that notification registry is never cleaned.

To avoid memory leak, clean the notification registry as soon as `leaveFn` function is called.


packages/components/src/Notification/Notification.component.js
```diff
@@ -172,6 +172,14 @@ class Registry {
 			this.timerRegistry[notification.id].cancel();
 		}
 	}
+
+	clean(notification) {
+		if (this.isRegistered(notification)) {
+			this.cancel(notification);
+			// Clean registry to avoid memory leak
+			delete this.timerRegistry[notification.id];
```
jsomsanith-tlnd: Shouldn't it be removed on cancel too ?
From what I saw in the code, on close, we call cancel, but the timer is not removed in this case.

So what about : 
- cancel: cancel the timer, and remove it from registry
- On close: cancel
- On register, the timer calls cancel instead of this clean function. If we move the delete in the cancel function, this clean function is not relevant anymore

packages/components/src/Notification/Notification.component.js
```diff
@@ -267,7 +278,7 @@ class NotificationsContainer extends React.Component {
 }
 
 const notificationShape = {
-	id: PropTypes.any.isRequired,
+	id: PropTypes.any.isRequired, // each notification must have a unique id
```
jsomsanith-tlnd: This comment is not really useful as it describe that the id proptypes is required, just next to the code.
I suggest you to remove it.

PR #2561, opened 2019-11-28
feat(components): richlayout with forwardref

jare-talend:

The popover is visible in the list while the popover has the focus.
A button lost his focus if we set it to disabled. So the focus is lost on the popover.

lost_focus

Add a function to add the focus on the popover

### DIFF
--- a/packages/components/src/RichTooltip/RichLayout/RichLayout.component.js
+++ b/packages/components/src/RichTooltip/RichLayout/RichLayout.component.js
@@ -37,33 +37,21 @@ const Footer = ({ id, children, className }) => (
 );
 Footer.propTypes = TooltipPropTypes;
 
-export default class RichLayout extends React.Component {
-	constructor(props) {
-		super(props);
-		this.richLayout = React.createRef();
-	}
+const RichLayout = React.forwardRef((props, ref) => (
+	<span id={props.id} ref={ref} tabIndex="-1">
+		<Header id={`${props.id}-header`}>
+			{Inject.getReactElement(props.getComponent, props.Header)}
+		</Header>
+		<Body id={`${props.id}-body`} className={props.className}>
+			{props.text && <p>{props.text}</p>}
+			{!props.text && Inject.getReactElement(props.getComponent, props.Content)}
+		</Body>
+		<Footer id={`${props.id}-footer`}>
+			{Inject.getReactElement(props.getComponent, props.Footer)}
+		</Footer>
+	</span>
+));
 
-	focusRichLayout() {
-		this.richLayout.current.focus();
-	}
-
-	render() {
-		return (
-			<span id={this.props.id} ref={this.richLayout} tabIndex="-1">
-				<Header id={`${this.props.id}-header`}>
-					{Inject.getReactElement(this.props.getComponent, this.props.Header)}
-				</Header>
-				<Body id={`${this.props.id}-body`} className={this.props.className}>
-					{this.props.text && <p>{this.props.text}</p>}
-					{!this.props.text && Inject.getReactElement(this.props.getComponent, this.props.Content)}
-				</Body>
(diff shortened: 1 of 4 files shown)


packages/components/src/RichTooltip/RichLayout/RichLayout.component.js
```diff
@@ -37,21 +37,32 @@ const Footer = ({ id, children, className }) => (
...
-			<Body id={`${props.id}-body`} className={props.className}>
-				{props.text && <p>{props.text}</p>}
-				{!props.text && Inject.getReactElement(props.getComponent, props.Content)}
-			</Body>
-			<Footer id={`${props.id}-footer`}>
-				{Inject.getReactElement(props.getComponent, props.Footer)}
-			</Footer>
-		</span>
-	);
+export default class RichLayout extends React.Component {
```
frassinier: why do you need a class here?

packages/components/src/RichTooltip/RichLayout/RichLayout.component.js
```diff
@@ -37,21 +37,32 @@ const Footer = ({ id, children, className }) => (
...
+		this.richLayout = React.createRef();
+	}
+
+	focusRichLayout() {
+		this.richLayout.current.focus();
+	}
+
+	render() {
+		return (
+			<span id={this.props.id} ref={this.richLayout} tabIndex="-1">
```
frassinier: It makes sense here to use FocusTrap I guess

frassinier: And do you know why a span was used here? I'm wondering?

packages/components/src/RichTooltip/RichLayout/RichLayout.component.js
jare-talend: https://fr.reactjs.org/docs/refs-and-the-dom.html#adding-a-ref-to-a-class-component.
https://fr.reactjs.org/docs/refs-and-the-dom.html#refs-and-function-components

My components where is set my RichLayout call a focusRichLayout when i lost the focus on the button.

packages/components/src/RichTooltip/RichLayout/RichLayout.component.js
jare-talend: Legacy I think to wrap the element.

packages/components/src/RichTooltip/RichLayout/RichLayout.component.js
frassinier: Thank you for the documentation but my comment was more about "why do you change the functional component in favor of a class-based component?"

packages/components/src/RichTooltip/RichLayout/RichLayout.component.js
frassinier: Could you profit of your PR to change it to a more convenient div, please?

packages/components/src/RichTooltip/RichLayout/RichLayout.component.js
jare-talend: If i want to access to the method instance, i need a class based component instead a function component.

packages/components/src/RichTooltip/RichLayout/RichLayout.component.js
jare-talend: done

romainseb:
Same as @jmfrancois could the focusManager make the job here ?
https://github.com/Talend/ui/blob/master/packages/components/src/FocusManager/FocusManager.component.js

jare-talend:
> Same as @jmfrancois could the focusManager make the job here ?
> https://github.com/Talend/ui/blob/master/packages/components/src/FocusManager/FocusManager.component.js

Like react-focus-lock, i need to force the focus on a element, Nope, I try and we lost the focus. We should force the focus on one element.

PR #2577, opened 2019-12-12
feat(Components): Dynamic enumeration height

nicomaligne:
The en enumration's height is by default 40vh.
This causes the list to grow to 40vh when the first item is added, and in some cases that's not very desirable.
We need to calculate dynamically the size of the list, so we are doing that in js, applying the height with inline style.
You need to pass a calculateListHeight props in itemsProp object, it's a function which take the list as first parameters.


packages/components/src/Enumeration/Items/Items.component.js
```diff
@@ -104,11 +104,16 @@ class Items extends React.PureComponent {
 	}
 
 	render() {
+		const calculateListHeight = get(this.props, 'itemsProp.calculateListHeight');
 		const actions = this.props.itemsProp && this.props.itemsProp.actionsDefault;
 		return (
 			<div
+				test-id={'enumeration-items-list'}
```
jsomsanith-tlnd: What is this hard coded `test-id` ?

packages/components/src/Enumeration/Items/Items.component.js
```diff
@@ -104,11 +104,16 @@ class Items extends React.PureComponent {
...
 				className={classNames(theme['tc-enumeration-items'], 'tc-enumeration-items')}
 				onScroll={this.scrollEnumeration}
+				style={{
+					height: calculateListHeight ? `${calculateListHeight(this.props.items)}px` : {},
+				}}
```
jsomsanith-tlnd:
```
style={ calculateListHeight ? { height: `${calculateListHeight(this.props.items)}px` } : {} }
```
This avoid having `style = {{ height: {} }}`

packages/components/src/Enumeration/Items/Items.scss
```diff
@@ -8,6 +8,7 @@ $tc-enumeration-height: 40vh !default;
 	list-style: none;
 	box-shadow: inset 0 21px 3px -20px $shadow, inset 0 -21px 3px -20px $shadow;
 	height: $tc-enumeration-height;
+	max-height: $tc-enumeration-height;
```
jsomsanith-tlnd: Since you set the max-height, is the height still relevant ?

packages/components/src/Enumeration/Items/Items.component.js
nicomaligne: It'is a practice to target easily element in TU. It's a comon practice used in TU testing.

packages/components/src/Enumeration/Items/Items.scss
nicomaligne: If I remove the height it will be a braking for the existing component.
And in the case of dynamic height, we have the max height to avoid unlimited list.

PR #2586, opened 2019-12-18
feat(components): review about modal style

romainseb:
Update the style
Can't find another way to get the fixed header on the services table, i'm open to suggestions


packages/components/src/AboutDialog/AboutDialog.scss
```diff
@@ -21,9 +21,24 @@
 				margin: 0 auto;
 				margin-top: 3rem;
 				text-align: justify;
+				table-layout: fixed;
 
-				th {
-					text-transform: uppercase;
+				th,
+				td {
+					padding: 5px;
```
frassinier: there is a var to achieve that

packages/components/src/AboutDialog/AboutDialogTable.component.js
```diff
@@ -0,0 +1,98 @@
...
+ */
+function getTbodyStyle(services) {
+	return {
+		height: Math.min(maxHeight, services.length * serviceHeight),
+	};
+}
+
+export function AboutDialogTable({
+	t,
+	definition = [
```
frassinier: It deserves to use a composition pattern, isn't it?

romainseb: I've tried to use composition on this dialog but i haven't find a elegant solution to make it work as the composition has to be done for the header and the body, there is a lot of elements to put. Not sure it works well, but do you have something in mind ? :)

output/components.eslint.txt
```diff
@@ -2,6 +2,9 @@
 
 The react/require-extension rule is deprecated. Please use the import/extensions rule from eslint-plugin-import instead.
 
+/home/travis/build/Talend/ui/packages/components/src/AboutDialog/AboutDialogTable.component.js
+  79:5  error  't' PropType is defined but prop is never used  react/no-unused-prop-types
```
jsomsanith-tlnd: Please fix that

packages/components/src/AboutDialog/AboutDialogTable.component.js
```diff
@@ -0,0 +1,80 @@
...
+import classNames from 'classnames';
+
+import { getI18nInstance } from '../translate';
+import Skeleton from '../Skeleton';
+
+import theme from './AboutDialog.scss';
+
+const i18n = getI18nInstance();
+
+export const getColumnHeaders = () => ({
```
jsomsanith-tlnd: This is used only once with `Object.values(getColumnHeaders())` --> can you just make it a list ?

frassinier: No because of the need to spread [here](https://github.com/Talend/ui/pull/2586/files#diff-150395fd55bae16ae173a484d3dfa842R49)

PR #2607, opened 2020-01-03
feat(ARCH-266): codemode import lib + apply

jmfrancois:

Current project are not friendly with treeshaking because most of components import are done from the root index. Which was nice and easy at the beging but now a little project import of UI components and you get 4.7MB of JS in your bundle.

We want to be treeshaking friendly to support dynamic import project side.
Doing it by hand can be lots of boring work so lets do it using a codemode.

But wait there is global import here and there in UI too. so it wont work even if you do it well in your project.

Lets fix it.

NOTE: No code addition in this PR except the codemode so do not expect new tests.


jmfrancois:
This has been published under the @talend/babel-plugin-transform-ui-imports.
There is now also the opposite plugin (do /lib/XX => root) @talend/babel-plugin-import-from-index

PR #2613, opened 2020-01-08
feat(component/grid): allow to disable drag n drop

ncomont:

It's not possible to disable drag n drop in the grid

Allow to disable drag n drop with a new `isDraggable` prop.

### DIFF
--- a/packages/components/src/GridLayout/Grid.component.js
+++ b/packages/components/src/GridLayout/Grid.component.js
@@ -6,7 +6,12 @@ import classNames from 'classnames';
 
 import Tile from './Tile';
 import { SKELETON_TILE_CONF } from './Tile/Skeleton/SkeletonTile.component';
+
 import css from './Grid.scss';
+import { getTheme } from '../theme';
+
+const theme = getTheme(css);
+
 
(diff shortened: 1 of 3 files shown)


packages/components/src/GridLayout/Grid.component.js
```diff
@@ -40,7 +43,9 @@ function Grid({
 }) {
 	return (
 		<ResponsiveGridLayout
-			className="layout"
+			className={classNames(css.layout, 'layout', {
```
romainseb: Just saying
- css.layout seems to do nothing as the class is not defined in the scss file
- You can use getTheme from the theme.js file to get rid of the double declaration (and the import of classnames here)

frassinier: > css.layout seems to do nothing as the class is not defined in the scss file

Yes, it does. 
https://github.com/Talend/ui/pull/2613/files#diff-78b74fcfd7d29cbea99b03ac9503e68fR1

frassinier: Perhaps a demo could be added @ncomont 😉

PR #2621, opened 2020-01-17
feat(cmf): customize sentry initialization

jare-talend:
We can't customize the sentry init and we can't call configureScope in sentry.

Customize the Sentry initialization for the hosted APP.
add a function configureSentryScope to customize the data send from the APP to Sentry. By this way, set this function on UI, we use the same Sentry instance.


packages/cmf/src/onError.js
```diff
@@ -71,7 +71,7 @@ function hasReportFeature() {
  * @param {Error} error instance of Error
  */
 function report(error, options = {}) {
-	if (ref.SENTRY_DSN) {
+	if (ref.SENTRY_DSN || getCurrentHub().getClient() !== undefined) {
```
jmfrancois: there are many places where we test ref.SENTRY_DSN to consider the sentry setup.
Not sure what is the best but here the fix is imcomplete. 
For example hasReportFeature will report false but sentry is available... so the Error Component will display wrong info.

packages/cmf/src/onError.js
```diff
@@ -139,13 +139,13 @@ function onJSError(event) {
  * init Sentry lib
  * @return {[type]} [description]
  */
-function setupSentry() {
+function setupSentry(option = {}) {
```
sgendre:
```suggestion
function setupSentry(options = {}) {
```

packages/cmf/src/onError.js
```diff
@@ -139,13 +139,13 @@ function onJSError(event) {
...
 	if (!ref.SENTRY_DSN) {
 		return;
 	}
 	window.removeEventListener('error', onJSError);
 	try {
-		init({ dsn: ref.SENTRY_DSN });
+		init({ dsn: ref.SENTRY_DSN, ...option });
```
sgendre:
```suggestion
		init({ dsn: ref.SENTRY_DSN, ...options });
```

packages/cmf/src/onError.md
```diff
@@ -65,15 +65,15 @@ in your own components in some key places.
 import { ErrorBoundary } from '@talend/react-cmf';
 
 function DoComplexComponent(props) {
-    // ... do some stuff
+	// ... do some stuff
```
sgendre: 4 spaces vs 8 spaces ?

jmfrancois: not it was tab and now it space, don t know why ... may be a prettier which do not understand this is javascript :trollface:

PR #2622, opened 2020-01-17
fix(forms/TDC-3368): trigger error JS when the component will unmount

jare-talend:

error JS happens when the component is umount. Cmfconnect in a parent component call deleteState, a child trigger componentWillReceiveProps but the state is empty and triggers an error JS

fix the error JS with a test if the property exists


packages/containers/src/ComponentForm/ComponentForm.component.js
```diff
@@ -88,7 +88,12 @@ export class TCompForm extends React.Component {
 
 	componentWillReceiveProps(nextProps) {
 		if (this.props.state.get('properties') !== nextProps.state.get('properties')) {
```
frassinier: I would like to suggest this instead 
```suggestion
		const nextProperties = nextProps.state.get('properties', Map());
		if (this.props.state.get('properties') !== nextProperties) {
			this.setState({ properties: nextProperties.toJS() });
		}
```

PR #2623, opened 2020-01-17
fix(datagrid/TDC-3343): remove the scroll on the pin column

jare-talend:

The latest version of ag-grid set always the overflow-x to scroll.

hide the scroll


packages/datagrid/src/components/DataGrid/DataGrid.scss
```diff
@@ -67,6 +67,10 @@ $td-datagrid-skeleton-height: 9.6rem !default;
 		font-family: inherit;
 	}
 
+	:global(.ag-horizontal-left-spacer) {
+		overflow-x: hidden;
```
frassinier: Why is not `auto`?

jare-talend: it is only the pin column, the pin column never overflows

PR #2646, opened 2020-02-10
feat(DateTimeRangePicker): ability to set default time value

vyshkov:

ability to set default time value user selects some date. So for example if i chose 12.12.2019 date in calendar, there should be possible to set 23:59 time value without user actions

added ability to pass props for the default time


packages/components/src/DateTimePickers/InputDateTimePicker/InputDateTimePicker.component.js
```diff
@@ -69,6 +70,17 @@ InputDateTimePicker.propTypes = {
...
+		PropTypes.shape({
+			hours: PropTypes.string,
+			minutes: PropTypes.string,
+			seconds: PropTypes.string,
+		}),
+		PropTypes.shape({
+			hours: PropTypes.string,
+			minutes: PropTypes.string,
+		}),
+	]),
```
jsomsanith-tlnd: You don't need to redefine that to set seconds as optional. By default, a prop is optional, unless you set `.isRequired`.
Define `defaultTimeValue ` it once as non required. And set the hours and minutes as required

PR #2649, opened 2020-02-11
feat(SubHeaderBar): add label display

tlnd-mhuchet:
We need to be able to use a label style to display the subtitle in the subheader.
See https://jira.talendforge.org/browse/TFD-3720 for details.

Add a possibility to customise the subtitle of the subheader bar.


tlnd-mhuchet:
> Could you add some stories, please ? Ideally, I would prefer two separate PRs, but I'll let other judge of the necessity of that

I have added two stories, one on SubHeaderBar component ("with label subtitle") and one on VirtualizedList component ("List > Table with label author").

I can split this into 2 distinct PR if it is needed. It is right that 2 different component are modified but it is for the same purpose: be able to use label style. So as you want.

tlnd-mhuchet:
> as -> has :)

It is to say "render it as a label", not to say "render with a label".

packages/components/src/SubHeaderBar/TitleSubHeader/TitleSubHeader.component.js
```diff
@@ -109,15 +109,22 @@ function SubTitle({ subTitleLoading, subTitle }) {
...
+		if (asLabel) {
+			subtitleClassnames = classNames('label', 'tc-subheader-details-text-subtitle-label', {
+				'label-info': labelType === 'INFO',
+			});
+		} else {
+			subtitleClassnames = classNames(
+				theme['tc-subheader-details-text-subtitle'],
+				'tc-subheader-details-text-subtitle',
+			);
+		}
```
romainseb: How do this part could render with conditional properties using classnames api instead of having the asLabel if ?

tlnd-mhuchet: I tried, but it was not really sexy to have all classes as conditionnal properties with `!asLabel` or `asLabel`. "label" classes are not an addition to the original style, they replace them.

If you see a sexy way to write with classname, i'll be glad to use it :)

packages/components/src/SubHeaderBar/TitleSubHeader/TitleSubHeader.component.js
```diff
@@ -109,15 +109,22 @@ function SubTitle({ subTitleLoading, subTitle }) {
 	}
 
 	if (subTitle) {
+		let subtitleClassnames;
+		if (asLabel) {
+			subtitleClassnames = classNames('label', 'tc-subheader-details-text-subtitle-label', {
```
ncomont: Please use css module class

packages/components/src/SubHeaderBar/TitleSubHeader/TitleSubHeader.component.js
```diff
@@ -109,15 +109,22 @@ function SubTitle({ subTitleLoading, subTitle }) {
...
-			<small
-				className={classNames(
-					theme['tc-subheader-details-text-subtitle'],
-					'tc-subheader-details-text-subtitle',
-				)}
-			>
-				{subTitle}
-			</small>
+			<div>
+				<small className={subtitleClassnames}>{subTitle}</small>
```
ncomont: Why don't you use `Label` from `react-bootstrap` as it is done in `CollapsiblePanel`, for example ?

image

packages/components/src/SubHeaderBar/TitleSubHeader/TitleSubHeader.component.js
tlnd-mhuchet: I used `getTheme` to simplify

packages/components/src/SubHeaderBar/TitleSubHeader/TitleSubHeader.component.js
tlnd-mhuchet: This part has been rewritten

PR #2652, opened 2020-02-11
feat(VirtualizedList): add CellLabel type

tlnd-mhuchet:
We need to be able to use a label style to display data in a list.
See https://jira.talendforge.org/browse/TFD-3720 for details.

Add a new cell type (label) to the list component.


packages/components/src/VirtualizedList/CellLabel/CellLabel.test.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+
+import CellLabel from './CellLabel.component';
+
+describe('CellLabel', () => {
+	it('should render label', () => {
```
romainseb: WDYT to add a UT with a mount and have a snapshot of the HTML ?

tlnd-mhuchet: The test was here to ensure that we really have the right classes, because they do 100% of the job.

PR #2663, opened 2020-02-20
label should not be required for link Action

jillyan:
Un-correct warning. It shows warning for a link Action, said label is undefined.
Though a 'hideLabel' could solve the warning, but with a 'label' and 'hideLabel', it will show a default tooltip, and it is not right for the ones who uses TooltipTrigger .

Remove required for label in ActionButton ProtoTypes

Though add a hideLabel={true} would hide the label, but it will shows a tooltip which is not good to the one who has a overlayComponent


jsomsanith-tlnd:
1.5 year old, I guess requirements could have changed, if it's really needed, let's see that with Design System team

PR #2667, opened 2020-02-25
fix(form/Datalist): Datalist display value in text mode when using groups

yyanwang:
In text mode, Datalist will display value (instead of name) of an option when there's groups in options.
image

Datalist should always display name of an option in text mode.


packages/forms/src/UIForm/fields/Datalist/displayMode/TextMode.component.js
```diff
@@ -34,9 +34,16 @@ class TextMode extends React.Component {
...
 		const titleMap = this.state.titleMap || this.props.schema.titleMap || [];
-		const titleEntry = titleMap.find(entry => entry.value === value);
+		let titleEntry = titleMap.find(entry => entry.value === value);
+
+		if (options && options.isMultiSection) {
+			options.titleMap.find(section => {
+				titleEntry = section.suggestions.find(entry => entry.value === value);
+				return !!titleEntry;
+			});
+		}
```
jsomsanith-tlnd: Can you wrap the non-multi-section part in an else please ?
There is no need to search the label in titleMap, then in multi-section mode, search again in option.titleMap

packages/forms/src/UIForm/fields/Datalist/displayMode/TextMode.component.js
```diff
@@ -36,13 +36,15 @@ class TextMode extends React.Component {
 		const { id, schema, value, t } = this.props;
 		const { title, options } = schema;
 		const titleMap = this.state.titleMap || this.props.schema.titleMap || [];
```
jsomsanith-tlnd: You can move this line into the `else` block too

packages/forms/src/UIForm/fields/Datalist/displayMode/TextMode.component.js
```diff
@@ -34,9 +34,18 @@ class TextMode extends React.Component {
...
+		const { title, options } = schema;
+		let titleEntry;
+
+		if (options && options.isMultiSection) {
+			options.titleMap.find(section => {
+				titleEntry = section.suggestions.find(entry => entry.value === value);
+				return !!titleEntry;
+			});
+		} else {
+			const titleMap = this.state.titleMap || this.props.schema.titleMap || [];
```
smouillour: use lodash get at least for this.props.schema.titleMap

PR #2670, opened 2020-02-26
fix(SidePanel): fit content width

jsomsanith-tlnd:
When an entry in side panel is too long, the text is troncated.
image

It worked before, but introducing simplebar as lib to simulate a scrollbar in side panel changed the behavior: the content is wrapped in a `simplebar` div in `absolute` position. It means that the content is out of the elements flow. The side panel has now its min-width 200px.

There is no css only solution with this kind of library. So here we handle it with javascript.
* make the ActionList fit its content. The SidePanel uses ActionList component.
* on undock change, we get the ActionList element real size, et set this width to the SidePanel.

Bonus
* introduce the SidePanel uncontrolled mode 🎉. It will use an internal state if there is no `props.onToggleDock`


packages/components/src/ActionList/ActionList.scss
```diff
@@ -3,11 +3,14 @@
...
 
-$tc-action-list-item-width: 200px !default;
-$tc-action-list-item-icons-size: 20px !default;
-$tc-action-list-item-border-size: 2px !default;
+$tc-action-list-item-width: 20rem !default;
+$tc-action-list-item-icons-size: 2rem !default;
+$tc-action-list-item-border-size: 0.2rem !default;
 
 .tc-action-list {
+	width: fit-content;
```
vbocquet: I know it will not be pleasant, but is it ok with IE11? I know that `fit-content` is not supported [https://caniuse.com/#search=fit-content](https://caniuse.com/#search=fit-content)
I try to check in it with Browserstack but Storybook was not ok with it.

packages/components/src/SidePanel/SidePanel.component.js
```diff
@@ -39,14 +43,36 @@ function SidePanel({
...
+	const [dockState, setDockState] = useState(dockedProp);
+	const docked = onToggleDock ? dockedProp : dockState;
+	const [width, setWidth] = useState(DEFAULT_MIN_WIDTH);
+	const ref = React.createRef();
+
+	useEffect(() => {
+		if (docked || minimised) {
+			setWidth(large ? LARGE_DOCKED_MIN_WIDTH : DOCKED_MIN_WIDTH);
+		} else if (ref) {
+			const actionList = ref.current.querySelector('.tc-action-list');
```
vbocquet: After re-check, `actionList.offsetWidth` will throw an error if we don't have any actions: `.tc-action-list` will not be in the DOM. What do you think?

PR #2674, opened 2020-03-02
fix(form/language): Custom error message bleeds over every form

yyanwang:
https://jira.talendforge.org/browse/TMC-19984
Custom validation message passed to one form will be applied to all the forms.
When using the `language` prop of forms to override error message, the overridden message(s) will then be used by the other forms. Example :
Form A : 
```
<Form {...props}
        language={
            { STRING_PATTERN: 'My error message when regexp pattern is not matched' }
        }
/>
```

Form B : 
```
<Form {...props} />
```

Once Form A has been rendered once, STRING_PATTERN overridden message will be applied to Form B if a field has a regexp pattern check that failed.

Every form instance will have its own language map identified by id


packages/forms/src/UIForm/UIForm.component.js
```diff
@@ -40,15 +41,16 @@ export class UIFormComponent extends React.Component {
 		this.setFormRef = this.setFormRef.bind(this);
 		// control the tv4 language here.
 		const language = getLanguage(props.t);
+		const languageId = props.languageId || props.id || uuid.v4();
```
jsomsanith-tlnd: A little suggestion: 
- let's not introduce a new prop since we can use the id to identify the the form, and isolate the language props.
- let's not generate a random id. Use the old hard coded `@talend` is sufficient
```suggestion
		const languageId =props.id || '@talend';
```

yyanwang: Hi @jsomsanith , thanks for review!
Just one word about your second point, that to use `@talend` if no id is set.
`id` is not mandatory for form. Imagine we have 2 forms, neither of them has id, then custom validation message passed to one form will be used by the other one still.

jsomsanith-tlnd: That's a little on purpose :D 
So it enforce the fact that we need an id, and it's best practice. But I agree that a hidden thing like that is unclear.
If you want isolation by default, let's generate an id like your did with uuid

yyanwang: OK. I see, so it's intended, to reuse the custom messages. then I will use '@talend' by default.

PR #2691, opened 2020-03-09
feat(TCCUI-130): folder tree view without icons

ybaskaran:
Currently the folder tree view component show the talend-folder as the default icon if the icon is not specified.
In some cases, we need to display folder and sub-folders without any icons.
https://jira.talendforge.org/browse/TCCUI-130
Ref:
https://jira.talendforge.org/browse/TMC-16719

Do not render Icon if specified with "icon: none".
http://2691.talend.surge.sh/components/?path=/story/data-tree-foldertreeview--without-icons
Any item in the tree structure can be displayed without icon by specifying the value 'none' for 'icon' property. 
{ 
...,
icon: 'none' 
}

### DIFF
--- a/packages/components/src/TreeView/FolderTreeView.stories.js
+++ b/packages/components/src/TreeView/FolderTreeView.stories.js
@@ -85,33 +85,33 @@ const structureWithoutIcons = [
 		name: 'hitmonlee',
 		children: [{ name: 'Hitmonchan' }],
 		isOpened: false,
-		icon: 'false',
+		icon: false,
 	},
 	{
 		name: 'pikachu',
 		children: [
 			{
 				name: 'raichu',
-				icon: 'false',
+				icon: false,
 			},
 		],
 		isOpened: true,
-		icon: 'false',
+		icon: false,
 	},
 	{
 		id: 'selected',
 		name: 'Abra',
-		icon: 'false',
+		icon: false,
 		isOpened: true,
 		children: [
 			{
 				name: 'Kadabra',
-				icon: 'false',
+				icon: false,
 				isOpened: true,
 				children: [
 					{
 						name: 'Alakazam',
-						icon: 'false',
+						icon: false,
 					},
 				],
 			},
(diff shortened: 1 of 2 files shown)


packages/components/src/TreeView/FolderTreeView.stories.js
```diff
@@ -80,6 +80,45 @@ const structureWithIcons = [
 	},
 ];
 
+const structureWithoutIcons = [
+	{
+		name: 'hitmonlee',
+		children: [{ name: 'Hitmonchan' }],
+		isOpened: false,
+		icon: 'none',
```
vbocquet: Can we make the icon key optional instead of an icon 'none'? We can check that key exist here: https://github.com/Talend/ui/blob/master/packages/components/src/TreeView/TreeViewItem/TreeViewItem.component.js#L271, it will also avoid instantiation of `<TreeViewIcon />` with a null return.

It can be tricky, if tomorrow an icon "none" exist.

ybaskaran: hi, it is already optional and existing applications use this TreeView without the property icon and we are using the default talend-folder icon if the property doesn't exist.
https://github.com/Talend/ui/blob/master/packages/components/src/TreeView/TreeViewItem/TreeViewItem.component.js#L23

vbocquet: Maybe i didn't understand, but why we can't do that here: https://github.com/Talend/ui/blob/master/packages/components/src/TreeView/TreeViewItem/TreeViewItem.component.js#L271 : 
```
{icon && <TreeViewIcon key="icon" icon={icon} isOpened={showOpenedFolder} />}
```

vbocquet: See in MP with @ybaskaran , my solution will break the existing cases without icon (=> folders).

jsomsanith-tlnd: In fact `props.icon` is optional. But there is a default icon which is the folder.
@vbocquet is right, passing a string in the icon prop can be annoying if we have an icon with this name.

Let's change this prop to accept string of boolean instead. If the prop is `false` then you can skip the icon rendering. Else if a string is provided, we render the adequat icon, and if not provided we render the default one.

PR #2693, opened 2020-03-11
fix(components/sidepanel): sidepanel docked initial width

romainseb:
We have some glitch using the overview and the sidepanel as the events used by react-grid-layout are only on the window resize, https://github.com/STRML/react-grid-layout/blob/master/lib/components/WidthProvider.jsx#L65
Apply a default value when the sidepanel is docked (as the labels are not visible)


packages/components/src/SidePanel/SidePanel.component.js
```diff
@@ -52,7 +61,7 @@ function SidePanel({
 }) {
 	const [dockState, setDockState] = useState(dockedProp);
 	const docked = onToggleDock ? dockedProp : dockState;
-	const [width, setWidth] = useState();
+	const [width, setWidth] = useState(getInitialWidth(docked, large));
```
jsomsanith-tlnd:
```suggestion
	const [width, setWidth] = useState(() => getInitialWidth(docked, large));
```

PR #2697, opened 2020-03-19
fix: eslint issue

romainseb:
We got an issue with eslint
image

remove the concerned part


.eslintrc
```diff
@@ -1,7 +1,6 @@
 {
 	"extends": "airbnb",
 	"parser": "babel-eslint",
-	"ignorePatterns": ["**/*.stories.js"],
```
tlnd-mhuchet: I think it is useful to avoid all lint errors from stories.

It seems we cannot use ignorePatterns when extending another config, see https://eslint.org/docs/user-guide/configuring#how-it-works

Should try .eslintignore maybe ?

PR #2699, opened 2020-03-20
poc(cmf): integrate LaunchDarkly

romainseb:
We don't have a feature flip solution.
As we are going to use LD, this PR introduce another bootstrap that is async (in order to use the async function of LD initialisation). If the LDClientSideID is passed to the bootstrap method, it will init LD and wrap it into the provider.

I'm not sure it's the place i want to it as this is cmf which is kinda free from external solutions but i'm not sure with the current implementation i could put it elsewhere. Feel free to discuss :)


romainseb:
This content is too talend specific. I think it's not the good place for it.

PR #2706, opened 2020-03-23
feat(cmf): handle blob type

jare-talend:

saga http don't handle the blob type


jare-talend:
> The code is ok, but there is no test.
> Please add an assert on the promise results.

Blob isn't handle by node. It is the best i can do

PR #2710, opened 2020-03-25
fix(components): cellDateTime when null is given

romainseb:
When null or undefined value is passed to CellDateTime, it fallback on 1970
image

Fallback on nothing when the data is not available
image


packages/components/src/VirtualizedList/CellDatetime/CellDatetime.component.js
```diff
@@ -13,6 +13,9 @@ import TooltipTrigger from '../../TooltipTrigger';
 const DATE_TIME_FORMAT = 'YYYY-MM-DD HH:mm:ss';
 
 export function computeValue(cellData, columnData, t) {
+	if (!cellData) {
+		return t('NOT_AVAILABLE', { defaultValue: 'Not available' });
```
jsomsanith-tlnd: I'm wondering, instead of setting a string in the cell, why not letting the cell empty ?

PR #2732, opened 2020-04-08
feat(TDC-3798): Implement containsIgnoreCase operator

vbocquet:
In faceted-search package, contains operator is used to check contains in string comparaison. This is great but this TQL operator is case sensitive, that may be annoying for end-user.

To avoid this, this PR implement the `containsIgnoreCase` in operator faceted search dictionary. It have the same label / icon that `contains` operator because at the end, we want to only serve one of these in app.

### DIFF
--- a/packages/faceted-search/src/dictionary/operator.dictionary.js
+++ b/packages/faceted-search/src/dictionary/operator.dictionary.js
@@ -1,5 +1,6 @@
 const operatorNames = {
 	contains: 'contains',
+	containsIgnoreCase: 'containsIgnoreCase',
 	equals: 'equals',
 	notEquals: 'notEquals',
 	in: 'in',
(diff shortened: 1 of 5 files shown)


packages/faceted-search/src/dictionary/operator.dictionary.js
```diff
@@ -31,6 +32,13 @@ const standardOperators = t => ({
 		name: 'contains',
 		iconName: 'contains',
 	},
+	[operatorNames.containsIgnoreCase]: {
+		label: t('OPERATOR_CONTAINS_LABEL', {
```
eguillossou, line 36: why keep same label than for contains?
Newbie question 🔰

vbocquet: Checked with PM: "For the user, it's the same", so no need to have a different i18N key 👍

eguillossou: Ok thanks 🥇

PR #2737, opened 2020-04-10
fix(AppGuidedTour):  Import content default value

gjou-tlnd:

Demo content should be imported by default on first time use only


packages/components/src/AppGuidedTour/AppGuidedTour.component.js
```diff
@@ -13,19 +14,28 @@ function AppGuidedTour({
 	isOpen,
 	appName,
 	steps,
+	localStorageKey,
```
romainseb, line 17: Could we give a default value if we don't want to redefine it ?

PR #2742, opened 2020-04-12
feat(cmf): async modules

romainseb:
With a current integration, one need would be to initialise a cmf module with an async function to create a provider to wrap the app.
It could be severals ways to do it, here, the main idea is to have a function called `asyncInit` that could be defined in a module like that :

```javascript
const asyncModule = {
	id: 'sentryModule',
	init: async () => {
		const configAnswer = await fetch(API.config);
		const config = await configAnswer.json();
		const sentryToken = config.sentryToken;

		const Component = ({ children, token = sentryToken }) => (
			<SentryProvider token={token}>
				{children}
			</SentryProvider>
		);

		return {
			RootComponent: Component,
		};
	},
};
```

- _Why using an `asyncInit` function instead of just having a async function to return the module ?_
- Glad you asked ! Here the main benefit would be to give the possibility to have to same async module declared more that one time in the apps / libraries and be initialised just one time. It could be really better if the module has to make some http call to init itself.

Anyway, i'll create a PR with the other way to do it with just an async function handle and we'll see which one we'll keep !


packages/cmf/src/cmfModule.js
```diff
@@ -1,21 +1,31 @@
...
-			if (acc.some(({ id }) => current.id === id)) {
-				console.warn(
-					`cmf.bootstrap: 2 modules have the same id ${current.id}. This duplicated module will be skipped.`,
-				);
-			} else {
-				acc.push(current);
-				find(current, buff);
+
+			// prevent module duplication
+			if (!buff.some(({ id }) => current.id === id)) {
```
jsomsanith-tlnd, line 23: I think we should let the console.warn, because module duplication is usually a mistake. If we don't warn the user, we don't offer a chance to fix it.

jsomsanith-tlnd:
run tests locally, skip ci

PR #2747, opened 2020-04-15
feat(Typeahead): opt-in value as data-feature

sgendre:
The list items are sometimes driven by backend payload which does not provide the data-feature property whereas it is interesting to track them
Provide a flag to fallback data-feature property on the item value

### DIFF
--- a/packages/components/src/Typeahead/Typeahead.component.js
+++ b/packages/components/src/Typeahead/Typeahead.component.js
@@ -15,11 +15,9 @@ import {
 import { Action } from '../Actions';
 import I18N_DOMAIN_COMPONENTS from '../constants';
 
-const emptyArray = [];
-
 function getItems(items, enableValueAsDataFeature) {
 	if (!items) {
-		return emptyArray;
+		return [];
 	}
 	if (!enableValueAsDataFeature) {
 		return items;
--- a/packages/components/src/Typeahead/Typeahead.component.js
+++ b/packages/components/src/Typeahead/Typeahead.component.js
@@ -15,7 +15,7 @@ import {
 import { Action } from '../Actions';
 import I18N_DOMAIN_COMPONENTS from '../constants';
 
-function getItems(items, enableValueAsDataFeature) {
+export function getItems(items, enableValueAsDataFeature) {
 	if (!items) {
 		return [];
 	}
--- a/packages/components/src/Typeahead/Typeahead.component.js
+++ b/packages/components/src/Typeahead/Typeahead.component.js
@@ -24,7 +24,7 @@ export function getItems(items, dataFeature) {
 	}
 
 	return items.map(item => ({
-		'data-feature': `${dataFeature}-${item.value}`,
+		'data-feature': `${dataFeature}.${item.value}`,
 		...item,
 	}));
 }
--- a/packages/components/src/Typeahead/Typeahead.component.js
+++ b/packages/components/src/Typeahead/Typeahead.component.js
@@ -15,7 +15,7 @@ import {
 import { Action } from '../Actions';
 import I18N_DOMAIN_COMPONENTS from '../constants';
 
-export function getItems(items, dataFeature) {
+function getItems(items, dataFeature) {
 	if (!items) {
 		return [];
 	}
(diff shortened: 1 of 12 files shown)


packages/components/src/Typeahead/Typeahead.component.renderers.js
```diff
@@ -258,7 +258,7 @@ export function renderItem(item, { value, ...rest }) {
 				[theme.selected]: value === title,
 			})}
 			title={title}
-			data-feature={item['data-feature'] || rest['data-feature']}
+			data-feature={item['data-feature'] || rest['data-feature'] || item.value}
```
jsomsanith-tlnd, line 261: With that, every datalist (even the ones that we don't want to get pendo stats) will have a data-feature, with only the item value. This will pollute the pendo data.
It's better to let the dev explicitly set the data-feature. You already set the item api to set a data-feature by item with `item['data-feature']`. Why having this `item.value` ?

frassinier: I agree with @jsomsanith-tlnd, our tag values are not enough specific today to authorize this kind of fallback, you can easily mess the tracking up!

sgendre: for what I know, pendo tracking is opt-in so if the rule to target the click is not create on pendo platform, it won't appear

`item['data-feature']` need the backend to send and store that data which is related to tck, otherwise mapping the collection somewhere like Datalist.component ```items={
                        this.state.suggestions &&
                        this.state.suggestions.map(suggestion => ({
                            ...suggestion,
                            'data-feature': suggestion.value,
                        }))
                    }``` that will end up being available in pendo as current fallback solution but will need a realloc that will slow down apps

packages/components/src/Typeahead/Typeahead.component.js
```diff
@@ -15,6 +15,20 @@ import {
 import { Action } from '../Actions';
 import I18N_DOMAIN_COMPONENTS from '../constants';
 
+export function getItems(items, dataFeature) {
```
jsomsanith-tlnd, line 18: You should not export that for tests.
Render a typeahead (mount) and check the attributes

packages/forms/src/UIForm/fields/Datalist/Datalist.component.js
```diff
@@ -199,6 +200,8 @@ if (process.env.NODE_ENV !== 'production') {
 				}),
 			),
 			autoFocus: PropTypes.bool,
+
```
frassinier, line 203:
```suggestion

```

packages/components/src/Typeahead/Typeahead.test.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import { mount } from 'enzyme';
-import Typeahead from './Typeahead.component';
+import Typeahead, { getItems } from './Typeahead.component';
```
jsomsanith-tlnd, line 3:
```suggestion
import Typeahead from './Typeahead.component';
```

packages/components/src/Typeahead/Typeahead.test.js
```diff
@@ -205,4 +216,75 @@ describe('Typeahead', () => {
...
+				items: null,
+				multiSection: false,
+			};
+			const typeahead = <Typeahead {...props} />;
+
+			// when
+			const typeaheadInstance = mount(typeahead);
+
+			// then
+			expect(typeaheadInstance.find('Item')).toMatchObject({});
```
jsomsanith-tlnd, line 236:
```suggestion
			expect(typeaheadInstance.find('Item').length).toBe(0);
```

packages/components/src/Typeahead/Typeahead.test.js
```diff
@@ -205,4 +216,75 @@ describe('Typeahead', () => {
...
+				items: undefined,
+				multiSection: false,
+			};
+			const typeahead = <Typeahead {...props} />;
+
+			// when
+			const typeaheadInstance = mount(typeahead);
+
+			// then
+			expect(typeaheadInstance.find('Item')).toMatchObject({});
```
jsomsanith-tlnd, line 253:
```suggestion
			expect(typeaheadInstance.find('Item').length).toBe(0);
```

PR #2748, opened 2020-04-15
chore: security scan

jsomsanith-tlnd:
Yarn audit through our travis was changing at every scan, making the build to loop (ci restrat after new ci audit result push).
We deactivated it.

Replace yarn audit with sourceclear scan via GH action
- add a symlink to yarn.lock in each monorepo package
- cron to run it every sunday at midnight

Note that unfortunately, I cannot test the cron on my branch. It just doesn't schedule it. I run it with a push event, the result is available on veracode.
This cron schedule `should` run it. If not, it will be reworked next monday.


.github/actions/sourceclear/entrypoint.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/sh -l
+echo "Download srcclr ci script ..."
+curl -sSL https://download.sourceclear.com/ci.sh --output srcclr.sh
+echo "Script saved to srcclr.sh"
+chmod a+x srcclr.sh
+echo "Script execution rights added"
+
+# monorepo
+echo "Starting scan on . (root) ..."
```
tlnd-mhuchet, line 9: Is it not possible to make a loop on every subfolder of `./packages` to scan them? It could avoid to have to add each package one by one.

.github/actions/sourceclear/entrypoint.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/sh -l
+echo "Download srcclr ci script ..."
+curl -sSL https://download.sourceclear.com/ci.sh --output srcclr.sh
```
vbocquet, line 3: Not too dangerous to curl this?

.github/actions/sourceclear/entrypoint.sh
jsomsanith-tlnd: I can change that to have a list of folder to scan, yes, but not all the folders. There are some of them that we don't need to scan, like generator.

.github/actions/sourceclear/entrypoint.sh
```diff
@@ -0,0 +1,68 @@
...
+echo "Starting scan on ./packages/forms ..."
+./srcclr.sh scan ./packages/forms
+echo "Scan completed on ./packages/forms"
+
+echo "Starting scan on ./packages/icons ..."
+./srcclr.sh scan ./packages/icons
+echo "Scan completed on ./packages/icons"
+
+echo "Starting scan on ./packages/saga ..."
+./srcclr.sh scan ./packages/saga
```
vbocquet, line 46:
```suggestion
./srcclr.sh scan ./packages/sagas
```

.github/actions/sourceclear/entrypoint.sh
tlnd-mhuchet: Oh ok, i was thinking that we wanted to scan all of them. My comment was only about to avoid to have to maintain a list of packages 😅 
So you can keep it if you want :D, good for me :)

.github/actions/sourceclear/entrypoint.sh
frassinier: Why?

packages/cmf-cqrs/yarn.lock
```diff
@@ -0,0 +1 @@
+../../yarn.lock
```
frassinier, line 1: Weird?

packages/icons/pom.xml
```diff
@@ -1,240 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
frassinier, line 1: Is the purpose of the PR?

.github/actions/sourceclear/entrypoint.sh
vbocquet: Just about directly down a third party .sh to execute it, but  it's not really relevant here

.github/actions/sourceclear/entrypoint.sh
jsomsanith-tlnd: It will be easier to have a simple list instead of searching into the code. I'll do the change

packages/cmf-cqrs/yarn.lock
jsomsanith-tlnd: I created them with `ln -s ../../yarn.lock yarn.lock` ¯\_(ツ)_/¯

packages/icons/pom.xml
jsomsanith-tlnd: It's not used anymore, and with that, sourceclear detect that it's a maven project

.github/actions/sourceclear/entrypoint.sh
jsomsanith-tlnd: I took the example from sourceclear repo, and adapted it to our repo/need.
The curl is from here : https://github.com/srcclr/example-github-actions/blob/master/.github/actions/sourceclear/entrypoint.sh

With this, we download the official sourceclear script, and it's always up to date.

The only risk I see is if sourceclear get hacked and the file is replaced by something malicious.
So our solutions:
- go with this solution, the risk is limited and not really important since the repo is open sourced.
- download the current version of the script, that should not change a lot, and push it in our repo. And if it requires a change, we'll do a PR.

Either way I think it's acceptable.
Any preference @vbocquet @frassinier ?

vbocquet: I'm ok with you in you're first solution, the risk is limited 👍

vbocquet: And you're right, we have an open source repo

PR #2759, opened 2020-04-20
feat(faceted-search): Add some data-feature for Pendo / e2e tests

vbocquet:
For pendo and end-to-end tests usage, we miss some data-feature attribute to target specific features.

Add some data-feature tag based on this two JIRA:
- [https://jira.talendforge.org/browse/TDC-3820](https://jira.talendforge.org/browse/TDC-3820)
- [https://jira.talendforge.org/browse/TDC-3859](https://jira.talendforge.org/browse/TDC-3859)

### DIFF
--- a/packages/faceted-search/CHANGELOG.md
+++ b/packages/faceted-search/CHANGELOG.md
@@ -26,6 +26,15 @@ Types of changes
 
 ## [unreleased]
 
+### Breaking change
+Rename two data-feature attributes for harmonization:
+- `faceted-advanced-search.cancel` become `filter.advanced.clear`
+- `faceted-advanced-search.submit` become `filter.advanced.apply`
+
+### Added
+
+- [feat](https://github.com/Talend/ui/pull/2759): Add somes data-feature attributes for Pendo tracking / E2E tests
+
 ## [0.7.0]
 
 ### Added
--- a/packages/faceted-search/CHANGELOG.md
+++ b/packages/faceted-search/CHANGELOG.md
@@ -33,7 +33,7 @@ Rename two data-feature attributes for harmonization:
 
 ### Added
 
-- [feat](https://github.com/Talend/ui/pull/2759): Add somes data-feature attributes for Pendo tracking / E2E tests
+- [feat](#2759): Add somes data-feature attributes for Pendo tracking / E2E tests
 
 ## [0.7.0]
 
--- a/packages/faceted-search/CHANGELOG.md
+++ b/packages/faceted-search/CHANGELOG.md
@@ -33,7 +33,7 @@ Rename two data-feature attributes for harmonization:
 
 ### Added
 
-- [feat](#2759): Add somes data-feature attributes for Pendo tracking / E2E tests
+- [feat](https://github.com/Talend/ui/pull/2759): Add somes data-feature attributes for Pendo tracking / E2E tests
 
 ## [0.7.0]
 
--- a/packages/faceted-search/src/components/AdvancedSearch/AdvancedSearch.component.js
+++ b/packages/faceted-search/src/components/AdvancedSearch/AdvancedSearch.component.js
@@ -10,6 +10,7 @@ import { getTheme } from '@talend/react-components/lib/theme';
 import { useFacetedSearchContext } from '../context/facetedSearch.context';
 
 import theme from './AdvancedSearch.scss';
+import { PENDO_TAGS } from '../../constants';
 
 const css = getTheme(theme);
 
@@ -26,7 +27,6 @@ AdvancedSearchError.propTypes = {
 
 // eslint-disable-next-line import/prefer-default-export
 export function AdvancedSearch({
-	dataFeature = 'faceted-advanced-search',
 	initialQuery = '',
 	onCancel,
 	onChange,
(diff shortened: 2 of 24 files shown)


vbocquet:
@romainseb Thanks for checking, i understand your point of view but i'm not sure it's a good think to add complexity for this here: i can take times to maintain without any real benefits for end-user, but i'm not sure how it's works on others Talend products. Maybe with have this behavior everywhere

romainseb:
I just want to tell this is not the usual way to do in on @talend/ui but as it's for the faceted search that bundle some logic / behaviour and as it should only have one faceted search per page, i guess it's ok for now.

Just later, we'll need to add some prefix customisation. For now, let's keep this to avoid complexity.

packages/faceted-search/CHANGELOG.md
```diff
@@ -26,6 +26,15 @@ Types of changes
...
 ## [unreleased]
 
+### Breaking change
+Rename two data-feature attributes for harmonization:
+- `faceted-advanced-search.cancel` become `filter.advanced.clear`
+- `faceted-advanced-search.submit` become `filter.advanced.apply`
+
+### Added
+
+- [feat](https://github.com/Talend/ui/pull/2759): Add somes data-feature attributes for Pendo tracking / E2E tests
```
jsomsanith-tlnd, line 36: you can just write `#2759` instead of the full url and github does the link

packages/faceted-search/src/components/Badges/BadgeCheckboxes/BadgeCheckboxesForm.component.js
```diff
@@ -51,15 +52,11 @@ const getCheckboxes = (checkboxes, value, filterValue) => {
...
-			get(checkbox, 'label', '')
-				.toLocaleLowerCase()
-				.includes(formatFilterValue),
-		)
+		.filter(checkbox => get(checkbox, 'label', '').toLocaleLowerCase().includes(formatFilterValue))
 		.map(createCheckboxEntity(value));
 };
 
-const BadgeCheckboxesForm = ({ checkboxValues, id, onChange, onSubmit, value, t }) => {
+const BadgeCheckboxesForm = ({ checkboxValues, id, onChange, onSubmit, value, label, t }) => {
```
jsomsanith-tlnd, line 59: `label` seems confusing since it's not a displayed and translated label. It conflicts in my head with the button label for example. It seems to be only used to get the data-feature, perhaps something like `dataFeatureKey`?

packages/faceted-search/src/components/AdvancedSearch/AdvancedSearch.component.js
```diff
@@ -10,6 +10,7 @@ import { getTheme } from '@talend/react-components/lib/theme';
 import { useFacetedSearchContext } from '../context/facetedSearch.context';
 
 import theme from './AdvancedSearch.scss';
+import { PENDO_TAGS } from '../../constants';
```
frassinier, line 13: We should avoid using "PENDO" everywhere, this repo is public and we don't need to link Pendo to our repo at all.
USAGE_TRACKING_TAGS should be enough generic for example

packages/faceted-search/src/components/Badges/BadgeCheckboxes/BadgeCheckboxesForm.component.js
frassinier: or just "feature"

packages/faceted-search/src/components/AdvancedSearch/AdvancedSearch.component.js
vbocquet: Totally agree, good point. Just made the change

packages/faceted-search/src/components/Badges/BadgeCheckboxes/BadgeCheckboxesForm.component.js
vbocquet: Understand, it can be confusing, just fix

packages/faceted-search/CHANGELOG.md
vbocquet: Just test, `[feat](#2759) not work: it take this for an anchor in Readme.md.
Think it can be work if i just do a `#2759` but we want to keep a label like "feat", "fix" etc..

Anyway, #2759 will only works in a Github context, can be good to keep a full link to copy-paste from IDE.

packages/faceted-search/src/components/Badges/BadgeNumber/BadgeNumberForm.component.js
```diff
@@ -1,15 +1,17 @@
...
 import { getTheme } from '@talend/react-components/lib/theme';
 import RichLayout from '@talend/react-components/lib/RichTooltip/RichLayout';
+import { getApplyDataFeature } from '../../../helpers/usage.helpers';
 
 import cssModule from './BadgeNumber.scss';
 
 const theme = getTheme(cssModule);
 
-const BadgeNumberForm = ({ id, onChange, onSubmit, value, t }) => {
+const BadgeNumberForm = ({ id, onChange, onSubmit, value, label, t }) => {
```
jsomsanith-tlnd, line 13: Don't forget the other components for the rename `label` --> `feature`

PR #2764, opened 2020-04-20
fix(GuidedTour): add loading message

jare-talend:

The demo content can take a lot a time to importent the content.

Add a message to inform them that this step can take a lot a time

### DIFF
--- a/packages/components/src/AppGuidedTour/DemoContentStep.component.js
+++ b/packages/components/src/AppGuidedTour/DemoContentStep.component.js
@@ -18,7 +18,7 @@ export default function DemoContentStep({ demoContentSteps }) {
 				<Icon name="talend-info-circle" />
 				{i18n.t('tui-components:DEMO_CONTENT_LOADING_MESSAGE', {
 					defaultValue:
-						'Loading may take a few minutes to complete. Bear with us. Content is loading. Isn\'t it time for tea or coffee?',
+						"Loading may take a few minutes to complete. Bear with us. Content is loading. Isn't it time for tea or coffee?",
 				})}
 			</p>
 			<Stepper steps={demoContentSteps} />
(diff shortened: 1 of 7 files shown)


packages/components/src/AppGuidedTour/DemoContentStep.component.js
```diff
@@ -0,0 +1,30 @@
...
+
+export default function DemoContentStep({ demoContentSteps }) {
+	if (!demoContentSteps.length) {
+		return null;
+	}
+
+	return (
+		<React.Fragment>
+			<p className={theme.info}>
+				{i18n.t('tui-components:DEMO_CONTENT_LOADING_MESSAGE', {
```
vbocquet, line 18: I'm not sure about the `tui-components` namespace, don't think you need to declare it?

jare-talend: tui-components is the namespace to use for the components. How can i declare it ? The extract is correct for the component

romainseb: Do we tend to use useTranslation with the namespace in components rather than using the i18n shared instance ?
I think the main difference would be to be connected to the react lifecycle, so this implementation will work of course, it's more a manner of what we want to do.
In general, i'm using useTtranslation in react components and global i18n instance when i can't use the hook (as used in saga or reducer for instance)

Do we have some guidelines about this ? cc @jsomsanith-tlnd

vbocquet: Ok good, thx for the precision anyway

jare-talend: I replaced by useTranslation, it's better indeed

PR #2770, opened 2020-04-21
chore: lerna publish skip access check

jsomsanith-tlnd:
Lerna publish doesn't publish.
It checks if the current npm user has the rights. In our case, it requests https://registry.npmjs.org/-/org/talend-frontend/package?format=cli.

The result indicates that `talend-frontend` npm user has only `read` access to some packages, which is wrong. I checked the npm org configuration, and the user has write access to all the org package.
Furthermore, the single publish works.

I sent a message to the npmjs support.
In the meantime, let's disable the check.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "start-stepper": "cd packages/stepper && yarn start",
     "start-forms": "cd packages/forms && yarn start",
     "start-theme": "cd packages/theme && yarn start",
-    "publish": "lerna publish --force-publish=*",
+    "publish": "lerna publish --force-publish=* --no-verify-access",
     "storybook": "start-storybook -p 6006",
     "build-storybook": "build-storybook",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -28,7 +28,7 @@
     "start-stepper": "cd packages/stepper && yarn start",
     "start-forms": "cd packages/forms && yarn start",
     "start-theme": "cd packages/theme && yarn start",
-    "publish": "lerna publish --force-publish=*",
+    "publish": "lerna publish --force-publish=* --no-verify-access",
```
frassinier, line 31: Since it's temporary, we can't use this instead?
```bash
$> yarn publish -- --no-verify-access
```

jsomsanith-tlnd: Hey that's really clever :D 
I close the PR

jsomsanith-tlnd: I initiated the PR because I changed the lerna.json to disable that, but it didn't work, and I found the cli option that worked. But yes it doesn't worth a PR for that then

PR #2797, opened 2020-04-28
feat(forms): parameterize id concatenation separator

sgendre:
The separator is hard-coded to _ and it is annoying when the key property path contains values with _ within it.

The separator can be configured so a non-problematic value can be provided.

### DIFF
--- a/packages/forms/__mocks__/data.js
+++ b/packages/forms/__mocks__/data.js
@@ -45,12 +45,9 @@ export const nestedData = {
 		type: 'object',
 		title: 'Comment',
 		properties: {
-			lastname: {
+			content: {
 				type: 'string',
 				minLength: 4,
-			},
-			firstname: {
-				type: 'string',
 			},
 			timestamp: {
 				type: 'object',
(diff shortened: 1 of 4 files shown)


jsomsanith-tlnd:
Why is it `annoying when the key property path contains values with _ within it` ? What is the issue behind ?

packages/forms/__mocks__/data.js
```diff
@@ -40,6 +40,65 @@ export const data = {
 	errors: {},
 };
 
+export const nestedData = {
```
jsomsanith-tlnd, line 43: I don't think you need that complicated schema to test an id separator. Can you put the minimum amount of schema please ? The easier it is, the more it is maintainable.

sgendre: > Why is it `annoying when the key property path contains values with _ within it` ? What is the issue behind ?

if fails to split id and replace to work with a new wrapper in TPD

sgendre: > I don't think you need that complicated schema to test an id separator. Can you put the minimum amount of schema please ? The easier it is, the more it is maintainable.

I don't get what complexity you are talking about, it is very similar to the data var that is in place

jsomsanith-tlnd: You need data for your test. Your test only check the timestamp part. But the data is a schema describing jsonSchema and uiSchema of timestamp, firstname, and lastname.

>  it is very similar to the data var that is in place
That's not the point, I haven't seen how the other data are used, but that doesn't mean it's good to do the same here.

PR #2798, opened 2020-04-28
chore(datagrid): lint errors

jsomsanith-tlnd:
WARNING: this is to merge to jsomsanith/chore/talend_scripts because of eslint rules at root. It conflicts with talend/scripts config and breaks the lint. Let's move all packages to @talend/scripts before the merge to master.

Eslint errors in datagrid

Fix them.
There is still a lint error on accessibility that needs more work.

This solves the datagrid package for talend/scripts.


packages/datagrid/src/components/DataGrid/DataGrid.component.js
```diff
@@ -269,9 +270,7 @@ export default class DataGrid extends React.Component {
 			`.${FOCUSED_COLUMN_CLASS_NAME}`,
 		);
 
-		for (const focusedCell of focusedCells) {
-			focusedCell.classList.remove(FOCUSED_COLUMN_CLASS_NAME);
-		}
+		focusedCells.forEach(({ classList }) => classList.remove(FOCUSED_COLUMN_CLASS_NAME));
```
ocamon, line 273: We need to check that IE has a polyfill for this method as it doesn't support forEach on NodeList. See: https://developer.mozilla.org/fr/docs/Web/API/NodeList/forEach

packages/datagrid/src/components/DataGrid/DataGrid.component.js
```diff
@@ -286,9 +285,7 @@ export default class DataGrid extends React.Component {
 			`[col-id="${colId}"]:not(.${FOCUSED_COLUMN_CLASS_NAME})`,
 		);
 
-		for (const columnCell of columnsCells) {
-			columnCell.classList.add(FOCUSED_COLUMN_CLASS_NAME);
-		}
+		columnsCells.forEach(({ classList }) => classList.add(FOCUSED_COLUMN_CLASS_NAME));
```
ocamon, line 288: Same here, we need to check IE compat.

packages/datagrid/src/components/DataGrid/DataGrid.component.js
jare-talend: we use babel polyfill to add this method. https://github.com/Talend/ui-scripts/blob/master/packages/config-babel/.browserslistrc

PR #2800, opened 2020-04-28
fix(TCCUI-135) - show date value if specified without time

ybaskaran:
1. Date value is not shown in the Date field if time is not specified.
2. Form submit doesn't show any error in such case.
The issue is only with the string DateTime input.
https://jira.talendforge.org/browse/TCCUI-135

show date value if specified without time and also throw error if validation fails
http://2800.talend.surge.sh/components/?path=/story/form-controls-datepicker-datetime--text-input

### DIFF
--- a/packages/components/src/DateTimePickers/DateTime/Manager/Manager.component.js
+++ b/packages/components/src/DateTimePickers/DateTime/Manager/Manager.component.js
@@ -33,6 +33,8 @@ function ContextualManager(props) {
 		if (props.value !== state.datetime) {
 			const nextState = extractParts(props.value, getDateOptions());
 			setState(nextState);
+			// Triggering onChange will propagate the error to the outside world when the
+			// new input is invalid and different then the default dateTime input.
 			if (nextState.errors && nextState.errors.length > 0) {
 				onChange(null, nextState);
 			}
--- a/packages/components/src/DateTimePickers/DateTime/Manager/Manager.component.js
+++ b/packages/components/src/DateTimePickers/DateTime/Manager/Manager.component.js
@@ -22,19 +22,23 @@ function ContextualManager(props) {
 	const initialState = extractParts(props.value, getDateOptions());
 	const [state, setState] = useState(initialState);
 
-	useEffect(() => {
-		if (props.value !== state.datetime) {
-			const nextState = extractParts(props.value, getDateOptions());
-			setState(nextState);
-		}
-	}, [props.value]);
-
 	function onChange(event, payload) {
 		if (props.onChange) {
 			const { datetime, textInput, errors, errorMessage } = payload;
 			props.onChange(event, { datetime, textInput, errors, errorMessage });
 		}
 	}
+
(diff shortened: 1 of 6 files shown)


packages/components/src/DateTimePickers/DateTime/Manager/Manager.component.js
```diff
@@ -22,19 +22,23 @@ function ContextualManager(props) {
...
+	}
+
 	useEffect(() => {
 		if (props.value !== state.datetime) {
 			const nextState = extractParts(props.value, getDateOptions());
 			setState(nextState);
+			if (nextState.errors && nextState.errors.length > 0) {
+				onChange(null, nextState);
+			}
 		}
```
VolodymyrKovalM, line 39: lgtm

and, with triggering onchange here, won't even be  necessary to implement separate validation for "datetime" widget in form, since now it validates datetime as simple input text, so only date without time (not empty string) is considered as valid, and the form can be submitted, which is incorrect

packages/components/src/DateTimePickers/DateTime/Manager/Manager.component.js
```diff
@@ -22,19 +22,23 @@ function ContextualManager(props) {
...
+		}
+	}
+
 	useEffect(() => {
 		if (props.value !== state.datetime) {
 			const nextState = extractParts(props.value, getDateOptions());
 			setState(nextState);
+			if (nextState.errors && nextState.errors.length > 0) {
+				onChange(null, nextState);
+			}
```
jsomsanith-tlnd, line 38: Can you add a comment above this code explaining the case please ?
To know that we want to propagate the error to the outside world when the new input is different (injected value) that is invalid.
I'm afraid that it's not that obvious and someone removes it in the future.

packages/components/src/DateTimePickers/DateTime/Manager/Manager.component.js
```diff
@@ -22,19 +22,23 @@ function ContextualManager(props) {
...
+			const { datetime, textInput, errors, errorMessage } = payload;
+			props.onChange(event, { datetime, textInput, errors, errorMessage });
+		}
+	}
+
 	useEffect(() => {
 		if (props.value !== state.datetime) {
 			const nextState = extractParts(props.value, getDateOptions());
 			setState(nextState);
+			if (nextState.errors && nextState.errors.length > 0) {
```
jsomsanith-tlnd, line 36: Can you add a test please ?

packages/components/src/DateTimePickers/DateTime/Manager/Manager.component.js
ybaskaran: added comment

packages/components/src/DateTimePickers/DateTime/Manager/Manager.component.js
ybaskaran: added tests

PR #2805, opened 2020-04-29
feat(components): Handle long values in RecordsViewer

tlnd-mhuchet:
In hierarchical view, if a value is too long for the line, it is shrink with an ellipsis and the full value is shown in a tooltip. The property name is shrink too in some cases.

We want to have another behaviour. The property name should never be truncated, and if the value is too long for the remaining space, it is shrink with an ellipsis. But we have a new button available on the line to expand the value and display it fully.

### DIFF
--- a/packages/components/src/DataViewer/Core/Tree/Tree.scss
+++ b/packages/components/src/DataViewer/Core/Tree/Tree.scss
@@ -14,6 +14,7 @@
 
 	&-node-border {
 		border-left: 1px solid $alto;
+		margin-left: -12px;
 		&:hover {
 			border-left-color: $silver-chalice;
 		}
--- a/packages/components/src/DataViewer/RecordsViewer/Branch/RecordsViewerBranch.component.js
+++ b/packages/components/src/DataViewer/RecordsViewer/Branch/RecordsViewerBranch.component.js
@@ -8,8 +8,8 @@ import { LengthBadge } from '../../Badges';
 import { TreeBranchIcon } from '../../Icons';
 import theme from '../RecordsViewer.scss';
 
-const ICON_MARGIN_PLUS = -12;
-const ICON_MARGIN_CARET = -10;
+const ICON_MARGIN_PLUS = -17;
+const ICON_MARGIN_CARET = -17;
 
 /**
  * Used with the lazy loading to allow the render of the skeleton.
--- a/packages/components/src/DataViewer/RecordsViewer/Branch/RecordsViewerBranch.component.js
+++ b/packages/components/src/DataViewer/RecordsViewer/Branch/RecordsViewerBranch.component.js
@@ -8,8 +8,6 @@ import { LengthBadge } from '../../Badges';
 import { TreeBranchIcon } from '../../Icons';
 import theme from '../RecordsViewer.scss';
 
-const ICON_MARGIN = -17;
-
 /**
  * Used with the lazy loading to allow the render of the skeleton.
  * @param {object} value
@@ -108,9 +106,6 @@ export class RecordsViewerBranch extends React.Component {
 						jsonpath={this.props.jsonpath}
 						onToggle={this.props.onToggle}
 						opened={opened}
-						style={{
-							marginLeft: ICON_MARGIN,
-						}}
 						useCustomIcon={type === 'object'}
 						value={value}
 					/>
--- a/packages/components/src/DataViewer/RecordsViewer/Branch/RecordsViewerBranch.component.js
+++ b/packages/components/src/DataViewer/RecordsViewer/Branch/RecordsViewerBranch.component.js
@@ -8,8 +8,7 @@ import { LengthBadge } from '../../Badges';
 import { TreeBranchIcon } from '../../Icons';
 import theme from '../RecordsViewer.scss';
 
-const ICON_MARGIN_PLUS = -17;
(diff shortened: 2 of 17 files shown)


packages/components/src/DataViewer/RecordsViewer/Branch/RecordsViewerBranch.component.js
```diff
@@ -110,7 +109,7 @@ export class RecordsViewerBranch extends React.Component {
 						onToggle={this.props.onToggle}
 						opened={opened}
 						style={{
-							marginLeft: type === 'object' ? ICON_MARGIN_PLUS : ICON_MARGIN_CARET,
+							marginLeft: ICON_MARGIN,
```
jsomsanith-tlnd, line 112: Is it possible to not place the icon the way it is ? :) 
Let's all do that in a simple sass file
```css
.tc-records-viewer-branch-content {
    position: relative;
}

.tc-tree-branch-icon {
    position: absolute;
    left: 0;
    top: 50%;
    transform: translate(-150%, -50%);
}
```

And remove the js style that is now constant.

Plus, I know it's not in your scope but this button is not accessible. :( There is no text alt.

packages/components/src/DataViewer/RecordsViewer/Leaf/RecordsViewerLeaf.component.js
```diff
@@ -11,20 +12,57 @@ export function RecordsViewerLeaf({
...
 			className={classNames(theme['tc-records-viewer-leaf'], 'tc-records-viewer-leaf', className, {
 				[theme['tc-records-viewer-leaf-highlighted']]: nodeHighlighted,
 				'tc-records-viewer-leaf-highlighted': nodeHighlighted,
 			})}
 		>
+			{isValueOverflown && (
+				<span
+					className={classNames(theme['tc-leaf-overflow-icon'], 'tc-leaf-overflow-icon', className)}
+				>
+					<Icon
```
jsomsanith-tlnd, line 42: The button is not accessible because of the alternative text here too.
But the issue is on Icon itself first, it doesn't allow a text. You can use ActionButton with the
* `icon` + `iconTransform` props to set the icon
* `label` + `hideLabel` props to have a hidden accessible label
* `link` props to have only the icon style. You surely will have to fix the margins/paddings, but you're already do that

packages/components/src/DataViewer/RecordsViewer/Leaf/RecordsViewerLeaf.component.js
```diff
@@ -11,20 +12,57 @@ export function RecordsViewerLeaf({
 	className,
 	nodeHighlighted,
 	displayTypes,
+	measure,
 }) {
+	const ref = React.createRef();
+	const [isValueOverflown, setIsValueOverflown] = useState(false);
+	const [isLongValueToggled, setIsLongValueToggled] = useState(false);
```
jsomsanith-tlnd, line 19: Can you change the name, because `toggled` doesn't indicate if it's opened or closed

packages/components/src/DataViewer/RecordsViewer/RecordsViewer.scss
```diff
@@ -114,3 +122,43 @@ $border-size: 0.1rem;
...
+@mixin chevron($color) {
+	border: 1px solid $scooter;
+	padding: 3px;
+	border-radius: 10px;
+	color: $color;
+}
+
+.tc-leaf-overflow-icon {
+	margin-left: -17px;
+	margin-right: $padding-smaller;
```
jsomsanith-tlnd, line 139:
````css
    position: absolute;
    left: -1.5rem;
    top: 50%;
    transform: translateY(-50%);

File: packages/components/src/DataViewer/Text/SimpleTextKeyValue/SimpleTextKeyValue.component.js [hunk 5]
```diff
@@ -59,6 +60,8 @@ export function AvroRenderer({ colDef, data }) {
 			return (
 				<DefaultValueRenderer
 					value={data.value}
+					isValueOverflown={isValueOverflown}
+					isLongValueToggled={isLongValueToggled}
```
jsomsanith-tlnd, line 64: Just wondering, instead of passing explicitly those 2 values, would it be better to just spread the rest of the props to each implementation ?
(don't have the answer to that)

packages/components/src/DataViewer/Core/Tree/Tree.scss
```diff
@@ -14,6 +14,7 @@
 
 	&-node-border {
 		border-left: 1px solid $alto;
+		margin-left: -12px;
```
jsomsanith-tlnd, line 17: Why 12px ?

packages/components/src/DataViewer/Text/SimpleTextKeyValue/SimpleTextKeyValue.component.js
tlnd-mhuchet: This component uses those props for some style so we already pick them individually.

packages/components/src/DataViewer/Core/Tree/Tree.scss
tlnd-mhuchet: It is to align the border below the branch icon, right in the middle.

PR #2809, opened 2020-05-01
feat(cmf): provide async http functions

romainseb:
On the current state of cmf package, our tooling allow us to do http request with the middleware or via saga.
If we need to use basic async function, we can't right now with the current exposed api

Put in common stuff in http at cmf root level and expose async functions to make theses calls


packages/cmf/src/http.js
```diff
@@ -0,0 +1,274 @@
...
+ *
+ * @param  {Response} response A response object
+ * @return {Promise}           A promise that reject with the result of parsing the body
+ */
+export function handleError(response) {
+	// in case of network issue
+	if (response instanceof Error) {
+		return new HTTPError({ response, data: response });
+	}
+	return handleBody(response).then(body => new HTTPError(body));
```
gjou-tlnd, line 130: can we stop swallowing errors?

packages/cmf/src/http.js
```diff
@@ -0,0 +1,274 @@
+import merge from 'lodash/merge';
+import get from 'lodash/get';
+import { mergeCSRFToken } from './middlewares/http/csrfHandling';
+import { HTTP_STATUS, testHTTPCode, HTTP_METHODS } from './middlewares/http/constants';
+
+/**
+ * Storage point for the doc setup using `setDefaultConfig`
+ */
+export const HTTP = {
```
jsomsanith-tlnd, line 9: Can you not export that please ?
There is a getter, that is part of the api. Let's keep it internal.

packages/cmf/src/http.js
romainseb: Sure we can but it will be a breaking change with the current API
At least, we can think about it.
It may be an idea to expose this tools in a separate package as it feels a bit weird to me too to import http from react-cmf, if we do so, we can keep the swallowing with the current implementation in cmf and let the error be dispatch in the new package.

WDYT @jsomsanith-tlnd  ?

packages/cmf/src/http.js
romainseb: Is it used in the tests ? (but even if, it's not a good reason :D)

packages/cmf/src/http.js
jsomsanith-tlnd: Let's keep the scope of this PR which is extracting the code to be usable. Agree with you about let the error be dispatched in the extracted code.
For the existing api, feel free to create a new PR and we'll schedule the breaking change properly

packages/cmf/src/http.js
jsomsanith-tlnd: it seems to be used in the tests to reset its state. You can expose a reset of the tests can get + mutate

jmfrancois:
must be rework according to ARCH-322

PR #2810, opened 2020-05-04
feat(forms): input wigdet + rhf input widget

jsomsanith-tlnd:
We want to have common inputs, to use with RHF.

Why do we want those inputs instead of using directly an `<input>` with RHF ?
- common markup to fit with the common style
- a11y comes by default (description and errors are linked to the input)

Folder hierarchy
```
src
    |_ widgets
        |_ fields
        |_ fieldsets
        |_ templates
    |_ rhf
        |_ fields
        |_ fieldsets
```

Framework-agnostic widgets
- `src/widgets/templates`: common templates to manage style, `description`, `error`, `inProgress` status
- `src/widgets/fieldsets` : framework-agnostic structural widgets.
- `src/widgets/fields` : framework-agnostic inputs widgets that has values.
  - wrap components or html elements, using the common templates (for description, error, ...)
  - manages a11y
  - pass down any other props to the element/component.

RHF wrappers
- `src/rhf/fieldsets`: rhf wrapper around structural widgets
- `src/rhf/fields`: rhf wrapper around framework-agnostic inputs widgets
  - register to rhf
  - get error from rhf

This is a start, with the simple input for now. How to use it ? Write a RHF form, and use the `<Input>` widget instead of `<input>` html element. See the story for more code.

TODO: tests


packages/forms/.storybook/config.js
```diff
@@ -40,4 +36,10 @@ addDecorator(
...
 addDecorator(withA11y);
 addDecorator(withFormLayout);
-configure(loadStories, module);
+configure(
+	[
+		require.context('../src', true, /\.stories\.js$/),
+		require.context('../stories-core', false, /index\.js$/),
+	],
+	module,
+);
```
jsomsanith-tlnd, line 45: Let's not stay in the past :D 
The stories for the widgets will be located next to the widget. So we still load `stories-core` folder for the old stories, but we load the `.stories.js` files too

packages/forms/src/widgets/templates/FieldTemplate/FieldTemplate.test.js
```diff
@@ -0,0 +1,49 @@
...
+			error: 'This is the error',
+		};
+		// when
+		const wrapper = mount(
+			<FieldTemplate {...props}>
+				<input name="test" />
+			</FieldTemplate>,
+		);
+		// then
+		expect(wrapper.html()).toMatchSnapshot();
```
jsomsanith-tlnd, line 22: Snapshotting a `.html()` make it difficult to read.
Can you use `toJson()` instead ?

PR #2812, opened 2020-05-05
feat(containers/forms): pass actual properties when change the definitionURL

jare-talend:
The container dispatches a event when the definitionURL changed. We want to save the data typed before by the user on the action ON_DEFINITION_URL_CHANGED.

pass the actual properties on the event 

### DIFF
--- a/packages/containers/src/ComponentForm/ComponentForm.test.js
+++ b/packages/containers/src/ComponentForm/ComponentForm.test.js
@@ -255,13 +255,13 @@ describe('ComponentForm', () => {
 			const dispatch = jest.fn();
 			const oldUrl = 'http://old';
 			const newUrl = 'http://new';
-			const properties = { properties: { name: 'old' } };
+			const componentState = { properties: { name: 'old' } };
 
 			// when
 			const wrapper = shallow(
 				<TCompForm state={state} definitionURL={oldUrl} dispatch={dispatch} />,
 			);
-			wrapper.setState(properties);
+			wrapper.setState(componentState);
 			wrapper.setProps({ definitionURL: newUrl });
 
 			// then
@@ -270,7 +270,7 @@ describe('ComponentForm', () => {
 				dispatch,
 				state,
 				type: TCompForm.ON_DEFINITION_URL_CHANGED,
-				...properties,
+				...componentState,
 			});
 		});
 	});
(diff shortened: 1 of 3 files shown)


jsomsanith-tlnd:
I don't understand your need. The definitionUrl is the url to fetch the schema, as a whole new form. So if it changes, it means that you display a new form. Can you describe your use case please ?

jare-talend:
> I don't understand your need. The definitionUrl is the url to fetch the schema, as a whole new form. So if it changes, it means that you display a new form. Can you describe your use case please ?

Yes I can. Our add dataset form allows to create a new connection during the dataset creation. We open a drawer with the connection creation and we valid this form, we returns on the dataset form and we change the definitionURL to get the tacokit parameters for this connection.

The user can start to type a dataset name, next ask to create a new connection. We lose the datset name typed by the user when it returns on the dataset form.
I added on your side after this event to restore the previous dataset name in the component state.

packages/containers/src/ComponentForm/ComponentForm.test.js
```diff
@@ -255,19 +255,22 @@ describe('ComponentForm', () => {
 			const dispatch = jest.fn();
 			const oldUrl = 'http://old';
 			const newUrl = 'http://new';
+			const properties = { properties: { name: 'old' } };
```
jsomsanith-tlnd, line 258:
```suggestion
			const state = { properties: { name: 'old' } };
```

packages/containers/src/ComponentForm/ComponentForm.test.js
```diff
@@ -255,19 +255,22 @@ describe('ComponentForm', () => {
...
 
 			// when
 			const wrapper = shallow(
 				<TCompForm state={state} definitionURL={oldUrl} dispatch={dispatch} />,
 			);
+			wrapper.setState(properties);
```
jsomsanith-tlnd, line 264:
```suggestion
			wrapper.setState(state);
```

packages/containers/src/ComponentForm/ComponentForm.test.js
```diff
@@ -255,19 +255,22 @@ describe('ComponentForm', () => {
...
 
 			// then
 			expect(dispatch).toBeCalledWith({
-				type: TCompForm.ON_DEFINITION_URL_CHANGED,
-				state,
 				definitionURL: newUrl,
 				dispatch,
+				state,
+				type: TCompForm.ON_DEFINITION_URL_CHANGED,
+				...properties,
```
jsomsanith-tlnd, line 273:
```suggestion
				properties: state.properties,
```

packages/containers/src/ComponentForm/ComponentForm.test.js
jare-talend: state is already take, I rename by componentState

PR #2824, opened 2020-05-15
feat(component): Allow to customize RecordsViewer types render

tlnd-mhuchet:
We can choose to display or not type of data. But we can't customise it.

Add the possibility to use a custom render to display types.

### DIFF
--- a/packages/components/src/DataViewer/Text/SimpleTextKeyValue/SimpleTextKeyValue.component.js
+++ b/packages/components/src/DataViewer/Text/SimpleTextKeyValue/SimpleTextKeyValue.component.js
@@ -123,7 +123,7 @@ const SimpleTextKeyValue = React.forwardRef(function SimpleTextKeyValue(
 	if (displayTypes && schema && value) {
 		types = (
 			<span className={classNames(theme['tc-simple-text-type'], 'tc-simple-text-type')}>
-				{typesRenderer ? typesRenderer(schema) : `- ${schema.type.type}`}
+				{typesRenderer(schema)}
 			</span>
 		);
 	}
@@ -180,6 +180,7 @@ SimpleTextKeyValue.propTypes = {
 
 SimpleTextKeyValue.defaultProps = {
 	displayTypes: false,
+	typesRenderer: schema => `- ${schema.type.type}`,
 };
 
 export default SimpleTextKeyValue;
--- a/packages/components/src/DataViewer/Text/SimpleTextKeyValue/SimpleTextKeyValue.component.js
+++ b/packages/components/src/DataViewer/Text/SimpleTextKeyValue/SimpleTextKeyValue.component.js
@@ -113,11 +113,21 @@ const SimpleTextKeyValue = React.forwardRef(function SimpleTextKeyValue(
 		style,
 		value,
 		displayTypes,
+		typesRenderer,
 		isValueOverflown,
 		isLongValueToggled,
 	},
 	ref,
 ) {
+	let types;
+	if (displayTypes && schema && value) {
+		types = (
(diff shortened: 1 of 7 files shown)


smouillour:
There is no dedicated test for this change?

packages/components/src/DataViewer/Text/SimpleTextKeyValue/SimpleTextKeyValue.component.js
```diff
@@ -128,11 +138,7 @@ const SimpleTextKeyValue = React.forwardRef(function SimpleTextKeyValue(
 				<span className={classNames(theme['tc-simple-text-key'], 'tc-simple-text-key')}>
 					{formattedKey}
 					{separator}
-					{displayTypes && schema && value && (
-						<span className={classNames(theme['tc-simple-text-type'], 'tc-simple-text-type')}>
-							- {schema.type.type || value.unionKey}
```
jare-talend, line 133: we lose value.unionKey ?

tlnd-mhuchet: I added it as a workaround to handle data from pipeline preview.
I may be wrong, but I think types are defined by the schema, not the value. So I think it is relevant to only use the schema for it. `unionKey` should not be used for that I think 🤔

packages/components/src/DataViewer/Text/SimpleTextKeyValue/SimpleTextKeyValue.component.js
```diff
@@ -113,11 +113,21 @@ const SimpleTextKeyValue = React.forwardRef(function SimpleTextKeyValue(
...
 		isValueOverflown,
 		isLongValueToggled,
 	},
 	ref,
 ) {
+	let types;
+	if (displayTypes && schema && value) {
+		types = (
+			<span className={classNames(theme['tc-simple-text-type'], 'tc-simple-text-type')}>
+				{typesRenderer ? typesRenderer(schema) : `- ${schema.type.type}`}
```
jsomsanith-tlnd, line 126: Instead of an if, could you set a default value to `typesRenderer ` props please ?

tlnd-mhuchet: Done

tlnd-mhuchet:
> There is no dedicated test for this change?

@smouillour @jsomsanith-tlnd You can now find tests for default render and custom one.

tlnd-mhuchet:
> Could you add a storie for this ?

@jare-talend I added a story with only the RecordsViewer displaying types.

packages/components/src/DataViewer/Text/SimpleTextKeyValue/SimpleTextKeyValue.test.js
```diff
@@ -49,4 +49,45 @@ describe('SimpleTextKeyValue', () => {
...
+			<SimpleTextKeyValue
+				className="myCLass"
+				formattedKey="myKey"
+				value={data}
+				schema={schema}
+				displayTypes
+				typesRenderer={s => <span>And the type is {s.type.type}</span>}
+			/>,
+		);
+		expect(wrapper.getElement()).toMatchSnapshot();
```
jsomsanith-tlnd, line 91:
```suggestion
		expect(wrapper.find('.tc-simple-text-type').text()).toBe('And the type is int');
```

Can we avoid the snapshot please ?

If you really want to check that the span is hear, perhaps you can snapshot just `wrapper.find('.tc-simple-text-type').getElement()`

PR #2827, opened 2020-05-19
fix: remove @babel/polyfill on the packages

jare-talend:

cmf-cqrs and router use a dependencies @babel/polyfill instead to use in devdependencies. This force the projects to use a old version core-js@2 instead core-js@3

core-js@3 brings the latest ECMAScript features.

Important! If the browser you are targeting doesn't support ES2015 generators, you must transpile them (i.e. with @talend/scripts-core or babel plugin) and provide a valid runtime (regenerator-runtime/runtime)

avoid to use @babel/polyfill and use core-js

* [] Tests for the changes have been added (for bug fixes / features) And [non reg](./screenshots.md) done before need review


packages/cmf-cqrs/package.json
```diff
@@ -31,9 +31,10 @@
   },
   "homepage": "https://github.com/Talend/ui/cmf-cqrs#readme",
   "dependencies": {
-    "@babel/polyfill": "^7.8.3",
+    "core-js": "3",
```
jsomsanith-tlnd, line 34: core-js and regenerator-runtime should be dev dependencies as they are build dependencies used by babel ?

jare-talend: https://babeljs.io/docs/en/babel-preset-env#usebuiltins

Since @babel/polyfill was deprecated in 7.4.0, we recommend directly adding core-js and setting the version via the corejs option.

The build generated by babel wraps the code with the import from core-js and regenerator-runtime like this : 

require("core-js/modules/es.object.keys");
require("core-js/modules/web.dom-collections.for-each");
require("regenerator-runtime/runtime");

// ... your code transpiled. So when the code is executed, we need core-js and regenerator-runtime/runtime, so it is why i have integrated dependencies and not devDependencies.

jsomsanith-tlnd: Dependencies and dev dependencies are available locally for any run and build. That's not the point. Babel transform your executed code, it's still a dev dep. Those libraries are used in the build process to enhance your code, not used directly in your code. That's why I suggested to put them as dev dep.

Moreover, as dependencies, don't you introduce again the problem you have today but with a new version of core-js ?

PR #2830, opened 2020-05-26
fix(component): Fix TabBar with Dropdown

tlnd-mhuchet:
When we use the TabBar component in responsive mode, it use a Dropdown instead of a classic Nav.
But in this mode, the tab content is no longer displayed.

A gif can be seen in this Jira: https://jira.talendforge.org/browse/TFD-10784

The issue was caused by the fact that we don't have the Tab.Container when using Dropdown. Without the active key management it provides, tabs were not aware of it.
So I modified the component to always have this container.

### DIFF
--- a/output/components.eslint.txt
+++ b/output/components.eslint.txt
@@ -59,9 +59,6 @@ The react/require-extension rule is deprecated. Please use the import/extensions
 /home/travis/build/Talend/ui/packages/components/src/SubHeaderBar/SubHeaderBar.test.js
   4:8  error  'Container' is defined but never used  no-unused-vars
 
-/home/travis/build/Talend/ui/packages/components/src/TabBar/TabBar.component.js
-  209:20  error  'tooltipPlacement' PropType is defined but prop is never used  react/no-unused-prop-types
-
 /home/travis/build/Talend/ui/packages/components/src/Toggle/Checkbox.js
   3:1  error  Prefer default export  import/prefer-default-export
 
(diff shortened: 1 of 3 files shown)


output/components.eslint.txt
```diff
@@ -59,6 +59,9 @@ The react/require-extension rule is deprecated. Please use the import/extensions
 /home/travis/build/Talend/ui/packages/components/src/SubHeaderBar/SubHeaderBar.test.js
   4:8  error  'Container' is defined but never used  no-unused-vars
 
+/home/travis/build/Talend/ui/packages/components/src/TabBar/TabBar.component.js
+  209:20  error  'tooltipPlacement' PropType is defined but prop is never used  react/no-unused-prop-types
```
jsomsanith-tlnd, line 63: Can you fix this lint error please ?

packages/components/src/TabBar/TabBar.component.js
```diff
@@ -120,24 +120,55 @@ function TabBar(props) {
...
+									{...icon}
+								/>
+							)}
+							{item.label}
+						</React.Fragment>
+					</TooltipTrigger>
+				</NavItem>
+			))}
+		</Nav>
+	);
```
jsomsanith-tlnd, line 155: Can you put that in the `else` of responsive condition ?

packages/components/src/TabBar/TabBar.scss
```diff
@@ -26,3 +26,7 @@
 		}
 	}
 }
+
+:global(.dropdown) + :global(.tab-content) {
+	padding-top: 20px;
```
jsomsanith-tlnd, line 31:
```suggestion
	padding-top: $padding-large;
```

PR #2841, opened 2020-06-02
fix(TCCUI-170) - use panel title if collapsibleFieldSet contains array

ybaskaran:
CollapsiblePanel title gets overridden with the last collpasibleFieldSet title if the widget contains arrays or multiple collpasibleFieldSet widgets
https://jira.talendforge.org/browse/TCCUI-170
http://talend.surge.sh/forms/?path=/story/uiform-v2-core-fieldsets--core-collapsiblefieldset

check if the collapsibleFieldSet contains arrays or multiple collapsible widgets and return the schema title
http://2841.talend.surge.sh/forms/?path=/story/uiform-v2-core-fieldsets--core-collapsiblefieldset

### DIFF
--- a/packages/forms/src/UIForm/fieldsets/CollapsibleFieldset/CollapsibleFieldset.component.js
+++ b/packages/forms/src/UIForm/fieldsets/CollapsibleFieldset/CollapsibleFieldset.component.js
@@ -29,31 +29,8 @@ function getDrillKey(key) {
 	}, []);
 }
 
-function isCollapsibleArray(element, schema) {
-	if (!schema && element.items && !element.items.some(item => item.type === 'array')) {
-		return true;
-	}
-	const hasCollapsible = (element.items || []).reduce((acc, item) => {
-		if (item.items && isCollapsibleArray(item, element)) {
-			acc.push(true);
-		}
-		if (
-			acc.length === 0 &&
-			schema &&
-			(schema.items || []).some(data => data.widget === 'collapsibleFieldset')
-		) {
-			acc.push(true);
-		}
-		return acc;
-	}, []);
-	return hasCollapsible.length > 0;
-}
-
 export function defaultTitle(formData, schema, options) {
-	if (!isCollapsibleArray(schema)) {
-		return schema.title;
-	}
-	const title = (schema.items || []).reduce((acc, item) => {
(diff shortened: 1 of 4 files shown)


packages/forms/src/UIForm/fieldsets/CollapsibleFieldset/CollapsibleFieldset.component.js
```diff
@@ -29,7 +29,30 @@ function getDrillKey(key) {
 	}, []);
 }
 
+function isCollapsibleArray(element, schema) {
```
tlnd-mhuchet, line 32: `isCollapsibleArray` take 2 params, and only one is provided, is it normal?

ybaskaran: the second parameter is the main schema passed on the recursive calls within the same function

tlnd-mhuchet: Oh! Missed that part 😅

jsomsanith-tlnd:
The idea of this widget is that you can customize the title depending on your need. There is a `createCollapsibleFieldset()` function that takes the title function.
If you need a different title, you can use it to create your collapsible fieldset that you pass as widgets to the form.

Is that an acceptable solution ?

packages/forms/src/UIForm/fieldsets/CollapsibleFieldset/CollapsibleFieldset.component.js
```diff
@@ -29,7 +29,30 @@ function getDrillKey(key) {
...
+		if (
+			acc.length === 0 &&
+			schema &&
+			(schema.items || []).some(data => data.widget === 'collapsibleFieldset')
+		) {
+			acc.push(true);
+		}
+		return acc;
+	}, []);
+	return hasCollapsible.length > 0;
```
yyanwang, line 49: IMHO `reduce` is not necessary here, can it be written like:
```
const hasCollapsible = get(element, 'items', []).some(
		item =>
			(item.items && isCollapsibleArray(item, element)) ||
			get(schema, 'items', []).some(data => data.widget === 'collapsibleFieldset'),
	);
return hasCollapsible;
```
?

PR #2843, opened 2020-06-04
fix(TCCUI-167): change link left margin according to guidelines

mkondzera:
https://jira.talendforge.org/browse/TCCUI-167
Link margin is too big, so was changed in guidelines
image

Change link margin-left from 1ch to 0.35rem

### DIFF
--- a/packages/components/src/InlineMessage/InlineMessage.scss
+++ b/packages/components/src/InlineMessage/InlineMessage.scss
@@ -26,9 +26,9 @@ $tc-inline-message-icon-offset: 0.4rem !default;
 		line-height: $tc-inline-message-min-height;
 		min-height: $tc-inline-message-min-height;
 
-		span {
+		&-item {
 			&:after {
-				content: " ";
+				content: ' ';
 			}
 		}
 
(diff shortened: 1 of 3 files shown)


packages/components/src/InlineMessage/InlineMessage.scss
```diff
@@ -47,7 +48,7 @@ $tc-inline-message-icon-offset: 0.4rem !default;
 	}
 
 	&-link {
-		margin-left: 1ch;
+		margin-left: $tc-inline-message-link-margin;
```
frassinier, line 51: I would suggest trying to update guidelines instead for using simple space

```scss
.css-classname-on-each-span {
  content: ' ';
}
```

packages/components/src/InlineMessage/InlineMessage.scss
```diff
@@ -27,6 +26,12 @@ $tc-inline-message-icon-offset: 0.4rem !default;
 		line-height: $tc-inline-message-min-height;
 		min-height: $tc-inline-message-min-height;
 
+		span {
```
frassinier, line 29: could you add a css class on all instead, please?
I'm not sure than a span will stay to display link in the future, so maybe a class is better than a specific element based selector!

frassinier: ok you are not targeting the link, I've read too fast, but yes a css class will be better imho

PR #2853, opened 2020-06-15
feat: http package

romainseb:
Our current XHR calls are tied up to the CMF package and we got only availability to make call using the http middleware or the generator way.

This package provide implementations to make xhr calls using generators and promised version.
It handle CSRF token stuff

Usage of this package in @talend/cmf would come after the release of this one


packages/http/src/http.common.js
```diff
@@ -0,0 +1,123 @@
...
+ * @return {Promise}                          A Promise that resolves to a Response object.
+ */
+export async function httpFetch(url, config, method, payload) {
+	const defaultHeaders = {
+		Accept: 'application/json',
+		'Content-Type': 'application/json',
+	};
+
+	/**
+	 * If the playload is an instance of FormData the body should be set to this object
```
vbocquet, line 97:
```suggestion
	 * If the payload is an instance of FormData the body should be set to this object
```
Minor typo

PR #2859, opened 2020-06-18
fix(TCCUI-178) - datalist onselect not triggered on mouse down event

ybaskaran:
https://jira.talendforge.org/browse/TCCUI-178
There is a regression with the datalist after the recent changes with the DataList component.
OnSelect not triggered on mousedown when the focus is already within the datalist.

Regression from,
https://github.com/Talend/ui/pull/2753
and
https://github.com/Talend/ui/pull/2741

datalist_focus_issue

Instead of event.preventDefault() we can use the same onSelect when the mouseDown event is triggered
http://2859.talend.surge.sh/components/?path=/story/form-controls-datalist--default-multisection

### DIFF
--- a/packages/components/src/Typeahead/Typeahead.component.js
+++ b/packages/components/src/Typeahead/Typeahead.component.js
@@ -140,8 +140,18 @@ function Typeahead({ onToggle, icon, position, docked, ...rest }) {
 		...inputProps,
 		items: getItems(rest.items, rest.dataFeature),
 		itemProps: ({ itemIndex }) => ({
-			onMouseDown: rest.onSelect,
-			onClick: rest.onSelect,
+			// onBlur shouldn't be called before onClick but the general order is (mouseDown, Blur, Click)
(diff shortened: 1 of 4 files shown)


packages/components/src/Typeahead/Typeahead.component.js
```diff
@@ -140,7 +140,7 @@ function Typeahead({ onToggle, icon, position, docked, ...rest }) {
 		...inputProps,
 		items: getItems(rest.items, rest.dataFeature),
 		itemProps: ({ itemIndex }) => ({
-			onMouseDown: event => event.preventDefault(),
+			onMouseDown: rest.onSelect,
```
frassinier, line 143: The `onClick` handler is needed for usage tracking reasons, if we handle `onMouseDown` before I'm not sure that the `onClick` will be fired.
Could you double-check, please? 

Related to https://github.com/Talend/ui/pull/2753

PR #2860, opened 2020-06-19
fix(form/Time): can't input `:` into time field

yyanwang:
Fix for the following issues of Time field in form:
1. Can't input `:` into time field: https://www.loom.com/share/8b69686be6ac4b96861af1676c69708a
2. Time is not displayed after selection: https://www.loom.com/share/a6601ddf13f64a9e9b7687097d94f0e5
3. The width of Time input keeps changing when user type in.

### DIFF
--- a/packages/components/src/DateTimePickers/DateTime/Manager/Manager.component.js
+++ b/packages/components/src/DateTimePickers/DateTime/Manager/Manager.component.js
@@ -47,7 +47,6 @@ function ContextualManager(props) {
 		const nextState = {
 			...state,
 			...newState,
-			time,
 		};
 		setState(nextState);
 		onChange(event, nextState);
--- a/packages/components/src/DateTimePickers/InputDatePicker/InputDatePicker.component.js
+++ b/packages/components/src/DateTimePickers/InputDatePicker/InputDatePicker.component.js
@@ -46,7 +46,7 @@ export default function InputDatePicker(props) {
 	});
 
 	const inputProps = omit(props, PROPS_TO_OMIT_FOR_INPUT);
-	const timePicker = [
+	const datePicker = [
 		<DatePicker.Input {...inputProps} id={`${props.id}-input`} key="input" inputRef={inputRef} />,
 		handlers.showPicker && (
 			<Popper
@@ -96,7 +96,7 @@ export default function InputDatePicker(props) {
 					handlers.onKeyDown(event, inputRef.current);
 				}}
 			>
-				{timePicker}
+				{datePicker}
 			</FocusManager>
 		</DatePicker.Manager>
 	);
--- a/packages/components/src/DateTimePickers/InputDateTimePicker/__snapshots__/InputDateTimePicker.component.test.js.snap
+++ b/packages/components/src/DateTimePickers/InputDateTimePicker/__snapshots__/InputDateTimePicker.component.test.js.snap
@@ -207,7 +207,7 @@ exports[`InputDateTimePicker should render 1`] = `
                     <span
                       style="font-weight: 400; font-size: 1.6rem; visibility: hidden; position: absolute;"
                     >
-                      15:27:00
+                      HH:mm:ss
                     </span>
                   </div>,
                 }
@@ -244,8 +244,7 @@ exports[`InputDateTimePicker should render 1`] = `
                   key="input"
                 >
                   <InputSizer
-                    inputText="15:27:00"
-                    placeholder="HH:mm:ss"
+                    inputText="HH:mm:ss"
                   >
                     <DebounceInput
                       autoComplete="off"
@@ -309,7 +308,7 @@ exports[`InputDateTimePicker should render 1`] = `
                         }
                       }
                     >
-                      15:27:00
+                      HH:mm:ss
                     </span>
                   </InputSizer>
                 </Time.Input>
--- a/packages/components/src/DateTimePickers/InputDateTimePicker/__snapshots__/InputDateTimePicker.component.test.js.snap
+++ b/packages/components/src/DateTimePickers/InputDateTimePicker/__snapshots__/InputDateTimePicker.component.test.js.snap
@@ -242,7 +242,6 @@ exports[`InputDateTimePicker should render 1`] = `
                     }
                   }
                   key="input"
-                  value="15:27:00"
                 >
                   <InputSizer
                     inputText="15:27:00"
--- a/packages/components/src/DateTimePickers/InputDateTimeRangePicker/__snapshots__/InputDateTimeRangePicker.component.test.js.snap
+++ b/packages/components/src/DateTimePickers/InputDateTimeRangePicker/__snapshots__/InputDateTimeRangePicker.component.test.js.snap
@@ -226,7 +226,7 @@ exports[`InputDateTimeRangePicker should render 1`] = `
                             <span
                               style="font-weight: 400; font-size: 1.6rem; visibility: hidden; position: absolute;"
                             >
-                              00:00:00
+                              HH:mm:ss
                             </span>
                           </div>,
                         }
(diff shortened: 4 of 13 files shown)


packages/components/src/DateTimePickers/DateTime/Manager/Manager.component.js
```diff
@@ -47,7 +47,6 @@ function ContextualManager(props) {
 		const nextState = {
 			...state,
 			...newState,
-			time,
```
yyanwang, line 50: This is moved to https://github.com/Talend/ui/pull/2860/files#diff-ba183129831b5094f5d0ea4646f4b316R234

packages/components/src/DateTimePickers/InputDatePicker/InputDatePicker.component.js
```diff
@@ -46,7 +46,7 @@ export default function InputDatePicker(props) {
 	});
 
 	const inputProps = omit(props, PROPS_TO_OMIT_FOR_INPUT);
-	const timePicker = [
+	const datePicker = [
```
yyanwang, line 49: just renaming

packages/components/src/DateTimePickers/InputTimePicker/InputTimePicker.component.js
```diff
@@ -54,7 +40,6 @@ export default function InputTimePicker(props) {
 		<Time.Input
 			{...inputProps}
 			id={`${props.id}-input`}
-			value={getTime(props.value, props.useSeconds)}
```
yyanwang, line 57: the value of input should only be provided by Manager's state.

packages/components/src/DateTimePickers/Time/Input/Input.component.js
```diff
@@ -10,10 +10,7 @@ export default function Input(props) {
 	const { time, inputManagement } = useContext(TimeContext);
 
 	return (
-		<InputSizer
-			placeholder={inputManagement.placeholder}
-			inputText={time.time ? time.textInput : ''}
-		>
+		<InputSizer inputText={inputManagement.placeholder}>
```
yyanwang, line 13: Fix: the time input width keeps changing as user type in.

packages/components/src/DateTimePickers/Time/Manager/Manager.component.js
```diff
@@ -1,84 +1,78 @@
...
-class ContextualManager extends React.Component {
-	static displayName = 'Time.Manager';
-	static propTypes = {
-		children: PropTypes.node,
-		onChange: PropTypes.func,
-		useSeconds: PropTypes.bool,
-		timezone: PropTypes.string,
-		value: PropTypes.oneOfType([PropTypes.string, PropTypes.object]),
-	};
+function TimeContextualManager(props) {
```
yyanwang, line 7: Make Time.Manager a functional component

packages/components/src/DateTimePickers/InputDateTimePicker/__snapshots__/InputDateTimePicker.component.test.js.snap
```diff
@@ -207,7 +207,7 @@ exports[`InputDateTimePicker should render 1`] = `
                     <span
                       style="font-weight: 400; font-size: 1.6rem; visibility: hidden; position: absolute;"
                     >
-                      15:27:00
```
yyanwang, line 210: this snapshot change is because now we only use placeholder text to measure the width of input.

packages/components/src/DateTimePickers/pickers/TimePicker/TimePicker.component.js
```diff
@@ -76,20 +76,17 @@ export class TimePicker extends React.Component {
 			this.scrollItemIntoView(this.props.textInput);
 		}
 	}
-	componentDidUpdate(prevProps) {
```
yyanwang, line 79: Remove this because TimePicker will not update. Usually it will create a new instance when toggle it.

packages/components/src/DateTimePickers/pickers/TimePicker/TimePicker.component.test.js
```diff
@@ -41,25 +41,5 @@ describe('TimePicker component', () => {
 					.hasClass('highlight'),
 			).toBe(true);
 		});
-		it('should scroll the first match into view when user inputs', () => {
```
yyanwang, line 44: Remove this test case because the `componentWillUpdate` has been removed here:
https://github.com/Talend/ui/pull/2860/files#diff-45f8e95db6daaa9d3e4125618e1d07bfL79

packages/components/src/DateTimePickers/InputDateTimeRangePicker/__snapshots__/InputDateTimeRangePicker.component.test.js.snap
```diff
@@ -226,7 +226,7 @@ exports[`InputDateTimeRangePicker should render 1`] = `
                             <span
                               style="font-weight: 400; font-size: 1.6rem; visibility: hidden; position: absolute;"
                             >
-                              00:00:00
+                              HH:mm:ss
```
yyanwang, line 229: changes in this snapshot file are because now we only use placeholder text to measure the width of input.

packages/forms/src/UIForm/fields/Date/Time.component.js
```diff
@@ -25,14 +25,14 @@ function TimeWidget({
...
 
-	function onTimeChange(event, { errorMessage: nextErrorMessage, textInput, time }) {
+	function onTimeChange(event, { errorMessage: nextErrorMessage, textInput }) {
 		setState({ errorMessage: nextErrorMessage });
 		const payload = {
 			schema,
 			value: textInput,
 		};
 		onChange(event, payload);
-		if (!nextErrorMessage && time) {
```
yyanwang, line 35: Simplify the condition, remove redundancy. "no nextErrorMessage" equals to "time is not null".

packages/components/src/DateTimePickers/InputTimePicker/InputTimePicker.component.test.js
```diff
@@ -72,6 +72,9 @@ describe('InputTimePicker', () => {
 				.first()
 				.simulate('click');
 
+			jest.runAllTimers();
+
+			wrapper.update();
```
jsomsanith-tlnd, line 77: Wondering why we need that now but not before ?

packages/components/src/DateTimePickers/Time/Manager/Manager.component.js
```diff
@@ -1,84 +1,78 @@
...
+	const [state, setState] = useState(extractTime(props.value, props.useSeconds));
```
jsomsanith-tlnd, line 8: The problem: `extractTime` is executed at each render, and passed to useState. But useState only uses it at first render.
To fix that, we can provide an initialisation function instead that will be called only at first render

```suggestion
	const [state, setState] = useState(() => extractTime(props.value, props.useSeconds));
```

packages/components/src/DateTimePickers/Time/Manager/Manager.component.js
```diff
@@ -1,84 +1,78 @@
...
 
-	onInputChange(event) {
+	function onInputChange(event) {
 		const textInput = event.target.value;
-		const nextState = extractTime(textInput, this.props.useSeconds);
-		this.setState(nextState, () => {
-			this.onChange(event, 'INPUT');
-		});
+		const nextState = extractTime(textInput, props.useSeconds);
+		setState(nextState);
```
jsomsanith-tlnd, line 27: You can put that `setState` in `onChange()` instead, and remove this line and the one in `onPickerChange()`.
That's not a big deal, but just a suggestion :)

(I like the idea that we have a common `onChange`, and 2 specific change fonctions, that only differs in the way to create the data we save)

packages/components/src/DateTimePickers/InputTimePicker/InputTimePicker.component.test.js
yyanwang: Indeed, they're not necessary. removed.

packages/components/src/DateTimePickers/Time/Manager/Manager.component.js
yyanwang: thanks :) suggestions applied.

PR #2875, opened 2020-06-30
feat(TDC-1501/components/slider): add support for new operator modes

olliew-talend:
The slider component currently only supports highlighting of its track to a specific index from a start point on left of the scale. In the future, the slider will be used to select values of a specific index only or starting from an index up to an end value. In these cases, the slider component needs to be able to highlight the relevant part of its track based on the usage/context scenario.

The slider will now support additional usage contexts through the addition of two parameters. These parameters will activate modifier CSS classes which change the visual appearance of the slider track. The change is a visual one only and the functional behaviour of the slider is unchanged. The new parameters are invisible unless passed as true. The do not require default values.

Style guide reference: http://guidelines.talend.com/document/92132#/specific-features/rating-sliders

- Add override styles for slider track style
- Add new parameters to enable new styles
- Include new styles in Storybook

### DIFF
--- a/packages/components/src/Slider/Slider.component.js
+++ b/packages/components/src/Slider/Slider.component.js
@@ -218,6 +218,8 @@ class Slider extends React.Component {
 		captionTextStepNumber: PropTypes.number,
 		min: PropTypes.number.isRequired,
 		max: PropTypes.number.isRequired,
+		rightTrack: PropTypes.bool,
+		noTrack: PropTypes.bool,
 		captionsFormat: PropTypes.func,
 		disabled: PropTypes.bool,
 	};
@@ -239,6 +241,8 @@ class Slider extends React.Component {
 			captionsFormat,
 			min,
 			max,
+			rightTrack,
+			noTrack,
 			onChange,
 			disabled,
 			...rest
@@ -254,7 +258,14 @@ class Slider extends React.Component {
 						min={min}
 						max={max}
 						handle={noValue ? undefined : this.state.handle}
-						className={classnames(theme['tc-slider-rc-slider'], 'tc-slider-rc-slider')}
+						className={classnames(
+							theme['tc-slider-rc-slider'],
+							{[theme['tc-slider-rc-slider--track-right']]: rightTrack},
+							{[theme['tc-slider-rc-slider--track-none']]: noTrack},
+							'tc-slider-rc-slider',
+							{'tc-slider-rc-slider--track-right': rightTrack},
+							{'tc-slider-rc-slider--track-none': noTrack}
+						)}
 						onChange={onChange}
 						disabled={disabled}
 						{...rest}
--- a/packages/components/src/Slider/Slider.component.js
+++ b/packages/components/src/Slider/Slider.component.js
@@ -48,6 +48,29 @@ export function getCaptionsValue(captionsLength, min, max) {
 }
 
 /**
+ * Return correct modifier styles for usage mode
+ * @param {string} mode
+ * @param {number} value
+ */
+export function getModeStyles(mode, value) {
+	if (Array.isArray(value)) {
+		if (mode === 'exclusive') {
+			return 'exclusive';
+		} else {
+			return false;
+		}
+	} else {
+		if (mode === 'greaterThan') {
+			return 'greaterThan';
+		} else if (mode === 'equals') {
+			return 'equals';
+		} else {
+			return false;
+		}
+	}
+}
+
+/**
  * This function allow to get the actions components
  * @param {array} actions
  * @param {number} value
@@ -218,8 +241,7 @@ class Slider extends React.Component {
 		captionTextStepNumber: PropTypes.number,
 		min: PropTypes.number.isRequired,
 		max: PropTypes.number.isRequired,
-		rightTrack: PropTypes.bool,
-		noTrack: PropTypes.bool,
+		mode: PropTypes.bool,
 		captionsFormat: PropTypes.func,
 		disabled: PropTypes.bool,
 	};
@@ -241,14 +263,14 @@ class Slider extends React.Component {
 			captionsFormat,
 			min,
 			max,
-			rightTrack,
-			noTrack,
+			mode,
 			onChange,
 			disabled,
 			...rest
 		} = this.props;
 		const noValue = value === null || value === undefined;
 		const Component = Array.isArray(value) ? Range : RcSlider;
+		const modeStyles = getModeStyles(mode, value);
 		return (
 			<div>
 				<div className={classnames(theme['tc-slider'], 'tc-slider')} key="slider">
@@ -260,11 +282,13 @@ class Slider extends React.Component {
 						handle={noValue ? undefined : this.state.handle}
 						className={classnames(
 							theme['tc-slider-rc-slider'],
-							{[theme['tc-slider-rc-slider--track-right']]: rightTrack},
-							{[theme['tc-slider-rc-slider--track-none']]: noTrack},
+							{[theme['tc-slider-rc-slider--track-equals']]: modeStyles === 'equals'},
+							{[theme['tc-slider-rc-slider--track-exclusive']]: modeStyles === 'exclusive'},
+							{[theme['tc-slider-rc-slider--track-greater-than']]: modeStyles === 'greaterThan'},
 							'tc-slider-rc-slider',
-							{'tc-slider-rc-slider--track-right': rightTrack},
-							{'tc-slider-rc-slider--track-none': noTrack}
+							{'tc-slider-rc-slider--track-equals': modeStyles === 'equals'},
+							{'tc-slider-rc-slider--track-exclusive': modeStyles === 'exclusive'},
+							{'tc-slider-rc-slider--track-greater-than': modeStyles === 'greaterThan'}
 						)}
 						onChange={onChange}
 						disabled={disabled}
--- a/packages/components/src/Slider/Slider.component.js
+++ b/packages/components/src/Slider/Slider.component.js
@@ -10,6 +10,14 @@ import theme from './Slider.scss';
 import Action from '../Actions/Action';
 
 const noFormat = value => value;
+/**
+ * Options for controlling slider operator display mode
+ */
+export const SLIDER_MODE = {
+	GREATER_THAN: 'greaterThan',
+	EQUALS: 'equals',
+	EXCLUSIVE: 'exclusive'
+};
 
 /**
  * this function check if we have icons to display
@@ -45,29 +53,6 @@ export function getCaptionsValue(captionsLength, min, max) {
 	const captionsValue = range(min, max, interval);
 	captionsValue.push(max);
 	return captionsValue;
-}
-
-/**
- * Return correct modifier styles for usage mode
- * @param {string} mode
- * @param {number} value
- */
-export function getModeStyles(mode, value) {
-	if (Array.isArray(value)) {
-		if (mode === 'exclusive') {
-			return 'exclusive';
-		} else {
-			return false;
-		}
-	} else {
-		if (mode === 'greaterThan') {
-			return 'greaterThan';
-		} else if (mode === 'equals') {
-			return 'equals';
-		} else {
-			return false;
-		}
-	}
 }
 
 /**
@@ -241,7 +226,7 @@ class Slider extends React.Component {
 		captionTextStepNumber: PropTypes.number,
 		min: PropTypes.number.isRequired,
 		max: PropTypes.number.isRequired,
-		mode: PropTypes.bool,
+		mode: PropTypes.string,
 		captionsFormat: PropTypes.func,
 		disabled: PropTypes.bool,
 	};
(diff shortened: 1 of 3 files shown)


packages/components/src/Slider/Slider.component.js
```diff
@@ -218,6 +218,8 @@ class Slider extends React.Component {
 		captionTextStepNumber: PropTypes.number,
 		min: PropTypes.number.isRequired,
 		max: PropTypes.number.isRequired,
+		rightTrack: PropTypes.bool,
```
frassinier, line 221: what do you think about a `trackPosition` (by default set to' Left') ['Right', 'Left', 'None'] instead?

packages/components/src/Slider/Slider.scss
```diff
@@ -86,4 +86,25 @@ $tc-slider-line-height: 24px;
...
+		:global(.rc-slider-rail),
+		:global(.rc-slider-track) {
+			background-color: $alto;
+		}
+
+		:global(.rc-slider-handle) {
+			box-shadow: $tc-slider-thumb-shadow;
+			background-color: $brand-primary;
+		}
+	}
```
frassinier, line 109:
```suggestion
&-rc-slider {
	&--track-right {
		:global {
			.rc-slider-rail {
				background-color: $brand-primary;
			}
			.rc-slider-track {
				background-color: $alto;
			}
		}
	}

	&--track-none {
		:global {
			.rc-slider-rail,
			.rc-slider-track {
				background-color: $alto;
			}

			.rc-slider-handle {
				box-shadow: $tc-slider-thumb-shadow;
				background-color: $brand-primary;
			}
		}
	}
}
```

packages/components/src/Slider/Slider.stories.js
```diff
@@ -78,6 +78,26 @@ storiesOf('Form/Controls/Slider', module).add('default', () => (
 				<p>By default</p>
 				<Slider onChange={action('onChange')} />
 			</div>
+			<div style={delimiterStyle}>
+				<p>Track on the right</p>
+				<Slider
+					max={10}
+					min={0}
+					rightTrack={true}
```
frassinier, line 86:
```suggestion
					rightTrack
```

packages/components/src/Slider/Slider.stories.js
```diff
@@ -78,6 +78,26 @@ storiesOf('Form/Controls/Slider', module).add('default', () => (
...
+					onChange={action('onChange')}
+					value={3}
+				/>
+			</div>
+			<div style={delimiterStyle}>
+				<p>No track</p>
+				<Slider
+					max={10}
+					min={0}
+					noTrack={true}
```
frassinier, line 96:
```suggestion
					noTrack
```

packages/components/src/Slider/Slider.component.js
jsomsanith-tlnd: Agree, instead of multiplying the props for 1 purpose, let's enable different modes via 1 prop.
I would prefer `props.mode` with different values (`lowerThan`, `between`, `greaterThan`, `equals`). We need to speak the same concept/language as the UX in the guideline

olliew-talend: Thank you both. I had considered this after and I agree, one prop is cleaner. In my opinion, since this is a visual change only, the styles should not have an implied context and should be generic. On that basis, I would prefer `left, right, none` instead of `lessThan, GreaterThan, equals`.

If we do this, `left` will be the default prop value but it will have no dedicated modifier class unless we want that? If we do, it will be a breaking change. 

For range, this was not in scope for my ticket so the behaviour is not currently modified. There is an issue where you could apply modifiers to a range but I will fix this. If we want to handle the range like `left, right`, we could have `inner, outer` to support an inclusive or exclusive range. The final 'mode' options would be:

```
- none
- left
- right
- inner
- outer
```

jsomsanith-tlnd: Can you explain the `since this is a visual change only` ?
What is plained for the behavior (non-visual part) ? Are you planning to return the value or the corresponding range depending on the mode ?
Either way, instead of mapping visual values to the real meaning (left = less than the selected value), I think it's clearer to have directly the meaning. And if there is some behavior change that are planned, we can directly use those values too, no mental mapping.
But if you both agree on the left/right values, no blocker, let's go with that.

olliew-talend: There will be no change to the functional behaviour of the slider. It will return a value or range as normal which can be interpreted by the app in whatever context is needed. 

Okay, so maybe we can have:
```
- lessThan
- equals
- greaterThan
- inclusive (for range)
- exclusive (for range)
```

frassinier: sure, go for it!

olliew-talend:
@frassinier @jsomsanith-tlnd: Can we see if this is looking better, at least for the basic approach? Thanks!

packages/components/src/Slider/Slider.stories.js
```diff
@@ -109,11 +136,22 @@ storiesOf('Form/Controls/Slider', module).add('default', () => (
...
+					allowCross={false}
+				/>
+			</div>
+			<div style={delimiterStyle}>
+				<p>with range (exclusive)</p>
 				<Slider
 					onChange={action('onChange')}
 					min={0}
 					max={100}
+					mode={'exclusive'}
```
frassinier, line 154: Please use exported constant here instead
```suggestion
					mode={SLIDER_MODE.EXCLUSIVE}
```

packages/components/src/Slider/Slider.scss
```diff
@@ -85,5 +85,41 @@ $tc-slider-line-height: 24px;
 				box-shadow: $tc-slider-thumb-shadow-focused;
 			}
 		}
+
+		&--track-greater-than {
```
frassinier, line 89:
```suggestion
		&--track-greater-than,
		&--track-exclusive {
```

packages/components/src/Slider/Slider.scss
```diff
@@ -85,5 +85,41 @@ $tc-slider-line-height: 24px;
...
+		&--track-exclusive {
+			:global {
+				.rc-slider-rail {
+					background-color: $brand-primary;
+				}
+				.rc-slider-track {
+					background-color: $alto;
+				}
+			}
+		}
```
frassinier, line 123:
```suggestion
```

packages/components/src/Slider/Slider.component.js
```diff
@@ -47,6 +47,29 @@ export function getCaptionsValue(captionsLength, min, max) {
...
+	} else {
+		if (mode === 'greaterThan') {
+			return 'greaterThan';
+		} else if (mode === 'equals') {
+			return 'equals';
+		} else {
+			return false;
+		}
+	}
+}
```
jsomsanith-tlnd, line 71: Not sure what this function is for ?
Can't you just use the mode value to set the classname ?

packages/components/src/Slider/Slider.component.js
```diff
@@ -218,6 +241,7 @@ class Slider extends React.Component {
 		captionTextStepNumber: PropTypes.number,
 		min: PropTypes.number.isRequired,
 		max: PropTypes.number.isRequired,
+		mode: PropTypes.bool,
```
frassinier, line 244: nope

packages/components/src/Slider/Slider.component.js
```diff
@@ -254,7 +280,16 @@ class Slider extends React.Component {
 						min={min}
 						max={max}
 						handle={noValue ? undefined : this.state.handle}
-						className={classnames(theme['tc-slider-rc-slider'], 'tc-slider-rc-slider')}
+						className={classnames(
+							theme['tc-slider-rc-slider'],
+							{[theme['tc-slider-rc-slider--track-equals']]: modeStyles === 'equals'},
```
frassinier, line 285: You can use a function instead to determinate the right class

packages/components/src/Slider/Slider.component.js
```diff
@@ -47,6 +47,29 @@ export function getCaptionsValue(captionsLength, min, max) {
 	return captionsValue;
 }
 
+/**
+ * Return correct modifier styles for usage mode
+ * @param {string} mode
+ * @param {number} value
+ */
+export function getModeStyles(mode, value) {
```
frassinier, line 55: Wow, can you simplify this? And use contants like `SLIDER_MODE`

packages/components/src/Slider/Slider.component.js
olliew-talend: 🤦

packages/components/src/Slider/Slider.component.js
olliew-talend: Something like `SLIDER_MODE` could be used alone but wouldn't stop someone calling a range slider and applying `equals` styling which would probably not be wanted. I don't know if we need to worry about protecting against this. If not then it can be simpler. 

We could alternatively put this protection in the CSS if we give the slider a .range class and use `:not`

jsomsanith-tlnd: I don't think this is necessary, it adds a level of complexity, to try to prevent a development usage error. But the dev error will be directly visible.
What do you think ?

olliew-talend: I agree. I'll refactor this and maybe just document in the stories that `inclusive/exclusive` are valid for range sliders.

packages/components/src/Slider/Slider.component.js
olliew-talend: We won't do the integrity check for mode type now so classes can be added inline by checking the value of `mode` in the normal way with classnames.

packages/components/src/Slider/Slider.component.js
olliew-talend: We'l remove this and just check `mode` inline with classnames.

olliew-talend:
@frassinier @jsomsanith-tlnd: Have another look, more comments welcome! Thanks.

packages/components/src/Slider/Slider.component.js
```diff
@@ -254,7 +264,16 @@ class Slider extends React.Component {
 						min={min}
 						max={max}
 						handle={noValue ? undefined : this.state.handle}
-						className={classnames(theme['tc-slider-rc-slider'], 'tc-slider-rc-slider')}
+						className={classnames(
```
romainseb, line 267: Could you use getTheme like in 
https://github.com/Talend/ui/blob/cbd4f128b72705c7a7f3ea98cdf8bb1a96941ca6/packages/components/src/PieChart/PieChartIcon.component.js#L95

It allows to duplicate the classnames declaration :)

packages/components/src/Slider/Slider.stories.js
```diff
@@ -2,8 +2,9 @@ import React from 'react';
 import { storiesOf } from '@storybook/react';
 import { action } from '@storybook/addon-actions';
 
-import Slider from './Slider.component';
+import Slider, {SLIDER_MODE} from './Slider.component';
 import IconsProvider from '../IconsProvider';
+// import SLIDER_MODE from './Slider.component';
```
romainseb, line 7:
```suggestion
```

packages/components/src/Slider/Slider.component.js
```diff
@@ -254,7 +264,16 @@ class Slider extends React.Component {
...
+							theme['tc-slider-rc-slider'],
+							{[theme['tc-slider-rc-slider--track-equals']]: mode === SLIDER_MODE.EQUALS},
+							{[theme['tc-slider-rc-slider--track-exclusive']]: mode === SLIDER_MODE.EXCLUSIVE},
+							{[theme['tc-slider-rc-slider--track-greater-than']]: mode === SLIDER_MODE.GREATER_THAN},
+							'tc-slider-rc-slider',
+							{'tc-slider-rc-slider--track-equals': mode === SLIDER_MODE.EQUALS},
+							{'tc-slider-rc-slider--track-exclusive': mode === SLIDER_MODE.EXCLUSIVE},
+							{'tc-slider-rc-slider--track-greater-than': mode === SLIDER_MODE.GREATER_THAN}
+						)}
```
frassinier, line 276: We need to see together how to refactor it

packages/components/src/Slider/Slider.stories.js
jsomsanith-tlnd, line 7: You can remove the comment

olliew-talend: 🤦

olliew-talend: 🤦

jsomsanith-tlnd: That happens to the bests

packages/components/src/Slider/Slider.component.js
olliew-talend: I'll take a look!

olliew-talend:
I will merge this now with approvals. I will continue to look for improvements based on the  outstanding suggestions.

PR #2881, opened 2020-07-03
fix(BasicSearch): set overflow prop from "auto" to "visible" to hide  scrollbar for IE and Edge.

hbhong:
The BasicSearch component shows an additional scrollbar for IE and Edge browser by default.
Jira link: https://jira.talendforge.org/browse/TMDM-14670

To update the corresponding CSS , set overflow prop from "auto" to "visible" 


packages/faceted-search/src/components/BasicSearch/BasicSearch.scss
```diff
@@ -2,7 +2,7 @@
 	align-items: center;
 	display: flex;
 	height: inherit;
-	overflow: auto;
+	overflow: visible;
```
jsomsanith-tlnd, line 5: This introduces a new behavior: no scroll anymore. If you have tons of badges, it will just take more place, instead of scrolling.
Is it something that is validated with UX ?

(Not related, but a question: why do we have a keight: inherit just above ?)

hbhong: I append some badges, and then reduce the width of the browser,  it wraps the badges to next line, is it acceptable?

jsomsanith-tlnd: ☝️ `Is it something that is validated with UX ?`

hbhong: I asked Maxime Nicoul, and waiting for his reponse.

hbhong: @jsomsanith  I confirmed with  Maxime Nicoul, when set it to "visible", with many badges, it would wrap to multi-line to display all filters,  this behavior is acceptable.

PR #2899, opened 2020-07-20
fix(components): ActionIconToggle now handle the ref to the button

tlnd-mhuchet:
Unlike `ActionButton` or `ActionDropdown`, the `ActionIconToggle` component does not manage the ref to the button.
But for example, if you want to use this kind of button with a popover, the ref is needed to display the popover at the right place.

Add the missing `buttonRef` prop to `ActionIconToggle`, keeping the name used in other components.

### DIFF
--- a/packages/components/src/Actions/ActionIconToggle/ActionIconToggle.component.js
+++ b/packages/components/src/Actions/ActionIconToggle/ActionIconToggle.component.js
@@ -19,6 +19,7 @@ function ActionIconToggle(props) {
 		id,
 		label,
 		tooltipPlacement,
+		buttonRef,
 		...rest
 	} = props;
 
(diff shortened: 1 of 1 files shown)


tlnd-mhuchet:
> I guess you mean tooltip, not popover right ? A popover needs a click to display the popover block, but the click the button click :)
> It's too bad that we don't just use a forwardRef, but to ensure that the code is homogenous, let's stay with that.

I really mean overlay/popover.

To be precise, it is related with the addition of an icon in the preview panel in Pipeline Designer. This icon open/close a menu. It is the menu that is handle in the popover. And to be correctly place aside the icon button, it needs the ref.

You can see it in this image: https://jira.talendforge.org/secure/attachment/236676/proposal%202.png
It is the "three vertical dots" icon in the upper right of the bottom panel, with the menu opened and containg one element with a toggle.

And yeah, it should be better with forwardRef, but i took the choice to stick with the same behaviour than other Action components for the moment :)

PR #2910, opened 2020-07-22
fix(faceted-search): cannot apply filter for a decimal field in IE11

htyin:
Can not apply filter for a decimal field
Update input type number to numberDecimal

### DIFF
--- a/packages/faceted-search/CHANGELOG.md
+++ b/packages/faceted-search/CHANGELOG.md
@@ -24,6 +24,10 @@ Types of changes
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [unreleased]
+
+- [fixed](https://github.com/Talend/ui/pull/2910): IE11 : update input type number to numberDecimal
+
 ## [0.10.3]
 
 - [fixed](https://github.com/Talend/ui/pull/2881): IE11: hide scrollbar for BasicSearch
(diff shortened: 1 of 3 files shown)


jsomsanith-tlnd:
Hi @htyin,
I updated the title of the PR to follow the guideline. Be careful next time please.
I don't understand your fix. The Text form widget takes the type and set it as html input element `type` attribute. `number` should work with decimal when you have an `<input type="number" />`, and `numberDecimal` doesn't exist.
You can try it on the number field [here](http://talend.surge.sh/forms/?path=/story/uiform-v2-core-fields--core-text)

And I don't see the relation with IE11 (that is present in the title)

htyin:
> Hi @htyin,
> I updated the title of the PR to follow the guideline. Be careful next time please.
> I don't understand your fix. The Text form widget takes the type and set it as html input element `type` attribute. `number` should work with decimal when you have an `<input type="number" />`, and `numberDecimal` doesn't exist.
> You can try it on the number field [here](http://talend.surge.sh/forms/?path=/story/uiform-v2-core-fields--core-text)
> 
> And I don't see the relation with IE11 (that is present in the title)

htyin:
Hi @jsomsanith 
The issue is that Decimal value is not valid for number type, only happens on IE.
You are right, numberDecimal doesn't work as expected. I need continue investigate.

PR #2923, opened 2020-08-06
feat(components/Enumeration): Support custom actions for Enumeration item

yyanwang:
In TMC we have a use case of Enumeration to custom actions base on item data. Currently Enumeration doesn't support custom actions, all items share actions props.
Let user pass in a function instead of array, which returns custom actions base on item data.
Story: http://2923.talend.surge.sh/components/?path=/story/form-controls-enumeration--with-custom-actions

### DIFF
--- a/packages/components/src/Enumeration/Items/Items.component.js
+++ b/packages/components/src/Enumeration/Items/Items.component.js
@@ -32,9 +32,13 @@ class Items extends React.PureComponent {
 
 		switch (item.displayMode) {
 			case DISPLAY_MODE_EDIT: {
+				let actions = this.props.itemsProp.actionsEdit;
+				if (typeof actions === 'function') {
+					actions = actions(itemWithIndex);
+				}
 				itemWithIndex.itemProps = {
 					key: this.props.itemsProp.key,
-					actions: this.props.itemsProp.actionsEdit,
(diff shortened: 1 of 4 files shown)


packages/components/src/Enumeration/Items/Items.component.js
```diff
@@ -51,9 +55,13 @@ class Items extends React.PureComponent {
...
 			default: {
+				let actions = this.props.itemsProp.actionsDefault;
+				if (typeof actions === 'function') {
+					actions = actions(itemWithIndex);
+				}
 				const itemPropDefault = {
 					key: this.props.itemsProp.key,
-					actions: this.props.itemsProp.actionsDefault,
+					actions,
 					onSelectItem: this.props.itemsProp.onSelectItem,
```
VolodymyrKovalM, line 65: looks like changes in both cases are completely the same and can be done once above switch

yyanwang: The difference is when in default mode, it uses `actionsDefault`; in edit mode, it uses `actionsEdit`.

VolodymyrKovalM: ah, yes, good

PR #2932, opened 2020-08-26
feat(facetedsearch/sliderfacet): pointer cursor on the slider facet value

ncomont:

the user does not understand that the value is clickable

pointer cursor on the slider facet value


packages/faceted-search/src/components/Badges/BadgeSlider/BadgeSlider.scss
```diff
@@ -22,6 +22,7 @@
 			}
 
 			.tc-badge-value-unit {
+				cursor: pointer;
```
jsomsanith-tlnd, line 25: This should be a button. This is not a pointer because it's a span, so it's not accessible (no pointer, no focus, no activation via keyboard)
The right fix is to turn it into a button and fix the style, you'll have your pointer by default

jsomsanith-tlnd (approved):

frassinier (approved):

PR #2934, opened 2020-08-26
feat(components/HeaderBar): Accept custom Intercom component

sgangnant:
We want to be able to inject a custom Intercom component in the HeaderBar.

Accept a new prop `Intercom` that receives the rendered Intercom component. This prop has precedence over the `intercom` configuration object.

### DIFF
--- a/packages/components/src/HeaderBar/HeaderBar.test.js
+++ b/packages/components/src/HeaderBar/HeaderBar.test.js
@@ -42,6 +42,16 @@ describe('HeaderBar', () => {
 		expect(element).not.toBeUndefined();
 	});
 
+	it('should render custom Intercom component', () => {
+		function Intercom() {
+			return null;
+		}
+
+		const wrapper = mount(<HeaderBarComponent Intercom={Intercom} />);
+		const element = wrapper.find(Intercom);
+		expect(element).not.toBeUndefined();
+	});
+
 	it('should render search', () => {
 		const search = {
 			id: 'search',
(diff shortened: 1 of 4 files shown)


packages/components/src/HeaderBar/HeaderBar.test.js
```diff
@@ -42,6 +42,16 @@ describe('HeaderBar', () => {
 		expect(element).not.toBeUndefined();
 	});
 
+	it('should render custom Intercom component', () => {
+		function Intercom() {
+			return null;
+		}
+
+		const wrapper = mount(<HeaderBarComponent Intercom={Intercom} />);
```
jsomsanith-tlnd, line 50: Here you pass the component and not the rendered component. But check my general comment first.

jsomsanith-tlnd (commented):
Question: is it possible to have the same api as app switcher in your case ? Instead of taking the rendered component in props, take the component, that the headerbar renders. My concern is to have the same api for the same purpose.

packages/components/src/HeaderBar/HeaderBar.test.js
sgangnant: Sure, I thought it was more convenient to have it already rendered, but it also makes sense to have the same API than any other injected component 👍

jsomsanith-tlnd (approved):

VolodymyrKovalM (approved):

frassinier (approved):

PR #2955, opened 2020-09-14
chore: update versions script and upgrade

jmfrancois:

versions script fail on non existing folder.
dependencies.js is a script that can t be shared easly outside.

* Update the script so it works
* launch yarn upgrade on UI so the lock file is aligned and up to date
* put the dependencies in a JSON file so it is a static file
* fix all tests

Note: the two snapshots updated are clearly a bug in the current master branch.
The new jest need to resolve mock of uuid in the current package so I have duplicate the mock. 


packages/components/package.json
```diff
@@ -64,7 +64,7 @@
     "simulant": "^0.2.2",
     "styled-components": "^4.1.3",
     "timezone-support": "^1.5.5",
-    "uuid": "^3.0.1"
+    "uuid": "^7.0.3"
```
frassinier, line 67: there was a breaking change for this one if I remember

frassinier (approved):

jsomsanith-tlnd (approved):

packages/components/package.json
jmfrancois: after dive into it. The only breaking change is uuid() do not exists anymore. 
All our code except ine one place use uuid.v4() which is compatible

PR #2973, opened 2020-09-29
feat(ARCH-203/IconsProvider): remove icons from react bundle

jmfrancois:

Currently we embed all icons in the bundle. A first effort has been done using ICON_BUNDLE env variable which has never be stable. Some project has tried without success.

We don t want the svg icons in the JS bundle.

Change the IconsProvider props to only accept an array of URLS.
URLS will be fetched and put as it is in the DOM.
This change is a breaking change but only for the IconsProvider component.
Icon component continue to work as it is.

In some case icons exists in atomic way and provided by the backend (a cms for example).
For this kind of case we have added the `remote-` prefix in the name option.
This will fetch the icon and put the result in the DOM using native browser API.
This leverage the http browser cache API.


jsomsanith-tlnd:
Replaced by #3061 that has 6.0 brach as target

PR #2990, opened 2020-10-09
fix(containers/notification): attach actions to Notification

jsomsanith-tlnd:
We added actions creators to Notification containers to be able to write `Notification.actions`.
But the content is undefined because we default import nothing as they are named exported.

`import * as actions from './Notification.actions'`

I changed the name from `actionCreators` to `actions` to align the names. It's not a breaking change because it never worked, so I guess never used, so it doesn't break any code.


jmfrancois (commented):
👍

jmfrancois (approved):

romainseb (approved):

PR #3005, opened 2020-10-16
fix(http): Return full error in http handle request

vbocquet:
There is an issue in `@talend/http` here: https://github.com/Talend/ui/blob/master/packages/http/src/http.common.js#L69
Xe can’t pass the response to Error, because it only take a string message as param. Behind, we can’t exploit this response (to get status for example).

In current CMF http sagas it works because it’s a promise, so we can pass anything: https://github.com/Talend/ui/blob/master/packages/cmf/src/sagas/http.js#L90

To fix this, we simply instance an error with the request status, and the complete response in a custom property and only throw the error after that.

I add some tests to cover the fix.

Thanks @jmfrancois for the help on this 👍

### DIFF
--- a/packages/http/src/http.common.test.js
+++ b/packages/http/src/http.common.test.js
@@ -117,6 +117,31 @@ describe('#handleHttpResponse', () => {
 			done();
 		});
 	});
+
+	it('should handle the response with an error http code', async () => {
+		await expect(
+			handleHttpResponse(
+				new Response('', {
+					status: HTTP_STATUS.INTERNAL_SERVER_ERROR,
+				}),
+			),
+		).rejects.toThrow(`${HTTP_STATUS.INTERNAL_SERVER_ERROR}`);
+	});
+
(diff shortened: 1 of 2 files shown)


jmfrancois (approved):

packages/http/src/http.common.test.js
```diff
@@ -117,6 +117,31 @@ describe('#handleHttpResponse', () => {
...
 	});
+
+	it('should handle the response with an error http code', async () => {
+		await expect(
+			handleHttpResponse(
+				new Response('', {
+					status: HTTP_STATUS.INTERNAL_SERVER_ERROR,
+				}),
+			),
+		).rejects.toThrow(`${HTTP_STATUS.INTERNAL_SERVER_ERROR}`);
```
jsomsanith-tlnd, line 128: you don't need the string literal here :D

jsomsanith-tlnd (approved):

PR #3014, opened 2020-10-22
feat(ListComposition): filter taking into account only visible columns

VolodymyrKovalM:
Currently ListComposition goes through all object fields when filtering collection, which results in not correct results, as examples:
- it goes through such field as createDate & updateDate, resulting incorrect filter results if entering digits in text filter
- it goes through 'actions' & 'persistentActions' on item object, which result all items shown if entering "object" in text filter (array.toString() - [object object])
- it goes through other object fields which are not displayed in raw format (only mapped to corresponding translations)
This results in big issues when filtering on all lists that are without lazy loading (with lazy loading filtering is performed on server side), because in such lists filter does not work good in many cases  

it's possible to form custom filterFunctions and return false for object fields that do not need to participate in filtering (so skip these fields), but this is not nice solution at all, rather workaround, especially if you have 3 or more types of entities with different models shown in one list, this way need to "skip" a lot of fields, moreover if the object model change in data base and backend (new attribute added) - there will be issues again

So, I would suggest to have possibility to filter taking into account only fields that are visible in list columns (this will be especially useful with column chooser, since filtering should be performed only on chosen columns that are currently visible), if not providing visible columns to list manager props everything will work as before

### DIFF
--- a/packages/components/src/List/ListComposition/Manager/ListManager.component.js
+++ b/packages/components/src/List/ListComposition/Manager/ListManager.component.js
@@ -9,14 +9,7 @@ import { useCollectionSort } from './hooks/useCollectionSort.hook';
 import { useCollectionFilter } from './hooks/useCollectionFilter.hook';
 import theme from '../List.scss';
 
-function Manager({
-	initialDisplayMode,
-	initialSortParams,
-	visibleColumns,
-	children,
-	t,
-	...rest
-}) {
+function Manager({ initialDisplayMode, initialSortParams, visibleColumns, children, t, ...rest }) {
 	let collection = rest.collection;
 
 	const [displayMode, setDisplayMode] = useState(initialDisplayMode || displayModesOptions[0]);
--- a/packages/components/src/List/ListComposition/Manager/ListManager.component.js
+++ b/packages/components/src/List/ListComposition/Manager/ListManager.component.js
@@ -9,7 +9,14 @@ import { useCollectionSort } from './hooks/useCollectionSort.hook';
 import { useCollectionFilter } from './hooks/useCollectionFilter.hook';
 import theme from '../List.scss';
 
-function Manager({ initialDisplayMode, initialSortParams, visibleColumnsKeys, children, t, ...rest }) {
+function Manager({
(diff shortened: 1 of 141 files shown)


lmaillet (approved):
Great improvement 👍

packages/components/src/List/ListComposition/Manager/ListManager.component.js
```diff
@@ -9,7 +9,14 @@ import { useCollectionSort } from './hooks/useCollectionSort.hook';
 import { useCollectionFilter } from './hooks/useCollectionFilter.hook';
 import theme from '../List.scss';
 
-function Manager({ initialDisplayMode, initialSortParams, children, t, ...rest }) {
+function Manager({
+	initialDisplayMode,
+	initialSortParams,
+	visibleColumnsKeys,
```
frassinier, line 15: `visibleColumns` should be enough

packages/components/src/Typeahead/Typeahead.component.js
```diff
@@ -177,6 +177,7 @@ Typeahead.propTypes = {
 	onToggle: PropTypes.func,
 	docked: PropTypes.bool,
 	icon: PropTypes.shape({
+		tooltipPlacement: PropTypes.string,
```
frassinier, line 180: not related to the PR but ok :)

packages/components/src/List/ListComposition/Manager/ListManager.component.js
```diff
@@ -22,7 +22,10 @@ function Manager({ initialDisplayMode, initialSortParams, children, t, ...rest }
 	collection = sortedCollection;
 
 	// Filter by text
-	const { filteredCollection, textFilter, setTextFilter } = useCollectionFilter(collection);
+	const { filteredCollection, textFilter, setTextFilter } = useCollectionFilter(
+		collection,
```
frassinier, line 26: The definition has changed ⚠️

```suggestion
		collection,
    undefined,
    undefined,
```

frassinier (approved):

PR #3030, opened 2020-10-29
feat(ARCH-203): add Icon support for remote svg

jmfrancois:

Current Icon do not support remote svg.
It support src to leverage `img` tag.

This PR is part of #2973 

add `remote-URL` into the name as src- but with a different behavior:
The component will fetch the content and if this is a svg it will add it to the DOM

```<Icon name="remote-/foo/bar.svg"/>```

### DIFF
--- a/packages/components/src/Icon/Icon.component.js
+++ b/packages/components/src/Icon/Icon.component.js
@@ -3,7 +3,6 @@ import React from 'react';
 import classnames from 'classnames';
 
 import theme from './Icon.scss';
-import IconsProvider from '../IconsProvider';
 
 const FA_TRANSFORMS = {
 	spin: 'fa-spin',
@@ -104,8 +103,6 @@ function Icon({ className, name, title, transform, onClick, ...props }) {
 				if (isRemote && content && content.includes('svg')) {
 					// eslint-disable-next-line no-param-reassign
 					ref.innerHTML = content;
-				} else if (!content && !isRemote) {
-					IconsProvider.injectIcon(name, ref);
 				}
 			}}
 			{...accessibility}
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: packages/components/src/Icon/Icon.test.js [hunk 1]
```diff
@@ -37,4 +50,50 @@ describe('Icon', () => {
...
+		// });
+		// expect(container.querySelector('svg')).toMatchSnapshot();
+		expect(global.fetch).toHaveBeenCalledTimes(1);
+		expect(global.fetch).toHaveBeenCalledWith('/assets/icons/my-icon.svg', {
+			headers: { Accept: 'image/svg+xml' },
+		});
+		const instance = wrapper.children().getElement();
+		const ref = {};
+		instance.ref(ref);
+		expect(ref.innerHTML).toMatchSnapshot();
```
jmfrancois, line 97: I didn t find any other way to make the ref callback called ... I am opened to any other option

romainseb (approved):
Don't know if it would work with the storybook sandbox, but it could be great to add a story about this 😉

packages/components/src/Icon/Icon.component.js
```diff
@@ -63,17 +92,41 @@ function Icon({ className, name, title, transform, onClick, ...props }) {
...
+			className={classname}
+			ref={ref => {
+				if (!ref) {
+					return;
+				}
+				if (isRemote && content && content.includes('svg')) {
+					// eslint-disable-next-line no-param-reassign
+					ref.innerHTML = content;
+				}
+			}}
```
jsomsanith-tlnd, line 107: Please use `useRef` and `useEffect` to set things in the ref if it is initialized

packages/components/src/Icon/Icon.test.js
jsomsanith-tlnd: The idea is to get the resulting html, why don't we use [Enzyme .html()](https://enzymejs.github.io/enzyme/docs/api/ShallowWrapper/html.html).

Something like `wrapper.find('svg').html()`

romainseb (approved):

jsomsanith-tlnd (approved):

PR #3071, opened 2020-11-16
fix(IconsProvider): make it work without context 

jsomsanith-tlnd:
The IconsProvider via context doesn't work in angular apps which has react->ng and ng->react layers. The context is broken.

Make it work without context
- the iconsProvider share a common list of fetching bundles
- the injectIcon() get the icon is it is already in dom, otherwise it waits for all bundles to be fetched to search again and inject the icon.

### DIFF
--- a/packages/components/6.0 - breaking changes.md
+++ b/packages/components/6.0 - breaking changes.md
@@ -174,21 +174,6 @@ And `context` is now attached to `Tile` component.
 +const href = IconsProvider.getIconHREF(iconName);
 ```
 
-The IconsProvider is now a context provider, it must wrap your app
-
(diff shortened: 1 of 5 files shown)


frassinier (approved):

packages/components/6.0 - breaking changes.md
```diff
@@ -174,21 +174,6 @@ And `context` is now attached to `Tile` component.
 +const href = IconsProvider.getIconHREF(iconName);
```
jmfrancois, line 174: that is not valid right anymore.

jmfrancois (approved):

PR #3076, opened 2020-11-17
fix(ColumnChooser): do not overwrite onSubmit

yyanwang:
internal onSubmit will be overwritten by props.onSubmit when spread props to ColumnChooserButton.
omit onSubmit when spread props.

### DIFF
--- a/packages/components/src/List/ListComposition/ColumnChooser/ColumnChooser.component.js
+++ b/packages/components/src/List/ListComposition/ColumnChooser/ColumnChooser.component.js
@@ -1,6 +1,5 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import pickBy from 'lodash/pickBy';
 import { useListContext } from '../context';
 import ColumnChooserButton from '../../Toolbar/ColumnChooserButton';
 
@@ -9,6 +8,7 @@ function ColumnChooser(props) {
 
 	return (
 		<ColumnChooserButton
+			{...props}
 			columns={columns.map(({ dataKey, label }, i) => ({
 				key: dataKey,
 				label,
@@ -21,7 +21,6 @@ function ColumnChooser(props) {
 					props.onSubmit(_, changes);
 				}
 			}}
-			{...pickBy(props, (_, key) => key !== 'onSubmit')}
 		/>
 	);
 }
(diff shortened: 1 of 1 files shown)


packages/components/src/List/ListComposition/ColumnChooser/ColumnChooser.component.js
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import omit from 'lodash/omit';
```
jmfrancois, line 3: lodash omit is deprecated

jmfrancois (changes requested):
you should use pickBy instead.

```javascript
import pickBy from 'lodash/pickBy';

pickBy(props, (v, key) => keyToOmit.indexOf(key) === -1);
```

packages/components/src/List/ListComposition/ColumnChooser/ColumnChooser.component.js
frassinier: Please remove that

packages/components/src/List/ListComposition/ColumnChooser/ColumnChooser.component.js
```diff
@@ -20,7 +21,7 @@ function ColumnChooser(props) {
 					props.onSubmit(_, changes);
 				}
 			}}
-			{...props}
+			{...omit(props, 'onSubmit')}
```
frassinier, line 24: What you want is to spread props before the onSubmit I guess
```suggestion
			{...props}
			onSubmit={(_, changes) => {
				setVisibleColumns(changes.filter(c => !c.hidden).map(c => c.key));
				if (props.onSubmit) {
					props.onSubmit(_, changes);
				}
			}}
```

packages/components/src/List/ListComposition/ColumnChooser/ColumnChooser.component.js
jsomsanith-tlnd: Instead, you can destructure the props
```
const { onSubmit, ...rest } = props;
```
So you use onSubmit and you pass `rest` to the button

jmfrancois (approved):

jsomsanith-tlnd (approved):

PR #2998, opened 2020-10-13
chore(DatePicker): add minWidth for inputs

jlame:
**What is the problem this PR is trying to solve?**
having issue with inputSizer when the input of the picker is contained in a not expanded collapsible at first render

chrome-capture

**What is the chosen solution to this problem?**
adding a minWidth props in order to be able to set a minimal width for the input

**Please check if the PR fulfills these requirements**

* [ ] The PR commit message follows our [guidelines](https://github.com/talend/tools/blob/master/tools-root-github/CONTRIBUTING.md)
* [ ] Tests for the changes have been added (for bug fixes / features) And [non reg](./screenshots.md) done before need review
* [ ] Docs have been added / updated (for bug fixes / features)
* [ ] Related design / discussions / pages (not in jira), if any, are all linked or available in the PR

**[ ] This PR introduces a breaking change**

PR #3081, opened 2020-11-18
chore(DateTime/Time): adding minWidth support for datetime range and time

jlame:
same as : https://github.com/Talend/ui/pull/2998 
but making it work for datetime range and time

### DIFF
--- a/packages/components/src/DateTimePickers/Time/Input/Input.component.js
+++ b/packages/components/src/DateTimePickers/Time/Input/Input.component.js
@@ -8,9 +8,10 @@ import theme from './Input.scss';
 
 export default function Input(props) {
 	const { time, inputManagement } = useContext(TimeContext);
+    const { minWidth } = props;
 
 	return (
-		<InputSizer inputText={inputManagement.placeholder}>
+		<InputSizer inputText={inputManagement.placeholder} minWidth={minWidth}>
 			{width => (
 				<DebounceInput
 					autoComplete="off"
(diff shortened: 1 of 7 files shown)


jmfrancois:
it doesn t seems to be related to your PR but when I try to pick a date with FF:
Screenshot 2020-11-20 at 08 45 12

packages/components/src/DateTimePickers/Time/Input/Input.component.js
```diff
@@ -8,9 +9,10 @@ import theme from './Input.scss';
 
 export default function Input(props) {
 	const { time, inputManagement } = useContext(TimeContext);
+	const { minWidth } = props;
```
jmfrancois, line 12: please destructure the minwidth so it is not spread to the debounce input and only consume by the sizer

jmfrancois (changes requested):
The size you put seems changed, not a big deal I hope.
Screenshot 2020-11-20 at 09 15 51

I request change for the attribute on the input only

packages/components/src/DateTimePickers/Time/Input/Input.component.js
jlame: done, concerning the question about the +5px when setting minwidth, this is because the inputSizer adds an extra space at the end.

jmfrancois (approved):

ncomont (approved):

PR #3094, opened 2020-11-26
fix(TDC-3764/File): Support base64 encoded file names with multi-byte characters

olliew-talend:
The current atob() method used to decode base64 file names for the file upload component does not support multi-byte characters (e.g. `ć, è`). This results in parsing errors when the file name is displayed in the input value (e.g. `cÌ.xlsx`).

The decoding function is improved based on the SO discussion linked below. This solution includes additional decoding steps to reduce the content to a byte stream and re-encode it to produce the original string. Initial tests indicate that this approach is flexible enough to service multiple base64 encoding approaches and as such should not result in a breaking change if the components is provided a string encoded using simply `window.btoa()`.

In addition, the story for this component has been enhanced to utilise an encoding approach that supports multi-byte characters.

[JIRA issue](https://jira.talendforge.org/browse/TDC-3674)
[SO Discussion](https://stackoverflow.com/questions/30106476)

### DIFF
--- a/packages/forms/src/UIForm/fields/File/File.component.js
+++ b/packages/forms/src/UIForm/fields/File/File.component.js
@@ -18,6 +18,22 @@ const BASE64_NAME = ';name=';
 const BASE64_PREFIX = ';base64,';
 
 /**
+ * Decode the base64 file name with wide character support
+ * @param {string} filename The base64 value of the file name
+ * @returns {string} The decoded file name
+ */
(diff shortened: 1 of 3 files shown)


ncomont (approved):

lmaillet:
Hi Ollie,
Can you add unit tests to be sure that `base64Decode()` works with different chars, single and multi bytes?

lmaillet (approved):

PR #3101, opened 2020-11-30
fix(forms): remove ie11 support to be umd compliant

jmfrancois:

with UMD build react-hook-form/dist/ie11 is not umd so it is embed in the UMD
on the project side this is the same.

but having rhf embed on both bundle make the provider different.

The side effect of this is useFormContext function call return null like if the context was not provided.

Remove ie11 dist import to rely on umd so the code is shared.


silvanocostanzo (commented):
LGTM!

PR #3105, opened 2020-12-01
fix(List/ColumnChooser): "Select all" and columns list filter

sgangnant:
"Select all" option in the lists' column choser should apply to shown columns (e.g. those shown when the user filters the list of columns)
image

Integrate filtering into the column chooser manager and update the manager's behavior accordingly, example :
- Given a list with 3 columns, "_First name_", "_Last name_", "_Login_", all 3 checked (so "_Select all_" is also checked)
- User types in "_name_" in the filter (only "_First name_" and "_Last name_" appear in the list) **→** Unchecks "_Select all_" **→** Apply
- Only "_Login_" column is shown in the list

### DIFF
--- a/packages/components/src/List/Toolbar/ColumnChooserButton/hooks/columnChooserManager.hook.js
+++ b/packages/components/src/List/Toolbar/ColumnChooserButton/hooks/columnChooserManager.hook.js
@@ -71,7 +71,7 @@ const prepareColumns = (columns, lockedLeftItems) =>
 		),
 	);
 
-export const changeColumnChooserAttribute = key => value => column => {
+const changeColumnChooserAttribute = key => value => column => {
 	if (!column.locked) {
 		return { ...column, [key]: value };
 	}
--- a/packages/components/src/List/Toolbar/ColumnChooserButton/hooks/columnChooserManager.hook.js
+++ b/packages/components/src/List/Toolbar/ColumnChooserButton/hooks/columnChooserManager.hook.js
@@ -1,4 +1,4 @@
-import { useState } from 'react';
+import { useState, useMemo } from 'react';
 import flow from 'lodash/flow';
 import cloneDeep from 'lodash/cloneDeep';
 import { compareOrder } from '../service';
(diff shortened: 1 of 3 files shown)


packages/components/src/List/Toolbar/ColumnChooserButton/hooks/columnChooserManager.hook.js
```diff
@@ -1,26 +1,7 @@
...
- * Helps to change the select all visible status. If all columns are visible, select all is checked.
- * @param {array} columns
- */
-const isEveryItemVisible = columns => columns.every(isItemVisible);
-
-const findColumnLabel = label => column => column.label === label;
-
-export const getColumn = label => columns => columns.find(findColumnLabel(label));
-
-export const setColumn = (column, label) => columns => {
```
jmfrancois, line 18: it looks like a breaking change but they do not seems to be used outside. could you please confirm ?

jmfrancois (approved):
The story is working as described and the tests looks good !

packages/components/src/List/Toolbar/ColumnChooserButton/hooks/columnChooserManager.hook.js
sgangnant: I couldn't find a place where it was used outside the components so I assume it's safe to remove it (don't know why it was exported prior this PR, tests ? 🤔)

packages/components/src/List/Toolbar/ColumnChooserButton/hooks/columnChooserManager.hook.js
```diff
@@ -89,33 +71,56 @@ const updateVisibilityAttr = changeColumnChooserAttribute('visible');
...
+export const useColumnChooserManager = (
+	initialColumns = [],
+	nbLockedLeftItems = 0,
+	initialFilterValue = '',
+) => {
+	const [columns, setColumns] = useState(() => prepareColumns(initialColumns, nbLockedLeftItems));
+
+	const [textFilter, setTextFilter] = useState(initialFilterValue);
+
+	const filteredColumns = useMemo(
```
frassinier, line 83: Ok not the best way to filter but it was just moved

frassinier (approved):

PR #3128, opened 2020-12-15
feat(cmf): add http error attributes from RFC 7807

sgendre:

Current way to handle errors from HTTP endpoint is completly custom on every projects.
CMF do not take part on that so if you have a custom error code you want to describe you have nothing.

Rely on RFC7807 to get something standardized
https://tools.ietf.org/html/rfc7807


jmfrancois (approved):

jmfrancois:
I am not sure for the custom attribute you called custom. in the RFC they talk about some custom attribute not one attribute named custom but ok

sgendre:
> I am not sure for the custom attribute you called custom. in the RFC they talk about some custom attribute not one attribute named custom but ok.

yes indeed, my goal is to avoid each project to add his own properties and do related PR, by this way it is generic and business attributes are left to be managed by product

jlame (approved):

PR #3129, opened 2020-12-16
feat(TDC-3961): Text mode for file input fields

olliew-talend:
The file input component does not currently register a text-only display mode which results in missing information when viewing pre-filled form fields in read-only contexts.

Register a new default text mode function for file input fields.

**Notes:**
Any forms currently used with a display mode of text, that include a file input, will now automatically show the contents of the file input in text mode. This results in more content being rendered to the screen than may have been intended or designed for. In addition, it will result in the display of the file name value of the input, which may not be desirable in some cases.


packages/forms/src/UIForm/utils/widgets.js
```diff
@@ -56,7 +56,7 @@ const widgets = {
 	code_text: CodeTextMode,
 	comparator_text: TextModeComparator,
 	datalist_text: DatalistTextMode,
-	file_text: () => null,
+	file_text: FileTextMode,
```
frassinier, line 59: Not a breaking change imho, you simply show what was previously hidden

frassinier (approved):

packages/forms/src/UIForm/utils/widgets.js
olliew-talend: Thanks buddy. I'm happy to not call it a BC if that's the consensus. I suppose I look at this kind of thing as if it's a new feature which forces an unrequested change in behaviour then it is breaking to the experience, even if it doesn't cause errors.

frassinier: Actually, imagine CSS changes, they could be UI/UX huge changes but the apps will continue to work without any developer manipulations 🤷

olliew-talend: Agree, which is why I consider CSS changes to be breaking too! :grin:

I suppose it's just the definition of 'breaking' as 'causes errors' or 'impacts the experience'. I suppose my view is a result of my background :shrug:

ncomont (approved):

PR #3164, opened 2021-01-15
feat(components): tags

frassinier:
Tags come from Bootstrap and it's time to get rid of Bootstrap

Tags come from the Design System


packages/components/package.json
```diff
@@ -34,6 +34,7 @@
     "url": "https://github.com/Talend/ui.git"
   },
   "dependencies": {
+    "@talend/design-system": "^1.0.0-beta.12",
```
jmfrancois, line 37: beta do not work with CDN

yarn.lock
```diff
@@ -2264,6 +2264,11 @@
   dependencies:
     "@types/node" ">= 8"
 
+"@popperjs/core@^2.5.4":
```
jmfrancois, line 2267: what about react-popper migration first ?

packages/components/package.json
```diff
@@ -72,6 +73,7 @@
     "@storybook/addons": "^5.3.1",
     "@storybook/react": "^5.3.1",
     "@talend/bootstrap-theme": "^6.9.0",
+    "@talend/design-system": "^1.0.2",
```
jmfrancois, line 76: it s already a dep

packages/components/src/TabBar/TabBar.component.js
```diff
@@ -164,41 +164,57 @@ function TabBar(props) {
...
-						eventKey={item.key}
-						componentClass="button"
-					>
-						<TooltipTrigger
-							label={badge && badge.label ? `${badge.label} ${item.label}` : item.label}
-							tooltipPlacement={tooltipPlacement}
+				{items.map(({ icon, badge, ...item }) => {
+					let StyledTag = Tag;
+					switch (badge?.bsStyle?.toLowerCase()) {
+						case 'danger':
```
Kporal, line 170: Any chance to use `Status.component.js` (talend/ui/packages/components/src/Status/Status.component.js) to compute the style and reduce hardcoded strings?

frassinier: Not really an option, `Status` and `Tag` are two different components

packages/components/__mocks__/@talend/design-system.js
```diff
@@ -0,0 +1,37 @@
...
+		mocks[name][variation].displayName = variationName;
+	};
+}
+
+function mockComponent(name) {
+	mocks[name] = getFakeJSX(name);
+	mocks[name].displayName = name;
+	Object.keys(Coral[name])
+		.filter(startsWithUpperCase)
+		.forEach(mockComponentVariation(name));
```
jsomsanith, line 30: This part creates the attached components mock but you don't attach them to the main mocked component

packages/components/src/Tag/Tag.component.js
```diff
@@ -0,0 +1,35 @@
...
+			StyledTag = CoralTag.Warning;
+			break;
+		case 'danger':
+			StyledTag = CoralTag.Destructive;
+			break;
+		default:
+			break;
+	}
+	return <StyledTag {...rest} />;
+};
```
jsomsanith, line 27: Why this `new` component ? Shouldn't it be a simple proxy to DS instead of offering a new api ? I mean it's not a component that exists today in tui, so we are not stucked with an existing api

frassinier: Ok let's get rid of it!

packages/components/__mocks__/@talend/design-system.js
frassinier: _What do you mean?_ 🎵

packages/components/src/Tag/Tag.component.js
frassinier: I can't, actually! If I don't have this component as a proxy, I need to duplicate the mapping bootstrap v3 -> Design System several times!

jsomsanith-tlnd (approved):

Kporal (approved):

yarn.lock
```diff
@@ -2264,11 +2278,16 @@
...
   version "2.6.0"
   resolved "https://registry.yarnpkg.com/@popperjs/core/-/core-2.6.0.tgz#f022195afdfc942e088ee2101285a1d31c7d727f"
   integrity sha512-cPqjjzuFWNK3BSKLm0abspP0sp/IGOli4p5I5fKFAzdS8fvjdOwDCfZqAaIiXd9lPkOWi3SUUfZof3hEb7J/uw==
 
+"@popperjs/core@^2.6.0":
+  version "2.7.0"
+  resolved "https://registry.yarnpkg.com/@popperjs/core/-/core-2.7.0.tgz#5f7760965170f7f1e2910d2498c5be88b47a2a8c"
+  integrity sha512-V3WyEc8ZyAuOQ2fpFuTuYYOd2tV4NePeSdxaHYgYAOs7ERLxlcFi2XsmgI5LJFdAUmJKXsg8jaIiVTkTHQygQw==
+
```
jmfrancois, line 2290: please take care to remove this lib.
I think a simple npx yarn-dedup should do the job but I am afraid it increase a lot the changes in the yarn lock :D

jmfrancois (approved):

PR #3178, opened 2021-01-21
fix(components/ListView): showToggleAll prop change

sgangnant:
When a parent changes the `showToggleAll` prop in its `ListView` component, nothing happens because actual effect is the result of prop computation that happens in component's construct, not at render time.

Move computation at render time. Also, expose the component's various display modes and remove an intermediate component that does nothing.

### DIFF
--- a/packages/components/src/ListView/ListView.component.js
+++ b/packages/components/src/ListView/ListView.component.js
@@ -25,7 +25,7 @@ function ListView(props) {
 		<div className={listviewClasses()}>
 			<HeaderListView {...props} />
 			{props.items.length ? (
-				<ItemsListView {...props} />
+				<Items {...props} />
 			) : (
 				<span className={theme['empty-message']}>{label}</span>
 			)}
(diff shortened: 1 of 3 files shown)


packages/components/src/ListView/ListView.component.js
```diff
@@ -33,6 +33,8 @@ function ListView(props) {
 	);
 }
 
+ListView.DISPLAY_MODES = { DISPLAY_MODE_DEFAULT, DISPLAY_MODE_SEARCH };
```
sgangnant, line 36: Caller can now use the actual component's display modes

packages/components/src/ListView/ListView.component.js
```diff
@@ -25,14 +25,16 @@ function ListView(props) {
 		<div className={listviewClasses()}>
 			<HeaderListView {...props} />
 			{props.items.length ? (
-				<ItemsListView {...props} />
+				<Items {...props} />
```
sgangnant, line 28: Directly call `Items` instead of the wrapper that only spreads the received props

frassinier (approved):

tlnd-mhuchet (approved):

PR #3185, opened 2021-01-22
feat(forms/CollapsibleFieldSet): allow definition of fallback title

sgendre:
TPD had his own implementation of title function, which had a hard coded fallback value: when computed title is empty text,  "Undefined" was returned.

Allow a fallback value through form definition or widget creation.


packages/forms/src/UIForm/fieldsets/CollapsibleFieldset/CollapsibleFieldset.component.js
```diff
@@ -55,6 +55,10 @@ export function defaultTitle(formData, schema, options) {
 	if (title.length > 0) {
 		return title.join(options?.separator || schema.options?.separator || ', ');
 	}
+
+	if (options?.emptyTitleFallback) {
+		return options.emptyTitleFallback || schema.options?.emptyTitleFallback;
```
romainseb, line 60: i think `options.emptyTitleFallback` will always be return with the current condition. Not sure this is intended

tlnd-mhuchet (approved):
Have you a lead to understand why the linter is crazy in the CI?

jmfrancois (approved):
really strange error in the lint if you can find the source it would be great

romainseb:
A fix for the linter issue is there : https://github.com/Talend/ui/pull/3192
The issue here is the fact that there is no lint anymore in this package and at some point (some commit before) the file has been updated this way cause there were some missing bracket or something like that

PR #3196, opened 2021-01-27
fix(CellDatetime): fix displaying if there is no date or it is invalid

VolodymyrKovalM:
format from date-fns does not throw error if date is not valid, it just returns string "Invalid date", so the error won't be caught in try catch block, and "Invalid date" will be directly displayed in list

parse the cell data, check if it's valid, and return cell data back if it's not valid date


jmfrancois (approved):
good. Could you also add a new line that show the behavior in the example on storybook ?
http://3196.talend.surge.sh/components/?path=/story/data-list-list--list-cell-renderer

VolodymyrKovalM (commented):
added two rows with invalid and no date
unit test that renders without date already exists in CellDatetime.test, so I added one more case with invalid date there

jmfrancois (approved):

jsomsanith-tlnd (approved):

PR #3201, opened 2021-02-01
feat(components): add data-test & data-feature props to reset btn

sgendre:
QA wants to target reset button easily

Add a data-test and a data-feature prop that will end up in data-test and data-feature attribute in html elements


jsomsanith-tlnd (commented):
Why don't you just spread rest props to the main element ?
ActionButton is a button so passing extra props should end up end up on the button.
The filter has an input as main element so, same here.

sgendre:
> Why don't you just spread rest props to the main element ?
> ActionButton is a button so passing extra props should end up end up on the button.
> The filter has an input as main element so, same here.

Because those specific props are kebab-case where as usual props are camelCase

kebab-case is used by html data-attributes, why those 2 concepts are mixed at the moment?

jsomsanith-tlnd (approved):

jmfrancois (approved):

PR #3090, opened 2020-11-23
feat(cmf): add bootstrap option to not render the app

jmfrancois:
**What is the problem this PR is trying to solve?**

Current non react project which use cmf, need to bootstrap all the needed libraries by hand and can t leverage cmfModule

**What is the chosen solution to this problem?**

add option to do not render project but still merge all cmfModules and provide all the needed internals to setup it in angular for example.

**Please check if the PR fulfills these requirements**

* [x] The PR commit message follows our [guidelines](https://github.com/talend/tools/blob/master/tools-root-github/CONTRIBUTING.md)
* [x] Tests for the changes have been added (for bug fixes / features) And [non reg](./screenshots.md) done before need review
* [x] Docs have been added / updated (for bug fixes / features)
* [ ] Related design / discussions / pages (not in jira), if any, are all linked or available in the PR

**[ ] This PR introduces a breaking change**

PR #3217, opened 2021-02-18
fix(cmf): export httpInterceptors

jmfrancois:

In the context of moving our http call from gateway to public API we need to be able to configure every clients at once with interceptors.
But they are not exported

export it


jmfrancois:
I prefer the approach of #3090 which is more consistent

PR #3221, opened 2021-02-22
feat(components/CellDatetime): allow formatting to specific timezone

VolodymyrKovalM:
Currently not possible to format to specific timezone using CellDatetime

Implement possibility to format to specific timezone in CellDatetime

### DIFF
--- a/packages/components/src/VirtualizedList/CellDatetime/CellDatetime.component.js
+++ b/packages/components/src/VirtualizedList/CellDatetime/CellDatetime.component.js
@@ -6,6 +6,7 @@ import pick from 'lodash/pick';
 import { distanceInWordsToNow, format } from 'date-fns';
 import isValid from 'date-fns/is_valid';
 import parse from 'date-fns/parse';
+import { formatToTimeZone } from 'date-fns-timezone';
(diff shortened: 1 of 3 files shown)


packages/components/src/VirtualizedList/CellDatetime/CellDatetime.component.js
```diff
@@ -45,24 +51,43 @@ export class CellDatetimeComponent extends React.Component {
...
 
-		return cell;
+		return (
+			<TooltipTrigger
+				label={columnData.tooltipLabel || computedValue}
+				tooltipPlacement={columnData.tooltipPlacement || 'bottom'}
+			>
+				{cell}
+			</TooltipTrigger>
+		);
```
VolodymyrKovalM, line 90: I think displaying tooltip with the same value as inside cell may be useful for cases when the date is not fully visible due to smaller column size
We have such case in TMC on new operations list (where we have 18 possible columns to display)

jmfrancois: Need UX approval.
cc @frassinier

PR #3230, opened 2021-02-25
fix(DatePicker): ref error

jsomsanith-tlnd:
On date selection in date picker, there is an error because the reference are not defined.
In fact the references in the component are not clear, there are duplications and only one of the duplications is set. That causes the problem of null object because all of them are used in parts of the code.

Remove the duplications, set only 1 ref for the input et 1 ref for the popper div.

### DIFF
--- a/packages/components/src/DateTimePickers/InputDatePicker/InputDatePicker.component.js
+++ b/packages/components/src/DateTimePickers/InputDatePicker/InputDatePicker.component.js
@@ -1,4 +1,4 @@
-import React, { useRef, useState } from 'react';
+import React, { useState } from 'react';
 import PropTypes from 'prop-types';
 import omit from 'lodash/omit';
 import uuid from 'uuid';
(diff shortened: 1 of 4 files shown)


packages/components/src/DateTimePickers/InputDatePicker/InputDatePicker.component.js
```diff
@@ -85,18 +83,17 @@ export default function InputDatePicker(props) {
...
 			value={props.value}
 			dateFormat={props.dateFormat}
-			onChange={(...args) => handlers.onChange(...args, inputRef.current)}
+			onChange={(...args) => handlers.onChange(...args, referenceElement)}
 			useUTC={props.useUTC}
 			timezone={props.timezone}
 		>
 			<FocusManager
 				className={classnames(theme['date-picker'], 'date-picker')}
-				divRef={containerRef}
```
jsomsanith-tlnd, line 94: This is a ref to set on the div containing the input and the popper. But we don't use it, so let's not set a ref on it here.

frassinier (approved):

smouillour (approved):

PR #3256, opened 2021-03-24
chore(cmf): upgrade sentry

jmfrancois:

sentry API is v5 and only for pure JS.
Since then Sentry has published a sentry for react in their v6 API.

upgrade CMF to v6

Crappy work: Because of the bundle impact sentry has, I propose to expose @sentry/react has a global variable `SentryReact`. Sentry is now really modular and Sentry variable refer to `@sentry/browser`. They propose bundle for all integration except react with only all in one bundles.

So you can use the doc and API just replace Sentry by SentryReact global.


jmfrancois:
Have an issue with code shared and CDN :/

PR #3267, opened 2021-04-06
feat(components/QualityBar): not applicable

jlame:
adding another metric in the quality bar which is not applicable see : https://jira.talendforge.org/browse/TDS-5711

adding it 

### DIFF
--- a/packages/components/src/QualityBar/QualityBar.component.js
+++ b/packages/components/src/QualityBar/QualityBar.component.js
@@ -5,6 +5,7 @@ import RatioBar from '../RatioBar';
 import {
 	QualityEmptyLine,
 	QualityInvalidLine,
+	QualityNotApplicableLine,
 	QualityType,
 	QualityValidLine,
 } from './QualityRatioBar.component';
@@ -72,6 +73,20 @@ export function QualityBar({ invalid, valid, empty, onClick, getDataFeature, dig
 						? e =>
 								onClick(e, {
 									type: QualityType.EMPTY,
+								})
+						: null
+				}
+				dataFeature={getDataFeature ? getDataFeature(QualityType.EMPTY) : null}
+				value={empty}
+				percentage={emptyPercentage}
+				t={t}
+			/>
+			<QualityNotApplicableLine
+				onClick={
+					onClick
+						? e =>
+								onClick(e, {
+									type: QualityType.QualityNotApplicableLine,
 								})
 						: null
 				}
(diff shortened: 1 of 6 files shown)


packages/components/src/QualityBar/QualityBar.component.js
```diff
@@ -14,17 +15,18 @@ import I18N_DOMAIN_COMPONENTS from '../constants';
...
  * @param {number} invalidRaw number of invalid raw
- * @param {number} emptyRaw number of invalid raw
- * @param {number} validRaw number of invalid raw
+ * @param {number} emptyRaw number of empty raw
+ * @param {number} validRaw number of valid raw
+ * @param {number} naRaw number of not applicable raw
  * @param {number} digits number of digits we want to keep
  */
-export function getQualityPercentagesRounded(invalid, empty, valid, digits = 0) {
+export function getQualityPercentagesRounded(invalid, empty, valid, na, digits = 0) {
```
vbocquet, line 23: maybe set a default value to na at 0, just to avoid regression?

vbocquet (approved):

packages/components/src/QualityBar/QualityBar.component.js
jlame: good catch

packages/components/src/QualityBar/QualityBar.component.js
```diff
@@ -14,17 +15,18 @@ import I18N_DOMAIN_COMPONENTS from '../constants';
...
  * @param {number} invalidRaw number of invalid raw
- * @param {number} emptyRaw number of invalid raw
- * @param {number} validRaw number of invalid raw
+ * @param {number} emptyRaw number of empty raw
+ * @param {number} validRaw number of valid raw
+ * @param {number} naRaw number of not applicable raw
  * @param {number} digits number of digits we want to keep
  */
-export function getQualityPercentagesRounded(invalid, empty, valid, digits = 0) {
+export function getQualityPercentagesRounded(invalid, empty, valid, na = 0, digits = 0) {
```
frassinier, line 23: Are you sure this fn is not used outside? By project? If so, be careful with breaking changes!

jlame: the default value covers that case to avoid breaking change ?

jlame: Concerning the function signature, doesn't seem like it's used in other project. so i guess we should be fine

PR #3271, opened 2021-04-12
fix(container): context.store.dispatch not a function

jmfrancois:

we have on some event handler the following error:
context.store.dispatch is not a function

This is due to the call to the old actionAPI script in a mapstatetoprops.
Yes we don t have store.dispatch in mapstatetoprops.

There is no risk here because the code were not working before so we just get ride of that uncaught JS error.
Check if we have context.store.dispatch before adding onclick

### DIFF
--- a/packages/containers/src/actionAPI.js
+++ b/packages/containers/src/actionAPI.js
@@ -1,4 +1,5 @@
 import cmf from '@talend/react-cmf';
+import get from 'lodash/get';
 
 /**
  * add support for expression in actions.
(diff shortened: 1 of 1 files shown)


packages/containers/src/actionAPI.js
```diff
@@ -36,19 +36,23 @@ export function getActionsProps(context, ids, model) {
...
-					model,
-					...info.payload,
-				});
-			}
-		},
-		...evalExpressions(info, context, { model }),
-	}));
+	const props = infos.map(info => {
+		const newprops = {};
+		if (context && context.store && context.store.dispatch) {
```
smouillour, line 41: is it possible to use lodash.get here?

smouillour (approved):

Kporal (approved):

PR #3281, opened 2021-04-29
fix(components/List/CellActions): fix safari compatibility issue for linear-gradient

yyanwang:
A piece of shadow appears when hover on selected row.
https://jira.talendforge.org/browse/TMC-18091
image

This is because safari doesn't support some format in linear-gradient. 
- use rgba color instead of string
- use `90deg` instead of `to right`

### DIFF
--- a/packages/components/src/VirtualizedList/CellTitle/CellTitle.scss
+++ b/packages/components/src/VirtualizedList/CellTitle/CellTitle.scss
@@ -101,7 +101,7 @@ $tc-list-title-icon-size: $svg-rg-size !default;
 	&:focus,
 	&:global(.ally-focus-within) {
 		:global(.tc-list-title .cell-title-actions) {
-			background: linear-gradient(to right, transparent, $tc-list-row-active-hover-bg $padding-large);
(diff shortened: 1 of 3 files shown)


jsomsanith-tlnd (approved):

ybaskaran (approved):

packages/components/src/VirtualizedList/CellTitle/CellTitle.scss
```diff
@@ -101,7 +101,7 @@ $tc-list-title-icon-size: $svg-rg-size !default;
 	&:focus,
 	&:global(.ally-focus-within) {
 		:global(.tc-list-title .cell-title-actions) {
-			background: linear-gradient(to right, transparent, $tc-list-row-active-hover-bg $padding-large);
+			background: linear-gradient(90deg, rgba(255, 255, 255, 0), rgba($tc-list-row-active-hover-bg, 100%) $padding-large);
```
frassinier, line 104: Only transparent should be replaced,  isn't it?

frassinier (approved):

packages/components/src/VirtualizedList/CellTitle/CellTitle.scss
yyanwang: indeed.

PR #3287, opened 2021-05-06
fix(forms/nestedListView): search functionality does not work properly

akaseka:
Search functionality is not working properly
https://jira.talendforge.org/browse/TPSVC-16547

Searching not only in the children elements


jmfrancois (changes requested):
Please add a testcase that prove the error which is fixed by your change

packages/forms/src/UIForm/fields/NestedListView/NestedListView.utils.js
```diff
@@ -15,25 +15,34 @@ function hasChildren(item) {
  * @returns {Array}
  */
 export function getDisplayedItems(items, value, searchCriteria) {
-	let textFilter;
-	if (searchCriteria) {
-		textFilter = ({ label }) => label.toLowerCase().includes(searchCriteria.toLowerCase());
-	}
+	const textFilter = searchCriteria ? searchCriteria.toLowerCase() : '';
```
jmfrancois, line 18: this look like a criteria more than a filter (naming issue)

packages/forms/src/UIForm/fields/NestedListView/NestedListView.utils.js
```diff
@@ -15,25 +15,34 @@ function hasChildren(item) {
...
+		}
+
+		const filteredChildren = item.children.filter(child =>
+			child.label.toLowerCase().includes(textFilter),
+		);
+
+		return filteredChildren.length > 0
+			? [...filtered, { ...item, children: filteredChildren }]
+			: filtered;
+	}, []);
```
jmfrancois, line 45: why do we need to reduce twice ? Look quite heavy to reduce a datastructure to then reduce it again.
I am looking forward to the test to look deeper into your algorithm here.
AFAIK we can have lots  of items here, so this is important to stay efficient in this kind of filter

jmfrancois:
your new test passed with code from master.
I would like a test that do not pass to show the issue
Screen Shot 2021-05-07 at 11 21 37

jmfrancois (approved):

PR #3301, opened 2021-05-20
feat(TDOPS-14/FolderTreeView): added disabled list item state

vyshkov:

The common component should be updated:

- to update the style of the active state, hover state
- and introduce the disabled state style

More details: https://jira.talendforge.org/browse/TDOPS-14
Link to frontify: https://company-57688.frontify.com/document/92132#/ui-controls/tree-view

Fixed css

### DIFF
--- a/packages/components/src/TreeView/TreeViewItem/TreeViewItem.component.js
+++ b/packages/components/src/TreeView/TreeViewItem/TreeViewItem.component.js
@@ -252,7 +252,7 @@ class TreeViewItem extends React.Component {
 				<div
 					className={classNames('tc-treeview-item', css['tc-treeview-item'], {
 						[css.disabled]: disabled,
-						'disabled': disabled,
+						disabled: disabled,
 					})}
 					style={{ paddingLeft }}
 				>
--- a/packages/components/src/TreeView/TreeViewItem/TreeViewItem.component.js
+++ b/packages/components/src/TreeView/TreeViewItem/TreeViewItem.component.js
@@ -236,7 +236,8 @@ class TreeViewItem extends React.Component {
 					return !disabled && onSelect(e, item);
 				}}
 				onKeyDown={e =>
-					!disabled && onKeyDown(e, this.containerRef, {
+					!disabled &&
+					onKeyDown(e, this.containerRef, {
 						...item,
 						hasChildren: children.length,
 						isOpened,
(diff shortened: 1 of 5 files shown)


packages/components/src/TreeView/TreeViewItem/TreeViewItem.component.js
```diff
@@ -213,6 +214,7 @@ class TreeViewItem extends React.Component {
 			actions,
 			icon,
 			counter = children.length,
+			disabled,
```
frassinier, line 217: `disabled` must be passed to the Action, isn't it?

vyshkov: I thought about it, disabling all actions limits our flexibility.  We can set actions to be disabled using action props, and with this approach, disabled list items may contain other clickable actions (info, warning etc) that may be better for the user experience.

jmfrancois: The onselect action is still fired on the demo so I don t understand it s only visual disabled right ?

jmfrancois (changes requested):
The definition said:
When an element cannot be selected the whole line is 0.54 opacity, with a not-allowed cursor.
But the props disabled only apply css style it do not prevent the onSelect to be triggered by a click event

packages/components/src/TreeView/TreeViewItem/TreeViewItem.component.js
vyshkov: Yes, you are right. We can also skip onSelect callback for disabled items. But onToggle should be available for the user. We use it on plan rerun dialog, and the user should be able to see what's inside disabled steps.

jmfrancois (approved):

packages/components/src/TreeView/TreeViewItem/TreeViewItem.component.js
frassinier: Using the keyboard I can still fire the onSelect on the demo, so we must consider modifying the other `onKey*`handlers.

Another thing, we need to get the info that this button is actually `[aria-disabled=true]` if we want to manage it manually!

vyshkov: good catch @frassinier !

frassinier (approved):

PR #3313, opened 2021-05-28
feat(forms/TimezoneList): TimezoneList form field

sgangnant:
There's no common way to build a timezone selection field in forms.

Add a new field.
Widget requires timezones from `cldr-dates-full` until we have a nice way to lazy load such JSON file (see README.md).


packages/forms/src/UIForm/fields/TimezoneList/TimezoneList.component.js
```diff
@@ -0,0 +1,32 @@
+import React, { useMemo } from 'react';
+import PropTypes from 'prop-types';
+
+import Datalist from '../Datalist';
+
+import { getTimezones } from './TimezoneList.utils';
+
+function TimezoneList(props) {
+	const { schema } = props;
+	const { lang, schema: { type } } = schema;
```
sgangnant, line 10: Maybe there's a way to get lang from parent `<Form />` component 💭

packages/forms/src/UIForm/fields/TimezoneList/TimezoneList.utils.js
```diff
@@ -0,0 +1,151 @@
+import get from 'lodash/get';
+
+import frTimezones from 'cldr-dates-full/main/fr/timeZoneNames.json';
+import enTimezones from 'cldr-dates-full/main/en/timeZoneNames.json';
+import jaTimezones from 'cldr-dates-full/main/ja/timeZoneNames.json';
+import deTimezones from 'cldr-dates-full/main/de/timeZoneNames.json';
```
sgangnant, line 6: This might not be the best idea, to load all translations files by default (uncompressed, this is 4 * ~70 = 240kb)

packages/forms/src/UIForm/fields/TimezoneList/TimezoneList.utils.js
```diff
@@ -0,0 +1,151 @@
...
+ */
+function formatUtcOffset(offset, separator) {
+	const sign = offset >= 0 ? '+' : '-';
+
+	const absoluteOffset = Math.abs(offset);
+	const min = absoluteOffset % 60;
+	const hours = (absoluteOffset - min) / 60;
+
+	return `${sign}${padTwoDigits(hours)}${separator}${padTwoDigits(min)}`;
+}
```
sgangnant, line 62: Those helpers already exist in UI/components, there might be a better way to share it
- An independant date utils package seems overkill
- Direct import is dirty
- ?

packages/forms/stories-core/json/fields/core-timezoneList.json
```diff
@@ -0,0 +1,24 @@
+{
+	"doc": "TimezoneList/README.md",
```
sgangnant, line 2: Will be written when solution is validated

packages/forms/stories-core/json/fields/core-timezoneList.json
```diff
@@ -0,0 +1,24 @@
...
+	"jsonSchema": {
+		"type": "object",
+		"properties": {
+			"timezone": {
+				"type": "string"
```
sgangnant, line 7: It could be useful to be able to specify `"type": "object"` and to get the selected timezone's full object (timezone, offset ...). Other way around is to expose timezone helpers.
There's a use case in TMC where the full object is managed in the form submit (might be the only use case though).

packages/forms/src/UIForm/fields/TimezoneList/TimezoneList.utils.js
jmfrancois: Do we need all the content of theses json ?
https://unpkg.com/cldr-dates-full@39.0.0/main/fr/timeZoneNames.json

packages/forms/src/UIForm/fields/TimezoneList/TimezoneList.utils.js
jmfrancois: forms depends on components already

packages/forms/src/UIForm/fields/TimezoneList/TimezoneList.utils.js
sgangnant: Exemplar city is the key we need actually, it's where the city name is translated

packages/forms/src/UIForm/fields/TimezoneList/TimezoneList.utils.js
sgangnant: @frassinier good point
@jmfrancois true, but utils are not exposed (and ... it's a components repo actually 😅 )

jmfrancois: If you create a package that is "shared" because we use UMD we need to make it a UMD to make it really shared.
Look at @talend/http it is embed every where so it is there multiple times.
So may be we should open a track on that on how to share code needed by components taking all constraints into account ?

packages/forms/src/UIForm/fields/TimezoneList/TimezoneList.utils.js
```diff
@@ -0,0 +1,151 @@
...
+
+		case 'ja':
+			zones = jaTimezones.main.ja.dates.timeZoneNames.zone;
+			break;
+
+		case 'en':
+		default:
+			zones = enTimezones.main.en.dates.timeZoneNames.zone;
+			break;
+	}
```
jmfrancois, line 90: const zones
```suggestion
	const zones = ZONES[lang].main[lang].dates.timeZoneNames.zone;
```

packages/forms/src/UIForm/fields/TimezoneList/TimezoneList.utils.js
```diff
@@ -0,0 +1,151 @@
+import get from 'lodash/get';
+
+import frTimezones from 'cldr-dates-full/main/fr/timeZoneNames.json';
+import enTimezones from 'cldr-dates-full/main/en/timeZoneNames.json';
+import jaTimezones from 'cldr-dates-full/main/ja/timeZoneNames.json';
+import deTimezones from 'cldr-dates-full/main/de/timeZoneNames.json';
+
```
jmfrancois, line 7:
```suggestion
const ZONES = {fr: frTimezones, de: deTimezones, en: enTimezones, ja: jaTimezones};
```

jmfrancois (changes requested):

packages/forms/src/UIForm/fields/TimezoneList/TimezoneList.component.js
sgangnant: No "lang"/"language" prop accepted by UI's `<Form />`, `t` not passed to forms widgets, assuming this is not possible atm

packages/forms/stories-core/json/fields/core-timezoneList.json
sgangnant: In the end, `@talend/utils` (which is already a dep needed for this widget) provides a way to get timezone offset from its name.

gjou-tlnd (approved):

jmfrancois (approved):
we will see to update this widget once we have the assets API

PR #3319, opened 2021-06-07
feat(components/DatePicker): Disabled dates can't be selected

yyanwang:
Currently we can select any date from date picker. In TMC we would like to disable certain days so user can't select them. For example, we want to disable past days when schedule a task.
guidelines: https://company-57688.frontify.com/document/92132#/ui-controls/date-time-picker-1561102408
Add disabled checker for date picker. You can pass in a func to decide whether a date is disabled.


jmfrancois (commented):
I prefer the positive approach: use `validate` instead of isDisabled if possible.

In terms of responsability because it is a validation process I would prefer to use form validation process instead of adding feature to the picker.

Also there is a validation package in ui, may be we could add check on date here.

packages/components/src/DateTimePickers/DateTime/datetime-extraction.js
```diff
@@ -217,7 +217,7 @@ function updatePartsOnDateChange(datePickerPayload, time, options) {
 		date: date || dateTextInput,
 		time,
 		datetime,
-		textInput: dateAndTimeToStr(date || dateTextInput, time, options),
+		textInput: dateAndTimeToStr(dateTextInput || date, time, options),
```
jmfrancois, line 220: I am worried about this swap. It looks to me extracted date is more important than text input.
Are you sure about this ?

jmfrancois (approved):

packages/components/src/DateTimePickers/DateTime/datetime-extraction.js
yyanwang: Yes, the swap solve a particular problem: when user manual input an invalid date (eg. '2021-06-xx'), we can still get the user input ('2021-06-xx 12:00'). date will be an invalid Date object so can't get the date string('Invalid Date 12:00').
image

ybaskaran (approved):

PR #3332, opened 2021-06-15
feat(containers): Support async multisection datalist

tlnd-mhuchet:
Multisection datalist are not supported in `resolveNameFortitleMap` because the `titleMap` has a different shape.

In "classic" datalist, the `titleMap` is an array of objects with this shape:
```javascript
{ name: "Readable label", value: "someUglyId" }
```

In a multisection datalist, the `titleMap` is an array of objects with this shape:
```javascript
{
  title: "Section title",
  suggestions: [
    { name: "Readable label", value: "someUglyId" },
    { name: "Another label", value: "someUglyId2" },
    ...
  ]
}
```

Add the support of multisection datalist by checking the presence of `suggestions`. If we find it, we assume it is a multisection and handle it. If not we keep the current behavior for classic datalists.


jmfrancois (commented):
please add a unit test

tlnd-mhuchet:
> please add a unit test

Not one, but three! 😄 
Tried to cover the sames cases as we do for no multi section lists.

romainseb (approved):

packages/containers/src/ComponentForm/ComponentForm.component.js
```diff
@@ -46,15 +46,25 @@ export function resolveNameForTitleMap({ schema, properties, value }) {
...
-	const names = uniformValue
-		.map(nextValue => schema.titleMap.find(titleMap => titleMap.value === nextValue))
-		.map(entry => entry && entry.name);
+	const names = schema.titleMap
+		.map(item => {
+			if (item.suggestions) {
+				const suggestionItem = item.suggestions.find(subItem =>
+					uniformValue.includes(subItem.value),
+				);
+				return suggestionItem ? suggestionItem.name : undefined;
```
gjou-tlnd, line 60:
```suggestion
				return suggestionItem?.name;
```

packages/containers/src/ComponentForm/ComponentForm.test.js
```diff
@@ -117,7 +117,7 @@ describe('ComponentForm', () => {
 			// then
 			expect(properties.my.awesome).toEqual({
 				value: ['correct value', 'neither'],
-				$value_name: ['Yes this is the name', 'Neither this one'],
+				$value_name: ['Neither this one', 'Yes this is the name'],
```
jmfrancois, line 120: why value name is reversed compared to the value ?

jmfrancois (changes requested):
the order of value name seems wrong

packages/containers/src/ComponentForm/ComponentForm.test.js
```diff
@@ -41,7 +41,7 @@ describe('ComponentForm', () => {
 		});
 	});
 
-	describe('#resolveNameForTitleMap', () => {
+	describe.only('#resolveNameForTitleMap', () => {
```
tlnd-mhuchet, line 44:
```suggestion
	describe('#resolveNameForTitleMap', () => {
```

packages/containers/src/ComponentForm/ComponentForm.test.js
tlnd-mhuchet: Yes you are right. Good catch.

tlnd-mhuchet:
> the order of value name seems wrong

Should be better now!

packages/containers/src/ComponentForm/ComponentForm.component.js
```diff
@@ -46,14 +46,31 @@ export function resolveNameForTitleMap({ schema, properties, value }) {
...
-
 	// Here we add a field side by side with the value
 	// to keep the title associated to the value
 	const valueIsArray = Array.isArray(value);
 	const uniformValue = valueIsArray ? value : [value];
 
 	const names = uniformValue
-		.map(nextValue => schema.titleMap.find(titleMap => titleMap.value === nextValue))
+		.map(nextValue => {
+			const titleMap = schema.titleMap;
```
sgendre, line 56: please use spread operator

tlnd-mhuchet: Don't be shy, use suggestions 😄 

```suggestion
			const { titleMap } = schema;
```

packages/containers/src/ComponentForm/ComponentForm.component.js
```diff
@@ -46,14 +46,31 @@ export function resolveNameForTitleMap({ schema, properties, value }) {
...
 	// Here we add a field side by side with the value
 	// to keep the title associated to the value
 	const valueIsArray = Array.isArray(value);
 	const uniformValue = valueIsArray ? value : [value];
 
 	const names = uniformValue
-		.map(nextValue => schema.titleMap.find(titleMap => titleMap.value === nextValue))
+		.map(nextValue => {
+			const { titleMap } = schema;
+			const isMultiSection = !!titleMap[0]?.suggestions;
```
jmfrancois, line 57: please move those two var outside of the map

jmfrancois (approved):

PR #3341, opened 2021-06-17
fix(TDS-6053) : Sorting on number not working

abdelillah-tech:
sorting on number is messed up because of one condition (0 = false)

checking that the value is not undefined

### DIFF
--- a/packages/containers/src/List/selector.js
+++ b/packages/containers/src/List/selector.js
@@ -65,7 +65,7 @@ export function configureGetFilteredItems(configure) {
 
 export function compare(sortBy) {
 	return (a, b) => {
-		if (a.get(sortBy)) {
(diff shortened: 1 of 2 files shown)


jlame (approved):

packages/containers/src/List/selector.js
```diff
@@ -65,7 +65,7 @@ export function configureGetFilteredItems(configure) {
 
 export function compare(sortBy) {
 	return (a, b) => {
-		if (a.get(sortBy)) {
+		if (a.get(sortBy) !== undefined) {
```
jsomsanith-tlnd, line 68: You'll have the same issue with null. In fact the comparison in this block seems to work only for string and number. What about testing that the value is a string or a number ?

jsomsanith-tlnd (approved):

PR #3345, opened 2021-06-21
fix(ActionButton): remove empty tooltip when no label

Kporal:

When using ActionButton component for displaying only an icon (without labels) with `hideLabel`, prevent an empty tooltip to be displayed.

image

Review the tooltip condition.

### DIFF
--- a/packages/components/src/Actions/ActionButton/ActionButton.component.js
+++ b/packages/components/src/Actions/ActionButton/ActionButton.component.js
@@ -206,7 +206,7 @@ function ActionButton(props) {
 		);
 	}
 
-	if (hideLabel || tooltip || tooltipLabel) {
+	if ((tooltipLabel || label) && (hideLabel || tooltip)) {
 		btn = (
 			<TooltipTrigger
 				label={tooltipLabel || label}
--- a/packages/components/src/Actions/ActionButton/ActionButton.test.js
+++ b/packages/components/src/Actions/ActionButton/ActionButton.test.js
@@ -226,6 +226,22 @@ describe('Action', () => {
 		expect(wrapper.getElement()).toMatchSnapshot();
 	});
 
+	it('should render tooltip when there is a label with hideLabel property', () => {
+		// when
+		const wrapper = shallow(<ActionButton name="custom_name" label="a label" hideLabel />);
+
+		// then
+		expect(wrapper.getElement()).toMatchSnapshot();
+	});
+
+	it('should NOT render tooltip when there is an empty label with hideLabel property', () => {
+		// when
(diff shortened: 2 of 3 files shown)


frassinier (changes requested):
This is an a11y issue actually if there is no label at all! 😬 
Please, make sure there is at least an aria-label on the button

Kporal:
> This is an a11y issue actually if there is no label at all! 😬
> Please, make sure there is at least an aria-label on the button

Good point, thanks Fabien!

packages/components/src/Actions/ActionButton/ActionButton.component.js
```diff
@@ -206,7 +206,8 @@ function ActionButton(props) {
 		);
 	}
 
-	if (hideLabel || tooltip || tooltipLabel) {
+	const hasLabel = tooltipLabel || label;
+	if (hasLabel && ((hideLabel && tooltip !== false) || tooltip)) {
```
frassinier, line 210: It's not enough?
```suggestion
	if (hasLabel && (tooltip || hideLabel)) {
```

frassinier (approved):

packages/components/src/Actions/ActionButton/ActionButton.component.js
Kporal: Nope because if you set explicitly the `tooltip` as `false` with `hideLabel` as `true` on your component then you don't want the tooltip whereas in your suggestion the tooltip will be displayed. I've added a test to cover that case 👍

packages/components/src/Actions/ActionButton/ActionButton.component.js
```diff
@@ -206,7 +206,8 @@ function ActionButton(props) {
...
 		btn = (
 			<TooltipTrigger
 				label={tooltipLabel || label}
```
ncomont, line 213: value of the const defined l. 209 could be used (if renamed of course)

packages/components/src/Actions/ActionButton/ActionButton.component.js
```diff
@@ -206,7 +206,8 @@ function ActionButton(props) {
 		);
 	}
 
-	if (hideLabel || tooltip || tooltipLabel) {
+	const hasLabel = tooltipLabel || label;
```
ncomont, line 209: i don't understand this condition. to me the following proposition is ok but maybe i miss something

image

ncomont: if i undestood, the only difference between your needs and the current code is to make `tooltip !== false` prior to the rest

ncomont (approved):

PR #3353, opened 2021-06-28
fix(TDS-6053) : Sorting on number not working

abdelillah-tech:
sort on number in this format (number1/number2) doesn't work

change the string to list of two number and sort by number1


jlame:
maybe that's a little bit to specific to our needs on TDS side, it shouldnt be in that repo. Instead we should look for a way to pass a custom sort function.

jlame (changes requested):
see comment

PR #3358, opened 2021-06-29
feat(Dates): translate date formats using Intl

lmaillet:
**What is the problem this PR is trying to solve?**
The dates are not well translated.
image
Could be great to have a common way of translating the dates.

**What is the chosen solution to this problem?**
Provide a common way to translate the dates using Intl

**Please check if the PR fulfills these requirements**

* [ ] The PR commit message follows our [guidelines](https://github.com/talend/tools/blob/master/tools-root-github/CONTRIBUTING.md)
* [ ] Tests for the changes have been added (for bug fixes / features) And [non reg](./screenshots.md) done before need review
* [ ] Docs have been added / updated (for bug fixes / features)
* [ ] Related design / discussions / pages (not in jira), if any, are all linked or available in the PR

**[ ] This PR introduces a breaking change**

PR #3364, opened 2021-07-07
feat(dates): export FORMAT

lmaillet:
I forgot to export the list of date formats :man_facepalming:

### DIFF
--- a/packages/utils/src/date/index.ts
+++ b/packages/utils/src/date/index.ts
@@ -208,7 +208,7 @@ const options: DateFormatOptions = {
  * @param lang {string} language
  * @returns The formated date
  */
-export function format (date: DateFnsFormatInput, dateOption: string, lang: string): string {
+export function format(date: DateFnsFormatInput, dateOption: string, lang: string): string {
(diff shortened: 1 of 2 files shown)


packages/utils/src/date/index.ts
```diff
@@ -217,7 +217,7 @@ export default {
 	convertToTimeZone,
 	convertToUTC,
 	format,
-	options,
+	FORMAT,
```
vbocquet, line 220: But you remove the options export?

lmaillet: I don't want [options](https://github.com/Talend/ui/blob/master/packages/utils/src/date/index.ts#L197) to be exposed, only [FORMAT](FORMAT).
See my previous comment [here](https://github.com/Talend/ui/pull/3358#discussion_r662147649).

vbocquet: understand 👍 seems fine if it's not used for now, if so, it will be a blocker

vbocquet (approved):

jsomsanith-tlnd (approved):

PR #3370, opened 2021-07-27
fix(utils/dates): fix formatTimeZoneTokens function for Safari

yyanwang:
Function `formatTimeZoneTokens` uses regex with lookbehind identifier, which Safari currently doesn't support.
Use lookahead instead.

### DIFF
--- a/packages/utils/src/date/index.ts
+++ b/packages/utils/src/date/index.ts
@@ -81,7 +81,7 @@ export function formatReadableUTCOffset(offset: number): string {
  * @see https://github.com/prantlf/date-fns-timezone/blob/master/src/formatToTimeZone.js#L131
  */
 function formatTimeZoneTokens(dateFormat: string, timeZone: string): string {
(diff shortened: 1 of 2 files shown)


packages/utils/src/date/index.ts
```diff
@@ -81,7 +81,7 @@ export function formatReadableUTCOffset(offset: number): string {
  * @see https://github.com/prantlf/date-fns-timezone/blob/master/src/formatToTimeZone.js#L131
  */
 function formatTimeZoneTokens(dateFormat: string, timeZone: string): string {
-	return dateFormat.replace(/(?<!\[)(z|ZZ?)/g, match => {
+	return dateFormat.replace(/(z|ZZ?)(?!\])/g, match => {
```
yyanwang, line 84: There's already a unit test for this logic:
https://github.com/Talend/ui/blob/master/packages/utils/src/date/date.test.ts#L119

tlnd-mhuchet (approved):

vyshkov (approved):

PR #3378, opened 2021-08-05
feat(forms/FieldTemplate): Support extra custom props for label

sgangnant:
There's no way to pass a prop to the label in the field template

Support `labelProps` in `Label` props — that can be passed via the schema, via field compnoent — that is spread into `Label`
Factorize tests default props

### DIFF
--- a/packages/forms/src/UIForm/fields/FieldTemplate/FieldTemplate.component.test.js
+++ b/packages/forms/src/UIForm/fields/FieldTemplate/FieldTemplate.component.test.js
@@ -4,18 +4,20 @@ import { shallow } from 'enzyme';
 import FieldTemplate from './FieldTemplate.component';
 
 describe('FieldTemplate', () => {
+	const defaultProps = {
+		isValid: true,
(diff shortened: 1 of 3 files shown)


packages/forms/src/UIForm/fields/FieldTemplate/FieldTemplate.component.test.js
```diff
@@ -4,18 +4,20 @@ import { shallow } from 'enzyme';
...
 describe('FieldTemplate', () => {
+	const defaultProps = {
+		isValid: true,
+		description: 'My awesome description',
+		descriptionId: 'myAwesomeField-description',
+		errorId: 'myAwesomeField-error',
+		errorMessage: 'This is wrong o_o',
+		id: 'myAwesomeField',
+		label: 'My awesome label',
+	};
```
sgangnant, line 15: `defaultProps` spread to avoid redundancy and show only what changes in each case

ncomont (approved):

vbocquet (approved):

PR #3392, opened 2021-08-18
feat(UIForm): Support custom root tag

gjou-tlnd:

It's not possible to use an UIform as part of another form as nesting form is not allowed in HTML

Support custom html tag when rendering the form

### DIFF
--- a/packages/forms/src/UIForm/UIForm.component.js
+++ b/packages/forms/src/UIForm/UIForm.component.js
@@ -348,7 +348,7 @@ export class UIFormComponent extends React.Component {
 				id={this.props.id}
 				method={this.props.method}
 				name={this.props.name}
-				noValidate={this.props.noHtml5Validate}
+				noValidate={this.props.as === 'form' ? this.props.noHtml5Validate : undefined}
 				onReset={this.props.onReset}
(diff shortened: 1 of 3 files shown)


frassinier (approved):

packages/forms/src/UIForm/UIForm.component.js
```diff
@@ -336,8 +336,10 @@ export class UIFormComponent extends React.Component {
...
 		};
 
+		const Element = this.props.as;
+
 		return (
-			<form
+			<Element
 				acceptCharset={this.props.acceptCharset}
 				action={this.props.action}
 				autoComplete={this.props.autoComplete}
```
jmfrancois, line 345: you should not pass not valid props to custom element

jmfrancois (changes requested):
If you use div for example the div tag will have accept-charset, action and auto-complete attribute which are not valid.

gjou-tlnd:
> If you use div for example the div tag will have accept-charset, action and auto-complete attribute which are not valid.

it's up to the developer to not provide these props when using a div

jmfrancois (approved):

PR #3418, opened 2021-09-03
fix(forms/NestedListView): Component should now detect props change

Gbacc:
When using a NestedListView component in a form, if we change the values passed as a prop outside of a user interaction, the component won't take it into account.

Add to the component _componentDidUpdate_ lifecycle a condition to check if the props have been changed and does not match the internal values

* [x] The PR have used `yarn changeset` to a request a release from the CI if wanted.

### DIFF
--- a/packages/forms/src/UIForm/fields/NestedListView/NestedListView.test.js
+++ b/packages/forms/src/UIForm/fields/NestedListView/NestedListView.test.js
@@ -82,10 +82,10 @@ describe('NestedListView component', () => {
 			});
 
 			// when : trigger a props update
-			const allValues = {};
-			props.schema.items.forEach(({ key, titleMap }) => {
-				allValues[key] = titleMap.map(titleMapItem => titleMapItem.value);
-			});
+			const allValues = props.schema.items.reduce((acc, item) => {
+				acc[item.key] = item.titleMap.map(titleMapItem => titleMapItem.value);
+				return acc;
+			}, {});
 			ReactDOM.render(<NestedListViewWidget {...props} value={allValues} />, node);
 
 			// then
--- a/packages/forms/src/UIForm/fields/NestedListView/NestedListView.test.js
+++ b/packages/forms/src/UIForm/fields/NestedListView/NestedListView.test.js
@@ -1,9 +1,9 @@
-import React from 'react';
 import { shallow } from 'enzyme';
 import keycode from 'keycode';
-
+import React from 'react';
+import ReactDOM from 'react-dom';
 import { NestedListViewWidget } from './NestedListView.component';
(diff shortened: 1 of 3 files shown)


packages/forms/src/UIForm/fields/NestedListView/NestedListView.test.js
```diff
@@ -68,6 +68,36 @@ describe('NestedListView component', () => {
...
+			previousItems.forEach(({ checked, children }) => {
+				expect(checked).toBe(false);
+				children.forEach(childrenItem => expect(childrenItem.checked).toBe(false));
+			});
+
+			// when : trigger a props update
+			const allValues = {};
+			props.schema.items.forEach(({ key, titleMap }) => {
+				allValues[key] = titleMap.map(titleMapItem => titleMapItem.value);
+			});
```
jmfrancois, line 88:
```suggestion
			const allValues = props.schema.items.reduce((acc, item) => {
				acc[item.key] = item.titleMap.map(titleMapItem => titleMapItem.value);
				return acc;
			}, {});
```

packages/forms/src/UIForm/fields/NestedListView/NestedListView.test.js
```diff
@@ -68,6 +68,36 @@ describe('NestedListView component', () => {
 		expect(wrapper.getElement()).toMatchSnapshot();
 	});
 
+	describe('componentDidUpdate', () => {
+		it('should update items on props.value change', () => {
+			// given
+			const node = document.createElement('div');
+			// eslint-disable-next-line react/no-render-return-value
+			const instance = ReactDOM.render(<NestedListViewWidget {...props} value={{}} />, node);
```
jmfrancois, line 76: Pretty interesting to see this eslint rule disabled: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-render-return-value.md
It would make the test less readble to fix the lint error

jmfrancois (commented):
I know it do not came from your PR but I found this bug playgin with the story:
The icon disappear  when I open part of the tree. Could please take a look ?
Screen Shot 2021-09-03 at 15 58 05

jsomsanith-tlnd (approved):

packages/forms/src/UIForm/fields/NestedListView/NestedListView.test.js
Gbacc: I find it more readable too  
Do we keep it like this ?

Gbacc:
> I know it do not came from your PR but I found this bug playgin with the story:
> The icon disappear when I open part of the tree. Could please take a look ?
> Screen Shot 2021-09-03 at 15 58 05

Sure i can , i'll make another dedicated PR

jmfrancois (approved):

PR #3426, opened 2021-09-09
chore(http): Drop lodash dependency

gjou-tlnd:

Lodash can't be used in service worker due to current ui-scripts config: lodash is expected as an external but it's not there
image

Config should be fixed but we can also remove lodash as it's not really needed

* [ ] The PR have used `yarn changeset` to a request a release from the CI if wanted.

### DIFF
--- a/packages/http/src/csrfHandling.js
+++ b/packages/http/src/csrfHandling.js
@@ -4,10 +4,6 @@
  * and then merge it if available into a http config.
  */
 
-import merge from 'lodash/merge';
-import flow from 'lodash/flow';
-import curry from 'lodash/curry';
-
 /**
  * @typedef {Object} HTTPConfig
  * @property {string} body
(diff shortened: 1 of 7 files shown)


jsomsanith-tlnd (approved):

frassinier (approved):
Ok if you don't introduce breaking changes

gjou-tlnd:
> Ok if you don't introduce breaking changes

Test suite is green, everything should be ok 😶

packages/http/src/csrfHandling.js
```diff
@@ -57,12 +53,12 @@ function parseCookie(cookie) {
  * @param {Object.CSRFTokenCookieKey} CSRFTokenCookieKey - default `csrfToken`
  * @param {Map.<string, string>} cookieValues
  */
-const findCSRFToken = curry(({ CSRFTokenCookieKey = 'csrfToken' }, cookieValues) => {
+const findCSRFToken = ({ CSRFTokenCookieKey = 'csrfToken' }) => cookieValues => {
```
jmfrancois, line 56: curry do not mean that.
You may break users of http here

jmfrancois (approved):
good but with a comment !

packages/http/src/csrfHandling.js
gjou-tlnd: do you have an example of what could go wrong? This function is not exported and I made sur the only usage is OK

romainseb (approved):

PR #3466, opened 2021-10-01
feat(TFD-13512) datalist component

silvanocostanzo:
**What is the problem this PR is trying to solve?**

**What is the chosen solution to this problem?**

**Please check if the PR fulfills these requirements**

* [ ] The PR have used `yarn changeset` to a request a release from the CI if wanted.
* [ ] The PR commit message follows our [guidelines](https://github.com/talend/tools/blob/master/tools-root-github/CONTRIBUTING.md)
* [ ] Tests for the changes have been added (for bug fixes / features) And [non reg](./screenshots.md) done before need review
* [ ] Docs have been added / updated (for bug fixes / features)
* [ ] Related design / discussions / pages (not in jira), if any, are all linked or available in the PR

**[ ] This PR introduces a breaking change**

PR #3506, opened 2021-10-29
feat(TFD-13512/datalist): add initialCheckValue

silvanocostanzo:
in the context of [TFD-13512](https://jira.talendforge.org/browse/TFD-13512) the Webapp needs to notify the user when a semantic type is no longer used for validation. 

**this PR aims to improve the previous one #3466 which contained a wrong computation for the 'isValid' value**

The solution chosen is to check at the component mount if the value is present in the given Titlemap and the check is only made if the props `initialCheckValue` is sent to the component

* [x] The PR have used `yarn changeset` to a request a release from the CI if wanted.


tlnd-mhuchet (approved):
Looks ok, did you tried it with copylib to check if the previous issue in dataset is ok?

silvanocostanzo:
> Looks ok, did you tried it with copylib to check if the previous issue in dataset is ok?

yes, it is exactly what I did.

jmfrancois:
if you set initialCheckValue but the titleMap is async it will not check the initial value right ?

silvanocostanzo:
> if you set initialCheckValue but the titleMap is async it will not check the initial value right ?

right, in case of an asyncTitleMap and restricted values there is no initial check

.changeset/heavy-schools-argue.md
```diff
@@ -0,0 +1,5 @@
+---
+'@talend/react-forms': major
```
jmfrancois, line 2: major release mean breaking change. I don t see it, may be an error ?
If no error please document the breaking change

jmfrancois (changes requested):

.changeset/heavy-schools-argue.md
```diff
@@ -1,5 +1,5 @@
 ---
-'@talend/react-forms': major
+'@talend/react-forms': patch
```
tlnd-mhuchet, line 2: A minor version looks more appropriate here. We don't fix a bug, we don't break an existing feature, we just add a new thing 😬 

```suggestion
'@talend/react-forms': minor
```

jmfrancois (approved):
please update tests to adapt with current french hours change

PR #3525, opened 2021-12-01
feat(design-system/Checkbox): Support controlled checked state

gjou-tlnd:

Changes to `checked` prop are not taken into account

This feels like a dirty fix and I guess we'll have the same issue in other components using the `useCheckboxState`... 
We could put the `useEffect` in `useCheckboxState` but it would break the "`useXXState` first param is a default state" convention

### DIFF
--- a/packages/design-system/src/components/Form/Field/Input/Input.Checkbox.tsx
+++ b/packages/design-system/src/components/Form/Field/Input/Input.Checkbox.tsx
@@ -99,7 +99,7 @@ export const SCheckbox = styled(InlineStyle)<{
 `;
 
 export type CheckboxProps = InputProps & {
-	checked: boolean | 'indeterminate' | (string | number)[];
+	checked?: boolean | 'indeterminate' | (string | number)[];
 };
 
 const Checkbox = React.forwardRef(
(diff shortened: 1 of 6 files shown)


packages/design-system/src/components/Form/Field/Input/Input.Checkbox.tsx
```diff
@@ -120,11 +120,16 @@ const Checkbox = React.forwardRef(
...
+		const state = (indeterminate && 'indeterminate') || defaultChecked || checked;
 		const checkbox = useCheckboxState({
-			state: (indeterminate && 'indeterminate') || defaultChecked || checked,
+			state,
 			readOnly,
 		});
 
+		useEffect(() => {
+			checkbox.setState(state);
+		}, [state]);
```
Benrajalu, line 131: I'm not sure this hook is valid, aren't we supposed to have the exhaustive dependancy list in the array? We'd be missing `checkbox` here which would also create a loop :/

gjou-tlnd: we can't have a loop here but yes adding the dependency makes sense

Benrajalu (approved):
Fine. But so much custom logic for what amounts to basic HTML components is SAD and I'm SAD.

frassinier (approved):

gjou-tlnd:
> Fine. But so much custom logic for what amounts to basic HTML components is SAD and I'm SAD.

So am I

PR #3554, opened 2021-12-10
feat(TDP-10570): Increaser Layout header z-index

vbocquet:
We have a complex z-index problem in Data Preparation: In some screens, we can't see dropdowns in header (App switcher, User profil).

Here is a screenshot of the problem:
zindex

This page have some specificity:

- A Layout with a stepper and a Drawer that contain a form with a Typeahead
- This Typeahead have a button that open another Drawer

To hide the typeahead opened popin, we need to have at least a `z-index` set to `1035`.
But with that, we hide the dropdown that have a `z-index` set to `110` (more additional infos at bottom)

With that in mind, we have a dilemma with our layout `z-index`:

- Be higher than `1035` to hide connection typehead
- Be smaller than `110` to show menu dropdown

The only solution that i see is this one:
- Increase header `z-index` to be at least `1036`
- Do not be higher than `1040` to prevent dialog conflict (https://github.com/Talend/ui/blob/master/packages/theme/src/theme/_guidelines.scss#L54)

Additional information:
In reality, header dropdown z-index is set to 1000 (class: .dropdown-menu). But one parent (header is set to 110). Because of stacking context feature, we can't have a child z-index higher that parent z-index (https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context)

### DIFF
--- a/packages/components/src/Layout/Layout.scss
+++ b/packages/components/src/Layout/Layout.scss
@@ -5,6 +5,8 @@
 
 /// Layout height of the header
 /// @type Number (Unit)
+
+// Need to be higher to typehead (1035)
 $tc-drawer-z-index: $zindex-navbar-fixed + 6 !default;
 $tc-layout-header-height: $navbar-height !default;
 $tc-layout-skip-links-z-index: $tc-drawer-z-index + 20 !default;
--- a/packages/components/src/Layout/Layout.scss
+++ b/packages/components/src/Layout/Layout.scss
@@ -5,7 +5,7 @@
 
 /// Layout height of the header
 /// @type Number (Unit)
-$tc-drawer-z-index: $drawer-z-index !default;
+$tc-drawer-z-index: $zindex-navbar-fixed + 6 !default;
 $tc-layout-header-height: $navbar-height !default;
 $tc-layout-skip-links-z-index: $tc-drawer-z-index + 20 !default;
 $tc-layout-header-z-index: $tc-drawer-z-index + 10 !default;


frassinier:
Out of curiosity? Why is the Typeahead still opened when you open the second drawer?

packages/components/src/Layout/Layout.scss
```diff
@@ -5,7 +5,7 @@
 
 /// Layout height of the header
 /// @type Number (Unit)
-$tc-drawer-z-index: $drawer-z-index !default;
+$tc-drawer-z-index: $zindex-navbar-fixed + 6 !default;
```
frassinier, line 8: Did we consider closing the typeahead menu while clicking its CTA instead?

ncomont (approved):

vbocquet:
@frassinier 
Thanks for checking,

That was my first idea, i try several approach to do that but not simple:

- All this form is on `ui-dataset`: https://github.com/Talend/ui-dataset/tree/develop/src/app/components/AddDatastoreForm 
- This form is render by a ComponentForm from `@talend/react-containers`: A black box that take URL of the service that return a `uiSpec` body (with `jsonSchema`, `uiSchema` & `properties`): https://github.com/Talend/ui-dataset/tree/develop/src/app/components/EnhancedDatalist 
- The component "Connection" is a TDP custom widget named "EnhancedDatalist" that is passed to `ComponentForm` in `widgets` props. This one use the Datalist field from `@talend/react-containers`, that use Datalist component from `@talend/react-forms`, that use Datalist from `@talend/react-components`... (https://github.com/Talend/ui/blob/master/packages/forms/src/UIForm/fields/Datalist/Datalist.component.js) 

We don't have access to a function that help use to close programmatically the Datalist, we only have a boolean in renderer function to see if it's showed. This renderer is in Typeahead (https://github.com/Talend/ui/blob/de74c0e04c0bc8823e155d2712031188607f0bb0/packages/components/src/Typeahead/Typeahead.component.renderers.js#L165 )

Anyway, even if we have that, we lost the focus on the field, no sure it will be works.


After spends several hours on this, i think the more simple way is this z-index tweak. The idea is to resolve that in a reasonable time

packages/components/src/Layout/Layout.scss
frassinier: At least, adding a comment on top of that magic number could be nice!

vbocquet: Good point, just add it

frassinier (approved):

PR #3571, opened 2021-12-15
feat(Checkbox): Support data-feature attribute

gjou-tlnd:

1. The data-feature attribute on the input, which has a 0px width.
2. data-feature should be suffixed by the action name

### DIFF
--- a/packages/design-system/src/components/Form/Field/Input/Input.Checkbox.tsx
+++ b/packages/design-system/src/components/Form/Field/Input/Input.Checkbox.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect } from 'react';
+import React from 'react';
 import { Checkbox as ReakitCheckbox, unstable_useId as useId } from 'reakit';
 import styled from 'styled-components';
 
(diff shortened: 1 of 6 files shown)


packages/design-system/src/components/Form/Field/Input/Input.Checkbox.tsx
```diff
@@ -100,6 +100,7 @@ export const SCheckbox = styled(InlineStyle)<{
 
 export type CheckboxProps = InputProps & {
 	checked?: boolean | 'indeterminate' | (string | number)[];
+	'data-feature'?: string;
```
Benrajalu, line 103: We should be able to pass any data attribute we want, it's a native HTML element we are trying to imitate here. `data-feature` is supposed to be spread within `...rest`.
A better approach would be to change these proptypes to reflect that.

gjou-tlnd: that's the point of this change: the data-feature has a custom behavior, it's not forwared to the label + it's suffixed with a state specific value

Benrajalu: Alright, I'm stil uncomfortable with it. Why would a random data-attribute duplicate what is essentially the state of that component and what could essentially be its value? 
Are we OK with shared components having such highly specific behaviours hidden behind data-attributes?

PR #3573, opened 2021-12-16
chore(ADR): Added ADR on using closed component APIs or composition

Benrajalu:
We need to document the direction we take on building components inside of a shared library.

We think components with closed APIs are a better way to ensure consistent UI and UX, which is our main objective. 

The reasons why composition may be chosen instead are also outlined in the ADR.

### DIFF
--- a/packages/design-system/adr/composition-vs-api.md
+++ b/packages/design-system/adr/composition-vs-api.md
@@ -1,0 +1,219 @@
+#  CSS Modules
+
+## Composition vs closed component APIs
+
+## Problems and definitions
+
+**Composition** is what we do when we assemble HTML or JSX elements to build a cohesive whole. For instance:
+
+```tsx
+<ul>
+	<li>This list</li>
+	<li>Exemplifies <a href="https://talend.com">composition</a></li>
+</ul>
+
+```
+
+The combination of `ul`, `li` and `a` gives us an unordered list with one of its items containing a hyperlink anchor. This is familiar, flexible and commonplace.
+
+A **closed component API model** does not rely on consumers performing the composition, but on them using  input / output parameters to achieve the same result. For instance:
+
+```tsx
+<ExampleList
+    entries=[
+        'This list',
+        <>Exemplifies <a href="https://talend.com">a closed API</a></>,
+    ]
+/>
+
+---
+// Where...
+type entry = string | React.ReactElement;
+
+type ListProps = {
+  entries: entry[];
+};
+```
+
+This would achieve the same unordered list with a hyperlink anchor.
+
+### Composition
+
+**Pros**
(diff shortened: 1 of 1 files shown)


jsomsanith-tlnd (approved):

frassinier (approved):

jmfrancois (approved):
Agree !

There is also an in between options for components like Forms where you can compose using provided sub components but as you mentionned the children are not typed so it means you can create fields. In this particular case this is super important to have also low level components to build custom fields and good documentation for that.

In any case composition and customisation means maintenance on both side so should be discouraged.

Also it means we will have to test all possible combination at some point as we already did in the past: Form in a drawer, Form in a Modal etc ...

cbaune-talend:
I find the arguments for the CCAPI convincing.
That said, it is impossible to predict all cases in advance.
This means that design teams will, from time to time, create mockups that will require modifications to the DS to be feasible.
This is, I think, the important point: detect these modifications in advance so that the team in charge of the implementation doesn't get stuck in its work.

ocamon:
This sound like a good idea but we would need a way to expand HTML attribute easily on some component.
Let's take the list exemple you provided:
As Webkit doesn't expose the list role while the `list-style` is set to none, in some case I would want to add a `role="list"` on the ul element to enforce the list to the enduser. 
We would need a way to customize html attribute on component without having a bloated API.
The limit on what can be extended is a bit tricky IMHO.

frassinier:
@ocamon Why do you need to provide that attribute from the product side? It can be smart to put that edge case directly in the DS' codebase, isn't it?

frassinier:
@cbaune-talend that's why the DS core is here at Talend, indeed!

ocamon:
> @ocamon Why do you need to provide that attribute from the product side? It can be smart to put that edge case directly in the DS' codebase, isn't it?

Because most of the time this attribute is not needed. If Safari changed it's behavior is because most web pages were too verbose.

Benrajalu:
> > @ocamon Why do you need to provide that attribute from the product side? It can be smart to put that edge case directly in the DS' codebase, isn't it?
> 
> Because most of the time this attribute is not needed. If Safari changed it's behavior is because web pages were too verbose.

Sure, and as you've said where we put the limit *is* tricky, but that's also the whole point. 

It's much easier to start with a closed system and open it bit by bit when necessary than it is to start with a fully opened one and close it later on. 

In the DS, we should be able to figure out that the current list component we're building relies on `list-style` set to `none`, and thus add the role on our end. We have no benefit in letting consumers set that value themselves (some might do it, some might forget). 

The ability to set `list-style` on components published by the DS isn't up to consumers either, it's done within the DS.

PR #3577, opened 2021-12-20
fix(components/Toggle): Default class

sgangnant:
`switch checkbox` classes are provided to the Toggle component as default value, but as soon as a `className` prop is provided, those disappear which breaks the component, rendering it as a checkbox.

Add them as base value for classes rather than default values

### DIFF
--- a/.changeset/tidy-guests-do.md
+++ b/.changeset/tidy-guests-do.md
@@ -1,0 +1,5 @@
+---
+'@talend/react-components': major
+---
+
+fix(components/Toggle): Default class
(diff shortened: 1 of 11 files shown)


.changeset/tidy-guests-do.md
```diff
@@ -0,0 +1,5 @@
+---
+'@talend/react-components': major
```
jmfrancois, line 2: fix is patch

jmfrancois (changes requested):

.changeset/tidy-guests-do.md
sgangnant: Might break some cases where it's already used, so I was hesitating

jmfrancois (approved):

Kporal (approved):

PR #3578, opened 2021-12-20
fix(design-system/Link): fix title related to target attribute

Gbacc:
The title "open in a new tab" should be related to the target attribute and not the detection of external link

### DIFF
--- a/.changeset/wet-elephants-fetch.md
+++ b/.changeset/wet-elephants-fetch.md
@@ -1,0 +1,5 @@
+---
+'@talend/design-system': minor
+---
+
+fix(design-system/Link): The title "open in a new tab" should be related to the target attribute and not the detection of external link
--- a/.changeset/wet-elephants-fetch.md
+++ b/.changeset/wet-elephants-fetch.md
@@ -1,5 +1,5 @@
 ---
-'@talend/design-system': minor
+'@talend/design-system': patch
 ---
 
 fix(design-system/Link): The title "open in a new tab" should be related to the target attribute and not the detection of external link
--- a/packages/design-system/src/components/Link/Link.spec.tsx
+++ b/packages/design-system/src/components/Link/Link.spec.tsx
@@ -27,6 +27,7 @@ context('<Link />', () => {
 		cy.mount(<TargetBlank />);
 		cy.get('.link')
 			.should('have.attr', 'title', 'Open in a new tab')
-			.should('have.attr', 'target', '_blank');
+			.should('have.attr', 'target', '_blank')
+			.should('have.attr', 'rel', 'noopener noreferrer');
 	});
 });
--- a/packages/design-system/src/components/Link/Link.spec.tsx
+++ b/packages/design-system/src/components/Link/Link.spec.tsx
@@ -1,11 +1,10 @@
 /// <reference types="cypress" />
 
-import React from 'react';
 import { composeStories } from '@storybook/testing-react';
-
+import React from 'react';
 import * as Stories from './Link.stories';
 
-const { Default, WithIcon, External } = composeStories(Stories);
+const { Default, WithIcon, External, TargetBlank } = composeStories(Stories);
 
 context('<Link />', () => {
 	it('should render', () => {
@@ -20,6 +19,11 @@ context('<Link />', () => {
 
 	it('should render external', () => {
 		cy.mount(<External />);
+		cy.get('.link .link__icon').should('have.attr', 'name', 'talend-link');
+	});
+
+	it('should render target blank', () => {
+		cy.mount(<TargetBlank />);
 		cy.get('.link')
 			.should('have.attr', 'title', 'Open in a new tab')
 			.should('have.attr', 'target', '_blank');
--- a/packages/design-system/src/components/Link/Link.spec.tsx
+++ b/packages/design-system/src/components/Link/Link.spec.tsx
@@ -1,7 +1,8 @@
 /// <reference types="cypress" />
 
-import { composeStories } from '@storybook/testing-react';
 import React from 'react';
+import { composeStories } from '@storybook/testing-react';
+
 import * as Stories from './Link.stories';
 
 const { Default, WithIcon, External, TargetBlank } = composeStories(Stories);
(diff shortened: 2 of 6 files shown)


packages/design-system/src/components/Link/docs/Link.stories.mdx
```diff
@@ -62,6 +62,14 @@ When the link targets a new domain, we add an external-link icon right after.
 	<Story story={Stories.External} />
 </Canvas>
 
+### Target blank
```
frassinier, line 65:
```suggestion
```

packages/design-system/src/components/Link/Link.tsx
```diff
@@ -60,13 +58,13 @@ const Link = React.forwardRef(
...
 			}
-			if (isExternal && isBlank && title) {
+			if (isBlank && title) {
 				return t('LINK_EXTERNAL_TITLE', {
 					title,
 					defaultValue: '{{title}} (open in a new tab)',
 				});
 			}
-			if (isExternal && isBlank) {
+			if (isBlank) {
```
frassinier, line 67:
```suggestion
			if (isExternal || isBlank) {
```

packages/design-system/src/components/Link/Link.tsx
```diff
@@ -60,13 +58,13 @@ const Link = React.forwardRef(
 					defaultValue: 'This link is disabled',
 				});
 			}
-			if (isExternal && isBlank && title) {
+			if (isBlank && title) {
```
frassinier, line 61:
```suggestion
			if ((isExternal || isBlank) && title) {
```

packages/design-system/src/components/Link/Link.spec.tsx
```diff
@@ -1,11 +1,10 @@
 /// <reference types="cypress" />
 
-import React from 'react';
 import { composeStories } from '@storybook/testing-react';
-
+import React from 'react';
```
frassinier, line 4: Please fix your IDE

packages/design-system/src/components/Link/Link.spec.tsx
```diff
@@ -20,6 +19,11 @@ context('<Link />', () => {
 
 	it('should render external', () => {
 		cy.mount(<External />);
+		cy.get('.link .link__icon').should('have.attr', 'name', 'talend-link');
```
frassinier, line 22: External links should continue to open in a new tab
```suggestion
		cy.get('.link .link__icon').should('have.attr', 'name', 'talend-link');
		cy.get('.link')
			.should('have.attr', 'title', 'Open in a new tab')
			.should('have.attr', 'target', '_blank');
```

packages/design-system/src/components/Link/docs/Link.stories.mdx
```diff
@@ -62,6 +62,14 @@ When the link targets a new domain, we add an external-link icon right after.
...
+
+When the link opens in a new window or tab, the title `open in a new tab` is added or appended to an existing title.
```
frassinier, line 67:
```suggestion
Even if it's the same domain, when the link opens in a new window or tab, the title `open in a new tab` is added or appended to an existing title.
```

packages/design-system/src/components/Link/Link.spec.tsx
Gbacc: But we don't force target="_blank" for external link , so we can have external link that does not open in a new tab

packages/design-system/src/components/Link/docs/Link.stories.mdx
```diff
@@ -62,6 +62,13 @@ When the link targets a new domain, we add an external-link icon right after.
 	<Story story={Stories.External} />
 </Canvas>
 
+
```
frassinier, line 65:
```suggestion
```

packages/design-system/src/components/Link/Link.spec.tsx
frassinier: Can be painful to get the distinction, I would suggest forcing external links into a new tab

Gbacc: I agree lets force it 👍

frassinier: I raised the question to our Product Designers. 
For now, forgot my suggestions.

frassinier (approved):

.changeset/wet-elephants-fetch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@talend/design-system': minor
```
jmfrancois, line 2: fix is patch
feat is minor
breaking change is major

packages/design-system/src/components/Link/Link.tsx
jmfrancois, line 61: please in that case could you also add a security fix ?
`rel="noopener noreferrer"`

jmfrancois (changes requested):

sgendre:
the meaning of isExternal is not about domain but application

ux rule accross talend is that a link to another webapp must open in a new tab

packages/design-system/src/components/Link/Link.tsx
Gbacc: We already have it as long as the target is set to _blank
```
rel={isBlank ? 'noopener noreferrer' : undefined}
``` 
Or maybe i misunderstood your point

jmfrancois: oh didn't see it thanks !

jmfrancois (approved):

ocamon:
Keep in mind that `_blank` is not the only value where the link will open in a new tab.
If the target attribute is set to `foo` this will create a new named browsing context that will be opened in a new tab.
I don't think this is something we want to support, but a typo can happen :)

frassinier:
@ocamon what do you suggest then? 
```ts
const isBlank = !['_self', '_parent', '_top'].some(targetValue => targetValue === target);
```

frassinier:
@Gbacc, can you take @ocamon's comment into account?

Gbacc:
> @Gbacc, can you take @ocamon's comment into account?

Sure it's a good idea

sgendre (approved):

PR #3586, opened 2021-12-22
chore(Design-System/Status): Updated using tokens and CSS modules

Benrajalu:
We're moving from Styled Components to CSS modules

Removing Styled Components and using CSS modules as well as design tokens. 
This also enables us to clean up our TypeScript APIs.


packages/design-system/src/components/ThemeProvider/ThemeProvider.tsx
```diff
@@ -13,6 +13,7 @@ const ThemeProvider = ({ theme = light, children }: ThemeProviderProps<any>) =>
 	const [selectedTheme, setSelectedTheme] = useState(theme);
 	// Handle nested Providers: parent Provider doesn't have context, child does
 	const context = useContext(ThemeContext);
+	document.body.dataset.theme = 'light';
```
frassinier, line 16:
```suggestion
React.useEffect(() => {
	document.body.dataset.theme = 'light';
}, []);
```

packages/design-system/src/components/Status/Primitive/StatusPrimitive.tsx
```diff
@@ -0,0 +1,42 @@
...
+		const text = <span className={styles.status__text}>{children}</span>;
+		const picto = (
+			<span className={styles.status__icon} aria-hidden>
+				{!inProgress && icon && <Icon name={icon} />}
+				{inProgress && <Loading />}
+			</span>
+		);
+
+		return (
+			<span className={classnames(styles.status, styles[variant])} ref={ref} {...rest}>
```
frassinier, line 33:
```suggestion
			<span  {...rest} className={classnames(styles.status, styles[variant])} ref={ref}>
```

packages/design-system/src/components/Status/Status.tsx
```diff
@@ -1,37 +1,42 @@
...
+	| StatusWarningProps
+	| StatusSuccessfulProps
+	| StatusInProgressProps
+	| StatusCanceledProps
 );
 
+const Status = React.forwardRef((props: StatusProps, ref: React.Ref<any>) => {
+	const { variant, ...rest } = props;
+	switch (variant) {
+		case 'failed':
```
frassinier, line 22: We may use `enum` instead?
https://github.com/Talend/ui/blob/master/packages/design-tokens/src/types.ts#L1-L13

packages/design-system/src/components/Status/variations/StatusInProgress.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import i18n from 'i18next';
+import StatusPrimitive, { StatusProps } from '../Primitive/StatusPrimitive';
+
+export type StatusInProgressProps = Omit<
+	StatusProps,
+	'icon' | 'variant' | 'inProgress' | 'inProgress'
```
frassinier, line 7:
```suggestion
	'icon' | 'variant' | 'inProgress'
```

packages/design-system/src/components/Status/variations/StatusWarning.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import i18n from 'i18next';
+import StatusPrimitive, { StatusProps } from '../Primitive/StatusPrimitive';
+
+export type StatusWarningProps = Omit<StatusProps, 'icon' | 'variant' | 'inProgress'>;
```
frassinier, line 5: variant seems always to have same props, isn't it?

Benrajalu: Yes, at this point it does. Not sure what the best practice would be.

packages/design-system/src/components/Status/Primitive/StatusPrimitive.tsx
```diff
@@ -0,0 +1,50 @@
...
+export const possibleVariants = [
+	'successful',
+	'failed',
+	'inProgress',
+	'warning',
+	'canceled',
+] as const;
+
+export type StatusProps = {
+	icon?: IconName;
```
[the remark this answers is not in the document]
Benrajalu: Fine by me for now. This is the primitive too, so no big deal IMO.

packages/design-system/src/components/Status/Status.tsx
Benrajalu: I don't really see the benefit?

packages/design-system/src/components/Status/variations/StatusWarning.tsx
Benrajalu: I like the explicitness of it. Each variant is a single file with types etc.. Yes those are the same types here in this context, but it may not be for other components. It feel better to enforce a single file pattern early on.

packages/design-system/src/components/Status/Status.stories.tsx
```diff
@@ -1,22 +1,22 @@
...
+	component: StatusInProgress,
 };
 
-export const InProgress = () => <Status.InProgress />;
-export const Successful = () => <Status.Successful />;
-export const Failed = () => <Status.Failed />;
-export const Warning = () => <Status.Warning />;
-export const Canceled = () => <Status.Canceled />;
-export const CustomSuccessful = () => <Status.Successful>Done</Status.Successful>;
+export const InProgress = () => <StatusInProgress />;
```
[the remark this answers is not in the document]
Benrajalu: In this case it's not necessary as we don't actually display the args underneath each story (they are packed together). we do have a dedicated story with args that does the job.

packages/design-system/src/components/Status/Status.tsx
frassinier: Avoid typo, not a huge fan of plain text

frassinier (approved):

packages/design-system/src/components/Status/Primitive/StatusPrimitive.tsx
```diff
@@ -0,0 +1,50 @@
...
+		const picto = (
+			<span className={styles.status__icon} aria-hidden>
+				{!inProgress && icon && <Icon name={icon} />}
+				{inProgress && <Loading />}
+			</span>
+		);
+
+		return (
+			<span {...rest} className={classnames(styles.status, styles[variant])} ref={ref}>
+				<StackHorizontal as="span" display="inline" gap="XXS" align="center" justify="start">
```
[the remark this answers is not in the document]
Benrajalu: Yes. I'm ok with this because IMO the two items represent two different things: 
The first span holds the responsibilities of the Stack's core styles. 
The second one (output of Stack) holds the layout. 

There is a way around that but it involves opening up `StackHorizontal` to classNames, which is something we're very cagey about.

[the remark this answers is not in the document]
cbaune-talend: The two nested spans aren't that bad IMO.
The styles are different so there's a purpose.
I think the tradeoff of opening `StackHorizontal`  isn't worth it.

cbaune-talend (approved):

packages/design-system/src/components/Status/variations/StatusInProgress.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import i18n from 'i18next';
+import StatusPrimitive, { StatusProps } from '../Primitive/StatusPrimitive';
+
+export type StatusInProgressProps = Omit<StatusProps, 'icon' | 'variant' | 'inProgress'>;
+
+const StatusInProgress = React.forwardRef(
+	(props: StatusInProgressProps, ref: React.Ref<HTMLSpanElement>) => {
+		return (
+			<StatusPrimitive inProgress variant="inProgress" {...props} ref={ref}>
```
tlnd-mhuchet, line 10: Wdyt about having `variants` as an enum to be able to use it in each component.

Benrajalu: We tried it, but it seems that TS then requires you to use said enum in your implem. It meant that for people who wanted to use `Status`, they would have had to import both `Status` and `StatusVariantEnum` in order to do something like `<Status variant={StatusVariantEnum.warning} />` which seemed tedious... 

Eventually here the typing is as strong (through keyof typeof the object) and the prop autocompletes fairly well. Is that an acceptable tradeoff to you?

tlnd-mhuchet (approved):

PR #3589, opened 2021-12-23
feat(Design-System/Link): CSS Modules

frassinier:

`Link` still uses Styled Components

Move to CSS Modules


packages/design-system/src/components/Link/Link.module.scss
```diff
@@ -0,0 +1,69 @@
...
+	&:hover,
+	&:active,
+	&:focus,
+	&:visited {
+		text-decoration: none;
+	}
+
+	.link__text {
+		border-bottom: tokens.$coral-border-solid-s transparent;
+		transition: 250ms ease-in-out;
```
frassinier, line 21: FIXME when the new version of the Design Tokens will be released

packages/design-system/src/components/Link/Link.module.scss
```diff
@@ -0,0 +1,69 @@
...
+
+.link {
+	font: tokens.$coral-paragraph-m;
+	color: tokens.$coral-color-accent-text;
+	background: none;
+	border: none;
+	border-bottom-color: currentColor;
+	cursor: pointer;
+
+	&,
```
Benrajalu, line 11:
```suggestion
```
Means nothing in CSS modules

frassinier: Yes, it does

Benrajalu: It does?! I've got to see how it compiles

frassinier: It should be if we do not load Bootstrap v3 at first 😇 
It's a bullet-proof solution

Benrajalu (approved):

packages/design-system/src/components/Link/Link.module.scss
jsomsanith-tlnd: For info, here is the result if we don't have this block
image

Bootstrap is loaded and takes precedence because it specifies a `hover` rule.

packages/design-system/src/components/Link/Link.module.scss
```diff
@@ -0,0 +1,70 @@
+@use '~@talend/design-tokens/lib/tokens';
+
+.link {
+	font: tokens.$coral-paragraph-m;
+	color: tokens.$coral-color-accent-text;
+	background: none;
+	border: none;
+	border-bottom-color: currentColor;
```
jsomsanith-tlnd, line 8: why the border-bottom-color, but no border ?

jsomsanith-tlnd (approved):

packages/design-system/src/components/Link/Link.module.scss
frassinier: for when there will be a border (on `:hover`, on `:focus` and on `:active`)

tlnd-mhuchet (approved):

PR #3608, opened 2022-01-05
fix(components): Toggle should reuse Checkbox

frassinier:

Any `Checkbox` without className ends up with a `Toggle` (aka `Switch`)

Instead of having `Checkbox` instantiating `Toggle`, do the opposite


packages/components/src/Toggle/Toggle.component.js
```diff
@@ -34,7 +34,7 @@ function Toggle({ id, label, className, intermediate, ...props }) {
 
 	return (
 		<div
-			className={classNames('checkbox tc-toggle switch checkbox', theme['tc-toggle'], className, {
+			className={classNames('checkbox tc-toggle', theme['tc-toggle'], className || 'switch', {
```
Benrajalu, line 37: In the case of `Checkbox` component, this will result in `checkbox tc-toggle checkbox` classnames. Which is weird. iof "checkbox" is the default, shouldn't this component be `Checkbox` and `Switch` call this component with a `switch` classname?

Benrajalu (approved):

sgangnant (approved):

smouillour (approved):

packages/components/src/Toggle/Toggle.component.js
frassinier: Too damn right

PR #3643, opened 2022-01-24
chore(design-system): Fixing linting errors on tests

Benrajalu:
We've got lint errors in tests since the move to react-testing-lib was made

Fix the errors we can fix

### DIFF
--- a/packages/design-system/src/components/Link/Link.tsx
+++ b/packages/design-system/src/components/Link/Link.tsx
@@ -89,7 +89,11 @@ const Link = React.forwardRef(
 			>
 				{icon &&
 					(typeof icon === 'string' ? (
-						<Icon className="link__icon link__icon--before" name={icon} />
+						<Icon
+							className="link__icon link__icon--before"
+							name={icon}
+							data-testid="link.icon.before"
+						/>
 					) : (
 						React.cloneElement(icon, {
 							className: `${icon.props?.className} link__icon link__icon--before`,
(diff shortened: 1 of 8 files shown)


jmfrancois (approved):

packages/design-system/src/components/Link/Link.tsx
```diff
@@ -89,7 +89,11 @@ const Link = React.forwardRef(
 			>
 				{icon &&
 					(typeof icon === 'string' ? (
-						<Icon className="link__icon link__icon--before" name={icon} />
+						<Icon
+							className="link__icon link__icon--before"
+							name={icon}
+							data-testid="link.icon.before"
```
frassinier, line 95: One of the last [ADR](https://github.com/Talend/ui/blob/8bdfe80afd0bc83c3aa5ec785fa9733903141c1e/packages/design-system/adr/data-test.md) was saying that we will prefer `data-test` instead of `data-testid` for DS components like Link.

- We can either think about adding a custom query when using render fn 
https://testing-library.com/docs/react-testing-library/api/#queries

- or adding it for once by re-exporting
https://testing-library.com/docs/dom-testing-library/api-custom-queries

I'm not a huge  fan of both solutions TBH

frassinier (approved):

packages/design-system/src/components/Link/Link.tsx
jmfrancois: This is documented here:
https://testing-library.com/docs/queries/about/#priority
you are supposed to use data-testid ( I do not see getByTest api but only getByTestId)

PR #3665, opened 2022-01-31
chore(ARCH-466): start sb via talend/scripts in components and containers

jsomsanith-tlnd:

storybook setup is replicate in every packages with some parts in common, but dependencies and so adds maintenance burden.

Move storybook setup to ui-scripts has been done to obtain a zero config setup that fits our needs.
Now we just use it.


packages/components/.storybook/i18n.js
```diff
@@ -0,0 +1,5 @@
+import { namespaces as tuiNamespaces } from '@talend/locales-tui-components/namespaces';
+import { locales as tuiLocales } from '@talend/locales-tui-components/locales';
+
+export const namespaces = [tuiNamespaces];
+export const locales = [tuiLocales];
```
jmfrancois, line 5: this is a regression from current master which load locales from unpkg and go againts #3692

packages/components/src/ActionBar/ActionBar.stories.js
```diff
@@ -1,5 +1,4 @@
 import React from 'react';
-import { storiesOf } from '@storybook/react';
 import { action } from '@storybook/addon-actions';
```
jmfrancois, line 2: this looks weid to me to import package not specified in devDependencies 🤔

packages/components/package.json
```diff
@@ -70,15 +70,11 @@
     "uuid": "^3.4.0"
   },
   "devDependencies": {
-    "@storybook/addon-a11y": "^6.4.19",
-    "@storybook/addon-actions": "^6.4.19",
```
jmfrancois, line 74: this one is required by stories, we should keep it

jsomsanith-tlnd: I find it hard if we upgrade sb in the future, libs and apps have to know that starting this version of talend/scripts we need to sync our addon-action version to a compatible version.
Just an idea: what about exposing that from @talend/scripts-config-storybook-lib ? Not sure it's a good one

packages/components/talend-scripts.json
```diff
@@ -1,3 +1,6 @@
 {
-  "preset": "@talend/scripts-preset-react-lib"
+  "preset": "@talend/scripts-preset-react-lib",
+  "dynamic-cdn-webpack-plugin": {
+    "resolver": "@talend/module-to-cdn"
+  }
```
jmfrancois, line 5: why do you need this ? It is supposed to be the default value.

packages/components/package.json
```diff
@@ -56,7 +56,7 @@
     "react-bootstrap": "0.32.4",
     "react-debounce-input": "3.2.0",
     "react-draggable": "^4.4.4",
-    "react-grid-layout": "0.16.6",
+    "react-grid-layout": "^1.3.3",
```
jmfrancois, line 59: Screenshot 2022-03-01 at 15 25 20

😂

jsomsanith-tlnd: yeah.... 😅

.github/workflows/pr-demo.yml
```diff
@@ -32,6 +32,7 @@ jobs:
       - name: Install and build
         run: |
           yarn install --frozen-lockfile
+          yarn pre-release
```
jmfrancois, line 35: 😬 comes from the common cdn config

jmfrancois (approved):
just question around API.
looks good to me to point to storybook actions.
just have to check duplicate at upgrade times but seems acceptable.
Big thanks for this hard work Jimmy !

frassinier:
http://3665.talend.surge.sh/icons is not fully exhaustive

.github/workflows/pr-demo.yml
frassinier: why does it matter here?

packages/components/package.json
frassinier: In fact, actions are already in essentials package

packages/components/src/ActionBar/ActionBar.stories.js
frassinier: I can live with an exclusion from eslint at this point

frassinier:
Amazing work! Well done!

frassinier (approved):

packages/components/.storybook/i18n.js
```diff
@@ -0,0 +1,8 @@
+import { namespaces as tuiNamespaces } from '@talend/locales-tui-components/namespaces';
+import { namespaces as dsNamespaces } from '@talend/locales-design-system/namespaces';
+
+export const namespaces = [...tuiNamespaces, ...dsNamespaces];
+export const remoteLocalesMap = {
+	'tui-components': 'https://unpkg.com/@talend/locales-tui-components/locales/{{lng}}/{{ns}}.json',
+	'design-system': 'https://unpkg.com/@talend/locales-design-system/locales/{{lng}}/{{ns}}.json',
+};
```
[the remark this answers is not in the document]
jsomsanith-tlnd: It's configured in preview (i18n conf was imported at preview time) now

PR #3669, opened 2022-02-01
chore(design-system): Cypress instead of Jest

frassinier:

Jest and Cypress can't easily coexist for a Typescript setup.
Because of the Jest setup, we can't inherit from Cy autocomplete at the same time.
image

We have some tests using Jest
```bash
Test Suites: 4 passed, 4 total
Tests:       29 passed, 29 total
Snapshots:   8 passed, 8 total
Time:        34.641 s
Ran all test suites.
✨  Done in 37.33s.
```

And a second test will give us -
```
✨  Done in 13.70s.
```

That was pretty fast compared to Cypress runs (~99s)

Now, the tests we had based on Jest can easily be replaced by Cypress ones, 
That's why I suggest solving that typing issue by keeping Cypress only, 
Even if, yes, tests are executed slower but can be part of real web browsers which is more convenient to test UX behaviors (and we got videos).

And autocomplete is now working
image


packages/design-system/src/components/Button/Button.spec.tsx
```diff
@@ -1,34 +1,37 @@
-/// <reference types="cypress" />
+/// <reference path="../../../cypress/support/index.d.ts" />
```
Benrajalu, line 1:
```suggestion
/// <reference path="cypress/support/index.d.ts" />
```
Wouldn't that work with the `rootDirt` setup we have because of `~docs`?

packages/design-system/src/components/Button/Button.stories.tsx
```diff
@@ -12,185 +13,187 @@ export default {
...
-export const Primary = (props: Story) => {
-	return <Button.Primary {...props}>Basic Button</Button.Primary>;
+export const Primary = {
+	render: (props: Story<ButtonProps>) => {
+		return <Button.Primary {...props}>Basic Button</Button.Primary>;
+	},
 };
 
-export const PrimaryIcon = (props: Story) => {
+export const PrimaryIcon = (props: Story<Story<ButtonProps>>) => {
```
Benrajalu, line 22:
```suggestion
export const PrimaryIcon = (props: Story<Story<ButtonProps>>) => {
```
What? If that's necessary then it's very bad. A generic loop is never a good sign.

packages/design-system/src/components/Link/Link.tsx
```diff
@@ -92,7 +92,7 @@ const Link = React.forwardRef(
 						<Icon
 							className="link__icon link__icon--before"
 							name={icon}
-							data-testid="link.icon.before"
+							data-test="link.icon.before"
```
Kporal, line 95: why this change? what is the difference between data-test and data-testid otherwise?

Benrajalu: https://github.com/Talend/ui/blob/master/packages/design-system/adr/data-test.md

`data-test` is for internal use and is the basic data-attribute for testing in the DS. `data-testid` is reserved to product teams (ie: `<Link href="/pouet" data-testid="link.ctaForPouet">Let's pouet</Link>`).

Kporal (approved):

Benrajalu (approved):

PR #3712, opened 2022-02-15
chore(Security): Add semgrep security scanner

Gby56:
Scans diffs on PRs, pushes to master, and weekly cron at midnight UTC sundays
Will also add scan findings into the Github Security dashboard

### DIFF
--- a/.github/workflows/semgrep.yml
+++ b/.github/workflows/semgrep.yml
@@ -1,5 +1,6 @@
 name: Semgrep
 on:
+  workflow_dispatch:
   pull_request:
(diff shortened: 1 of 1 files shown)


.github/workflows/semgrep.yml
```diff
@@ -0,0 +1,28 @@
+name: Semgrep
+on:
+  pull_request: {}
```
frassinier, line 3:
```suggestion
  pull_request:
```

frassinier:
Prior to #3713

PR #3764, opened 2022-03-11
fix(ARCH-465/design-tokens): umd are not well defined

jmfrancois:

Current playground is broken.
UMD of design-tokens was not used before DS 2.0 (with Input field)
The current UMD reprensent two modules (index.js + index.scss) which is not supported by UMD.

fix design-token umd by using a single entry.
add css import to obtain the good output.
UMD mean one global variable which reprensent the src/index.js
add tests to not repeat this in the future

### DIFF
--- a/packages/design-tokens/scripts/umd.test.js
+++ b/packages/design-tokens/scripts/umd.test.js
@@ -21,4 +21,10 @@ describe('umd', () => {
 		expect(min.default.coralColorNeutralText).toBeDefined();
 		expect(dev.default.coralColorNeutralText).toBeDefined();
 	});
+	it('should expose dark and light css properties in css', () => {
+		const content = fs.readFileSync(cssDev).toString();
+		expect(content).toContain('[data-theme=light]');
+		expect(content).toContain('[data-theme=dark]');
+		expect(content).toContain('--coral-color-neutral-text');
+	});
 });
--- a/packages/design-tokens/scripts/umd.test.js
+++ b/packages/design-tokens/scripts/umd.test.js
@@ -1,0 +1,24 @@
+/* eslint-disable global-require */
+/* eslint-disable import/no-dynamic-require */
+const fs = require('fs');
+const path = require('path');
+
+const umdMinified = path.join(__dirname, '..', './dist/TalendDesignTokens.min.js');
+const umdDev = path.join(__dirname, '..', './dist/TalendDesignTokens.js');
+const cssDev = path.join(__dirname, '..', './dist/TalendDesignTokens.css');
+
+describe('umd', () => {
+	it('should exists', () => {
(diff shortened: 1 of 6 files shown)


packages/design-tokens/scripts/umd.test.js
```diff
@@ -0,0 +1,24 @@
...
+
+describe('umd', () => {
+	it('should exists', () => {
+		expect(fs.existsSync(umdMinified)).toBeTruthy();
+		expect(fs.existsSync(umdDev)).toBeTruthy();
+		expect(fs.existsSync(cssDev)).toBeTruthy();
+	});
+	it('should expose default and values', () => {
+		const min = require(umdMinified);
+		const dev = require(umdDev);
```
jmfrancois, line 18: thanks umd i can require you without a browser :)

tlnd-mhuchet (approved):

packages/design-tokens/src/index.umd.ts
```diff
@@ -0,0 +1,5 @@
+import './light/_index.scss';
+import './dark/_index.scss';
```
Benrajalu, line 2:
```suggestion
import './index.scss';
```
That should do the trick?

packages/design-tokens/webpack.config.js
```diff
@@ -1,5 +1,3 @@
 module.exports = {
-	entry: {
-		TalendDesignTokens: ['./src/index.ts', './src/index.scss'],
-	},
+	entry: './src/index.umd.ts',
```
frassinier, line 2: Why not use the previous version with the entry array?

jmfrancois: * What could it means ?
* UMD is supposed to reprensent one js file with externals... not a simple script, so what could we export.
* current umd config is one entry point: https://unpkg.com/@talend/scripts-config-react-webpack@12.0.0/config/webpack.config.umd.js
* webpack doc specify it: https://webpack.js.org/configuration/entry-context/#entry
> https://webpack.js.org/configuration/entry-context/#entry

Looking at the current result if you try to use the produced umd you have

```javascript
> require('@talend/design-tokens/dist/TalendDesignTokens').default
undefined
> require('@talend/design-tokens/dist/TalendDesignTokens')
Object [Module] {}
```

So this multiple entrypoint (array or object) do not make sens with UMD.

packages/design-tokens/src/index.umd.ts
jmfrancois: strange but it works.

packages/design-tokens/webpack.config.js
frassinier: I feel it weird to get that `index.umd` into that `src/` folder but ok
Maybe it's the current index which makes no sense actually, let's see that in another PR

PR #3768, opened 2022-03-14
fix(Link): Href attribute is ignored

gjou-tlnd:

Href is spread here: https://github.com/Talend/ui/blob/master/packages/design-system/src/components/Linkable/Linkable.tsx#L31
but not forwarded to the actual `a`  https://github.com/Talend/ui/blob/master/packages/design-system/src/components/Linkable/Linkable.tsx#L71

### DIFF
--- a/.changeset/bluh-bluh-bluh.md
+++ b/.changeset/bluh-bluh-bluh.md
@@ -1,0 +1,5 @@
+---
+'@talend/design-system': patch
+---
+
+Link: href is ignored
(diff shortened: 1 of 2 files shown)


Benrajalu (approved):
I am a silly silly man

frassinier (approved):

Kporal (commented):
Don't we need to add it for the `as` element case too?

Benrajalu:
> Don't we need to add it for the `as` element case too?

Yes we do, well spotted.

Benrajalu (approved):

PR #3786, opened 2022-03-21
fix(components): shrink column headers to fit with cells width

Nigui:

List column headers may have a bigger width than column cells, leading to unaligned headers or hidden columns when grid was resized horizontally. 

Hide header overflow and add ellipsis.

- [X] The PR have used `yarn changeset` to a request a release from the CI if wanted.


romainseb (approved):

packages/components/src/VirtualizedList/VirtualizedList.scss
```diff
@@ -25,6 +25,15 @@
 	}
 }
 
+:global(div[role='columnheader']) {
```
Benrajalu, line 28:
```suggestion
:global(div[role='columnheader']) {
```
Why do we need :global for `div` and a role? Are they not global by nature?

packages/components/src/VirtualizedList/VirtualizedList.scss
```diff
@@ -25,6 +25,15 @@
...
+	overflow: hidden;
+
+	> span {
+		text-overflow: ellipsis;
+		overflow: hidden;
```
frassinier, line 33: Overflow is already managed by the parent, isn't it?
```suggestion
```

packages/components/src/VirtualizedList/VirtualizedList.scss
```diff
@@ -25,6 +25,15 @@
 	}
 }
 
+:global(.tc-list-table [role='columnheader']) {
```
Benrajalu, line 28:
```suggestion
:global(.tc-list-table [role='columnheader']) {
```
`.tc-list-table` is global? Isn't the whole project on CSS modules? That's weird.

Benrajalu: It looks like it is indeed. What do we feel is the benefit of this compound selector versus adding a classname, albeit global if needed, on the column headers? Wouldn't that be more straightforward?

.changeset/tasty-walls-serve.md
```diff
@@ -0,0 +1,5 @@
+---
+'@talend/react-components': minor
```
frassinier, line 2: this is not even minor 
```suggestion
'@talend/react-components': patch
```

packages/components/src/VirtualizedList/VirtualizedList.scss
```diff
@@ -9,20 +9,28 @@
...
 		cursor: pointer;
 	}
 	:global(.ReactVirtualized__Table__headerTruncatedText) {
 		align-self: center;
-		overflow: hidden;
 		padding-left: $padding-small;
+	}
+}
+
+:global(.ReactVirtualized__Table__headerColumn) {
```
frassinier, line 22: 👍🏻 that's nice, a great improvement!

You might use a real CSS module instead of using a global style, to be really delightful.

frassinier (approved):
Awesome!

PR #3797, opened 2022-03-23
chore(ARCH-482): refactor datalist to remove deprecated lifecycle

jsomsanith-tlnd:
Datalist uses `componentWillReceiveProps` to 
- compute derived state from props
- set recompute suggestions after the derived state has been set as state

Because of the second point, we cannot use `getDerivedStateFromProps` as we cannot call a callback.

Let's refactor this component to modernize the logic with react function components and hooks


packages/components/src/Datalist/Datalist.component.js
```diff
@@ -14,111 +15,295 @@ export function escapeRegexCharacters(str) {
...
+	// suggestions: filter value, display flag, current hover selection
+	const [filterValue, setFilterValue] = useState('');
+	const [displaySuggestion, setDisplaySuggestion] = useState(DISPLAY.NONE);
+	const [selection, setSelection] = useState({
+		focusedItemIndex: undefined,
+		focusedSectionIndex: undefined,
+	});
+
+	// suggestions computation
+	const suggestions = useMemo(
```
Benrajalu, line 123:
```suggestion
	const suggestions = useMemo(
```
Is there an observable benefit to useMemo here?

jsomsanith-tlnd: hmm indeed, it changes at every change in fact.
With you comment I even try to think about possible optimisation like decoupling the hide/display from the suggestion computation, but when we hide the suggestions, the next open needs either
-  to have the whole liste, not filtered one, in case of newly focus/click, or key down/up
- to update the filtered suggestions in case we already are in input but closes suggestions and type in it again

jsomsanith-tlnd: Removed it

tlnd-mhuchet (approved):

packages/components/src/Datalist/Datalist.component.js
```diff
@@ -14,111 +15,285 @@ export function escapeRegexCharacters(str) {
...
+	// suggestions: filter value, display flag, current hover selection
+	const [filterValue, setFilterValue] = useState('');
+	const [displaySuggestion, setDisplaySuggestion] = useState(DISPLAY.NONE);
+	const [selection, setSelection] = useState({
+		focusedItemIndex: undefined,
+		focusedSectionIndex: undefined,
+	});
+
+	// suggestions computation
+	// we don't memo bacause it changes at EVERY modification
```
Benrajalu, line 123:
```suggestion
	// we don't memo because it changes at EVERY modification
```

Benrajalu (approved):
I did not pull this, so there's some trust in your testing here

PR #3845, opened 2022-04-06
chore(ARCH-482/EnumerationWidget): remove deprecated lifecycle and use RTL for tests

jsomsanith-tlnd:
Deprecated lifecycle in old EnumerationWidget

Remove deprecated lifecycle and migrate tests to RTL


packages/forms/src/deprecated/widgets/EnumerationWidget/EnumerationWidget.js
```diff
@@ -252,8 +253,10 @@ class EnumerationForm extends React.Component {
 		};
 	}
 
-	componentWillReceiveProps(nextProps) {
-		this.setState(prevState => ({ ...prevState, items: nextProps.formData }));
+	componentDidUpdate(prevProps) {
+		if (this.props.formData !== prevProps.formData) {
```
jmfrancois, line 257: formData is an array.
An array  may not change by reference but still have new items in it.
I believe we may lost re render here.

jsomsanith-tlnd: This widget is from the deprecated forms and I don't want to change any behavior.
The previous behavior was componentWillReceiveProps, this function is generally called if the props passed to the component change. This is done by reference. 
If you modify the array keeping the same reference, first it's a bad practice as we should prefer immutability, then it will not work as expected, like before.
Additional support for mutable array is not needed as this is deprecated.

packages/forms/src/deprecated/widgets/EnumerationWidget/EnumerationWidget.js
```diff
@@ -252,8 +253,10 @@ class EnumerationForm extends React.Component {
...
+			this.setState({ items: this.props.formData });
```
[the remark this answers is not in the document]
jsomsanith-tlnd: Done
(But for the record, it's exactly the same :) )

jmfrancois (approved):

romainseb (approved):

PR #3865, opened 2022-04-07
fix(ComponentForm): infinite rendering because of new Map ref everytime

jsomsanith-tlnd:
componentDidUpdate set states when properties have changed. But the comparison is ko when there is no properties passed, as it takes a new Map by default.

Compare the properties properly without default values to check if they have changed

### DIFF
--- a/packages/containers/src/ComponentForm/ComponentForm.component.js
+++ b/packages/containers/src/ComponentForm/ComponentForm.component.js
@@ -9,6 +9,8 @@ import memoizeOne from 'memoize-one';
 import kit from './kit';
 import tcompFieldsWidgets from './fields';
 
+const emptyMap = Map();
+
 const TO_OMIT = [
 	'definitionURL',
 	'uiSpecPath',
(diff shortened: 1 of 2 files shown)


gjou-tlnd (approved):

romainseb (approved):

jmfrancois (approved):

jmfrancois (changes requested):
datalist seems broken http://3865.talend-playground.surge.sh/playground/

packages/containers/src/ComponentForm/ComponentForm.component.js
```diff
@@ -102,9 +102,9 @@ export class TCompForm extends React.Component {
 	}
 
 	componentDidUpdate(prevProps) {
-		const nextProperties = this.props.state.get('properties', Map());
```
sgendre, line 105: extracting the allocation outside of the function would also work and bring less noise isn't it?

jsomsanith-tlnd: what ?

jmfrancois (approved):
datalist use triggers sorry

PR #3871, opened 2022-04-08
test(ARCH-497/module-to-cdn): fix tests and move them to cron

jmfrancois:

[module-to-cdn modules.json](https://unpkg.com/@talend/module-to-cdn/modules.json) file is crucial for us.
Every new release may break a build and can be hard for developers to understand what happens.

Current tests are runned only on next versions not on range edge versions which mean we do not cover the current range configuration.

* Fix the tests to be runned
* Move them to a dedicated file
* use a dedicated test command for it because thoses takes 2 minutes to run with cache and can be 30 minutes without cache.
* create a cron workflow that will create an issue if test fails

Screenshot 2022-04-08 at 16 56 41
...
Screenshot 2022-04-08 at 16 56 51

### DIFF
--- a/.github/workflows/tests-cron.yml
+++ b/.github/workflows/tests-cron.yml
@@ -23,7 +23,7 @@ jobs:
           scope: "@talend"
           cache: "yarn"
 
-      - name: Upgrade dependencies
+      - name: Install and tests
         run: |
           yarn --frozen-lockfile --ignore-scripts
           yarn test:cron
--- a/.github/workflows/tests-cron.yml
+++ b/.github/workflows/tests-cron.yml
@@ -9,7 +9,6 @@ jobs:
   tests:
     name: Long running Tests
     runs-on: ubuntu-latest
-    continue-on-error: true # we want the issue to popup even if sth goes wrong
 
     steps:
       - name: Checkout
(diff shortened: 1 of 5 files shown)


.github/workflows/tests-cron.yml
```diff
@@ -0,0 +1,44 @@
...
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 9 * * 1-5" # At 09:00 AM, Monday through Friday
+
+jobs:
+  tests:
+    name: Long running Tests
+    runs-on: ubuntu-latest
+    continue-on-error: true # we want the issue to popup even if sth goes wrong
```
smouillour, line 12: why do you need this?
You already added the if condition to manage the case of failure. https://github.com/Talend/ui/pull/3871/files#diff-d3782acde429bd2a63f81d47345c3de6f798bad3d04f757d2813bb77ece283a4R32

jmfrancois, line 12:
```suggestion
```

jmfrancois: indeed, let's remove it

smouillour (approved):

jsomsanith-tlnd (approved):
Thank you for that, I got a test fail twice on PRs because of that, but it wasn't related to my changes :D

PR #3897, opened 2022-04-14
feat(components/QualityBar): rework quality bar with design tokens

AdelineMrtn:
- Redesign style of the QualityBar component
- Use the QualityBar component within datagrid package

- Import design tokens in order to use new colors for each ratio-bar-line of QualityBar component
- Adapt css for space between each ratio-bar-line
- Import QualityBar component from '@talend/react-components' in DefaultHeaderRenderer component and adapt props
- Delete all QualityBar files in /packages/datagrid 


Nigui (approved):

jmfrancois:
Colors intensity is really high !
before
Screenshot 2022-04-28 at 15 09 57
after
Screenshot 2022-04-28 at 15 09 54

jmfrancois (changes requested):
Colors intensity is really high !
before
Screenshot 2022-04-28 at 15 09 57
after
Screenshot 2022-04-28 at 15 09 54

This is really nice refactoring but unfinished: the color in the cell is not the same anymore
Screenshot 2022-04-28 at 15 11 51

gjou-tlnd:
Looks like the new header height is not taken into account in SB: 
Screenshot 2022-04-28 at 15 17 50

.changeset/wise-chicken-divide.md
```diff
@@ -0,0 +1,9 @@
+---
+'@talend/react-datagrid': major
+'@talend/react-components': patch
```
jmfrancois, line 3: one changeset per changelog. changeset is not able to split the comments below.
This is ok if you want the same changelog in every packages whcih is not the case here.

jmfrancois (approved):

packages/components/src/RatioBar/RatioBar.scss
```diff
@@ -13,21 +9,21 @@ $ratio-bar-hover-height: 8px;
 .tc-ratio-bar {
 	padding: $padding-small 0;
 	display: flex;
-	height: $ratio-bar-hover-height;
-	width: $ratio-bar-width;
+	height: tokens.$coral-spacing-xs;
```
Benrajalu, line 12: Please use `sizing` tokens for sizes 🙏

Benrajalu: If they're not in your bundle, you're using an outdated version of the design tokens :)

Kporal: Oh yes, you're right, I got confused!
But `tokens.$coral-spacing-xs` = 0.8rem and the smaller size we have is `tokens.$coral-sizing-minimal` = 1.2rem 🤔 
What do you suggest to do then? keep it as it is for now? revert it to a custom value?

Benrajalu: If I remember correctly, @AnLefevre did specify this would be an exception to the token scale. Nobody likes them, but they may occur... When they do, we'd advise for custom values clearly labeled as such (`$custom-ratio-bar-height: 0.8rem`) so we can track all `custom` stuff down the line.

Kporal: Ok thanks for the explanation, I'll go for that then 👍

AnLefevre:
Style looks good. Just need to review the transition/animation
Please use -> transition/instant 100ms ease-out.

PR #3914, opened 2022-04-22
feat(design-system/Modal): Modal component

sgangnant:
https://jira.talendforge.org/browse/TUX-1157
Modal guidelines have been done, this is the implementation.

Implement it.


packages/design-system/src/components/Modal/Modal.scss
```diff
@@ -0,0 +1,99 @@
...
+	position: absolute;
+	top: 0;
+	bottom: 0;
+	left: 0;
+	right: 0;
+	z-index: $modal-backdrop-z-index;
+}
+
+.modal {
+	position: absolute;
```
Benrajalu, line 18:
```suggestion
	position: fixed;
```
With this you can use percentages instead of "vw" and "vh". "vh" specifically is a bitch.

packages/design-system/src/components/Modal/Modal.scss
```diff
@@ -0,0 +1,99 @@
...
+		flex-shrink: 0;
+		height: 60px;
+		padding: tokens.$coral-spacing-xs tokens.$coral-spacing-l;
+		border-bottom: tokens.$coral-border-s-solid tokens.$coral-color-neutral-border-weak;
+
+		display: flex;
+		align-items: center;
+		column-gap: tokens.$coral-spacing-m;
+
+		&--with-tabs {
```
Benrajalu, line 56: There's currently no tab support in this modal AFAIK. If you need them now, then we need to update the Figma specs.

sgangnant: Saw it in the examples, but indeed it didn't looked finished yet, removed the code related to that.

packages/design-system/src/components/Modal/Modal.tsx
```diff
@@ -1,88 +1,84 @@
...
-	(
-		{ disclosure, children, icon, title, subtitle, onClose, onValidate, ...props }: ModalProps,
-		ref: React.Ref<HTMLDivElement>,
-	) => {
-		const dialog = useDialogState({ animated: true });
-		const { t } = useTranslation();
+export type ModalPropsType = {
+	title: ReactNode;
+	description?: string;
+	icon?: ModalIcon;
```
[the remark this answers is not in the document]
sgangnant: You mean `title` should be an object ?

```
<Modal
   title={{
    text: 'Modal title',
    description: 'Modal description',
    icon: 'talend-cog',
  }}
/>
```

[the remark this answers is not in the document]
Benrajalu: For instance yes. Could be called `header`: 

```
type header = {
    title: string;
    icon?: IconName;
    description?: string;
}
```

packages/design-system/src/components/Modal/Modal.scss
```diff
@@ -0,0 +1,99 @@
...
+
+	&__title {
+		font: tokens.$coral-heading-l;
+	}
+}
+
+.modal-icon {
+	flex-grow: 0;
+	width: 24px;
+	height: 24px;
```
Benrajalu, line 94:
```suggestion
	width: tokens.$coral-sizing-xs;
	height: tokens.$coral-sizing-xs;
```

packages/design-system/src/components/Modal/Modal.tsx
```diff
@@ -1,88 +1,121 @@
...
-const Modal = React.forwardRef(
-	(
-		{ disclosure, children, icon, title, subtitle, onClose, onValidate, ...props }: ModalProps,
-		ref: React.Ref<HTMLDivElement>,
-	) => {
-		const dialog = useDialogState({ animated: true });
-		const { t } = useTranslation();
+function Modal(props: ModalPropsType): ReactElement {
+	const { header, onClose, primaryAction, secondaryAction, preventEscaping, children } = props;
+	const dialog = { visible: true };
```
Benrajalu, line 40:
```suggestion
      const dialog = useDialogState({ visible: true });
```
For now. But we can't skip the Reakit state management.

packages/design-system/src/components/Modal/Modal.tsx
```diff
@@ -1,88 +1,121 @@
...
+}
+
+export type ModalPropsType = {
+	header: {
+		title: ReactNode;
+		description?: string;
+		icon?: ModalIcon;
+	};
+	onClose: Function;
+	primaryAction?: ButtonPrimaryPropsType & { destructive?: boolean };
```
Benrajalu, line 32: This only works for now because ButtonPrimaryPropsType and ButtonDestructivePropTypes have the same type, more or less. It will break as soon as those evolve. 

```suggestion
	primaryAction?: ButtonPrimaryPropsType || { destructive?: boolean & ButtonDestructivePropTypes };
```

packages/design-system/src/components/Modal/Modal.scss
```diff
@@ -0,0 +1,99 @@
...
+	box-shadow: tokens.$coral-elevation-shadow-neutral-m;
+	border-radius: tokens.$coral-radius-s;
+	background: tokens.$coral-color-neutral-background;
+	display: flex;
+	flex-direction: column;
+	overflow: hidden;
+
+	> :first-child {
+		overflow: hidden;
+	}
```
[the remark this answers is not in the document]
sgangnant: In that case, it's the root `StackVertical`, it was the only way to get the content wrapper working with `overflow: auto` (I had hard times making it work, mainly due to the dynamic sizing of the main `.modal` component.
If you see how I can improve this without setting fixed dimensions, I'd be glad to get your insights

packages/design-system/src/components/Modal/Modal.tsx
sgangnant: Let me have a look so that I don't leave traps in the code for our future selves

packages/design-system/src/components/Modal/Modal.tsx
```diff
@@ -1,88 +1,126 @@
...
 
-		const onValidateHandler = () => {
-			if (onValidate) {
-				onValidate();
-			}
-			dialog.hide();
-		};
+	let primaryActionRendered;
+	if (primaryAction) {
+		const dataTest = 'modal.buttons.primary-action';
```
frassinier, line 48:
```suggestion
		const dataTest = 'modal.buttons.primary';
```

packages/design-system/src/components/Modal/Modal.tsx
```diff
@@ -1,88 +1,126 @@
...
+		<DialogBackdrop
+			{...dialog}
+			className={modalStyles['modal-backdrop']}
+			data-test="modal.backdrop"
+		>
+			<Dialog
+				{...dialog}
+				data-test="modal"
+				aria-title={header.title}
+				className={modalStyles['modal']}
```
frassinier, line 70:
```suggestion
				className={style.modal}
```

packages/design-system/src/components/Modal/Modal.scss
```diff
@@ -0,0 +1,101 @@
...
+
+	&__content,
+	&__buttons {
+		padding: 0 tokens.$coral-spacing-xl;
+	}
+
+	&__header {
+		flex-shrink: 0;
+		height: 60px;
```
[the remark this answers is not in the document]
sgangnant: What's the rule about this ? "Never px, always rem" ? Should the max-width be handled the same ?

packages/design-system/src/components/Modal/Modal.scss
```diff
@@ -0,0 +1,101 @@
...
+	background: tokens.$coral-color-neutral-background;
+	display: flex;
+	flex-direction: column;
+	overflow: hidden;
+
+	> :first-child {
+		overflow: hidden;
+	}
+
+	&__header,
```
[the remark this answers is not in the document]
sgangnant: ?

packages/design-system/src/components/Modal/Modal.spec.tsx
```diff
@@ -0,0 +1,58 @@
...
+const { Basic, WithNonClosingBackdrop } = composeStories(Stories);
+
+context('<Modal />', () => {
+	it('should render', () => {
+		cy.mount(<Basic />);
+		cy.getByTest('open-modal').click();
+		cy.getByTest('modal').should('be.visible');
+	});
+
+	it('should close the modal on cancel/close action', () => {
```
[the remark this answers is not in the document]
sgangnant: Ah yes good call, didn't think about it but now it's possible (stories are no longer rendering the modal open by default but rely on a button created for this)

packages/design-system/src/components/Modal/Modal.scss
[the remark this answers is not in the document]
Benrajalu: Yep, `rem` for every value that isn't a proportion.

packages/design-system/src/components/Modal/Modal.stories.mdx
```diff
@@ -0,0 +1,98 @@
...
+
+<FigmaImage
+	src="https://www.figma.com/file/0Jolh2prAAdfO5224n3OU3/Modal?node-id=738%3A12433"
+	alt="Zoning image for Modal"
+/>
+
+## Style
+
+<FigmaImage
+	src="https://www.figma.com/file/0Jolh2prAAdfO5224n3OU3/Modal?node-id=0%3A43"
```
Benrajalu, line 30:
```suggestion
	src="https://www.figma.com/file/0Jolh2prAAdfO5224n3OU3/Modal?node-id=941%3A9946"
```

Benrajalu: This is a better illustration for now

packages/design-system/src/components/Modal/Modal.spec.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+		// when
+		cy.mount(<WithNonClosingBackdrop onClose={onCloseStub} />);
+		cy.getByTest('open-modal').click();
+		cy.getByTest('modal.buttons.close').click();
+		cy.getByTest('modal').type('{esc}');
+
+		// then
+		expect(onCloseStub).not.to.be.called;
+	});
```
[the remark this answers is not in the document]
sgangnant: Found a bug in Reakit, initial focus is not well managed, I'll add it in our project for now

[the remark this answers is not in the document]
sgangnant: Fixed the issue and extended initial test case (cypress doesn't support `tab` interaction and I couldn't make the cypress-plugin-tab to work for this case https://github.com/kuceb/cypress-plugin-tab)

packages/design-system/src/components/Modal/Modal.scss
```diff
@@ -0,0 +1,112 @@
...
+		text-overflow: ellipsis;
+	}
+
+	&__title {
+		font: tokens.$coral-heading-l;
+	}
+}
+
+.modal-icon {
+	flex-grow: 0;
```
Benrajalu, line 105:
```suggestion
	flex-grow: 0;
	flex-shrink: 0;
```
Or is that a default? I never remember.

packages/design-system/src/components/Modal/Modal.stories.mdx
```diff
@@ -0,0 +1,131 @@
...
+import { FigmaImage, FigmaLink, Use } from '~docs';
+
+import Modal from './';
+
+import * as Stories from './Modal.stories';
+
+<Meta
+	title="Components/Modal"
+	parameters={{
+		status: { figma: 'ok', storybook: 'ok', react: 'ok', i18n: 'na' },
```
Benrajalu, line 11:
```suggestion
		status: { figma: 'ok', storybook: 'wip', react: 'ok', i18n: 'na' },
```
We'd still need to illustrate, but I wouldn't block a release for this.

packages/design-system/src/components/Modal/Modal.scss
sgangnant: According to Google 1 is the default for shrink

packages/design-system/src/components/Modal/Modal.stories.mdx
```diff
@@ -0,0 +1,131 @@
...
+Actions: Up to two, one primary, one secondary. Ideally can’t have a secondary if there’s no primary.
+Destructive: The primary action can be destructive.
+
+### Backdrop
+
+Transparent layer behind every modal. Clicking it closes the modal by default, but this behavior can be disabled by a prop.
+
+### Variations
+
+**Basic**
```
Benrajalu, line 63:
```suggestion
**No disclosure (controlled state)**
```

packages/design-system/src/components/Modal/Modal.stories.tsx
```diff
@@ -1,39 +1,249 @@
...
-);
-Confirm.parameters = {};
-
-export const Invite = () => (
+function ModalStory(props: Partial<ModalPropsType>) {
+	const [modalOpen, setModalOpen] = useState(false);
+
+	return (
+		<>
+			<button onClick={() => setModalOpen(true)} data-test="open-modal">
```
Benrajalu, line 17:
```suggestion
			<ButtonPrimary onClick={() => setModalOpen(true)} data-test="open-modal">
```
#corporate

packages/design-system/src/components/Modal/Modal.stories.tsx
```diff
@@ -1,39 +1,249 @@
...
+						setModalOpen(false);
+					}}
+					{...props}
+				/>
+			)}
+		</>
+	);
+}
+
+function QuicheRecipe(): ReactElement {
```
Benrajalu, line 36:
```suggestion
function QuicheRecipe() {
```
Do we really need to set the type?

packages/design-system/src/components/Modal/Modal.stories.tsx
```diff
@@ -1,39 +1,249 @@
...
+);
+
+export const WithDisclosure: ComponentStory<typeof Modal> = props => (
 	<Modal
-		baseId=""
+		{...props}
+		header={{ title: 'With disclosure' }}
 		disclosure={
-			<ButtonPrimary onClick={action('Clicked disclosure')}>Send invitation email</ButtonPrimary>
+			<ButtonPrimary data-test="modal-disclosure" onClick={() => {}}>
```
Benrajalu, line 149:
```suggestion
			<ButtonPrimary data-testid="modal-disclosure" onClick={() => {}}>
```
IIRC (@frassinier) for things that would come from the project's side, then we use `data-testid`. This mimics what would happen in a product. 

https://github.com/Talend/ui/blob/master/packages/design-system/adr/data-test.md

packages/design-system/src/components/Modal/Modal.tsx
```diff
@@ -1,88 +1,142 @@
...
+			primaryActionRendered = (
+				<ButtonPrimary {...(primaryAction as ButtonPrimaryPropsType)} data-test={dataTest} />
+			);
+		} else {
+			const { destructive, ...buttonProps } = primaryAction;
+			primaryActionRendered = (
+				<ButtonDestructive {...(buttonProps as ButtonDestructivePropsType)} data-test={dataTest} />
+			);
+		}
+	}
```
Benrajalu, line 70:
```suggestion
function PrimaryAction() {
      
	if (primaryAction) {
		const dataTest = 'modal.buttons.primary';

		if (!('destructive' in primaryAction) || !primaryAction.destructive) {
			return (
				<ButtonPrimary {...(primaryAction as ButtonPrimaryPropsType)} data-test />
			);
		} else {
			const { destructive, ...buttonProps } = primaryAction;
			return (
				<ButtonDestructive {...(buttonProps as ButtonDestructivePropsType)} data-test={dataTest} />
			);
		}
	}
}
```

packages/design-system/src/components/Modal/Modal.tsx
```diff
@@ -1,88 +1,142 @@
...
+
+	return (
+		<>
+			{disclosure && (
+				<DialogDisclosure {...dialog}>
+					{disclosureProps => React.cloneElement(disclosure, disclosureProps)}
+				</DialogDisclosure>
+			)}
+			{dialog.visible && (
+				<DialogBackdrop {...dialog} className={styles['modal-backdrop']} data-test="modal.backdrop">
```
Benrajalu, line 80:
```suggestion
				<DialogBackdrop {...dialog} className={styles.modal-backdrop} data-test="modal.backdrop">
```
Does that work? That's the way we handle styles in the other files and we'd prefer coding styles to be aligned.

packages/design-system/src/pages/blocks/HeaderBar.js
```diff
@@ -65,7 +65,7 @@ export default function HeaderBarBlock() {
 						<Dropdown
 							aria-label="Apps switcher"
 							items={[
-								{ onClick: () => {}, label: 'About', type: 'button' },
+								{ onClick: () => setAboutOpen(true), label: 'About', type: 'button' },
```
Benrajalu, line 68: Well that's not great 🤔 We may have to open `Dropdown`'s `items` to ReactElements so you can use `<Modal disclosure=....` here. Something to keep in mind.

packages/design-system/src/components/Modal/Modal.stories.tsx
sgangnant: Not really needed (TS inference would probably work here)

packages/design-system/src/components/Modal/Modal.tsx
sgangnant: Nope because it would evaluate the mathematical expression (_a minus b_)

packages/design-system/src/pages/blocks/HeaderBar.js
sgangnant: Yes that's what I also thought, it would be far more convenient. Out of the scope of this IMO though.

Benrajalu (approved):

romainseb (approved):

PR #3935, opened 2022-04-28
feat: add preinstall scripts

jmfrancois:

current mono repo start to be quite big.
CI take times
dev experience can be quite cumbersome

we use changeset so we know which packages has been changed and not released.
From this point it means we can compute the workspace before doing installation.

* add preinstall hook to compute the workspace from changeset files
* find a way to keep the demo from builted packages


package.json
```diff
@@ -6,9 +6,12 @@
     "cross-spawn": "^7.0.3",
     "husky": "^7.0.4",
     "i18next-scanner": "^3.1.0",
-    "prettier": "^2.6.2"
+    "prettier": "^2.6.2",
+    "eslint": "^7.32.0",
+    "lint-staged": "^12.3.8"
```
jmfrancois, line 11: interesting to see we use this from root package but they comes from underlying packages :)

PR #3969, opened 2022-05-04
fix(assetApi): prevent getCdnUrl to start with double slash

Kporal:

`Talend.getCDNUrl` is doing a concatenation of both `base` url and `CDN_URL` env variable.
In most of our apps, base is set to `/` and `CDN_URL` is set to `/cdn` which leads to `//cdn/etc` 
The browser then fails to call the resource due to double slashes

Prevent the getCDNUrl to return an url starting with a double slash.

### DIFF
--- a/packages/assets-api/src/index.ts
+++ b/packages/assets-api/src/index.ts
@@ -186,7 +186,7 @@ if (!window.Talend.getCDNUrl) {
 				const baseTag = document.querySelector('base');
 				if (baseTag) {
 					const root = baseTag.getAttribute('href') || '';
-					if (`${root}${CDN_URL}`.substr(0, 2) !== '//') {
+					if (root !== '/') {
 						return `${root}${CDN_URL}/${info.name}/${info.version}${info.path}`;
 					}
 				}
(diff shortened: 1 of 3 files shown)


jmfrancois (approved):

packages/assets-api/src/index.ts
```diff
@@ -186,7 +186,9 @@ if (!window.Talend.getCDNUrl) {
 				const baseTag = document.querySelector('base');
 				if (baseTag) {
 					const root = baseTag.getAttribute('href') || '';
-					return `${root}${CDN_URL}/${info.name}/${info.version}${info.path}`;
+					if (`${root}${CDN_URL}`.substr(0, 2) !== '//') {
```
jsomsanith-tlnd, line 189:
```suggestion
					if (root !== '/') {
```

jsomsanith-tlnd: @Kporal Have I missed a case ? ☝️

Kporal: No it seems good too, thanks 👍

jsomsanith-tlnd (approved):

PR #3984, opened 2022-05-06
fix(Dropdown): Close on button or link click

romainseb:
Currently, we have no way to close the dropdown other than click outside.
While the user has made his selection in the dropdown, the menu should close itself when actionned

Close automatically the menu while clicking on a button or a link.
For the link part, i was asking myself if it's required but for some use cases of internal routing, if the dropdown stays in the view, it could have some purprose.

### DIFF
--- a/packages/design-system/src/components/Dropdown/Dropdown.spec.tsx
+++ b/packages/design-system/src/components/Dropdown/Dropdown.spec.tsx
@@ -28,6 +28,14 @@ context('<Dropdown />', () => {
 		cy.getByTest('dropdown.menu').should('not.be.visible');
 	});
 
+	it('should hide the menu clicking on a link', () => {
+		cy.mount(<WithIcons />);
+		cy.getByTest('dropdown.button').click();
+		cy.getByTest('dropdown.menu').should('be.visible');
+		cy.get('a[data-test="dropdown.menuitem"]').eq(0).invoke('removeAttr', 'href').click();
+		cy.getByTest('dropdown.menu').should('not.be.visible');
+	});
+
 	it('should display menu with keyboard', () => {
 		cy.mount(<WithIcons />);
 		cy.getByTest('dropdown.button').type(' ');
--- a/packages/design-system/src/components/Dropdown/Dropdown.spec.tsx
+++ b/packages/design-system/src/components/Dropdown/Dropdown.spec.tsx
@@ -20,6 +20,14 @@ context('<Dropdown />', () => {
 		cy.getByTest('dropdown.menu').should('not.be.visible');
 	});
 
+	it('should hide the menu clicking on a button', () => {
+		cy.mount(<WithIcons />);
+		cy.getByTest('dropdown.button').click();
+		cy.getByTest('dropdown.menu').should('be.visible');
+		cy.get('button[data-test="dropdown.menuitem"]').click();
+		cy.getByTest('dropdown.menu').should('not.be.visible');
+	});
+
 	it('should display menu with keyboard', () => {
 		cy.mount(<WithIcons />);
 		cy.getByTest('dropdown.button').type(' ');
--- a/packages/design-system/src/components/Dropdown/Dropdown.tsx
+++ b/packages/design-system/src/components/Dropdown/Dropdown.tsx
@@ -1,4 +1,4 @@
-import React, { cloneElement, forwardRef, MouseEventHandler, ReactElement, Ref } from 'react';
+import React, { cloneElement, forwardRef, ReactElement, Ref } from 'react';
 import { Menu, MenuButton, useMenuState } from 'reakit';
 import { IconName } from '@talend/icons';
 import DropdownButton from './Primitive/DropdownButton';
--- a/packages/design-system/src/components/Dropdown/Dropdown.tsx
+++ b/packages/design-system/src/components/Dropdown/Dropdown.tsx
@@ -93,8 +93,11 @@ const Dropdown = forwardRef(
 								{...menu}
 								key={`${label}-${index}`}
 								id={`${label}-${index}`}
-								onClick={() => {
+								onClick={event => {
 									menu.hide();
+									if (entry.onClick) {
+										entry.onClick(event);
+									}
 								}}
 								data-test="dropdown.menuitem"
 							>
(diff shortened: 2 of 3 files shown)


packages/design-system/src/components/Dropdown/Dropdown.spec.tsx
```diff
@@ -20,6 +20,14 @@ context('<Dropdown />', () => {
 		cy.getByTest('dropdown.menu').should('not.be.visible');
 	});
 
+	it('should hide the menu clicking on a button', () => {
+		cy.mount(<WithIcons />);
+		cy.getByTest('dropdown.button').click();
+		cy.getByTest('dropdown.menu').should('be.visible');
+		cy.get('button[data-test="dropdown.menuitem"]').click();
```
frassinier, line 27: Isn't enough?
```suggestion
		cy.getByTest("dropdown.menuitem").click();
```

packages/design-system/src/components/Dropdown/Dropdown.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { cloneElement, forwardRef, ReactElement, Ref } from 'react';
+import React, { cloneElement, forwardRef, MouseEventHandler, ReactElement, Ref } from 'react';
```
frassinier, line 1: Is it really in use?

Nigui (approved):

packages/design-system/src/components/Dropdown/Dropdown.spec.tsx
romainseb: The first item is a link, not the button

packages/design-system/src/components/Dropdown/Dropdown.tsx
romainseb: Doh, i add it as some point to type a bit the events in my iteration but there is no use for it now.
I remove it

packages/design-system/src/components/Dropdown/Dropdown.tsx
```diff
@@ -89,6 +93,9 @@ const Dropdown = forwardRef(
 								{...menu}
 								key={`${label}-${index}`}
 								id={`${label}-${index}`}
+								onClick={() => {
+									menu.hide();
```
Benrajalu, line 97: We could still have `onClick` for this link though, if I'm not mistaken.

packages/design-system/src/components/Dropdown/Dropdown.tsx
```diff
@@ -57,6 +57,10 @@ const Dropdown = forwardRef(
 								<DropdownButton
 									{...entryRest}
 									{...menu}
+									onClick={event => {
+										menu.hide();
+										entry?.onClick(event);
```
jsomsanith-tlnd, line 62: `entry` can't be null/undefined  or it breaks before.

```suggestion
										entry.onClick(event);
```

packages/design-system/src/components/Dropdown/Dropdown.spec.tsx
frassinier: Could be nice to test both :)

packages/design-system/src/components/Dropdown/Dropdown.tsx
romainseb: That's not what indicate the type we want for the link
https://github.com/Talend/ui/blob/2119fdf1432dfaeb521dccd8a610a2e5ab3ac72a/packages/design-system/src/components/Dropdown/Dropdown.tsx#L19

But i can alter it in this PR if you want so :)

packages/design-system/src/components/Dropdown/Dropdown.tsx
romainseb: Fair point

packages/design-system/src/components/Dropdown/Dropdown.tsx
Benrajalu: Yup `LinkableType = Omit<AnchorHTMLAttributes<HTMLAnchorElement>, 'style'>...`

Benrajalu: And `AnchorHTMLAttributes<HTMLAnchorElement>` can take `onClick`

jsomsanith-tlnd: I explored the typing too and it doesn't seem possible

jsomsanith-tlnd (approved):

packages/design-system/src/components/Dropdown/Dropdown.tsx
jsomsanith-tlnd: image

packages/design-system/src/components/Dropdown/Dropdown.spec.tsx
romainseb: If we click on the link, we are redirected and the test fall into space, do you have a trick to prevent the navigation or something ? (without altering the story)

packages/design-system/src/components/Dropdown/Dropdown.tsx
Benrajalu: `onClick` is valid on all HTMLAttributes. The screenshot only shows the subset that's valid only for Anchor.
Links can have `onClick` handlers (for tracking, for this instance etc...), but buttons can't have `href` attributes.

jmfrancois: just tried out and it works, so you must add the entry?.onClick here

romainseb: I can add the on click if wanted so
I don't have really strong strong opinion about this.
I would tend to think like Jimmy and if you want to have a callback, it's more a button to place but i can add it in order to be less restrictive

jmfrancois (approved):

packages/design-system/src/components/Dropdown/Dropdown.tsx
romainseb: (added 🥳 )

Benrajalu: Let's not get confused here :D 

- If you want a callback, use a button 
- If you want navigation, use a link. If you need a callback upon navigation, HTML says you can.

packages/design-system/src/components/Dropdown/Dropdown.spec.tsx
frassinier: 🤷🏻 
```js
cy.getByTest("dropdown.menuitem").eq(0).invoke('removeAttr', 'target').click();
```

romainseb: Thanks for the trick 👍🏼 ❤️

Benrajalu (approved):

PR #4028, opened 2022-05-18
fix(ARCH-515/cdn): use findPackage instead of require.resolve

jmfrancois:

in the case you have storybook and react-cmf-router yarn may install you the history into different possible context.

require.resolve(history) will return you the one closest to the provided webpack context.

```
const resolvePkg = require('@talend/dynamic-cdn-webpack-plugin/src/resolve-pkg');
resolvePkg('history', {cwd: './node_modules/@talend/react-cmf-router'})
// '/Users/jmfrancois/github/talend/data-processing-pipeline-designer/webapp/node_modules/history/main.js'
```

but that is history 5 !! and not history 3 (so the one for storybook).
so we ends with 404 on the cdn request.

* add check on semver.major from the dependencies.json and throw exception if it do not match
* use findPackage on this particular part to be sure we get the right version here
* ensure compatibility with the rest of the code

### DIFF
--- a/fork/dynamic-cdn-webpack-plugin/src/find.js
+++ b/fork/dynamic-cdn-webpack-plugin/src/find.js
@@ -13,7 +13,8 @@ function findPackage(info) {
 		const {
 			packageJson: { version },
 		} = readPkgUp.sync({ cwd });
-		return semver.major(version) === semver.major(info.version);
+		// check we are at least upper or equal using caret range syntax
+		return semver.satisfies(`^${info.version}`, version);
 	});
 }
 
--- a/fork/dynamic-cdn-webpack-plugin/src/find.js
+++ b/fork/dynamic-cdn-webpack-plugin/src/find.js
@@ -10,10 +10,9 @@ function findPackage(info) {
 		[scope, name] = info.name.split('/');
 	}
 	return findPackages(scope, name).find(cwd => {
-		const pkg = readPkgUp.sync({ cwd });
 		const {
 			packageJson: { version },
-		} = pkg;
+		} = readPkgUp.sync({ cwd });
 		// check we are at least upper or equal using caret range syntax
 		return semver.satisfies(version, `^${info.version}`);
 	});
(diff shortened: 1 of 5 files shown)


fork/dynamic-cdn-webpack-plugin/src/resolve-pkg.js
```diff
@@ -6,6 +8,13 @@
  * @returns {string|undefined} the full path of the module id requested
  */
 function resolve(moduleId, options) {
+	if (options.version) {
+		let scope;
+		if (moduleId.startsWith('@')) {
+			scope = moduleId.split('/')[0];
+		}
```
jsomsanith-tlnd, line 15: What is it for ? This is not used

fork/dynamic-cdn-webpack-plugin/src/find.js
```diff
@@ -13,7 +13,7 @@ function findPackage(info) {
 		const {
 			packageJson: { version },
 		} = readPkgUp.sync({ cwd });
-		return version === info.version;
+		return semver.major(version) === semver.major(info.version);
```
jsomsanith-tlnd, line 16: This is less strict as before, comparing only the major version. Wouldn't it introduce possible side effects if 2 versions are present (not deduped) and the one that is selected doesn't contain the used feature for example ?

fork/dynamic-cdn-webpack-plugin/src/resolve-pkg.js
jmfrancois: oh thanks I have first use findPackages which is ask you a scope... removed

jmfrancois: oh thanks I have first use findPackages which is ask you a scope... removed

fork/dynamic-cdn-webpack-plugin/src/find.js
jmfrancois: yes exact, it could create bugs, I have fixed by using caret syntax so now we check against semver compatible with the version it has been compiled with.

jsomsanith-tlnd (approved):

tlnd-mhuchet (approved):

PR #4034, opened 2022-05-23
fix(components): datalist should show selected item by id

Gbacc:
On a datalist , when having a titlemap with multiple items with the same name, the selected item is not correctly displayed

We can now pass a `valueId` prop to the Typeahead component to be able to check a selected item based on its id/value instead of title/name. It is still possible to show selected item on a Typeahead component by title/name if no `valueId` prop is provided.


packages/components/src/Typeahead/Typeahead.snapshot.test.js
```diff
@@ -153,6 +173,37 @@ describe('Typeahead', () => {
...
+				value: 'le title 1',
+				valueId: 'letitle1copy',
+				items: itemsObjectWithId,
+			};
+
+			// when
+			const wrapper = renderer.create(<Typeahead {...props} />).toJSON();
+
+			// then
+			expect(wrapper).toMatchSnapshot();
```
tlnd-mhuchet, line 190: Are we sure it is this one selected? Because we only see the title `le title 1` in the snapshot, not the associated id. 🤔

tlnd-mhuchet (approved):

packages/components/src/Typeahead/Typeahead.snapshot.test.js
jsomsanith-tlnd: Please avoid snapshots. I know there are lots of snapshots in this file but we need to get rid of that.
The reason is that, looking at your test, it's really hard to check if the snapshot is correct. Don't snapshot tons of html to check only if the first option is selected.
Instead, you can query element with id `react-autowhatever-my-search-section-0-item-0` and assert that aria-selected is true

packages/components/src/Typeahead/Typeahead.test.js
```diff
@@ -177,6 +198,41 @@ describe('Typeahead', () => {
...
+				items: itemsObjectWithId,
+			};
+
+			// when
+			render(<Typeahead {...props} />);
+
+			// then
+			const titleList = screen.getAllByTitle('le title 1');
+			expect(titleList[1]).toHaveClass('theme-selected');
+			expect(titleList[1]).toHaveClass('theme-selected');
```
jsomsanith-tlnd, line 234:
```suggestion
			expect(titleList[0]).not.toHaveClass('theme-selected');
			expect(titleList[1]).toHaveClass('theme-selected');
```

jsomsanith-tlnd (approved):

packages/components/src/Typeahead/Typeahead.test.js
Gbacc: Thanks , didnt notice
Pushed a new commit about this , in fact selection by title should select both items like it's already to case

PR #4046, opened 2022-05-27
feat(design-system/TUX-1081): private Field component

Benrajalu:
Field is primitive wrapper component that handles the label, description, link and error logic for all text components. Previous `Field` did not use the tokens not Stack. 

It's only going to be used to build the public components. You won't use `Field` in your code. You will use `InputText` for instance that underneath is a composition of `Field` and `Input`. 

Use the tokens and stack.

### DIFF
--- a/packages/design-system/src/components/Form/Primitives/Checkbox/Checkbox.tsx
+++ b/packages/design-system/src/components/Form/Primitives/Checkbox/Checkbox.tsx
@@ -8,7 +8,7 @@ import Label from '../Label/Label';
 import styles from './Checkbox.module.scss';
 
 type CheckboxType = Omit<CheckboxProps, 'type' | 'prefix'> & {
-	label: string | ReactElement;
+	label: string;
 	id: string;
 	indeterminate?: boolean;
 };
--- a/packages/design-system/src/components/Form/Primitives/Field/Field.tsx
+++ b/packages/design-system/src/components/Form/Primitives/Field/Field.tsx
@@ -5,13 +5,18 @@ import Label, { LabelProps } from '../Label/Label';
 import { InlineMessageDestructive, InlineMessageInformation } from '../../../InlineMessage';
 import VisuallyHidden from '../../../VisuallyHidden';
 
-export type FieldStatusProps = {
-	isError?: boolean;
-};
+export type FieldStatusProps =
+	| {
+			hasError: true;
+			description: string;
+	  }
+	| {
+			hasError?: false;
+			description?: string;
+	  };
 
 type FieldProps = {
 	link?: LinkProps;
-	description?: string;
 	hideLabel?: boolean;
 	label: LabelProps;
 	id: string;
@@ -28,8 +33,8 @@ const Field = forwardRef(
 			id,
 			label,
 			name,
+			hasError = false,
 			hideLabel = false,
-			isError = false,
 			description,
 			...rest
 		} = props;
@@ -42,22 +47,22 @@ const Field = forwardRef(
 			<Label {...label} htmlFor={id} />
 		);
 
-		const Description = ({ descriptionText }: { descriptionText: string }) => {
-			const descProps = {
-				description: descriptionText,
-			};
-			if (isError) {
-				return <InlineMessageDestructive {...descProps} />;
+		const Description = () => {
+			if (hasError && description) {
+				return <InlineMessageDestructive description={description} />;
 			}
-			return <InlineMessageInformation {...descProps} />;
+			if (!!description) {
+				return <InlineMessageInformation description={description} />;
+			}
+			return null;
 		};
 
 		return (
 			<StackVertical gap="XXS" align={'stretch'} justify={'start'}>
 				{LabelComponent}
-				{cloneElement(children, { id, isError, name, rest }, ref)}
+				{cloneElement(children, { id, hasError, name, rest }, ref)}
 				{link && <Link {...link} />}
-				{description && <Description descriptionText={description} />}
+				{description && <Description />}
 			</StackVertical>
 		);
 	},
--- a/packages/design-system/src/components/Form/Primitives/Field/Field.tsx
+++ b/packages/design-system/src/components/Form/Primitives/Field/Field.tsx
@@ -2,19 +2,21 @@ import React, { forwardRef, ReactElement, Ref, cloneElement } from 'react';
 import Link, { LinkProps } from '../../../Link/Link';
 import { StackVertical } from '../../../Stack';
 import Label, { LabelProps } from '../Label/Label';
-import { VisuallyHidden } from '../../../../index';
+import { InlineMessageDestructive, InlineMessageInformation } from '../../../InlineMessage';
+import VisuallyHidden from '../../../VisuallyHidden';
+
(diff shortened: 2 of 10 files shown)


packages/design-system/src/components/Form/Primitives/Checkbox/Checkbox.tsx
```diff
@@ -8,7 +8,7 @@ import Label from '../Label/Label';
 import styles from './Checkbox.module.scss';
 
 type CheckboxType = Omit<CheckboxProps, 'type' | 'prefix'> & {
-	label: string | ReactElement;
+	label: string;
```
Benrajalu, line 11: Thanks to @sgendre for pointing out this mistake

tlnd-mhuchet (approved):

packages/design-system/src/components/Form/Primitives/Field/Field.tsx
```diff
@@ -0,0 +1,68 @@
+import React, { forwardRef, ReactElement, Ref, cloneElement } from 'react';
+import Link, { LinkProps } from '../../../Link/Link';
+import { StackVertical } from '../../../Stack';
+import Label, { LabelProps } from '../Label/Label';
+import { InlineMessageDestructive, InlineMessageInformation } from '../../../InlineMessage';
+import VisuallyHidden from '../../../VisuallyHidden';
+
+export type FieldStatusProps = {
+	isError?: boolean;
```
sgangnant, line 9: I'm not too keen on that wording, sounds wrong to me (not the first time I see it so I might be wrong)
```suggestion
	hasError?: boolean;
```

packages/design-system/src/components/Form/Primitives/_mixins.scss
```diff
@@ -82,9 +82,13 @@ $standardInputHeight: calc(#{tokens.$coral-sizing-m} - 0.2rem);
     box-shadow: 0 0 0 0.1rem tokens.$coral-color-accent-border;
   }
 
-  &.border_error {
+  &_borderError {
```
sgangnant, line 85: Feels like a BEM modifier, also I mostly saw hyphenated classes in Talend's projects 

```suggestion
  &--border-error {
```

Benrajalu: yes, but CSS modules and BEM are not great friends. I want to keep autocompletion of classNames in the CSS. Using the `styles["style--name"]` syntax robs me of that.

packages/design-system/src/components/Form/Primitives/Field/Field.tsx
Benrajalu: It's just a standard I'm used to follow for boolean. All booleans regardless of what they are start with `is`. I wish there was a Talend coding styleguide to settle on that.

jmfrancois: The best is linter for this. https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/boolean-prop-naming.md

packages/design-system/src/components/Form/Primitives/Field/Field.tsx
```diff
@@ -0,0 +1,68 @@
...
+			if (isError) {
+				return <InlineMessageDestructive {...descProps} />;
+			}
+			return <InlineMessageInformation {...descProps} />;
+		};
+
+		return (
+			<StackVertical gap="XXS" align={'stretch'} justify={'start'}>
+				{LabelComponent}
+				{cloneElement(children, { id, isError, name, rest }, ref)}
```
jsomsanith-tlnd, line 58: This doesn't seem compatible for inputs with label after the input such as checkboxes.
Do we have to support that here ?

packages/design-system/src/components/Form/Primitives/Field/Field.tsx
Benrajalu: That linter allows for both. If you guys are ok with both, fine by me!

sgendre:
> You won't use Field in your code. You will use InputText for instance that underneath is a composition of Field and Input.

I would have guess it will be named InputTextField or Field.Input.Text rather than only InputText

What do you think about that naming? I am not fan of implicit stuff :)

Benrajalu:
> > You won't use Field in your code. You will use InputText for instance that underneath is a composition of Field and Input.
> 
> I would have guess it will be named InputTextField or Field.Input.Text rather than only InputText
> 
> What do you think about that naming? I am not fan of implicit stuff :)

Sure. I can do `InputTextField`. The idea is not to provide you with an unopinionated `input type="text"` though, but only with an input that _requires_ a label, an id, a name etc... It's a way to enforce consistency and accessibility.

packages/design-system/src/components/Form/Primitives/Field/Field.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+		const Description = ({ descriptionText }: { descriptionText: string }) => {
+			const descProps = {
+				description: descriptionText,
+			};
+			if (isError) {
+				return <InlineMessageDestructive {...descProps} />;
+			}
+			return <InlineMessageInformation {...descProps} />;
+		};
```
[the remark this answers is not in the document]
Benrajalu: I replicated the current Field component's API but I did ask myself the same question. 
My reasoning in keeping it was that I don't like having two `string` props but no way to display both values.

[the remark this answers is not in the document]
jsomsanith-tlnd: I see it as 2 possible strings, and a switch `has/isError` to select the one we want.
I won't block the PR for that, it's a choice to make, and if you think the api is better like that, that's not a big deal 🤷‍♂️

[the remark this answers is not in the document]
jsomsanith-tlnd: Feel free to resolve the conversation if you decide to keep it as is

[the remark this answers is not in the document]
Benrajalu: I'm currently trying the implem to see how it feels. So far it feels like extra work for everybody involved :D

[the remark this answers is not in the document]
Benrajalu: I've opted for a single `description` props, sorry. But thanks for the suggestion though because it made me rework the typing. Now `description` is mandatory in `hasError` cases. I've updated the story to showcase that.

packages/design-system/src/components/Form/Primitives/Field/Field.tsx
```diff
@@ -0,0 +1,73 @@
...
+		);
+
+		const Description = () => {
+			if (hasError && description) {
+				return <InlineMessageDestructive description={description} />;
+			}
+			if (!!description) {
+				return <InlineMessageInformation description={description} />;
+			}
+			return null;
```
gjou-tlnd, line 57:
```suggestion
			if (description) {
			        if (hasError) {
				        return <InlineMessageDestructive description={description} />;
			        }
				return <InlineMessageInformation description={description} />;
			}
			return null;
```

PR #4063, opened 2022-06-01
feat(design-system): InlineMessage - Refactoring with CSS modules

Benrajalu:
InlineMessage is using StyledComponents and legacy tokens

Rewrite them with CSS modules a new component approach


tlnd-mhuchet (approved):

packages/design-system/src/components/InlineMessage/InlineMessage.stories.tsx
```diff
@@ -1,148 +1,105 @@
...
-		withBackground: true,
-	},
-	render,
-};
+export const WithRouterLink = () => (
+	<BrowserRouter>
+		<InlineMessage
+			withBackground
+			description="Inline message with a Router Link"
+			variant="information"
```
sgangnant, line 98: Wondering if this is really useful since you'll most likely want to change all the other props altogether rather than only the type (wording, etc). Also feels **a bit** like an anti-pattern (sort of) to what's mainly done in the DS (i.e exposing variations as components).
It won't prevent me from sleeping though.

packages/design-system/src/components/Form/FieldGroup/FieldGroup.tsx
```diff
@@ -41,24 +46,27 @@ const FieldGroup = React.forwardRef(
...
+				return <InlineMessageWarning {...descriptionProps} />;
 			}
 			if (hasSuccess) {
-				return <InlineMessage.Success {...descriptionProps} />;
+				return <InlineMessageSuccess {...descriptionProps} />;
 			}
 			if (hasInformation) {
-				return <InlineMessage.Information {...descriptionProps} />;
+				return <InlineMessageInformation {...descriptionProps} />;
 			}
```
sgangnant, line 68: `variant` could be used here 👼

packages/design-system/src/components/InlineMessage/InlineMessage.stories.mdx
```diff
@@ -1,22 +1,23 @@
 import { ArgsTable, Meta, Canvas, Story } from '@storybook/addon-docs';
 import { IconItem, IconGallery } from '@storybook/components';
 import { FigmaImage, Use } from '~docs';
-import * as Stories from '../InlineMessage.stories';
+import * as Stories from './InlineMessage.stories';
 
-import { Icon } from '../../Icon';
-import InlineMessage from '../';
+import { Icon } from '../Icon';
+import * as InlineMessage from './index';
```
sgangnant, line 7:
```suggestion
import { InlineMessage } from './index';
```

packages/design-system/src/components/InlineMessage/InlineMessage.stories.tsx
```diff
@@ -1,148 +1,105 @@
...
-};
-export const DestructiveSmall = {
-	args: {
-		...Destructive.args,
-		small: true,
-	},
-	render,
-};
+export const BackgroundDemo = () => (
+	<StackVertical gap={'XS'}>
```
sgangnant, line 47:
```suggestion
	<StackVertical gap="X">
```

packages/design-system/src/components/InlineMessage/InlineMessage.tsx
```diff
@@ -1,67 +1,58 @@
...
+		}
+
+		case 'beta': {
+			const { variant, ...rest } = props;
+			return <InlineMessageBeta {...rest} ref={ref} />;
+		}
+
+		default: {
+			return null;
+		}
```
sgangnant, line 52: This default case would most likely be the result of an error, maybe throw or make variant mandatory?

packages/design-system/src/components/InlineMessage/InlineMessage.tsx
```diff
@@ -1,67 +1,58 @@
...
+
+		case 'warning': {
+			const { variant, ...rest } = props;
+			return <InlineMessageWarning {...rest} ref={ref} />;
+		}
+
+		case 'beta': {
+			const { variant, ...rest } = props;
+			return <InlineMessageBeta {...rest} ref={ref} />;
+		}
```
sgangnant, line 48:
```suggestion
	const { variant, ...rest } = props;

	switch (props.variant) {
		case 'information':
			return <InlineMessageInformation {...rest} ref={ref} />;

		case 'destructive': 
			return <InlineMessageDestructive {...rest} ref={ref} />;

		case 'success': {
			const { variant, ...rest } = props;
			return <InlineMessageSuccess {...rest} ref={ref} />;
		}

		case 'warning': 
			return <InlineMessageWarning {...rest} ref={ref} />;

		case 'beta':
			return <InlineMessageBeta {...rest} ref={ref} />;
```

packages/design-system/src/components/InlineMessage/InlineMessage.tsx
sgangnant, line 52:
```suggestion
		default:
			return null;
```

packages/design-system/src/components/InlineMessage/Primitive/InlineMessagePrimitive.module.scss
```diff
@@ -0,0 +1,55 @@
...
+      display: inline;
+      margin-right: tokens.$coral-spacing-xxs;
+
+      &:last-child {
+        margin-right: 0;
+      }
+    }
+  }
+
+  &_withBackground {
```
sgangnant, line 42:
```suggestion
  &--withBackground {
```

packages/design-system/src/components/InlineMessage/InlineMessage.stories.tsx
sgangnant: (erratum, this pattern is also used in Skeleton)

Benrajalu: Yeah. About this. 
I personally don't like this pattern. I think it's a maintenance trap. Standalone components are easier to make evolve without breaking things. But we asked around and people told us they wanted the collection ones with props. 
We provided.

packages/design-system/src/components/InlineMessage/InlineMessage.tsx
Benrajalu: Variant is mandatory. In TS you cannot reach this default case. But in JS you could and linters won't let me explain :D

I could throw though, I'll consider it!

packages/design-system/src/components/InlineMessage/InlineMessage.tsx
Benrajalu: Good instincts but this doesn't work with TS. 
If you destructure outside of the `case`, TS cannot infer the type of `...rest` anymore.

.changeset/yellow-donuts-pump.md
```diff
@@ -0,0 +1,7 @@
+---
+'@talend/react-components': major
+'@talend/design-system': major
+'@talend/react-forms': patch
+---
+
+InlineMessage component is now only coming from design system. No more classnames. No more Link as components.
```
gjou-tlnd, line 7: maybe a line about the breaking API change in DS?

packages/design-system/src/components/InlineMessage/InlineMessage.stories.tsx
gjou-tlnd: > people told us they wanted

you can use my name

packages/design-system/src/components/InlineMessage/Primitive/InlineMessagePrimitive.module.scss
```diff
@@ -0,0 +1,55 @@
...
+    > * {
+      display: inline;
+      margin-right: tokens.$coral-spacing-xxs;
+
+      &:last-child {
+        margin-right: 0;
+      }
+    }
+  }
+
```
gjou-tlnd, line 41: do we really need these content styling overrides?

packages/design-system/src/components/InlineMessage/InlineMessage.stories.tsx
Benrajalu: Not just you.

packages/design-system/src/components/Form/FieldGroup/FieldGroup.tsx
Benrajalu: Kinda not though, we'd have to map the various `hasState` to `variant` props. That's sort of the same code down the line.

sgangnant (approved):

PR #4121, opened 2022-06-15
feat(design-system): allow to control popover

romainseb:
We can have button into a popover that close the popover at the end of the day (like setting up stuff inside the popover)
To allow the popover content to close the popover, we would like to access to the popover state that is contained into the component.

Here the integration is allowing to pass children as before if we don't have to handle this kind of use case and we can also provide children as function if we want to access the popover in the content

### DIFF
--- a/packages/design-system/src/components/Popover/Popover.stories.tsx
+++ b/packages/design-system/src/components/Popover/Popover.stories.tsx
@@ -3,10 +3,25 @@ import React from 'react';
 import Popover from './Popover';
 import { ButtonPrimary } from '../Button';
 import { action } from '@storybook/addon-actions';
+import { PopoverStateReturn } from 'reakit/ts';
+import { StackVertical } from '../Stack';
 
 export default {
 	component: Popover,
 };
+
+function EasyPopover() {
+	return <StackVertical gap="S">Hello hello</StackVertical>;
+}
+
+function ControlledPopover({ popover }: { popover?: PopoverStateReturn }) {
+	return (
+		<StackVertical gap="S">
+			There is some content
+			<ButtonPrimary onClick={() => popover?.hide()}>Close Me please</ButtonPrimary>
+		</StackVertical>
+	);
+}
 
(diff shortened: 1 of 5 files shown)


packages/design-system/src/components/Popover/Popover.stories.tsx
```diff
@@ -3,11 +3,26 @@ import React from 'react';
...
 
 export default {
 	component: Popover,
 };
 
+function EasyPopover() {
+	return <StackVertical gap="S">Hello hello</StackVertical>;
+}
+
+function ControlledPopover({ popover }: { popover?: PopoverStateReturn }) {
```
Benrajalu, line 17:
```suggestion
function ChildrenAsFunctionPopoverContent({ popover }: { popover?: PopoverStateReturn }) {
```

Benrajalu: The Popover is still uncontrolled, when we think about it. Its state is still entirely self-contained.

packages/design-system/src/components/Popover/Popover.tsx
```diff
@@ -13,24 +15,36 @@ import style from './Popover.module.scss';
...
 	return (
 		<>
 			<ReakitPopoverDisclosure {...popover}>
-				{disclosureProps => React.cloneElement(disclosure, disclosureProps)}
+				{disclosureProps => cloneElement(disclosure, disclosureProps)}
 			</ReakitPopoverDisclosure>
-			<ReakitPopover {...popover} {...props}>
+			<ReakitPopover
+				{...popover}
+				style={{ zIndex: tokens.coralElevationLayerStandardFront, ...props.style }}
```
Benrajalu, line 37: `style` is not going to remain as an acceptable prop. `zIndex` would be a standalone optional props.

Benrajalu (approved):

jsomsanith-tlnd (approved):

PR #4141, opened 2022-06-21
feat(components): new icons in old buttons

romainseb:
Allow to use the new icons in ord buttons (mainly for existing migration and sidepanel)

Allow to pass a props to mount a SizedIcon instead of an Icon

### DIFF
--- a/.changeset/old-socks-obey.md
+++ b/.changeset/old-socks-obey.md
@@ -1,0 +1,5 @@
+---
+'@talend/react-components': minor
+---
+
+feat(components): allow to use new icons in legacy buttons
--- a/packages/components/src/ActionList/ActionList.scss
+++ b/packages/components/src/ActionList/ActionList.scss
@@ -1,10 +1,12 @@
+@use '~@talend/design-tokens/lib/_tokens.scss' as tokens;
+
 ////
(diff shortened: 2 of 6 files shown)


jsomsanith-tlnd (approved):

jmfrancois (changes requested):
icons are not the same size.
16x16 against 2020

Benrajalu:
> Every sidepanel will break, i would like to avoid that, in my mind, the best way to handle the migration would be to have the sidepanl in the DS using only the new icons

Babysteps though. It'll break either way. I say let it break now for icons. If not we'll have some sidepanels with a salad of icons. There's also the possibility of contributing a Sidepanel component to the DS now, if you're game!

> Same as above, should be a major breaking change, i would not intend to bring that here to just use new icons in a sidepanel

OK, let's not change the button then. That's fine for now.

> Don't know how we can use a single prop, maybe if it's prefixed by `talend-` go for the old one but it fell a bit unstable to be based only on that
We can if we remove the possibility to use `talend-` stuff :)

Benrajalu:
A compromise would be forcing all icons in the sidebar to follow `tokens.$coral-sizing-xxxs`

PR #4161, opened 2022-06-24
chore(design-system/Dropdown): set a baseId

Nigui:

When using snapshots to test composants DOM with DS Dropdown (reakit Menu), `id` and `aria-controls` are not stable (generated during build). This fix should make it stable via a mock. 

Manually call id generator to be able to mock it.

- [X] The PR have used `yarn changeset` to a request a release from the CI if wanted.

### DIFF
--- a/packages/design-system/src/components/Dropdown/Dropdown.tsx
+++ b/packages/design-system/src/components/Dropdown/Dropdown.tsx
@@ -1,5 +1,5 @@
 import React, { cloneElement, forwardRef, MouseEvent, ReactElement, Ref } from 'react';
-import { Menu, MenuButton, useMenuState } from 'reakit';
+import { Menu, MenuButton, useMenuState, unstable_useId as useId } from 'reakit';
 // eslint-disable-next-line @talend/import-depth
 import { IconName } from '@talend/icons/dist/typeUtils';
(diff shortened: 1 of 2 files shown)


packages/design-system/src/components/Dropdown/Dropdown.tsx
```diff
@@ -44,6 +43,7 @@ const Dropdown = forwardRef(
 			animated: 250,
 			gutter: 4,
 			loop: true,
+			baseId: useId().id,
```
Benrajalu, line 46: Wouldn't that be different each time as well? If you need to set a definite id for your snapshots, wouldn't we need this?  

```suggestion
			baseId: useId().id || props.id,
```

Nigui: yes it would, but we already mock `useId` in tests to return a fixed number. 
And I wanted to prevent exposition of new properties if not necessary.

jmfrancois (commented):
I tend to think snapshot is not testing the DOM. I prefer the way to use RTL and came to check if few DOM elements are present.
Also ds is mocked (even if it should not...) so you should not be impacted by that.

PR #4170, opened 2022-06-28
feat(design-system/icons): Typing for icons in buttons

Benrajalu:
Components that used to rely on `Icon` were based in `IconName`, a type that lists several legacy icons but also takes `string`s without any actual check. 

The new component `Sizedicon` does not support this. It asks for a size and a name, and only accepts the name of icons available at the given size. 

Removing all uses of `Icon` would be the right way to go for the design-system, but it would also be a major breaking change for our consumers. 

Enable a mixture of `IconName` and the correct `IconNameWithSize` subset for the major `icon` props we could find (critical buttons etc...). This is a large compromise. Since `IconName` contains `string` in its union, TS will accept any string, which breaks the promise of a strongly-typed `SizedIcon` component. 

This PR also removes `<Icon />` components call where it's non-breaking.

### DIFF
--- a/packages/design-system/src/components/Button/Button.spec.tsx
+++ b/packages/design-system/src/components/Button/Button.spec.tsx
@@ -9,11 +9,25 @@ const { Loading } = composeStories(Stories);
 context('<Button />', () => {
 	describe('default', () => {
 		it('should be focusable', () => {
-			cy.mount(<ButtonPrimitive onClick={() => {}}>button</ButtonPrimitive>);
+			cy.mount(
+				<ButtonPrimitive size="M" onClick={() => {}}>
+					button
+				</ButtonPrimitive>,
+			);
 			cy.get('button').focus();
 		});
 		it('should accept data-feature', () => {
-			cy.mount(<ButtonPrimitive data-testid="my.button" data-feature="my.feature" onClick={() => {}}>button</ButtonPrimitive>);
+			cy.mount(
+				<ButtonPrimitive
+					size="M"
+					data-testid="my.button"
+					data-feature="my.feature"
+					onClick={() => {}}
+				>
+					button
+				</ButtonPrimitive>,
+			);
+			// eslint-disable-next-line testing-library/prefer-screen-queries
 			cy.getByTestId('my.button').should('have.attr', 'data-feature', 'my.feature');
 		});
 	});
@@ -21,6 +35,7 @@ context('<Button />', () => {
 	describe('loading state', () => {
 		it('should load', () => {
 			cy.mount(<Loading data-testid="my.button" />);
+			// eslint-disable-next-line testing-library/prefer-screen-queries
 			cy.getByTestId('my.button')
 				.should('have.attr', 'aria-busy', 'false')
 				.click()
@@ -30,6 +45,7 @@ context('<Button />', () => {
 
 		it('should have a tooltip', () => {
 			cy.mount(<Loading data-testid="my.button" />);
+			// eslint-disable-next-line testing-library/prefer-screen-queries
 			cy.getByTestId('my.button')
 				.focus()
 				.should('have.attr', 'aria-describedby')
--- a/packages/design-system/src/components/Button/Primitive/ButtonPrimitive.tsx
+++ b/packages/design-system/src/components/Button/Primitive/ButtonPrimitive.tsx
@@ -7,7 +7,7 @@ import Clickable, { ClickableProps } from '../../Clickable';
 import { DataAttributes, DeprecatedIconNames } from '../../../types';
 import { StackHorizontal } from '../../Stack';
 import Loading from '../../Loading';
-import { parseDeprecatedIcon } from '../../Icon/DeprecatedIconHelper';
+import { getIconWithDeprecatedSupport } from '../../Icon/DeprecatedIconHelper';
 
 import styles from './ButtonStyles.module.scss';
 import { SizedIcon } from '../../Icon';
@@ -63,7 +63,7 @@ function ButtonPrimitiveInner<S extends AvailableSizes>(
 				)}
 				{!isLoading && icon && (
 					<span className={styles.button__icon}>
-						{parseDeprecatedIcon({ iconSrc: icon, size: size === 'S' ? 'S' : 'M' })}
+						{getIconWithDeprecatedSupport({ iconSrc: icon, size: size === 'S' ? 'S' : 'M' })}
 					</span>
 				)}
 				{children}
--- a/packages/design-system/src/components/Button/Primitive/ButtonPrimitive.tsx
+++ b/packages/design-system/src/components/Button/Primitive/ButtonPrimitive.tsx
@@ -1,4 +1,4 @@
-import React, { forwardRef, Ref, ReactElement } from 'react';
+import React, { forwardRef, ReactElement, Ref } from 'react';
 // eslint-disable-next-line @talend/import-depth
 import { IconNameWithSize } from '@talend/icons/dist/typeUtils';
 import classnames from 'classnames';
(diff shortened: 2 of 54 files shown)


packages/design-system/src/components/Button/Primitive/ButtonPrimitive.tsx
```diff
@@ -1,78 +1,84 @@
...
-export type SharedButtonTypes = {
-	size?: AvailableSizes;
-	icon?: IconName | React.ReactElement;
+export type SharedButtonTypes<S extends AvailableSizes> = {
 	isLoading?: boolean;
 	isDropdown?: boolean;
+	size?: S;
+	icon?: S extends 'S'
+		? IconNameWithSize<'S'>
+		: DeprecatedIconNames | ReactElement | IconNameWithSize<'M'>;
```
Benrajalu, line 27: For legacy reasons, we still have to support the rest. This translate to `ReactElement | string` since `DeprecatedIconNames` is a union of string with `string`. 

Historically this was done so icons could be fetched through passing the name `remote-balbla` or even images by passing `src-blabla`.

I would love to deprecate this but it would be a large breaking change.

packages/design-system/src/components/Button/Primitive/ButtonPrimitive.tsx
```diff
@@ -1,78 +1,84 @@
...
+					</span>
+				)}
+			</StackHorizontal>
+		</Clickable>
+	);
+}
+
+const ButtonPrimitive = forwardRef(ButtonPrimitiveInner) as <S extends AvailableSizes>(
+	props: BaseButtonProps<S> & { ref?: Ref<HTMLButtonElement> },
+) => ReturnType<typeof ButtonPrimitiveInner>;
```
Benrajalu, line 82: TS needs a little help at this stage. `forwardRef` makes it lose the `S` generic without the typecasting.

packages/design-system/src/components/Button/Primitive/ButtonStyles.module.scss
```diff
@@ -36,6 +36,13 @@
 	&.size-S {
 		height: tokens.$coral-sizing-s;
 		padding: tokens.$coral-spacing-xxs tokens.$coral-spacing-xs;
+
+		.button__icon {
+			display: inline-flex;
+			width: tokens.$coral-sizing-minimal;
+			height: tokens.$coral-sizing-minimal;
+			align-items: center;
+		}
```
Benrajalu, line 45: Superfluous when the old icons are removed

packages/design-system/src/components/Button/Button.spec.tsx
```diff
@@ -9,18 +9,33 @@ const { Loading } = composeStories(Stories);
...
+				<ButtonPrimitive
+					size="M"
+					data-testid="my.button"
+					data-feature="my.feature"
+					onClick={() => {}}
+				>
+					button
+				</ButtonPrimitive>,
+			);
+			// eslint-disable-next-line testing-library/prefer-screen-queries
```
Benrajalu, line 30: Not the point of this PR. Also I don't think this package is configured for RTL

packages/design-system/src/components/Icon/SizedIcon.tsx
```diff
@@ -18,6 +18,11 @@ const SizedIcon = React.forwardRef(
 	) => {
 		const numericSize = getNumericSize(size);
 		const fullName = size ? `${name}:${size}` : name;
+		// eslint-disable-next-line @typescript-eslint/ban-ts-comment
```
Benrajalu, line 21: Helps debugging. Since on some of the TS implementations, a `string` can reach this component, this log will help point out why an icon could not be displayed.

packages/design-system/src/components/Icon/DeprecatedIconHelper.tsx
```diff
@@ -0,0 +1,21 @@
+import React, { ReactElement } from 'react';
+import { Icon } from './Icon';
+import { SizedIcon } from './SizedIcon';
+import { IconNameWithSize } from '@talend/icons';
+
+export function parseDeprecatedIcon({
```
Benrajalu, line 6: A helper function that helps decide whether something should be a new icon or a deprecated one, at runtime. 

When (if 😭 ) the old icons finally die their deserved death, this function could be easily tracked down and replaced with `SizedIcon` imports.

packages/design-system/src/components/Button/variations/ButtonSecondary.tsx
```diff
@@ -1,14 +1,25 @@
...
-const ButtonSecondary = forwardRef(
-	(props: ButtonSecondaryPropsType, ref: Ref<HTMLButtonElement>) => {
-		return <ButtonPrimitive {...props} ref={ref} className={styles.secondary} />;
-	},
-);
+function Secondary<S extends AvailableSizes>(
+	props: ButtonSecondaryPropsType<S>,
+	ref: Ref<HTMLButtonElement>,
+) {
+	return <ButtonPrimitive {...props} ref={ref} className={styles.secondary} />;
```
romainseb, line 18: Don't we need to 
`const { size = 'M', ...rest } = props;` on this button too ?

packages/design-system/src/components/Button/variations/ButtonTertiary.tsx
```diff
@@ -1,12 +1,25 @@
...
+	'className' | 'size'
+> & {
+	size?: S;
+};
 
-const ButtonTertiary = forwardRef((props: ButtonTertiaryPropsType, ref: Ref<HTMLButtonElement>) => {
+function Tertiary<S extends AvailableSizes>(
+	props: ButtonTertiaryPropsType<S>,
+	ref: Ref<HTMLButtonElement>,
+) {
```
romainseb, line 17: Don't we need to 
`const { size = 'M', ...rest } = props;` on this button too ?

packages/design-system/src/components/Icon/DeprecatedIconHelper.tsx
romainseb: Just in case, the name of this function looks a bit weird for me.
I don't understand that we get an icon returned.
Wdyt about something like `getDeprecatedIcon` (not 100% convinced anyway but still)

packages/design-system/src/components/Icon/DeprecatedIconHelper.tsx
```diff
@@ -0,0 +1,21 @@
...
+	iconSrc,
```
romainseb, line 7: Why not having just name as prop ?

packages/design-system/src/components/Button/variations/ButtonTertiary.tsx
Benrajalu: TS didn't tell me to 🤔 I might remove the destruct on the other ones if they are overkill. Thanks for spotting it.

packages/design-system/src/components/Link/Link.spec.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable testing-library/prefer-screen-queries */
```
Benrajalu, line 1: Again, not why we're here. Cypress is the DS test runner.

packages/design-system/src/components/Button/variations/ButtonTertiary.tsx
Benrajalu: Yep, not needed, thenks

packages/design-system/src/components/Icon/DeprecatedIconHelper.tsx
Benrajalu: `getIconWithDeprecatedSupport` is my suggestion 🤷

jmfrancois:
I found one aligment issue:

image

jmfrancois (commented):
I found one aligment issue:

image

Benrajalu:
> I found one aligment issue:
> 
> image

Thanks! Let me fix thiiis

romainseb (approved):

tlnd-mhuchet (approved):

PR #4178, opened 2022-06-29
fix(component): don't pass rowdata to cell title

romainseb:
RowData can contain children prop 🙄 

### DIFF
--- a/.changeset/pink-gorillas-hug.md
+++ b/.changeset/pink-gorillas-hug.md
@@ -1,0 +1,5 @@
+---
+'@talend/react-components': patch
+---
+
+fix(component): don't spread row data to cellTitle action
--- a/packages/components/src/VirtualizedList/CellTitle/CellTitleSelector.component.js
+++ b/packages/components/src/VirtualizedList/CellTitle/CellTitleSelector.component.js
@@ -40,7 +40,6 @@ function CellTitleSelector(props) {
 		return (
 			<Action
 				{...columnData}
-				{...rowData}
 				id={id && `${id}-btn`}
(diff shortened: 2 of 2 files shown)


Nigui (approved):

.changeset/pink-gorillas-hug.md
```diff
@@ -0,0 +1,5 @@
+---
+'@talend/react-components': patch
```
Benrajalu, line 2: This is an API break though, with possible breaking side-effects. You guys are sure?

Benrajalu: Sure, I can believe it's not necessary. But if it has been done somewhere, then this PR becomes breaking, not patch.

romainseb: I'm confident that this property is not used.
Everything already need is passed aside, spreading all the row data to the "title" button has no usage in our app afaik.

Thanks for raising this point.

jsomsanith-tlnd: Yes, this was an error at first.
The action represents the title of the tile that is restricted. This idea is not to open it to customisation.
And here we spread all the list's item properties as props, which doesn't make sense. In practice we end up with tons of unexpected props in Action component input, that won't be used.

jsomsanith-tlnd (approved):

.changeset/pink-gorillas-hug.md
Nigui: I found a side-effect : it removes the `name` attribute of CellTitle ActionButton button tag leading to break selectors.

PR #4201, opened 2022-07-08
fix(design-system): Allow additional bundles for IconsProvider

tlnd-mhuchet:
In Pipeline Designer (at least), we load additional icons bundles. We should not declare default bundle anymore, but if we only keep additional ones, it overrides the default config and we lost default icons.

Add a new `additionalBundles` props to the `IconsProvider` component and concat it with the `bundle` props.

### DIFF
--- a/.changeset/unlucky-trees-look.md
+++ b/.changeset/unlucky-trees-look.md
@@ -1,0 +1,5 @@
+---
+'@talend/design-system': patch
+---
+
+fix(design-system): Allow additional bundles for IconsProvider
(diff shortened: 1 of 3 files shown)


jmfrancois (approved):

romainseb (approved):

Benrajalu (approved):
As we discussed: if we need to be able to handle "unknown" icons (in the sense that the product design team has no ownership of them), down the line we will still need to find a way to ensure those icons fit some of their design guidelines (sizes at least). 

For now, the single `Icon` component will keep doing the job, but a dedicated component with a better ruleset should be on our minds!

PR #4205, opened 2022-07-11
fix(APIC-827): reset input copy state when input value change

cbaune-talend:
When the inputValue of the inputCopy change, the state of the hook `useCopyToClipboard` is not reset.
Preventing errors and copy message to be hidden.

Since we cannot reset the state of the hook manualy, the solution is to transfer the state of the hook to a custom local state.
Se we can reset this state when the value of the input change.

### DIFF
--- a/.changeset/modern-ladybugs-sell.md
+++ b/.changeset/modern-ladybugs-sell.md
@@ -1,0 +1,5 @@
+---
+'@talend/design-system': minor
+---
+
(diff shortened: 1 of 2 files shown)


Benrajalu (approved):

Benrajalu:
Adding a test would be better though

jmfrancois (approved):

PR #4250, opened 2022-07-26
fix(Datalist/TDOPS-2632): update data list entry if name is not the same

VolodymyrKovalM:
The entry in Data list is updated only if value has changed
When the title map is asynchronously loaded, for example, depending on value of another Data list, 
the name of entry can be the same as value (when the title map has changed, but the value is still old one, and there is not entry with this value in new title map, by default, for some time, the name will be the same as value)
So, this condition won't be true https://github.com/Talend/ui/blob/master/packages/components/src/Datalist/Datalist.component.js#L150 , as the value won't equal to filter value, and so, it won't update filter value, and previous filter value (from previous title map) will be displayed, although there will be a new value selected from new title map (the first entry)

https://user-images.githubusercontent.com/20522845/181039400-b735536c-e476-4028-84fb-d09e6d390079.mp4

Update data list entry is name is not the same

### DIFF
--- a/.changeset/thirty-llamas-act.md
+++ b/.changeset/thirty-llamas-act.md
@@ -1,0 +1,5 @@
+---
+'@talend/react-components': patch
+---
+
+fix(Datalist/TDOPS-2632): update data list entry if name is not the same
--- a/packages/components/src/Datalist/Datalist.component.js
+++ b/packages/components/src/Datalist/Datalist.component.js
@@ -143,9 +143,10 @@ function Datalist(props) {
 			return;
 		}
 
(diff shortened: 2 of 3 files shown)


jmfrancois (changes requested):
you need a changeset in the PR

gjou-tlnd (changes requested):
we had lots of issue on this component, please add a test case

jmfrancois (approved):
Agree with @gjou-tlnd please add a test case.

gjou-tlnd (approved):

PR #4293, opened 2022-08-29
feat(CDM-237) allow to customize subtitle tag

jmainguytalend:
Drawer's subtitle tag is not customizable (style)

Allow to specify which component tag from DS we can use

### DIFF
--- a/.changeset/twelve-kangaroos-carry.md
+++ b/.changeset/twelve-kangaroos-carry.md
@@ -1,0 +1,17 @@
+---
+'@talend/react-components': major
+---
+
+Allow to customize drawer's subtitle tag
+
+**Breaking change :**
+Props `subtitleTagLabel` and `subtitleTagTooltip` are replaced by a props `subtitleTag`.
+
+Props `subtitleTag` has following shape :
+```
+{
+    label: PropTypes.string,
+    tooltip: PropTypes.string,
+    component: PropTypes.element,
+}
+```
--- a/.changeset/twelve-kangaroos-carry.md
+++ b/.changeset/twelve-kangaroos-carry.md
@@ -12,6 +12,6 @@ Props `subtitleTag` has following shape :
 {
     label: PropTypes.string,
     tooltip: PropTypes.string,
-    component: PropTypes.element,
+    variant: PropTypes.oneOf(SubtitleTagVariants),
 }
 ```
--- a/packages/components/src/Drawer/Drawer.component.js
+++ b/packages/components/src/Drawer/Drawer.component.js
@@ -5,7 +5,7 @@ import omit from 'lodash/omit';
 import noop from 'lodash/noop';
 import { Transition } from 'react-transition-group';
 import classnames from 'classnames';
-import { StackHorizontal, Tag, TagDefault, Tooltip } from '@talend/design-system';
+import { StackHorizontal, Tag, Tooltip } from '@talend/design-system';
 import ActionBar from '../ActionBar';
 import Action from '../Actions/Action';
 import TabBar from '../TabBar';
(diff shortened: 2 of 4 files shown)


.changeset/twelve-kangaroos-carry.md
```diff
@@ -0,0 +1,17 @@
...
+
+**Breaking change :**
+Props `subtitleTagLabel` and `subtitleTagTooltip` are replaced by a props `subtitleTag`.
+
+Props `subtitleTag` has following shape :
+```
+{
+    label: PropTypes.string,
+    tooltip: PropTypes.string,
+    component: PropTypes.element,
```
Benrajalu, line 15:
```suggestion
    tagType: keyof typeof TagVariant,
```

We don't actually want *any* type of component here. We just want to be able to pick a type of tag. The API should reflect that.

packages/components/src/Drawer/Drawer.component.js
```diff
@@ -102,37 +102,47 @@ export function cancelActionComponent(onCancelAction, getComponent) {
...
+function renderSubtitleTag(
+	subtitleTagLabel,
+	subtitleTagTooltip,
+	subtitleTagComponent = TagDefault,
+) {
 	if (subtitleTagTooltip) {
 		return (
 			<Tooltip placement="top" title={subtitleTagTooltip}>
-				<TagDefault>{subtitleTagLabel}</TagDefault>
+				{React.createElement(subtitleTagComponent, null, subtitleTagLabel)}
```
Benrajalu, line 113:
```suggestion
				<Tag variant={subtitleTagVariant}>{subtitleTagLabel}</Tag>
```

.changeset/twelve-kangaroos-carry.md
Benrajalu: This could also be optional and defaulted to... 'default'

Benrajalu (changes requested):
We agree on the need, but API does not best describe the intention. The Tag component provides a prop that does just so.

.changeset/twelve-kangaroos-carry.md
jmainguytalend: Two problems with that :

1. We are in Javascript here, not typescript
2. TagVariant is not exported by the module (see ui/packages/design-system/src/components/Tag/index.ts)

:/

packages/components/src/Drawer/Drawer.component.js
jmainguytalend: TagVariant is not exported by Tag module. Could I change it to add it ?

Benrajalu: Yes absolutely. You can even switch it from an enum to a union of strings. Enums un TS have little value.

.changeset/twelve-kangaroos-carry.md
jmainguytalend: I'm talking about the component that is in JS. So i'm not able to specify `keyof typeof TagVariant`. Same thing about typing `subtitleTagVariant` with value-limited string list

Benrajalu: Oh yeah, my bad. Drawer is also a DS component so I assumed you were working on the newer one. 

You're free to export a `export const TagTypesArray = ['default', 'success'...]` if that helps. You could then do `PropTypes.oneOf(tagsTypesArray)`

jmainguytalend: OK thank you

packages/components/src/Drawer/Drawer.component.js
jmainguytalend: Updated, if you can review again please

packages/components/src/Drawer/Drawer.component.js
```diff
@@ -25,6 +25,15 @@ const STYLES = {
 	exited: { transform: 'translateX(100%)' },
 };
 
+export const SubtitleTagVariants = [
```
Benrajalu, line 28: It's up to you but if the DS changes its Tag variants someday, you would create a breaking change here. That's why creating this in the DS and tying it to the types would be safer.

PR #4301, opened 2022-09-01
feat(CDM-237) Add const to export Tag variations through string array

jmainguytalend:
To allow JS component to add customizable Tag as part of it, it requires to be able to define available tag variations in component prop type definition

Export available tag variations as string array

### DIFF
--- a/packages/design-system/src/components/Tag/Tag.tsx
+++ b/packages/design-system/src/components/Tag/Tag.tsx
@@ -19,6 +19,9 @@ export enum TagVariant {
 	beta = 'beta',
 }
 
+// This const allows JS component to type props with restricted list of tag variations
+export const TagVariants = Object.values(TagVariant);
+
 type TagProps = Omit<PrimitiveTagProps, 'className'> & {
 	/**
 	 * Tag variation depending on its semantic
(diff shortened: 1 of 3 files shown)


packages/design-system/src/components/Tag/Tag.tsx
```diff
@@ -19,6 +19,9 @@ export enum TagVariant {
 	beta = 'beta',
 }
 
+// This const allows JS component to type props with restricted list of tag variations
+export const TagVariants = Object.values(TagVariant);
```
Benrajalu, line 23:
```suggestion
export const TagVariantsNames = Object.values(TagVariant);
```

packages/design-system/src/components/Tag/index.ts
```diff
@@ -1,4 +1,4 @@
-import Tag from './Tag';
+import Tag, { TagVariants } from './Tag';
```
Benrajalu, line 1:
```suggestion
import Tag, { TagVariantsNames } from './Tag';
```

packages/design-system/src/components/Tag/index.ts
```diff
@@ -9,4 +9,13 @@ import {
...
-export { Tag, TagDefault, TagInformation, TagSuccess, TagWarning, TagDestructive, TagBeta };
+export {
+	Tag,
+	TagDefault,
+	TagInformation,
+	TagSuccess,
+	TagWarning,
+	TagDestructive,
+	TagBeta,
+	TagVariants,
```
Benrajalu, line 20:
```suggestion
	TagVariantsNames,
```

Benrajalu (approved):
Not actually exposing the variants, just their names. Otherwise LGTM!

PR #4358, opened 2022-09-27
fix(ARCH-631): use css module filename

jmfrancois:

our styles use css module but the filename does not reflect it.
This block us to use create-react-app kind of tools

* create a script to parse js, ts, tsx and rename found scss files.
* apply script on the repository

The script is available on gist: https://gist.github.com/jmfrancois/402c32c22fba98f1e35599f1e0dab2c2


toscssModule.js
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env node
```
Benrajalu, line 1: Are you sure you want to commit this?

Benrajalu (approved):
`tocssModule.js` seems like it shouldn't be pushed but otherwise OK

toscssModule.js
jmfrancois: not sure may be just put the script on a gist would be enough

sgendre:
what do you think about embedding in the PR the script that did the work to share knowledge?

Kporal (approved):

tlnd-mhuchet (approved):

romainseb (approved):

PR #4360, opened 2022-09-28
fix(ARCH-631): explicit use sass-data

jmfrancois:

our current tools add sass-data so scss file rely on defined variables, but developer write variables without include them.
This block us to use tools like create-react-app

* create script to do that: https://gist.github.com/jmfrancois/402c32c22fba98f1e35599f1e0dab2c2
* explicitly add use of the sass data

to test it, I have applied it in /lib folder of create-react-app and did the job to fix all build errors.

image

### DIFF
--- a/.changeset/silly-fireants-run.md
+++ b/.changeset/silly-fireants-run.md
@@ -1,5 +1,9 @@
 ---
 '@talend/react-components': patch
+'@talend/react-containers': patch
+'@talend/react-dataviz': patch
+'@talend/react-faceted-search': patch
+'@talend/react-forms': patch
 ---
 
 fix: explicit import of sass-data
--- a/.changeset/silly-fireants-run.md
+++ b/.changeset/silly-fireants-run.md
@@ -1,0 +1,5 @@
+---
+'@talend/react-components': patch
+---
+
+fix: explicit import of sass-data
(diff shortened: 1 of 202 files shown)


Benrajalu (commented):
This might be a pain but hear me out while you're scripting stuff: 
 
 `@use '~@talend/bootstrap-theme/src/theme/guidelines' as bootstrap;` instead.
 
And prefix all usages then with `bootstrap.` (that's clearly the painful part)
 
When you move to remove bootstrap, you'll be able to figure out exactly what remains to be cleaned up. And devs having to write `bootstrap.` should make it obvious there's something wrong :D

jmfrancois:
I am agree that could be nice to track down usage. But yeah not painful but have to spend time for that. Let's see that as a bonus, it could be done later.

Gbacc:
The idea is to have this import on all scss files in TUI ? 
So it could happen we add a new file and forget about it. How could we prevent that except during reviews ?

smouillour (approved):

Benrajalu (approved):

tlnd-mhuchet (approved):

tlnd-mhuchet:
> The idea is to have this import on all scss files in TUI ? So it could happen we add a new file and forget about it. How could we prevent that except during reviews ?

Adding a custom (or if one exist already) stylelint rule ? But linter are not really taken seriously in projects 😅

PR #4369, opened 2022-10-03
feat(DS): add width to StackHorizontal component

romainseb:
For some layout, i find the need to wrap this component with a container that have a width of 100%

For now, i'm not sure it's needed to have all the sizes for the component or if just 100% is enough 🤔 

### DIFF
--- a/packages/design-system/src/components/Stack/Primitive/StackPrimitive.tsx
+++ b/packages/design-system/src/components/Stack/Primitive/StackPrimitive.tsx
@@ -28,7 +28,7 @@ export const alignContentOptions = {
 	baseline: 'align-content-baseline',
 };
 
-export const sizeOptions = {
+export const spacingOptions = {
 	0: 'NONE',
 	XXS: 'XXS',
 	XS: 'XS',
@@ -38,7 +38,7 @@ export const sizeOptions = {
 	XL: 'XL',
 };
 
-export const heightOptions = {
+export const sizeOptions = {
 	'100%': '100',
 	XXXS: 'XXXS',
 	XXS: 'XXS',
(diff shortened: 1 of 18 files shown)


Benrajalu (approved):

sgangnant (approved):

tlnd-mhuchet (approved):

jmfrancois (commented):
it feels strange to me. stack is flex, and flex size is driven by content.
I have the feeling you are trying to get a grid component using flex.

packages/design-system/src/components/Stack/Primitive/StackPrimitive.tsx
```diff
@@ -38,7 +38,7 @@ export const sizeOptions = {
 	XL: 'XL',
 };
 
-export const heightOptions = {
+export const sizeOptions = {
```
Nigui, line 41: isn't a breaking change ?

Benrajalu:
> it feels strange to me. stack is flex, and flex size is driven by content. I have the feeling you are trying to get a grid component using flex.

Not a bad point, however from this library's standpoint, Stack is Stack. It happens to use flex as a box-model, but its main purpose is to force us into using spacing tokens for gaps in vertical or horizontal simple layouts. 

I did not provide height and width props originally because I also felt like it was the context's job. But we got several report of that decision actually discouraging the use of the Stacks. So I'd rather let it happen and have people use Stack( and thus the spacing tokens "for free" and optimizing their CSS with a lot less repetition) than keep it as a "pure" `FlexComponent`.

packages/design-system/src/components/Stack/StackHorizontal.stories.tsx
```diff
@@ -49,7 +49,11 @@ export const manualStackArgs = {
 		defaultValue: 'nowrap',
 	},
 	height: {
-		options: Object.keys(heightOptions),
+		options: Object.keys(sizeOptions),
+		control: { type: 'select' },
+	},
+	with: {
```
Nigui, line 55: `width`

packages/design-system/src/components/Stack/Primitive/StackPrimitive.tsx
romainseb: I don't think it's a breaking change as i dough the fact it's meant to be use in the host project 👀

Benrajalu: It's not exported in the lib yeah

romainseb:
Dropped for this new one : 
https://github.com/Talend/ui/pull/4374

PR #4370, opened 2022-10-03
fix(sidepanel): reversed mode

romainseb:
CleanShot 2022-10-03 at 10 42 45@2x
CleanShot 2022-10-03 at 10 41 36@2x

### DIFF
--- a/packages/components/src/SidePanel/SidePanel.module.scss
+++ b/packages/components/src/SidePanel/SidePanel.module.scss
@@ -69,7 +69,7 @@ $large-docked-width: 7rem;
 				width: $tc-side-panel-icons-size;
 				height: $tc-side-panel-icons-size;
 			}
-			:global(svg){
+			:global(svg) {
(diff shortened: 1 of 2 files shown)


Nigui (approved):

packages/components/src/SidePanel/SidePanel.module.scss
```diff
@@ -209,7 +209,7 @@ $large-docked-width: 7rem;
 		}
 	}
 
-	&.reverse[role='navigation'][aria-expanded] {
```
jmfrancois, line 212: do you know when we have loosed this expanded attribute ?

jmfrancois (approved):

packages/components/src/SidePanel/SidePanel.module.scss
romainseb: Not at all 👀 
But this attribute check has not to be check here as we want the style to be apply regardless of the expanded or not status

PR #4390, opened 2022-10-14
chore(design-system): remove styled component

jmfrancois:

we don't want styled-component anymore.

rewrite using css module

### DIFF
--- a/packages/design-system/src/components/Icon/Icon.module.scss
+++ b/packages/design-system/src/components/Icon/Icon.module.scss
@@ -1,0 +1,57 @@
+@use '~@talend/design-tokens/lib/tokens' as tokens;
+
+.border {
+	circle,
+	path,
+	polygon,
+	polyline {
+        // only if border
+	    transform: translate(25%, 25%);;
+	}
+    :global(.ti-border) {
+        stroke: currentColor; fill: none; transform: none;
+    }
+}
+
+.svg {
+    fill: currentColor;
+	width: tokens.$coral-sizing-xxs;
+	height: tokens.$coral-sizing-xxs;
+	transform-origin: center;
+
+
+
+	&.spin {
+		animation-name: svg-spin;
+		animation-duration: 2s;
+		animation-iteration-count: infinite;
+		animation-timing-function: linear;
+	}
+	&.rotate-45 {
+		transform: rotate(45deg);
+	}
+	&.rotate-90 {
+		transform: rotate(90deg);
+	}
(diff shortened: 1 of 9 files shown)


packages/design-system/src/components/Icon/Icon.module.scss
```diff
@@ -0,0 +1,57 @@
...
+        // only if border
+	    transform: translate(25%, 25%);;
+	}
+    :global(.ti-border) {
+        stroke: currentColor; fill: none; transform: none;
+    }
+}
+
+.svg {
+    fill: currentColor;
```
Benrajalu, line 17: Indent issues? 🤔 Linter should be mad

packages/design-system/src/components/Icon/Icon.module.scss
```diff
@@ -0,0 +1,57 @@
...
+
+
+
+	&.spin {
+		animation-name: svg-spin;
+		animation-duration: 2s;
+		animation-iteration-count: infinite;
+		animation-timing-function: linear;
+	}
+	&.rotate-45 {
```
Benrajalu, line 30: This is the legacy icon component right? 

If you guys still want these, then they should be wrapped in a :global declaration. Otherwise you can't get them through the opened `className` props, CSS module would not know what's what by then.

packages/design-system/src/components/Icon/Icon.tsx
```diff
@@ -183,15 +131,14 @@ export const Icon = React.forwardRef(
...
 
 		return (
-			<SVG
+			<svg
 				{...rest}
 				name={!(isImg || isRemote) ? name : null}
 				{...accessibility}
-				className={classnames('tc-svg-icon', classname)}
-				border={border}
+				className={classnames('tc-svg-icon', classname, {})}
```
gjou-tlnd, line 138: why the empty obj ? 🤔

jmfrancois: no idea

packages/design-system/src/components/Icon/Icon.module.scss
jmfrancois: True this is legacy, and it could be done at the app level if needed.
But even if it is not used it will create a breaking change :thinking: 
I would go one step at a time and prefer to keep it as it is, just drop styled components first

gjou-tlnd (approved):

Benrajalu (approved):

PR #4198, opened 2022-07-07
feat(design tokens/TUX-1175): Tokens doc redesign

Benrajalu:
**What is the problem this PR is trying to solve?**
The design tokens were not documented within the DS but in a separate storybook.

**What is the chosen solution to this problem?**
Document them in the DS and update the look and feel of the documentation to make it more readable. 

**Please check if the PR fulfills these requirements**

- [ ] The PR have used `yarn changeset` to a request a release from the CI if wanted.
- [x] The PR commit message follows our [guidelines](https://github.com/talend/tools/blob/master/tools-root-github/CONTRIBUTING.md)
- [ ] Tests for the changes have been added (for bug fixes / features) And [non reg](./screenshots.md) done before need review
- [ ] Docs have been added / updated (for bug fixes / features)
- [ ] Related design / discussions / pages (not in jira), if any, are all linked or available in the PR

**[ ] This PR introduces a breaking change**

PR #4403, opened 2022-10-25
chore(TUX-1212): remove styled-components deps

jmfrancois:

We still have the dependency and components

* remove layout is not done at design level AFAIK (breaking change)
* remove deprecated tokens (breaking change)
* move switch component to css module
* remove all "pages" from the doc
* remove all templates from the doc

We need to keep styled-components in react-components because it is a peer dependency of guided tour


packages/design-system/src/components/ThemeProvider/ThemeProvider.tsx
```diff
@@ -1,32 +1,32 @@
...
-	DefaultTheme,
-	ThemeProvider as StyledThemeProvider,
-	ThemeProviderProps,
-} from 'styled-components';
 
-const ThemeProvider = ({ theme = light, children }: ThemeProviderProps<any>) => {
+export interface ThemeProviderProps {
+	theme: string;
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any
+	children: any;
```
smouillour, line 10: should be ReactNode instead of any?

packages/design-system/src/deprecatedTokens/animations.tokens.js
```diff
@@ -1,12 +0,0 @@
-import { css, keyframes } from 'styled-components';
```
smouillour, line 1: about all deprecated things, did you check if it's used somewhere before?

jmfrancois: no, at least not everywhere, they are deprecated since #4198. This is the purpose of this PR to remove them.
I m almost sure no one use theses tokens but design-tokens is already quite used.

smouillour (approved):

romainseb (approved):

PR #4417, opened 2022-11-03
feat(PRR-282): add locales-codeshift.js script

sbouyer:
**Context and reasoning**

- Currently, 2 versions of the English UI exist in the Product. One is the fallback UI written by the devs in the code. The other one is the reviewed English delivered with all the locales.

- Should anything go wrong with the locales, the incorrect fallback English displays.

- In the future, locale packages will not longer be included in the product builds. They will be delivered asynchronously via the CDN.

(Related task are tracked here: https://jira.talendforge.org/browse/I18N-278)

This means that the product builds on their own (ie., without the CDN) will display UI that is incorrect (fallback UI) and very different from what users see in real life (which will be coming from the CDN).
Eventually, no locale files will be committed in the product repos. Changes to the UI will not surface to the devs anymore.

We need to reintegrate correct UI copy, coming from the UX-writing workflow, back into the apps repos to:

1. Have a complete and correct copy in the builds for testing, if using the products without the CDN (unlikely but...), for cleanliness/good practice
2. Make changes in the UI copy visible to devs and QA
There is an initial effort to set this up and catch up on past evolutions of the copy.
Then this should be an on-going process of update, to be repeated at each edit of the labels.

A poc ran at the December 21 hackathon and was continued by TMC and DOC teams.
The code we want to add to be used by applications is the content of this PR
Readme link : https://github.com/Talend/Hackathons/tree/main/TheLocalizers/codemods


jmfrancois (changes requested):
let's integrate this better than just copy a script here it will not be easy to access.
dev will have to type a quite long command.

`talend-scripts locales:codeshift` command could be the one.

an other option is to distribute it as a dedicated script.

But at the moment this way is not usable

An other important point: The generated patch is not valid :grimacing: 
image

jmfrancois (approved):

ybaskaran (approved):

PR #4438, opened 2022-11-10
fix(ARCH-658): make the types optional

jmfrancois:

Current projects suffer to update to latest react-components because inferred types are not useful

move types in a side folder until we have correct typing coverage.


packages/components/package.json
```diff
@@ -3,13 +3,13 @@
   "description": "Set of react components.",
   "main": "lib/index.js",
   "mainSrc": "src/index.js",
-  "types": "lib/index.d.ts",
```
jmfrancois, line 6: this is a breaking change

packages/datagrid/src/types/types.d.ts
```diff
@@ -4,3 +4,6 @@ declare module '*.scss' {
 }
 
 declare module '@talend/react-bootstrap';
+declare module '@talend/react-components' {
+	export * from '@talend/react-components/types';
+}
```
jmfrancois, line 9: without it, it do not compile. But it means you can just declare it as empty if you want

Nigui (approved):

romainseb (approved):

PR #4452, opened 2022-11-17
fix(TFD-14788) retrieve display name

jmainguytalend:
Datagrid : in column definition, use display name when available for column name.
Display name is real name handling accents and special characters for field names instead of using technical name where special chars are escaped.

**Breaking change**
Update `getColumnDefs` function from `datasetSerializer` : attribute `headerName` in returned object is now valued by display name (prop named `talend.component.label` in avro schema) in prior to technical name (prop named `name`).

So priority to value attribute `headerName` is now :

1. `doc`
2. `talend.component.label`
3. `name`

To allow each project to customize headerName, props `doc`, `talend.component.label` and `name` are now returned in object returned by function `getColumnDefs`.


packages/datagrid/src/types/index.ts
```diff
@@ -18,7 +23,11 @@ export type Sample = {
 };
 
 export type AvroField = {
+	name: string;
+	doc: string;
+	[CUSTOM_FIELD_DISPLAY_NAME]?: string;
 	type: any;
+	[otherProp: string]: unknown;
```
jmfrancois, line 30: I don't get the point of this one

jmainguytalend: I don't master well this schema, but apparently is it possible to have some properties like
```
'@talend-quality@': {
	'0': 0,
	'1': 38,
	'-1': 62,
	total: 100,
}
```
So I add this generic properties to handle them.

Example in the UT :
https://github.com/Talend/ui/pull/4452/files#diff-a9c7df2f71f4da3a9ad08343ecbb5b00561565b6b29275d192551fd652ea4018R20-R25

sgendre (approved):

PR #4459, opened 2022-11-24
feat(ARCH-661/scripts): toggle dynamic-cdn-webpack-plugin

jmfrancois:

In the context of simplify tooling we can't test without changing the complete toolchain.

There is already an option to configure the plugin.
Add support to put `false` as value and propagate this everywhere:

* htmlwebpack plugin receive false on inject option (turn off the need for INITIATOR)
* do not copy assets in /cdn folder
* remove the dynamic-cdn-webpack-plugin from the config
* keep meta in the header to let assets-api simple use cases works

Note: playground fails with this option on the datagrid lazy load page. It is because getUMD('ag-grid-community') fails to load as globals like React do not exists.


packages/playground/talend-scripts.json
```diff
@@ -2,9 +2,7 @@
   "preset": "@talend/scripts-preset-react",
   "cmf": true,
   "html": {},
-  "dynamic-cdn-webpack-plugin": {
-    "exclude": []
-  },
+  "dynamic-cdn-webpack-plugin": false,
```
jmfrancois, line 5: we will not keep that, just the time to be sure it works

yyanwang (approved):

smouillour (approved):

PR #4467, opened 2022-11-28
fix(TFD-14833) InlineEditing : fix enter key always handled

jmainguytalend:
Component InlineEditing always handles key enter event even when not editing text

Check if component is in edition mode

### DIFF
--- a/.changeset/long-days-vanish.md
+++ b/.changeset/long-days-vanish.md
@@ -1,0 +1,5 @@
+---
+'@talend/design-system': patch
+---
+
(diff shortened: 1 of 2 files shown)


jmfrancois (dismissed):
this doesn't looks good to me.
The input is not even inside a form tag.
I would suggest to wrap the input with a form tag and use an onSubmit property to avoid the need to have an event handler to manage the Enter key.
Edit: also a fix always come with a test

romainseb (approved):

PR #4472, opened 2022-11-29
chore(ARCH-662/design-system): remove react-use

jmfrancois:

We have only one occurence of react-use in ds.
This is the copyToClipboard

At the same time react-use is deprecated.

* rewrite the copyToClipboard using a more simple API.
* create @talend/react-hooks for the purpose
* remove react-use from @talend/design-system


packages/hooks/src/hooks/useCopyToClipboard.ts
```diff
@@ -0,0 +1,31 @@
+import { useState } from 'react';
+
+type CopiedValue = string | null;
+type CopyFn = (text: string) => Promise<boolean>; // Return success
+
+export function useCopyToClipboard(): [CopiedValue, CopyFn] {
+	const [copiedText, setCopiedText] = useState<CopiedValue>(null);
+
+	const copy: CopyFn = async text => {
```
aamalric-talend, line 9: Isn't is possible to use React.useCallback to define this copy function ?

packages/hooks/package.json
```diff
@@ -0,0 +1,33 @@
...
+  "main": "lib/index.js",
+  "types": "lib/index.d.ts",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/Talend/ui.git"
+  },
+  "author": "Talend Frontend <frontend@talend.com> (http://www.talend.com)",
+  "license": "Apache-2.0",
+  "scripts": {
+    "build:lib": "talend-scripts build:ts:lib",
```
Nigui, line 15:
```suggestion
    "build:lib": "talend-scripts build:lib",
```

packages/hooks/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "watch": "talend-scripts build:ts:lib --watch",
```
Nigui, line 16:
```suggestion
    "watch": "talend-scripts build:lib --watch",
```

packages/hooks/src/hooks/useCopyToClipboard.test.ts
```diff
@@ -0,0 +1,30 @@
...
+		const writeText = jest.fn((data: string) => Promise.resolve());
+		Object.assign(navigator, {
+			clipboard: {
+				writeText,
+			},
+		});
+	});
+	afterEach(() => {
+		// global.navigator.clipboard.writeText = oldClipboard;
+	});
```
Nigui, line 15:
```suggestion
```

packages/hooks/src/hooks/useCopyToClipboard.ts
```diff
@@ -0,0 +1,31 @@
...
+		if (!navigator?.clipboard) {
+			console.warn('Clipboard not supported');
+			return false;
+		}
+
+		try {
+			await navigator.clipboard.writeText(text);
+			if (text !== copiedText) {
```
Nigui, line 17: not sure this condition is required, because react shouldn't trigger re-render if you set state with the same primitive value.

packages/design-system/src/components/Form/Field/Input/Input.Copy.tsx
```diff
@@ -68,7 +48,6 @@ const InputCopy = forwardRef(
 			<FieldPrimitive
 				label={label}
 				description={getDescriptionMessage()}
-				hasError={!!copyError}
```
Nigui, line 71: As a user, you lost information that copy didn't work

jmfrancois: yes :disappointed: but I have not been able to reproduce. Do you have a use case ?
AFAIK we do not support browser that do not have the feature.

PR #4481, opened 2022-12-07
fix(TDP-11600): allow to add new value in datalist

romainseb:
If the value is not existing, we have only "no result found" and no way to understand that we can add a value in the datalist

On the datalist, new props called 
- `allowAddNewElements` can be passed to let the user create new value with a hint to tell him that the value was not existing in the current titlemap.
- `allowAddNewElementsSuffix` allow to override the "(new)" suffix by another one.

### DIFF
--- a/packages/components/src/Datalist/Datalist.component.js
+++ b/packages/components/src/Datalist/Datalist.component.js
@@ -155,9 +155,7 @@ function Datalist(props) {
 		filterValue,
 		multiSection: props.multiSection,
 		allowAddNewElements: props.allowAddNewElements,
-		allowAddNewElementsSuffix: props.allowAddNewElementsSuffix
-			? props.allowAddNewElementsSuffix
-			: t('NEW', { defaultValue: '(new)' }),
+		allowAddNewElementsSuffix: props.allowAddNewElementsSuffix ?? t('NEW', '(new)'),
 	});
 
 	// in uncontrolled mode, props.value acts as an initial value, then Datalist handles state, props.value never changes.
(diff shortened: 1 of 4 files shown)


packages/components/src/Datalist/Datalist.component.js
```diff
@@ -129,6 +154,10 @@ function Datalist(props) {
 		titleMap: props.titleMap,
 		filterValue,
 		multiSection: props.multiSection,
+		allowAddNewElements: props.allowAddNewElements,
+		allowAddNewElementsSuffix: props.allowAddNewElementsSuffix
+			? props.allowAddNewElementsSuffix
+			: t('NEW', { defaultValue: '(new)' }),
```
Nigui, line 160:
```suggestion
		allowAddNewElementsSuffix: props.allowAddNewElementsSuffix ?? `(${ t('NEW', 'new' ) })`,
```
- I've used `??` operator
- I've replaced suffix by wrapping "new" translation with parenthesis

Nigui (commented):
👍

packages/components/src/Datalist/Datalist.component.js
romainseb: Agree on the ?? operator
But for the parentheses, i would tend to say that it may depends of the locale if we want maybe some spacing or something

Nigui: Then IMO `NEW` i18n key is too generic to specify "new word with parenthesis"

Nigui (approved):

jmfrancois (approved):

PR #4488, opened 2022-12-12
fix(Password): use the InputPrimitive for password field [same but different]

jmfrancois:

#4487 
primitives are meant to be internal.
I continue to think we need to have field without the div but don't know what API for that ?
* a props <Password fieldOnly />
* a dedicated component for that but which name

I propose here to just do not use the field template from TUI.

If we are ok here is what should be done:
* [x] fix the test
* [ ] ensure all use cases are working well (error, disabled, etc..)

### DIFF
--- a/packages/forms/src/UIForm/fields/Text/Text.component.js
+++ b/packages/forms/src/UIForm/fields/Text/Text.component.js
@@ -2,8 +2,8 @@
 
 import PropTypes from 'prop-types';
 import React from 'react';
-import { get, omit } from 'lodash';
-import { Link, Form } from '@talend/design-system';
+import { get } from 'lodash';
+import { Form } from '@talend/design-system';
 import FieldTemplate from '../FieldTemplate';
 import { generateDescriptionId, generateErrorId } from '../../Message/generateId';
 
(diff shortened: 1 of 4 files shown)


packages/forms/src/UIForm/fields/Text/Text.component.js
```diff
@@ -48,7 +47,22 @@ export default function Text(props) {
...
+				{...fieldProps}
+				id={id}
+				label={title}
+				required={schema.required}
+				hasError={!!errorMessage}
+				description={description}
+				aria-invalid={!isValid}
+				aria-required={get(schema, 'required')}
+				aria-describedby={`${descriptionId} ${errorId}`}
+				link={link && <Link {...omit(link, ['label'])}> {link.label} </Link>}
```
lmaillet, line 62: Not sure about this part as the link prop will create a Link component in the design-system 🤔 
https://github.com/Talend/ui/blob/master/packages/design-system/src/components/Form/Primitives/Field/Field.tsx#L73

jmfrancois: It is on master I put it to not change the behavior. I don't know :shrug:  what is the use case for it.

romainseb:
@lmaillet is this PR still relevant ?

lmaillet:
Too old

PR #4511, opened 2022-12-29
fix(webpack): improve performance for dev experience

jmfrancois:

Working on a small project I already suffer from huge lag when change a bit of code.
The page is reloaded after 5 to 10 secondes.

Also looking at the index.html page which is generated I found all meta are present twice.

Latest issue I have a cdn folder copied even if I set the option `dynamic-cdn-webpack-plugin` to false.

Also eslint was broken makes it impossible to commit with husky pre commit hook.

* fix eslint config on the repo and the husky config

* Playing with the configuration I have discovered the analytics webpack plugins are the root cause.
unplug them but let them available with just the `--env analyze` flag as they are still usefull.

* Do not copy assets in /cdn if no cdn plugin.

* remove meta from the custom header.

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,10 @@
     "fork/*"
   ],
   "lint-staged": {
-    "*.{js,jsx,ts,tsx}": [
+    "*.{js,jsx}": [
+      "eslint --quiet --fix"
+    ],
+    "*.{ts,tsx}": [
       "eslint --quiet --fix"
     ],
     "*.{json,md,mdx,html,js,jsx,ts,tsx}": [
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -55,7 +55,10 @@
     "fork/*"
   ],
   "lint-staged": {
-    "*.{js,jsx,ts,tsx}": [
+    "*.{js,jsx}": [
+      "eslint --quiet --fix"
+    ],
+    "*.{ts,tsx}": [
```
jmfrancois, line 61: I dunno why but husky was always in TS mode. Do that fix it

packages/design-system/src/components/WIP/Accordion/Primitive/CollapsiblePanel.tsx
```diff
@@ -44,7 +44,7 @@ const CollapsiblePanel = forwardRef(
 		}: CollapsiblePanelPropsType,
 		ref: Ref<HTMLButtonElement>,
 	) => {
-		const [localExpanded, setLocalExpanded] = useState(false);
+		const [localExpanded, setLocalExpanded] = useState(!!expanded);
```
jmfrancois, line 47: By default if I put expanded but without manage the state it do not work

Gbacc (approved):

tlnd-mhuchet (approved):

PR #4512, opened 2022-12-29
chore(List): adding skeleton when default cell is loading

jlame:
we want to be able to add a skeleton on a single cell when data is being retrieved.
make the default renderer accept an object


packages/components/src/VirtualizedList/Content.component.js
```diff
@@ -2,6 +2,16 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { Column } from 'react-virtualized';
 import TooltipTrigger from '../TooltipTrigger';
+import Skeleton from '../Skeleton';
+
+function renderDefaultCell(cellData) {
+	const value = typeof cellData === 'object' ? cellData.value : cellData;
```
romainseb, line 8: What's weird for me is the fact for an entire line to be "loading", we will have a row with all attributes to be rendered in the list to { isLoading: true }.

Sure it works, i have no dough on that, but it feel strange for me.
Obviously, now it's easy to say it without having any proposal.

What happened with the title cell for instance as it's not using this cell ?
Just curious, if it's only related to a given column of the list, should it be a specific cell renderer for this type of data that needs to be loaded ?

Like the users in the dataset list :
https://github.com/Talend/ui-dataset/blob/develop/src/app/components/UserCellRenderer/UserCellRenderer.tsx
(not the best implementation possible but still, you understand the idea 😉)

jlame: thank you :D

PR #4523, opened 2023-01-05
feat(talend-scripts): move to pure ESM

jmfrancois:

knowing this https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c
more and more packages goes to pure ESM.

our current talend-scripts is not able to update some libs in major which are now ESM.
But pure ESM module can import CJS.

This means migration to EMS is possible

let's try on scripts-core to see the impact

lesson learned:
* import must have js extension (our linter is not happy with that)
* json is not supported by import(path) => `JSON.parse(fs.readFileSync(path))`;
* cjs can't require esm
* jest has to be upgraded before rewrite tests to ESM


tools/scripts-core/scripts/build-lib-umd.js
```diff
@@ -1,19 +1,24 @@
...
-		['--config', hereRelative(__dirname, '../config/webpack.config.js')].concat(options),
+		['--config', hereRelative(dirname, '../config/webpack.config.js')].concat(options),
 		{ stdio: 'inherit', env },
 	);
 }
 
-module.exports = function build(env, presetApi, options) {
-	const packageJSON = require(getUserConfigFile(['package.json']));
+export default function build(env, presetApi, options) {
+	const packageJSON = JSON.parse(fs.readFileSync(getUserConfigFile(['package.json'])));
```
sgendre, line 21: I usually do try catch around JSON.parse

tools/scripts-core/scripts/test.js
```diff
@@ -1,15 +1,16 @@
-const spawn = require('cross-spawn');
-const { resolveBin } = require('../utils/path-resolver');
-const { getPreset } = require('../utils/preset');
-const { getUserConfigFile } = require('../utils/env');
+/* eslint-disable import/extensions */
+import spawn from 'cross-spawn';
+import { resolveBin } from '../utils/path-resolver.js';
```
sgendre, line 3: those 3 imports are often needed, 
what about having an index.Js exposing them for being imported as a oneliner for convenience?

tools/scripts-core/scripts/build-lib-umd.js
jmfrancois: If I know what to do in the catch me too.
But in this case and others I prefer to have a stack trace and exit the process as something goes wrong. Which was the case also with require on JSON file

PR #4556, opened 2023-01-17
fix(ARCH-662/scripts): win32 support

jmfrancois:

ENOENT on win32 plateform is thrown.

This is due to resolveBin which is doing lots of useless things which may work if we fix them only on linux.


package.json
```diff
@@ -17,7 +17,6 @@
     "release": "yarn pre-release && yarn changeset publish",
     "lint-staged": "lint-staged",
     "lint": "cross-env WORKSPACE_RUN_FAIL=no-bail node workspace-run.js lint",
-    "build": "yarn workspaces run build",
```
jmfrancois, line 20: has never existed

tools/scripts-utils/src/path.js
```diff
@@ -23,22 +23,10 @@ function resolveBin(modName, { executable = modName, cwd = process.cwd() } = {})
...
-		const modPkgPath = require.resolve(`${modName}/package.json`);
-		const modPkgDir = path.dirname(modPkgPath);
-		const mod = JSON.parse(fs.readFileSync(modPkgPath));
-		const { bin } = mod.bin;
-		const binPath = typeof bin === 'string' ? bin : bin[executable];
-		const fullPathToBin = path.join(modPkgDir, binPath);
-		if (fullPathToBin === systemCommandPath) {
-			return executable;
-		}
-		return fullPathToBin.replace(cwd, '.');
```
jmfrancois, line 36: this code is supposed to look into package.json bin attribute but nodejs is creating a .CMD file for that on win32 so the path will never been equal and the output will never be useful

lmaillet (approved):

tools/scripts-core/src/config/webpack.config.js
```diff
@@ -28,7 +28,10 @@ export default async (env = {}) => {
 	// User configuration file
 	const userConfigPath = presetApi.getUserConfig(['webpack', 'config', presetApi.mode]);
 	if (userConfigPath) {
-		const userConfigAbsolutePath = utils.path.getAbsolutePath(userConfigPath);
+		let userConfigAbsolutePath = utils.path.getAbsolutePath(userConfigPath);
+		if (process.platform === 'win32') {
+			userConfigAbsolutePath = `file:///${userConfigAbsolutePath}`;
```
jmfrancois, line 33: as we use `await import(userConfigAbsolutePath)` we need a protocol supported by windows.
I have found on stackoverflow this trick :point_up: and it works

tools/scripts-utils/src/path.js
```diff
@@ -23,22 +23,10 @@ function resolveBin(modName, { executable = modName, cwd = process.cwd() } = {})
...
-	} catch (error) {
-		if (systemCommandPath) {
-			return executable;
-		}
-		throw error;
+	if (process.platform === 'win32' && systemCommandPath) {
+		return systemCommandPath;
```
jmfrancois, line 27: cross-spawn was doing this in case of ENOENT (bin not found).
as which command return us the full path (so with `.CMD` at the end ie (node_modules\\.bin\\webpack.CMD) we rely on this now

smouillour (approved):

PR #4595, opened 2023-02-13
feat(TDOPS-421): faceted search new operators and select all function

Gbacc:
Faceted search new operators and select all function

### DIFF
--- a/packages/faceted-search/src/components/Badges/BadgeCheckboxes/BadgeCheckboxesForm.component.js
+++ b/packages/faceted-search/src/components/Badges/BadgeCheckboxes/BadgeCheckboxesForm.component.js
@@ -25,6 +25,8 @@ const BadgeCheckbox = ({ checked, id, label, onChange }) => {
 				id={`${id}-checkbox`}
 				label={label}
 				checked={checked}
+				data-testid={`badge-checkbox-form-checkbox-${id}`}
+				data-test={`badge-checkbox-form-checkbox-${id}`}
 			/>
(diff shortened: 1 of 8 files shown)


romainseb (approved):

packages/faceted-search/src/components/Badges/BadgeCheckboxes/BadgeCheckboxesForm.component.js
```diff
@@ -86,24 +89,49 @@ const BadgeCheckboxesForm = ({
 			checkboxes.filter(c => c.checked),
 		);
 	};
+	const onToggleAll = event => {
+		const checked = event.target.checked;
+		checkboxes.forEach(entity => (entity.checked = checked));
```
jmainguytalend, line 94: Is it safe to update `checkboxes` ? This doesn't trigger any new render but this const is used for rendering component.
How new render is triggered ? Through "onChange" call ? If yes, it is more reliable to give some new array with updated properties

Gbacc: I understand your point 
In this case it will both update checkboxes internally , and will send the updated list with a "onChange". 
Most of the time "onChange" will update the checkbox on parent component and then the parent should send props back to the children, but it's not mandatory so it could cause a desync of the rendering. So i've used the same approach as "onChangeCheckBoxes" , i agree with you it's bugging me a bit too

jmainguytalend (approved):

PR #4613, opened 2023-03-01
feat(TDOPS-3583): data tracking attribute in checkbox

VolodymyrKovalM:
In TMC we use separate "data-tracking" attribute to track the element in Pendo, and avoid impact on Pendo when using another attributes (for example "id", "data-test" may be changed, and "data-tracking" is used only for Pendo)
For checkbox we need to track it's checked and unchecked state
Would be nice to have it in TUI component, otherwise we will need to create wrapper component, one more level, just to have access to "checked" property and handle "data-tracking" attribute for different checkbox state

Having "data-tracking" attribute can be useful for other Teams who may need to do similar tracking with Pendo, so, with this change we will be able to pass "data-tracking" in ui schema, and the component will append "on" or "offf" depending on checkbox state

https://jira.talendforge.org/browse/TDOPS-3583

### DIFF
--- a/.changeset/tidy-years-wonder.md
+++ b/.changeset/tidy-years-wonder.md
@@ -1,0 +1,5 @@
+---
+'@talend/react-components': minor
+---
+
+feat(TDOPS-3583): data tracking attribute in checkbox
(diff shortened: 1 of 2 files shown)


jmfrancois (commented):
data-feature seems to do the exact same thing

VolodymyrKovalM:
> > data-feature seems to do the exact same thing
> 
> I will check with PO, if the data-feature will be enough, as recently we agreed to have separate attribute for Pendo tracking

After checking, we will go with "data-feature" attribute for this case, to not implement specific attribute for Pendo
So, closing the PR

PR #4629, opened 2023-03-13
feat(TDOPS-421): Pass data attributes to faceted search components

Gbacc:
Pass data attributes to faceted search components

### DIFF
--- a/packages/faceted-search/src/components/Badges/BadgeDate/BadgeDate.component.js
+++ b/packages/faceted-search/src/components/Badges/BadgeDate/BadgeDate.component.js
@@ -38,6 +38,7 @@ const BadgeDate = ({
 	readOnly,
 	removable,
 	displayType,
+	...rest
 }) => {
 	const currentOperators = useMemo(() => operators || createDefaultOperators(t), [operators, t]);
 	const currentOperator = operator || currentOperators[0];
@@ -70,6 +71,7 @@ const BadgeDate = ({
 					value={badgeValue}
 					feature={category || label}
 					t={t}
+					{...rest}
 				/>
 			)}
 		</BadgeFaceted>
--- a/packages/faceted-search/src/components/BadgesGenerator/BadgesGenerator.component.js
+++ b/packages/faceted-search/src/components/BadgesGenerator/BadgesGenerator.component.js
@@ -3,15 +3,21 @@ import PropTypes from 'prop-types';
 import get from 'lodash/get';
 
 import { badgesFacetedPropTypes } from '../facetedSearch.propTypes';
+import { pick } from 'lodash';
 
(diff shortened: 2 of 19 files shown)


packages/faceted-search/src/components/Badges/BadgeDate/BadgeDate.component.js
```diff
@@ -38,6 +38,7 @@ const BadgeDate = ({
 	readOnly,
 	removable,
 	displayType,
+	...rest
```
ybaskaran, line 41: can we pass only the required props ? like dataAttributes ?

ybaskaran (approved):

packages/faceted-search/src/components/Badges/BadgeDate/BadgeDate.component.js
Gbacc: Well why not it's up to us i can do the change to restrict a bit

packages/faceted-search/src/components/BadgesGenerator/BadgesGenerator.component.js
```diff
@@ -3,15 +3,21 @@ import PropTypes from 'prop-types';
...
+		const dataAttributesKeys = Object.keys(metadata).filter(objectKey =>
+			objectKey.startsWith('data-'),
+		);
+		const dataAttributes = pick(metadata, dataAttributesKeys);
 		if (BadgeComponent) {
 			acc.push(
 				<BadgeComponent
 					{...metadata}
 					{...properties}
+					dataAttributes={dataAttributes}
```
jmfrancois, line 20: why not spread dataAttributes and use rest in others ?

jmfrancois (changes requested):
I don't really like the `dataAttributes`
I would let a more global function getDataAttributeFrom(props) where you can read it.
So we can use spread
```{...getDataAttributeFrom(rest)}``` If you think there is other stuff in rest you want.

or [dataAttributes, others] = filterDataAttributesFrom(props) so you can keep the other.
Even better I think
WDYT ?

Gbacc:
> I don't really like the `dataAttributes` I would let a more global function getDataAttributeFrom(props) where you can read it. So we can use spread `{...getDataAttributeFrom(rest)}` If you think there is other stuff in rest you want.
> 
> or [dataAttributes, others] = filterDataAttributesFrom(props) so you can keep the other. Even better I think WDYT ?

Ok let's say , in the badge generator i remove the filter on `data-`, you are right it's already spread with metadata.
In the upper badge component we just spread `{...rest}` like it's already done,  and in the badgeForm component we use some kind of function to filter data attributes only `{...getDataAttributeFrom(rest)}` on the Action ?

jmfrancois (approved):

PR #4643, opened 2023-03-17
feat(components): quality bar update

romainseb:
When the quality is not applicable, we don't have a way to disable it with an easy way

Pass a prop to disable it


packages/components/src/QualityBar/QualityBar.component.tsx
```diff
@@ -0,0 +1,100 @@
+import React from 'react';
+
+import { EnrichedQualityType, QualityBarPercentages, QualityCommonProps } from './QualityBar.types';
+import { QualityBarRatioBars } from './QualityBarRatioBars.component';
+import { SplitQualityBar } from './SplitQualityBar.component';
+
+export type QualityBarProps = QualityCommonProps & {
+	disabled?: boolean;
```
ncomont, line 8: could this be part of QualityCommonProps ?

romainseb: Could be part of it, but it would required to change other stuff regarding to the typing.
It's used for instance in `getQualityPercentagesRounded` with a `Required` so the property is meant to be present as the others.

Otherwise, it looks more a props rather that a quality attribute to me

packages/components/src/QualityBar/SplitQualityBar.component.tsx
```diff
@@ -0,0 +1,58 @@
...
+	valid,
+	percentages,
+	disabled,
+}: SplitQualityBarProps) {
+	const totalValues = (empty || 0) + (invalid || 0) + (na || 0) + (valid || 0);
+	const usedValues = { empty, invalid, na, valid };
+	const fwd = { getDataFeature, onClick };
+
+	if (disabled) {
+		return null;
```
ncomont, line 35: imo we should return the same kind of qb as we do for the others
i think that hiding it is more a project concern

romainseb: allright, i can make it this way 👍🏼

ncomont (approved):

PR #4644, opened 2023-03-17
feat(components): use DS Error State in Stepper

romainseb:
Migrate from the old error feedback to the ErrorState from the DS

CleanShot 2023-03-17 at 17 32 29@2x


packages/components/src/Stepper/Stepper.component.js
```diff
@@ -170,6 +172,16 @@ function Stepper({ steps, title, renderActions, children }) {
 		}
 	}, [steps]);
 
+	if (isInError) {
```
jmfrancois, line 175: The isInError is still used after, could you remove the usage on the code after this if branch ?

jmfrancois (changes requested):

packages/components/src/Stepper/Stepper.component.js
romainseb: Yes 👍🏼

packages/components/src/Stepper/Stepper.component.test.js
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import { shallow } from 'enzyme';
+import { render } from '@testing-library/react';
 import Stepper from './Stepper.component';
```
jmfrancois, line 4: may be ?
```suggestion
jest.unmock('@talend/design-system')
```

packages/components/src/Stepper/__snapshots__/Stepper.component.test.js.snap
```diff
@@ -1,179 +1,126 @@
...
-               (Aborted)
-            </div>
-          </li>
-        </ol>
-      </div>
-    </div>
-  </StepperTransition>
-</React.Fragment>
+      <span
+        class="CoralErrorState"
```
jmfrancois, line 13: this is a mock of ds

packages/components/src/Stepper/Stepper.component.test.js
romainseb: I wasn't sure of if we want to unmock the DS in this package.
I add it so

packages/components/src/Stepper/Stepper.component.js
```diff
@@ -170,25 +172,28 @@ function Stepper({ steps, title, renderActions, children }) {
...
+		const errorStep = getStepInError(steps);
```
aamalric-talend, line 176: Can it be possible that we get 'undefined' here, let's say there's no status LOADING_STEP_STATUSES.FAILURE ??

aamalric-talend (approved):

jmfrancois (approved):

PR #4650, opened 2023-03-22
fix(ARCH-662): request security issue

jmfrancois:
**What is the problem this PR is trying to solve?**

request 2.88.0 in Talend/​ui/​tools/​scripts-publish-local. Comes from verdaccio 5.20.1. There is no fixed version

**What is the chosen solution to this problem?**

* use npx so verdaccio is installed in its latest version only if the script is used

**Please check if the PR fulfills these requirements**

- [x] The PR have used `yarn changeset` to a request a release from the CI if wanted.
- [x] The PR commit message follows our [guidelines](https://github.com/talend/tools/blob/master/tools-root-github/CONTRIBUTING.md)
- [x] Tests for the changes have been added (for bug fixes / features) And [non reg](./screenshots.md) done before need review
- [x] Docs have been added / updated (for bug fixes / features)
- [x] Related design / discussions / pages (not in jira), if any, are all linked or available in the PR

**[ ] This PR introduces a breaking change**

PR #4652, opened 2023-03-22
chore(ARCH-662): upgrade cypress 9 -> 12

jmfrancois:

following #4650 here we target `unset-value` which has a security issue.

unset-value 1.0.0 Fixed version is 2.0.1.

* upgrade cypress
* rewrite test using RTL integration
* use the new default file pattern (*.cy.tsx)

image


packages/design-system/src/components/Dropdown/Dropdown.cy.tsx
```diff
@@ -69,75 +72,83 @@ const WithDividers = () => (
 );
 
 context('<Dropdown />', () => {
+	beforeEach(() => {
+		cy.configureCypressTestingLibrary({ testIdAttribute: 'data-test' });
```
smouillour, line 76: Can we do without it?

packages/design-system/src/components/Tag/Tag.cy.tsx
```diff
@@ -0,0 +1,12 @@
+/* eslint-disable testing-library/await-async-query */
+/* eslint-disable testing-library/prefer-screen-queries */
+import React from 'react';
+
+import { Tag } from '.';
+
+context('<Tag />', () => {
+	it('should render', () => {
+		cy.mount(<Tag data-testid="my.tag" />);
```
smouillour, line 9: Note for later: it could be good to test all variants and props OR put test directly on variant files

packages/design-system/src/components/Dropdown/Dropdown.cy.tsx
```diff
@@ -1,8 +1,11 @@
+/* eslint-disable testing-library/prefer-screen-queries */
+/* eslint-disable testing-library/await-async-query */
```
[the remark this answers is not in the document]
jmfrancois: better let's fix the props (I have removed it)

packages/design-system/src/components/Tag/Tag.cy.tsx
jmfrancois: I have added some but because it is css module and the variant only apply a class, I am not able to check the class (it is hashed)

packages/design-system/cypress/support/commands.ts
```diff
@@ -0,0 +1,37 @@
+/// <reference types="cypress" />
```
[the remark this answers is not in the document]
jmfrancois: I have moved commands like mount and integration of RTL here.

packages/design-system/cypress/support/commands.ts
```diff
@@ -0,0 +1,29 @@
...
+// declare global {
+//   namespace Cypress {
+//     interface Chainable {
+//       login(email: string, password: string): Chainable<void>
+//       drag(subject: string, options?: Partial<TypeOptions>): Chainable<Element>
+//       dismiss(subject: string, options?: Partial<TypeOptions>): Chainable<Element>
+//       visit(originalFn: CommandOriginalFn, url: string, options: Partial<VisitOptions>): Chainable<Element>
+//     }
+//   }
+// }
```
romainseb, line 27: Maybe the comments in this file are not required anymore :)

romainseb (approved):

PR #4666, opened 2023-04-06
feat(icons): add new svg icons

romainseb:
For some product feature, we would like to be able to access this external product icons
I know this package is "deprecated" when it comes to use the icons in the product but as it's externals logos, we don't want to put them in the new icons with size folder (which is our package)

Add them


tlnd-mhuchet (approved):

jmfrancois (approved):
So I would like to remove the deprecated from the doc. Both doesn't serve the same goal.

PR #4732, opened 2023-05-22
feat(lint): add --fix option

jmfrancois:

we have a lint command but adding `--fix` do not work.

Indeed as the script as to reconciliate options between stylelint and eslint we need to manage options.

Add --fix support.

### DIFF
--- a/.changeset/mighty-taxis-exist.md
+++ b/.changeset/mighty-taxis-exist.md
@@ -1,0 +1,5 @@
+---
+'@talend/scripts-core': minor
+---
+
+feat: add --fix support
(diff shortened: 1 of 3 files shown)


tlnd-mhuchet (approved):
lgtm

just one thought, as we already pass through the options, can't we always take all passed options with the old `args = args.concat(options);` but detect if we pass a path or not to add it or not ? 🤔 

This pass though was intended to go with husky which is passing all files modifed as parameter.
So I can't pass src or husky will commit a global --fix

smouillour (approved):

romainseb (approved):

PR #4733, opened 2023-05-23
feat(TFD-15699): make http package compatible with nodejs

sgendre:
document does not exists on node env

prevent access to document if not browser env


jmfrancois (changes requested):
It starts small and seems easy, but at the begining we have identify a lots of complexity.

* browser rely on DOM api (here fetch)
* nodejs rely on it's own API (http/https)
* ESM looks different on node (import ends with a dot js)
* Compilation (so tooling) must target something understood by both plateform which is may not possible regarding the previous point.
* you will not be able to do streaming structure on http request which is a best practice in nodejs world

So I would recommend not to try this approach for all theses reasons.

PR #4720, opened 2023-05-12
feat(TFD-15534): Add status to accordion header

aamalric-talend:
**What is the problem this PR is trying to solve?**
Adding the status variant to an Accordion header instead of simple string title
Figma:
- https://www.figma.com/file/o8mVcWTKrRF99ze5Y59r3C/Control-Flows?type=design&node-id=2236-62870&t=UsV5THoQtIjRLffH-0
- https://www.figma.com/file/CDfr4jLz1m6Ud2RNi4qpQJ/Accordion?type=design&node-id=2006-49458&t=C94f4Sg0w7YXIpOV-0

**What is the chosen solution to this problem?**
Add an optionnal status property that takes variant values ('successful', 'failed', 'inProgress', 'warning', 'canceled')

**Please check if the PR fulfills these requirements**

- [x] The PR have used `yarn changeset` to a request a release from the CI if wanted.
- [x] The PR commit message follows our [guidelines](https://github.com/talend/tools/blob/master/tools-root-github/CONTRIBUTING.md)
- [ ] Tests for the changes have been added (for bug fixes / features) And [non reg](./screenshots.md) done before need review
- [x] Docs have been added / updated (for bug fixes / features)
- [ ] Related design / discussions / pages (not in jira), if any, are all linked or available in the PR

**[ ] This PR introduces a breaking change**

PR #4760, opened 2023-06-13
feat(TFD-15573) InlineEditing : update value from default value prop

jmainguytalend:
InlineEditing component don't allow to update its initial value while mounted.
So actually, it requires some workaround to not display component while initial value to set in doesn't exist.

Add useEffect with dependency on defaultValue prop to update internal component value with it when updated


packages/design-system/src/components/InlineEditing/Primitives/InlineEditingPrimitive.tsx
```diff
@@ -84,6 +86,14 @@ const InlineEditingPrimitive = forwardRef(
 			}
 		}, [hasError]);
 
+		useEffect(() => {
+			// Update state's value when default value is updated
+			if (value !== defaultValue) setValue(defaultValue);
```
jmfrancois, line 91: nop the idea of a defaultValue is to uncontrolled.
If you want to control the value we should use a new value props and use this one here.

jmfrancois (changes requested):

packages/design-system/src/components/InlineEditing/Primitives/InlineEditingPrimitive.tsx
```diff
@@ -1,5 +1,11 @@
+import { ButtonIcon } from '../../ButtonIcon';
```
tlnd-mhuchet, line 1: Hum, is it wanted to change the order of imports ? Looks like the prettier plugin was not running 🤔

romainseb: Import order looks really weird indeed

packages/design-system/src/components/InlineEditing/Primitives/InlineEditingPrimitive.tsx
```diff
@@ -38,16 +36,40 @@ export type OnEditEvent =
...
+	 * To set to use component as controlled one.
+	 */
+	value?: string;
+
+	/**
+	 * (Optional) In controlled way, handle value update.
+	 * It must be set to use this component as controlled component.
+	 * @param newValue new value filled
+	 */
+	onChangeValue?: (newValue: string) => void;
```
romainseb, line 72: When it comes to define an API for a component that have 2 exclusives mode, i would prefer make it explicit with the typescript interface.

Like this way:
https://github.com/Talend/ui/pull/4720/files#diff-65104284714bc80639dde175162db58c4d8cb948751dd62e72377ce9c6906941R38

Cause if 2 sets of props is only working together, it's easier to understand rather than trying to pick the right ones

packages/design-system/src/components/InlineEditing/Primitives/InlineEditingPrimitive.tsx
```diff
@@ -57,22 +79,37 @@ const InlineEditingPrimitive = forwardRef(
...
+		const [inProgressValue, setInProgressValue] = useState<string | undefined>(defaultValue);
+
+		// Both modes (controlled & uncontrolled) - store a snapshot value
+		const [snapshotValue, setSnapshotValue] = useState<string | undefined>(defaultValue);
+
 		const toggleEditionMode = (isEditionMode: boolean) => {
+			if (isEditionMode) {
+				// Both modes - initiate inProgressValue with stored value
+				setInProgressValue(snapshotValue);
+			}
```
romainseb, line 111: Isn't `inProgressValue` already to `defaultValue`

packages/design-system/src/components/InlineEditing/Primitives/InlineEditingPrimitive.tsx
```diff
@@ -84,18 +121,31 @@ const InlineEditingPrimitive = forwardRef(
...
-				const sentValue = value || '';
-				onEdit(event, sentValue);
+
+			if (onChangeValue) {
+				// Controlled mode - call external controller to handle validating new filled value
+				onChangeValue(inProgressValue || '');
+			} else if (onEdit) {
+				// Uncontrolled mode - save input value as snapshot and notify
+				setSnapshotValue(inProgressValue);
+				onEdit(event, inProgressValue || '');
```
romainseb, line 140: We don't know on the parent side if it's a submit or just an onChange like altering the field

romainseb (dismissed):
Looks not like a `controlled` component here.
On a controlled approach, we would update the inline editing component with the current "value" set by the parent to be rendered as a value in the inlineEditing.
OnChange would be called for each alteration of the value and the parent has to deal with the cancel/submit part to setup the good value in the inline editing component.

Here, on my opinion, it's looks like a new uncontrolled mode but a different one.

packages/design-system/src/components/InlineEditing/Primitives/InlineEditingPrimitive.tsx
tlnd-mhuchet: It should, as `defaultValue` is the default value of the state 🤔 
Maybe it is for successive toggle of the edit mode ?

packages/design-system/src/components/InlineEditing/Primitives/InlineEditingPrimitive.tsx
jmainguytalend: problem is from prettierrc file i think. It doesn't import generic one and has no rule about import order. I will update prettierrc file

packages/design-system/src/components/InlineEditing/Primitives/InlineEditingPrimitive.tsx
jmainguytalend: regardless your opinion about controlled mode / uncontrolled mode here, defaultValue is designed for uncontrolled mode and is not used when user provider value/onChangeValue

packages/design-system/src/components/InlineEditing/Primitives/InlineEditingPrimitive.tsx
```diff
@@ -84,18 +121,27 @@ const InlineEditingPrimitive = forwardRef(
 			}
 		}, [hasError]);
 
+		useEffect(() => {
```
tlnd-mhuchet, line 124: I think the useEffect is no longer useful with this implem, as the internalValue is not used at all when controlled 🤔

jmainguytalend: Removed!

tlnd-mhuchet (approved):

jmfrancois (approved):
approved in advance:
why do you add placeholder in cypress test ?
there is an unused import, please fix eslint :pray:

PR #4774, opened 2023-06-29
feat(http): add global interceptors

romainseb:
We don't have a global interceptor stuff yet
For some use cases, it could be nice to handle some behaviour at only one place, like logout for instance

Add a local context to http package

### DIFF
--- a/.changeset/good-radios-push.md
+++ b/.changeset/good-radios-push.md
@@ -1,0 +1,5 @@
+---
+'@talend/http': minor
+---
+
+feat(http): add the possibility to add global interceptors for every calls that got through @talend/http calls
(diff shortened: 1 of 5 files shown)


aamalric-talend (approved):
LGTM

itenger-talend (commented):
Nice one. Request interceptors could be also useful in the future - for instance if we need to inject an authorization header (bearer token) into every request.

romainseb:
> Nice one. Request interceptors could be also useful in the future - for instance if we need to inject an authorization header (bearer token) into every request.

It's already available through the `setDefaultConfig` function i think where you can setup headers and so on that are passed to every request 🤞🏼 
https://github.com/Talend/ui/blob/39fb412eadf28b45a9dd215b6aa30e656c12a609/packages/http/src/config.ts#L37

itenger-talend (approved):

jmfrancois (commented):
you close the door to request interceptor ?
I think naming should follow request/response

PR #4788, opened 2023-07-07
feat(TDC-7254): Stepper progress updating according to current step

dlcaldeira:
Stepper's progress was not being updated as the user moved to the next step

Added a parameter to control which is the active step allowing the progress to be according to where the user is 

### DIFF
--- a/.changeset/pink-llamas-cheer.md
+++ b/.changeset/pink-llamas-cheer.md
@@ -2,4 +2,6 @@
 '@talend/design-system': minor
 ---
 
-feat(TDC-7254): Stepper progress updating according to current step
+feat(TDC-7254/Stepper): Add a new props `currentStepIndex` to control current step. Set to zero by default.
+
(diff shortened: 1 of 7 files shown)


jmfrancois (commented):
Great. Could you add a test file with a basic use case and also case over the limit?

.changeset/pink-llamas-cheer.md
```diff
@@ -0,0 +1,5 @@
+---
+'@talend/design-system': minor
+---
+
+feat(TDC-7254): Stepper progress updating according to current step
```
jmfrancois, line 5:
```suggestion
feat(TDC-7254/Stepper): Add a new props `currentStepIndex` to control current step. Set to zero by default.

This fix an issue as the previous code is based on react ref which is not updated when dom is changed. As we don't want to observe mutation on the DOM, let's go back on classic react patterns, make it pure and ask for a state
```

jmfrancois (approved):
with a change on the changelog

PR #4825, opened 2023-07-31
fix(TDOPS-4851):  No copied message when switching back

hbhong:
[TDOPS-4851 No copy info after switch project](https://jira.talendforge.org/browse/TDOPS-4851)
It is easy to reproduce:
1. Select one record A, clicking the copy button, and it shows a "Copied to clipboard" message.
2. Switch to another record B, and do nothing.
3. Switch back to record A. and click the copy button, there is no "Copied to clipboard" message.

When switching back the first record and clicking copy button, because the `clipboardValue` is not changed, so no calling for the `setCopiedValue`, copiedValue is not changed.
```
		useEffect(() => {
			setCopiedValue(clipboardValue as string);
		}, [clipboardValue]);
```
the inputValue is changed and copiedValue is an empty string, so it returns"", and no message pops.
```
		const getDescriptionMessage = () => {
			if (copyError) {
				return copyError.message;
			} else if (copiedValue && copiedValue === inputValue) {
				return t('FORM_COPY_COPIED_TO_CLIPBOARD', 'Copied to clipboard');
			}
			return '';
		};
```

 update the copiedValue at same time when clicking copy button


VolodymyrKovalM (approved):

.changeset/pretty-yaks-joke.md
```diff
@@ -0,0 +1,5 @@
+---
+'@talend/design-system': minor
```
ybaskaran, line 2: minor or patch ?

packages/design-system/src/components/Form/Field/Input/Input.Copy.tsx
```diff
@@ -63,7 +63,11 @@ const InputCopy = forwardRef(
 			}
 			return '';
 		};
-
+		const doCopy = () => {
+			const inputValue = inputRef.current?.value || '';
```
ybaskaran, line 67: also please rename inputValue 
>> 'inputValue' is already declared

.changeset/pretty-yaks-joke.md
hbhong: update to patch

ybaskaran (approved):

PR #4961, opened 2023-10-27
chore(components/PlainTextTitle): pass data attributes to edit button

yyanwang:
We would like to add data-tracking attribute to edit button inside PlainTextTitle for tracking.

Pass data attributes (props starts with 'data-') to the edit button inside.


inna-i (approved):

packages/components/src/EditableText/PlainTextTitle.component.tsx
```diff
@@ -15,6 +15,11 @@ type PlainTextTitleProps = {
 	onEdit: () => void;
 	text: string;
 	t: TFunction;
+	[key: string]: any;
+};
+
+type DataProps = {
```
tlnd-mhuchet, line 21: In the design system, we made a type with the list of allowed properties like this:

```
export type DataAttributes = {
	'data-feature'?: string;
	'data-test'?: string;
	'data-testid'?: string;
};
```

Don't know if it covers all your needs here. It's true that if we want to allow any `data-` property the generic way is better.

tlnd-mhuchet (approved):

jmfrancois (approved):

packages/components/src/EditableText/PlainTextTitle.component.tsx
yyanwang: Hi @tlnd-mhuchet ,
Thanks for info! I was not aware of that.
In TMC we need to attach `data-tracking` attribute for pendo. I would like to add this to DataAttributes, but I found this component is not in design system, but in react-components package 😅 , also it's the only component written in ts in the package.
Let's keep this now and maybe later remove the types from this file to keep consistency.

PR #5127, opened 2024-01-22
fix(TDOPS-5926): set fetch credentials include

lmaillet:
Since TMC is served by Gloo, as soon as the maxAge of the session cookie is exceeded a workflow is started to request a new cookie.
It fails because when calling `iam.int.cloud.talend.com/oidc/idp/authorize` from `app.int.cloud.talend.com`, fetch does not send the credentials (IAMSESSION cookie).

Use `include` instead of `same-origin` in the credentials option of fetch.


romainseb (approved):

packages/cmf/src/sagas/http.js
```diff
@@ -144,7 +144,7 @@ export function httpFetch(url, config, method, payload) {
 
 	const params = merge(
 		{
-			credentials: 'same-origin',
+			credentials: 'include',
```
lmaillet, line 147: Need to send the credentials (cookies) on cross-origin request. From `iam.*.cloud.talend.com` to  to `app.*.cloud.talend.com`.
See https://developer.mozilla.org/en-US/docs/Web/API/Request/credentials

Gbacc (approved):

PR #5211, opened 2024-02-27
chore(TUX-1228): update CollapsiblePanel to support multiple actions

andrewmoreno:
This PR updates `CollapsiblePanel` to support multiple actions passed as prop instead of just one.
For more info: https://jira.talendforge.org/browse/TUX-1228

Update the component props and their type.

### DIFF
--- a/.changeset/blue-frogs-rush.md
+++ b/.changeset/blue-frogs-rush.md
@@ -1,26 +1,5 @@
 ---
-'@talend/design-system': major
+'@talend/design-system': minor
 ---
 
 chore(TUX-1228): update CollapsiblePanel to support multiple actions
-
-## BREAKING CHANGES:
-
-`CollapsiblePanel` now supports multiple actions. The `action` property has been replaced by `actions` which is an array of objects with the same shape used before.
-
-```diff
-  <CollapsiblePanel
-    ...
--   action={{
-+   actions={[
-      {
-        icon: 'plus',
-        tooltip: 'action tooltip',
-        callback: () => window.alert('action callback'),
-      },
-    ]}
--   }}
-  >
-    ...
-  </CollapsiblePanel>
-```
--- a/.changeset/blue-frogs-rush.md
+++ b/.changeset/blue-frogs-rush.md
@@ -1,0 +1,5 @@
+---
+'@talend/design-system': minor
+---
+
+chore(TUX-1228): update CollapsiblePanel to support two actions
--- a/.changeset/blue-frogs-rush.md
+++ b/.changeset/blue-frogs-rush.md
@@ -1,5 +1,26 @@
 ---
-'@talend/design-system': minor
+'@talend/design-system': major
 ---
 
-chore(TUX-1228): update CollapsiblePanel to support two actions
+chore(TUX-1228): update CollapsiblePanel to support multiple actions
+
+## BREAKING CHANGES:
+
+`CollapsiblePanel` now supports multiple actions. The `action` property has been replaced by `actions` which is an array of objects with the same shape used before.
+
+```diff
+  <CollapsiblePanel
+    ...
+-   action={{
++   actions={[
+      {
+        icon: 'plus',
+        tooltip: 'action tooltip',
+        callback: () => window.alert('action callback'),
+      },
+    ]}
+-   }}
+  >
+    ...
+  </CollapsiblePanel>
+```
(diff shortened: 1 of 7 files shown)


jmainguytalend (changes requested):
Please update packages/design-system/src/stories/navigation/Accordion.mdx so storybook for this component will not be broken anymore ;)

`<Canvas of={Stories.WithAction} />`
v
`<Canvas of={Stories.WithActions} />`

(See http://5211.talend.surge.sh/design-system/?path=/docs/navigation-accordion--docs)

Don't hesitate to ping me on slack when updated so I can do visual review. Thanks

packages/design-system/src/components/Accordion/Primitive/CollapsiblePanel.tsx
```diff
@@ -1,20 +1,21 @@
...
+
 import styles from './CollapsiblePanel.module.scss';
 
 type CollapsiblePanelCommonPropsType = {
 	children: ReactChild;
 	managed?: boolean;
 	expanded?: boolean;
 	index?: number;
-	action?: PanelHeaderAction;
+	actions?: [PanelHeaderAction, PanelHeaderAction?];
```
jmainguytalend, line 18: What do you think about allowing unlimited number of actions ?
```suggestion
	actions?: PanelHeaderAction[];
```

jmainguytalend: I think it's already supported by `actions.map(` you did in CollapsiblePanelHeader

packages/design-system/src/components/Accordion/Primitive/CollapsiblePanelHeader.tsx
```diff
@@ -22,7 +22,7 @@ export type CollapsiblePanelHeaderPropsType = {
 	title?: ReactChild;
 	status?: keyof typeof variants;
 	metadata?: ReactChild[];
-	action?: PanelHeaderAction;
+	actions?: [PanelHeaderAction, PanelHeaderAction?];
```
jmainguytalend, line 25: If we allow any number of actions
```suggestion
	actions?: PanelHeaderAction[];
```

packages/design-system/src/components/Accordion/Primitive/CollapsiblePanel.tsx
andrewmoreno: I like the idea however i was just following what was required in the ticket... but I will update as you said 🙂

.changeset/blue-frogs-rush.md
```diff
@@ -0,0 +1,5 @@
+---
+'@talend/design-system': minor
```
romainseb, line 2: This looks to be a breaking change for design-system.
Also, there is changes also in form packages but without a changeset

andrewmoreno: Not really a change there, correct? I only updated the way the prop is passed.

.changeset/blue-frogs-rush.md
```diff
@@ -0,0 +1,26 @@
+---
+'@talend/design-system': major
+---
+
+chore(TUX-1228): update CollapsiblePanel to support multiple actions
+
+## BREAKING CHANGES:
+
+`CollapsiblePanel` now supports multiple actions. The `action` property has been replaced by `actions` which is an array of objects with the same shape used before.
```
romainseb, line 9: Just to be sure, do we have some JSON form that use theses fields ?
Cause a breaking change for the TCK form should be avoided

jmainguytalend: Maybe we can avoid breaking changes by don't updating props name and allowing single or multiple values :

`action?: PanelHeaderAction | PanelHeaderAction[];`

And you have to manage then if single or multiple values in CollapsiblePanelHeader.tsx

andrewmoreno: I'm ok with that if @romainseb agrees too

jmainguytalend (approved):

.changeset/blue-frogs-rush.md
romainseb: I think it's a safer approach indeed

andrewmoreno: Alright, thank you both for your review and suggestions 👍

PR #5226, opened 2024-03-11
chore: Remove ally.js

mhuchet:
We use `ally.js` to have a polyfill for `:focus-within` in css.

It is not really needed anymore as this pseudo class in now widely supported.

Remove `ally.js` polyfill and switch the usage of the lib css class to the pseudo class.


Gbacc (approved):

fork/module-to-cdn/modules.json
```diff
@@ -221,15 +221,6 @@
...
   },
-  "ally.js": {
-    "var": "ally",
-    "versions": {
-      ">= 1.0.0": {
-        "development": "/ally.min.js",
-        "production": "/ally.min.js"
-      }
-    }
-  },
```
jmfrancois, line 232: I would let it here in terms of backward compatiliby, in case you still have it in the code base it can break your app.
Remove it from the code and you should not have it anymore in the final application without this change.

jmfrancois (changes requested):
Also you need to add coresponding changeset

jmfrancois (approved):

PR #5319, opened 2024-05-28
fix(TMC-28581): UI Form display fix with hint and required asterisk

Gbacc:
UI Form display fix with hint and required asterisk

### DIFF
--- a/packages/forms/src/UIForm/utils/labels.js
+++ b/packages/forms/src/UIForm/utils/labels.js
@@ -1,6 +1,6 @@
 import { ButtonIcon, Popover, StackHorizontal } from '@talend/design-system';
 
-export const getLabelProps = (title, labelProps, hint) => {
+export const getLabelProps = (title, labelProps, hint, required) => {
 	if (!hint) {
 		return {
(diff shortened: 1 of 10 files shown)


yyanwang (approved):

packages/forms/src/UIForm/utils/labels.js
```diff
@@ -9,8 +9,9 @@ export const getLabelProps = (title, labelProps, hint) => {
 	}
 	return {
 		children: (
-			<StackHorizontal gap="XS" align="center">
+			<StackHorizontal gap="XXS" align="center">
 				{title}
+				{required && '*'}
```
jmfrancois, line 14: I prefer if you can use a classic css approach.
Here you can't for example make it red

```
.required label:after {
    color: #e32;
    content: ' *';
    display:inline;
}
```

Gbacc: Ok good point 
I could also change it on design system side , the idea was to have the same behavior

aamalric-talend (approved):

PR #5364, opened 2024-06-28
fix(): default value when add item in array widget

jmainguytalend:


packages/forms/src/UIForm/fieldsets/Array/Array.component.test.js
```diff
@@ -243,6 +243,40 @@ describe('Array component', () => {
...
+				name: 'default name',
+				email: 'default email',
+				comment: 'default comment',
+			};
+
+			const enhancedSchema = {
+				...schema,
+				schema: {
+					...schema.schema,
+					default: [defaultValue],
```
jmfrancois, line 258: this looks like a default value for the entire array, not only when adding an item

jmainguytalend: I used default value defined to initialize component with one single value to create new entry with same default value.
So here the default value for first element and that will be used for next added ones.

tlnd-mhuchet (approved):

romainseb (approved):

yyanwang (approved):

PR #5783, opened 2026-04-17
feat(http): add httpDeleteWithPayload

VolodymyrKovalM:
There is DELETE API in IAM which requires application names in request body
According to specification, it may not be perfect to have body in DELETE request, and may not be intended to do this way,
but, this API is used in TMC

It worked before with CMF http, as config was spread when calling interceptors https://github.com/Talend/ui/blob/master/packages/cmf/src/sagas/http.js#L179 , so, it was possible to pass payload in the config, and it would override undefined payload.
Probably this was not intended to pass payload in config and send body in delete request, most probably, it was kind of bug / "back door", but it worked for one API in TMC )

Now, as CMF has been removed from TMC, @talend/http package is used for now, and in @talend/http there is not such kind of "back door" to send body in delete request, so, the TMC part which call that API from IAM fails with error

For now I see few solutions
1.  Modify current [httpDelete](https://github.com/Talend/ui/blob/master/packages/cmf/src/sagas/http.js#L179) function, so, it will accept payload as argument
2. Implement separate function, something like "httpDeleteWithPayload", to not modify existing function and impact existing code
3. Use fetch API directly for that one DELETE request in TMC
4. Export "httpFetch" function, to allow specific implementation for httpDelete, which will be done in TMC

(maybe in future the DELETE API in IAM will be refactored, to accept application names as GET parameters instead of request body, in this case we will not need to send the body)

### DIFF
--- a/.changeset/ninety-mice-work.md
+++ b/.changeset/ninety-mice-work.md
@@ -2,4 +2,4 @@
 '@talend/http': minor
 ---
 
-Add function to allow sending HTTP DELETE with request body
+Add `httpDeleteWithPayload` function to allow sending HTTP DELETE with request body
--- a/.changeset/ninety-mice-work.md
+++ b/.changeset/ninety-mice-work.md
@@ -1,0 +1,5 @@
+---
+'@talend/http': minor
+---
+
+Export httpFetch to allow sending payload in DELETE request
--- a/.changeset/ninety-mice-work.md
+++ b/.changeset/ninety-mice-work.md
@@ -2,4 +2,4 @@
 '@talend/http': minor
 ---
 
-Export httpFetch to allow sending payload in DELETE request
+Add function to allow sending HTTP DELETE with request body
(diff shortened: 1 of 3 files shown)


packages/http/src/index.ts
```diff
@@ -1,5 +1,6 @@
 // eslint-disable-next-line import/prefer-default-export
 export { http } from './async';
+export { httpFetch } from './http.common';
```
smouillour: IMO, better to not expose private api. 
Better to add new public function to manage the edge case

smouillour (approved):

lmaillet (approved):

jmfrancois (approved):

.changeset/ninety-mice-work.md
```diff
@@ -0,0 +1,5 @@
+---
+'@talend/http': minor
+---
+
+Add function to allow sending HTTP DELETE with request body
```
jmfrancois:
```suggestion
Add `httpDeleteWithPayload` function to allow sending HTTP DELETE with request body
```

PR #5849, opened 2026-09-04
chore: migrate from yarn to pnpm

jmfrancois:
## Summary
- Migrate the monorepo from Yarn Classic to pnpm (isolated/strict `node_modules`), converting root config, all `tools/scripts-*` CLIs that shelled out to yarn, and CI workflows.
- Delete `tools/scripts-config-react-webpack` and `tools/scripts-config-cdn` (dead, zero real consumers) instead of patching their yarn-hoisting-coupled code, in line with the ongoing webpack→vite / jest→vitest migration.
- Fix phantom-dependency issues that only worked under yarn's flat hoist (`scripts-core`'s prettier/stylelint resolution, per-package `resolutions` in `packages/components`) and a critical one caught during validation: internal `@talend/*` deps don't use the `workspace:` protocol, so plain `pnpm install` silently resolved them from the npm registry instead of linking local workspace code — fixed via `linkWorkspacePackages: true`.
- Retire `.github/workflows/yarn-deduplicate.yml` (pnpm dedupes natively).
- Drop `@talend/scripts-publish-local` entirely (unused local-registry helper, no consumers) and its pending changeset.

## Test plan
- [x] `pnpm install` completes clean with no phantom-dependency errors
- [x] `pnpm build:lib` succeeds across all 26 workspace packages
- [x] `talend-scripts extends` resolves all 5 generated config branches (prettier/stylelint/eslint/babel/typescript) end-to-end
- [x] Rewritten `talend-yarn-workspace` CLI runs correctly against `pnpm -r run --if-present`
- [x] ESLint config loads and runs under pnpm (only pre-existing lint debt surfaced, no resolution crashes)
- [ ] CI green on this PR (workflows converted but not yet run)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jmfrancois:
@copilot resolve the merge conflicts in this pull request by droping the yarn.lock file

